PR #3, opened 2015-07-18
[RFC] Building ld-logger main with the appropriate arch switch.

labuwx:
From now ld-logger pack64bit can be built on 64bit machines without multilib support.

Tested on Arch linux x64 without gcc-multilib.

### DIFF
--- a/tests/test_helper/thrift_client_to_db.py
+++ b/tests/test_helper/thrift_client_to_db.py
@@ -102,7 +102,9 @@ class CCViewerHelper(ThriftAPIHelper):
     def __init__(self, host, port, uri, auto_handle_connection=True):
         # import only if necessary; some tests may not add this to PYTHONPATH
         from codeCheckerDBAccess import codeCheckerDBAccess
+        from codeCheckerDBAccess.constants import MAX_QUERY_SIZE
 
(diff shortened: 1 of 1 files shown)


labuwx:
So the reason I marked this pull request RFC:
When parallel jobs are enabled, it is possible that `ldlogger` is executed before `MAIN_ARCH` gets set to 32.

However `build_package.py` does not enable multiple build jobs so Codechecker users won't experience any unwanted behavior.
 
Secondly this makefile already contains problems like this: (`packbin` assumes that `ldlogger` is built)
```
pack32bit: 32bit packbin
32bit: ldlogger ldlogger_32.so

packbin:
	cp ldlogger $(BIN_DIR)
```

Please tell me if I should change something.

gyorb:
I was thinking about a cleanup in the make file, to be more explicit what is built when.
- create a separate rule to build 32bit and 64bit ld logger binaries (lib already exists)
- all: rule should only build (no packaging)
- create separate packager rules for 32bit only (bin and lib), 64bit only (bin and lib)
- create a new packager rules for mixed packages (32bit and 64bit libraries, but only 32bit or 64bit binary)

What do you think?

PR #19, opened 2015-07-28
Some fixes in thrift_client_to_db.py

labuwx:


gyorb:
Thanks for the fix, and update.

Issue #31, opened 2015-08-03
whitelist for directories and files


dkrupp:
sometimes it is useful to specify in a file the directories
from where files need to be checked.

this is especially problematic if the project takes file from many places of the filesystem.

Issue #41, opened 2015-08-04
allow for patterns in skip file - or add whitelist strategy


torbjoernk:
Thanks for this tool! Looking awesome right now :+1: 

My aim is to provide a skip file for CodeChecker within my project's source tree. But I want to only parse my own code and not optional and potential 3rd-party libraries. Thus, I can not put in absolute paths such as `/usr/include`. As well it's not possible to exclude specific files in the project's source tree (e.g. `<PROJECT>/include/3rdparty`) as I can not make any assumptions on the location of the project tree.

I tried specifying patterns such as `eigen3/`, `*/eigen3`, `/boost` or `*/boost` in a skip file. But that did not work. Only absolute paths such as `/usr/include/eigen3` are working.

It would be awesome being able to exclude paths based on patterns or - even better - whitelist only specific files or paths only in the project's source tree.

igalex:
Hi!

The skip file format is just changed. Now you can use glob patterns and allow/reject signs in skip files. You can find its documentation [here](https://github.com/Ericsson/codechecker/blob/master/docs/user_guide.md#skip-file).

PR #43, opened 2015-08-06
Skip file improvements

igalex:
- Add whitelist support
- Glob based pattern matching in skip files

Resolves Ericsson/codechecker#31 and Ericsson/codechecker#41


gyorb:
This request should be merged together with clean report filtering #44 
Both modifications modify database schema in an incompatible way, version should be increased.

PR #63, opened 2015-09-08
Catch the exception 'TypeError' when parse_compile_commands_json fails

sylvestre:
Exemple:
coercing to Unicode: need string or buffer, file found


igalex:
Thanks for your contribution. It seems OK.

PR #64, opened 2015-09-08
Flake8 consistency

sylvestre:
If you are ok with these kind of changes, I will fix the rest and make sure they don't come back.

### DIFF
--- a/codechecker_lib/host_check.py
+++ b/codechecker_lib/host_check.py
@@ -40,7 +40,6 @@ def check_zlib():
 # -----------------------------------------------------------------------------
 def check_psycopg2():
     try:
-        import psycopg2
         return True
     except Exception as ex:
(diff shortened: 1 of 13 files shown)


codechecker_lib/host_check.py
```diff
@@ -40,7 +40,6 @@ def check_zlib():
 # -----------------------------------------------------------------------------
 def check_psycopg2():
     try:
-        import psycopg2
```
igalex: This line is required. Please add it back.

igalex:
Please fix the issue above. Otherwise it's OK. Thank you again :-)

sylvestre:
@igalex any issue? Thanks

igalex:
Merged with a modification in codechecker_lib/host_check.py. Thanks for the patch :-)

PR #71, opened 2015-09-10
Work on test helper

labuwx:
I've been putting `run_performance_test.py` in shape, for which I need unlimited versions (aka getAll) of `getRunResults, getNewResults, getResolvedResults, getUnresolvedResults`.
Since these functions are very similar I added an universal `getAll` function to the test helper to minimize redundant code.


igalex:
Looks good but it has style issues. Please run flake8 against your code and reformat it according to the warning messages.

labuwx:
Thanks, I fixed them.

Issue #68, opened 2015-09-09
WEBGUI: index column should show the index of the report


dkrupp:
and not id field

PR #87, opened 2015-09-22
FIXES Ericsson/codechecker#68

bobszi:
Also removed deprecated images and css entries.


bobszi:
'Index' column omitted entirely as per design decision.

bobszi:
Fixed.

PR #92, opened 2015-09-25
Database migration support

igalex:
- Add new dependency: alembic
- Add initial schema script
- Set sqlalchemy naming conventions
- Use auto-generated sequences for table ids


dkrupp:
Please describe in a developers_guide.md file how one should perform schema upgrade according to this change.

Issue #148, opened 2015-11-16
support sqlite


dkrupp:
CodeChecker should be supporting sqlite as the bug storage database besides postgres. The advantage would be that for smaller projects, that do not store many bugs, the configuration would be simpler as sqlite consist of a single db file, and does not need a separate db server process.

PR #163, opened 2015-11-24
SQLite support

igalex:
- Move all database server specific code to database_handler.py
- Change check constraint naming convensions to be compatible with SQLite. This is backward compatible because currently there is no check constraint in the database schema.
- Fix Run.mark_finished() to store integer
- Fix suppress queries to use True and False (boolean) instead of strings.
- Fix supress -> suppress typos
- Add sqlite support to package_test
- Add sqlite test to Travis

This resolves Ericsson/codechecker#148

### DIFF
--- a/codechecker_lib/database_handler.py
+++ b/codechecker_lib/database_handler.py
@@ -136,14 +136,15 @@ class SQLServer(object):
 
 
     @classmethod
-    def from_cmdline_args(cls, args, context, env=None):
+    def from_cmdline_args(cls, args, workspace, migration_root, env=None):
         '''
         Normally only this method is called form outside of this module in
         order to instance the proper server implementation.
 
         Parameters:
             args: the command line arguments from CodeChecker.py
-            context: a Context object
+            workspace: path to the CodeChecker workspace directory
+            migration_root: path to the database migration scripts
             env: a run environment dictionary.
         '''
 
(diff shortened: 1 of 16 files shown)


codechecker_lib/database_handler.py
```diff
@@ -4,74 +4,379 @@
...
+        self.host = host
+        self.port = port
+        self.user = user
+        self.database = database
+        self.password = password
+        self.run_env = run_env
+
+        self.proc = None
+
+    def _call_command(self, command):
```
gyorb: call_command could be moved out to util module so other modules can use it
(for example get_checker_list function in the analyzer module).
If output and error messages required only in a blocking way the subprocess.check_output function can be used.

codechecker_lib/database_handler.py
```diff
@@ -4,74 +4,379 @@
...
+        Returns the connection string for SQLAlchemy.
+
+        DO NOT LOG THE CONNECTION STRING BECAUSE IT MAY CONTAIN THE PASSWORD
+        FOR THE DATABASE!
+        '''
+        pass
+
+
+    @classmethod
+    def from_cmdline_args(cls, args, context, env=None):
```
gyorb: arg:
should we use the whole arg here? 

context:
try not to depend too much on the context object, because it makes later testing/mocking harder.
right now both the postgresql and the sqlite uses only the codechecker_workspace, and migration_root values from it

codechecker_lib/database_handler.py
```diff
@@ -4,74 +4,379 @@
...
+    def _initialize_database_data(self):
+        '''Initialize a PostgreSQL instance with initdb. '''
+        LOG.debug('Initializing database at ' + self.path)
+
+        init_db = ['initdb', '-U', self.user, '-D', self.path, '-E SQL_ASCII']
+
+        err, code = self._call_command(init_db)
+        # logger -> print error
+        return code == 0
+
```
gyorb: pylint missing newline

codechecker_lib/database_handler.py
igalex: Should we use the whole arg here?
Yes, because we don't know what arguments will be required by a new implementation. The alternative would be to add 5 more parameter (and some more in the future) which looks terrible.

context:
OK, I'll change this.

PR #187, opened 2015-12-01
Assigned severity to more checks.

Xazax-hun:


gyorb:
These checkers will be available when the clang tidy integration is done.

PR #219, opened 2015-12-29
Add the incremental build example

sylvestre:


gyorb:
Thanks for extending the user guide.

I know last time I sad development is done on the development branch, but a few weeks ago we changed it and the main development branch is the master. We try to keep it as stable as possible.
I will merge the develop to the master (and delete the develop), so you do not have to do anything.
I plan to create a contribution guide so it will be more obvious how to contribute.

bruntib:
The update mode of CodeChecker what this pull request is about has become the default behavior of checking process. If the intention of the user is to clean the previous results and recheck the project then he or she can provide the `--force` flag instead of `--update` which is now deprecated. One can read about this flag in the [User guide](https://github.com/Ericsson/codechecker/blob/master/docs/user_guide.md) and there is also an example in [Usage examples](https://github.com/Ericsson/codechecker/blob/master/docs/usage.md) for update mode. Do you think these to be sufficient to describe the update behavior, or is further documentation needed?

gyorb:
We will delete the develop branch soon.
I will close this pull request please create a new issue or pull request.
Thanks!

Issue #234, opened 2016-02-05
extend compiler logging env variable


gyorb:
The problem is with the default cmake behaviour (calls c++)

Extend the CC_LOGGER_GCC_LIKE environment variable with c++, cpp

PR #237, opened 2016-02-09
Increment version number to 5.2

bruntib:


Xazax-hun:
I think this modification should only be merged after #234 is fixed.

PR #243, opened 2016-02-12
Dependency addition plus fix of ordering of instructions to run during setup

AlexTelon:
Added dependency wget which is needed during the installation of python modules. /`external/source-deps/ext_source_deps_config.json` declares that `wget `is to be used to get dependencies.

Also the pip install command  should be executed after the codechecker repo is cloned so that it can read the requirements and install them.


gyorb:
thanks for the update

AlexTelon:
My first pull request to an open source project, feels good :)

PR #267, opened 2016-03-17
Quickcheck migration to nosetests + a little unit_test cleanup

bobszi:
- Quickcheck migration to nosetests
  - Included moving of test files to a directory for neater dir structure
- A little unit_test cleanup:
  - More sensible directory names, __init__.py removed
- Env variable setter bash script added, tests are now using the variables set by this script


tests/quickcheck_test/test_quickcheck.py
```diff
@@ -0,0 +1,52 @@
...
+        os.chdir(cls.test_dir)
+
+    def __check_one_file(self, path):
+        """Test quickcheck output on a ".output" file."""
+        with open(path, 'r') as ofile:
+            lines = ofile.readlines()
+
+        options, correct_output = (lines[0].strip(), ''.join(lines[2:]))
+
+        command = 'CodeChecker quickcheck --analyzers clangsa ' + options
```
gyorb: Please move the whole check command to the top of the test output files

tests/quickcheck_test/test_quickcheck.py
```diff
@@ -0,0 +1,52 @@
...
+            os.path.join(cls.package_dir, 'bin') + ':' + cls.env['PATH']
+
+        cls.test_dir = os.path.join(
+            os.path.dirname(__file__), 'quickcheck_test_files')
+
+        # Change working dir to testfile dir so CodeChecker can be run easily.
+        os.chdir(cls.test_dir)
+
+    def __check_one_file(self, path):
+        """Test quickcheck output on a ".output" file."""
```
gyorb: add some additional description about the output test file format which is processed here

tests/quickcheck_test/test_quickcheck.py
```diff
@@ -0,0 +1,52 @@
...
+    """This class tests the CodeChecker quickcheck feature."""
+
+    @classmethod
+    def setup_class(cls):
+        """Setup the class."""
+        cls.package_dir = \
+            os.path.realpath(os.environ.get('TEST_CODECHECKER_DIR'))
+
+        # Put CodeChecker/bin to PATH so CodeChecker command becomes available.
+        cls.env = os.environ
```
gyorb: use os.environ.copy() to create a new environment which is used only for the current tests, without the copy it will modify the running process environment

tests/quickcheck_test/test_quickcheck.py
```diff
@@ -0,0 +1,52 @@
...
+        command = 'CodeChecker quickcheck --analyzers clangsa ' + options
+
+        output = subprocess.check_output(
+            ['bash', '-c', command], env=self.env, cwd=self.test_dir)
+
+        self.assertEqual(output, correct_output)
+
+    def test_quickcheck_files(self):
+        """Iterate over the test directory and run all tests in it."""
+        for ofile in glob.glob(self.test_dir + os.sep + '*.output'):
```
gyorb: use os.path.join() instead of adding with os.sep

test_env_setup.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
gyorb: It might be a good idea to allow sourcing this script from other directories not only from the current one.
The TEST_CODECHECKER_PACKAGE_DIR might be set to a wrong directory.
See the removed init.sh for example https://github.com/Ericsson/codechecker/commit/d4ab4c94e4fc301b2cbed4e0ab4a0696f36511b8

tests/quickcheck_test/test_quickcheck.py
bobszi: I do not follow. The corresponding build commands must be appended after the 'CodeChecker quickcheck --analyzers clangsa ', but these build commands become available only after 40th line.

PR #275, opened 2016-03-29
Nosetests migration - new directory structure of tests

bobszi:


gyorb:
please wait and update the merge request after the previous pull request #272 is accepted, it will remove the bzip2 test project so less files need to be moved
also pay attention to remove the generated object files by the build, before committing

PR #295, opened 2016-04-28
Simplify adding reports.

Xazax-hun:
Do the filtering in one step instead of two.

### DIFF
--- a/storage_server/report_server.py
+++ b/storage_server/report_server.py
@@ -356,20 +356,18 @@ class CheckerReportHandler(object):
         '''
         '''
         try:
-            # TODO: perfomance issues when executing the following query on large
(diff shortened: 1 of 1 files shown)


storage_server/report_server.py
```diff
@@ -357,19 +357,17 @@ def addReport(self,
         '''
         try:
             # TODO: perfomance issues when executing the following query on large
-            #       databaseses
-            reports = self.session.query(self.report_ident) \
-                                  .filter(Report.bug_id == bug_hash)
-
+            #       databaseses?
```
gyorb: please move the comment to the relevant query

Xazax-hun: Done.

PR #300, opened 2016-05-04
SQLite version server start bugfix

### LINKED ISSUE #297: server start failed with default sqlite database
CodeChecker server -w workspace command fails if the workspace directory does not exists.

bruntib:
This commit fixes #297. The workspace directory has to exist so SQLite
database can be created there. When CodeChecker server is started
without a previously created workspace, then it causes an error.

### DIFF
--- a/codechecker_lib/arg_handler.py
+++ b/codechecker_lib/arg_handler.py
@@ -98,6 +98,8 @@ def handle_server(args):
     if not args.workspace and util.is_localhost(args.dbaddress):
         LOG.info("Workspace is required when database server run on localhost.")
         sys.exit(1)
+    elif args.workspace and not os.path.exists(args.workspace):
+        os.makedirs(args.workspace)
 
     if args.suppress is None:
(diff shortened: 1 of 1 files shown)


codechecker_lib/database_handler.py
```diff
@@ -457,6 +457,9 @@ def __init__(self, workspace, migration_root, run_env=None):
         self.dbpath = os.path.join(workspace, 'codechecker.sqlite')
         self.run_env = run_env
     
+        if not os.path.exists(workspace):
```
gyorb: I think creating the workspace should be done in the arg_handler.py [handle_check](https://github.com/Ericsson/codechecker/blob/master/codechecker_lib/arg_handler.py#L202).
Why did this [L212](https://github.com/Ericsson/codechecker/blob/master/codechecker_lib/arg_handler.py#L212) in handle_check failed to create the workspace for sqlite?

bruntib:
handle_check function runs when checking a project. The error happens when a server is started without checking, i.e. workspace directory has not been created.

PR #301, opened 2016-05-05
Various cleanups and small bug fixes.

Xazax-hun:
Various cleanups and small bug fixes.

### DIFF
--- a/storage_server/report_server.py
+++ b/storage_server/report_server.py
@@ -47,64 +47,58 @@ class CheckerReportHandler(object):
     '''
 
     def __sequence_deleter(self, table, first_id):
-        """Delete points of sequnce in a general way."""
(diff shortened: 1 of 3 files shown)


storage_server/report_server.py
```diff
@@ -239,14 +230,14 @@ def needFileContent(self, run_id, filepath):
         f = self.session.query(File) \
                         .filter(and_(File.run_id == run_id,
                                      File.filepath == filepath)) \
-                        .first()
+                        .one_or_none()
```
gyorb: Please update the minimal required SQLAlchemy version in the documentation too to a higher one if a new query is used which is only available from 1.0.9 and handle the possible sqlalchemy.orm.exc.MultipleResultsFound exception.

PR #310, opened 2016-05-25
add new unknown options to the list

gyorb:
add new unknown option to the list unsupported by clang right now


bruntib:
Wouldn't we like to cover the `no-` counterparts as well, like in case of `'^-m(no-)?spe.*': 0`?

gyorb:
I'd like to keep this list as minimal as possible and only extend with elements which came up during analysis. I do not want to skip possible flags which could be handled by clang.
What specific flags do you have in mind?

bruntib:
There are some duplicates in the list:
```
^-fno-var-tracking-assignments
^-fconserve-stack
^-fcond-mismatch
```
Some alphabetic or lexical ordering might help.

gyorb:
thanks for pointing out the duplicates,

PR #314, opened 2016-05-30
Import checker results from PList

### LINKED ISSUE #288: PList import
It would be great to be able to use the codechecker as a viewer without using it as an analyzer wrapper script.

The best way to do it would be the ability to import plist files generated by scan-build or other analyzer runner script.

bruntib:
This commit fixes #288. One can now import checker results from PList
files too.


codechecker/CodeChecker.py
```diff
@@ -335,6 +335,38 @@ def signal_handler(sig, frame):
...
+                                  dest="name", required=True,
+                                  default=argparse.SUPPRESS,
+                                  help='Name of the analysis.')
+        plist_parser.add_argument('-d', '--directory', type=str,
+                                  dest="directory", required=True,
+                                  help='Path of a directory containing plist files to parse.')
+        plist_parser.add_argument('-j', '--jobs', type=int, dest="jobs",
+                                  default=1, required=False,
+                                  help='''\
+Number of jobs. Start multiple processes for faster analisys''')
```
gyorb: We could use global help messages for the arguments which are used at multiple places. Later modification will be easier.

codechecker_lib/analyzers/result_handler_plist_to_db.py
```diff
@@ -33,6 +33,127 @@ def __init__(self, buildaction, workspace, run_id):
...
         self.__run_id = run_id
 
+    def __store_bugs(self, files, bugs, connection, analisys_id):
+        file_ids = {}
+        # Send content of file to the server if needed
+        for file_name in files:
+            file_descriptor = connection.need_file_content(self.__run_id,
+                                                           file_name)
+            file_ids[file_name] = file_descriptor.fileId
+            if file_descriptor.needed:
```
gyorb: This part should be hardened. What if the original source file is not available. It might be still useful to store the results without the source file content. Viewing/filtering the result list for statistics could be possible, just viewing the file content with the bug will fail.  Further improvement might be needed for the web GUI #320

gyorb:
A new pull request was open for this issue #322

PR #317, opened 2016-06-01
--saargs and --tidyargs can contain env vars

### LINKED ISSUE #302: environment variables in config files
saargs and tidyargs config files should handle environment variables

for example in an saargs config file:
`-I/home/$$MY_PROJ$$/inc`

bruntib:
This commit fixes #302. If HELLO variable is set in the environment,
then in the config files given by --saargs and --tidyargs all occurences
of $(HELLO) are replaced to its value.

### DIFF
--- a/codechecker_lib/analyzers/analyzer_types.py
+++ b/codechecker_lib/analyzers/analyzer_types.py
@@ -7,6 +7,7 @@
 """
 supported analyzer types
 """
+import os
 import re
 import sys
 
@@ -173,6 +174,14 @@ def initialize_checkers(config_handler,
             else:
                 config_handler.disable_checker(checker_name)
 
+def __replace_env_var(cfg_file):
+    def replacer(matchobj):
+        env_var = matchobj.group(1)
+        if matchobj.group(1) not in os.environ:
+            LOG.error(env_var + ' environment variable not set in ' + cfg_file)
+            return ''
+        return os.environ[env_var]
+    return replacer
 
 def __build_clangsa_config_handler(args, context):
     """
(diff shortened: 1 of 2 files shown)


codechecker_lib/analyzers/analyzer_types.py
```diff
@@ -173,6 +174,12 @@ def initialize_checkers(config_handler,
             else:
                 config_handler.disable_checker(checker_name)
 
+def __replace_env_var(matchobj):
+    env_var = matchobj.group(1)
+    if matchobj.group(1) not in os.environ:
+        LOG.error(env_var + ' environment variable not found')
```
gyorb: Could we provide a bit more specific error message to the users, that the problem is in the provided sa or tidy configuraton file.

codechecker_lib/analyzers/analyzer_types.py
```diff
@@ -189,7 +198,10 @@ def __build_clangsa_config_handler(args, context):
     config_handler.includes = context.extra_includes
     try:
         with open(args.clangsa_args_cfg_file, 'rb') as sa_cfg:
-            config_handler.analyzer_extra_arguments = sa_cfg.read().strip()
+            config_handler.analyzer_extra_arguments = \
+                re.sub('\$\((.*?)\)',
```
Xazax-hun: What does the question mark here?

bruntib: This controls laziness: `abc$(def)ghi$(jkl)mno` would capture `def)ghi$(jkl` instead of `def` and `jkl`.

PR #318, opened 2016-06-01
Test framework update

gyorb:
* fix missing workspace cleanup
* extend test documentation
* extend logging information during the tests
* cleanup test files


run_tests.sh
```diff
@@ -1,15 +1,18 @@
...
-./build_package.py -o $TEST_CODECHECKER_PACKAGE_DIR -v
+./build_package.py -o $TEST_CODECHECKER_PACKAGE_DIR -v || exit 1
 
 # run all the tests
 nosetests tests/test_packages/
 
 # cleanup the environment variables and temporary
 # directories
-clean_tests.sh
+./clean_tests.sh
```
bruntib: This file should be placed in `tests` directory, like `test_env_setup.sh`. As this is in the root like `run_tests.sh`, it seems that there is some reason for the user to run it manually.

tests/test_packages/functional/package_test/test_regression_diff.py
```diff
@@ -89,18 +90,34 @@ def test_get_diff_res_count_unresolved(self):
         base_run_id = runs[0].runId
         new_run_id = runs[1].runId
 
+        base_count = self._cc_client.getRunResultCount(base_run_id, [])
```
gyorb: This functional test tests only the diff api. The result count api has different tests cases.
If this test asserts (getDiffResultCount) it can be useful to see the result count for the two separate runs which can help debugging what caused the test to fail.

PR #322, opened 2016-06-13
Import checker results from PList

### LINKED ISSUE #288: PList import

### LINKED ISSUE #320: handle missing source file content
The GUI should handle if the original source file content could not be stored during the analysis or processing the plist result files from Clang Static analyzer.
Viewing and filtering the results should be still possible only viewing a concrete result should fail with an error message.

bruntib:
This commit fixes #288 and fixes #320. One can now import checker
results from PList files too.


codechecker/CodeChecker.py
```diff
@@ -189,6 +189,10 @@ def signal_handler(sig, frame):
 
         workspace_help_msg = """Directory where the codechecker can store analysis related data."""
 
+        name_help_msg = """Name of the analysis."""
+
+        jobs_help_msg = '''Number of jobs. Start multiple processes for faster analisys''';
```
Xazax-hun: Use the same """ everywhere.

codechecker/CodeChecker.py
```diff
@@ -335,6 +338,37 @@ def signal_handler(sig, frame):
         debug_parser.set_defaults(func=arg_handler.handle_debug)
 
         # --------------------------------------
+        # plist parser
+        plist_parser = subparsers.add_parser('plist',
+                                             formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+                                             help='Parse plist files in the given directory.')
+        plist_parser.add_argument('-w', '--workspace', type=str,
```
Xazax-hun: Postgres is not supported with this feature? Is that intentional?

bruntib: Database connection information are added in line 360.

codechecker_lib/analyzers/result_handler_plist_to_db.py
```diff
@@ -33,6 +34,133 @@ def __init__(self, buildaction, workspace, run_id):
...
+        file_ids = {}
+        # Send content of file to the server if needed
+        for file_name in files:
+            file_descriptor = connection.need_file_content(self.__run_id,
+                                                           file_name)
+            file_ids[file_name] = file_descriptor.fileId
+
+            # sometimes the file doesn't exist, e.g. when the input of the
+            # analysis is pure plist files
+            if not os.path.isfile(file_name):
```
Xazax-hun: In what case can this happen? Shouldn't we rely on that the files mentioned by the plist are exist in the file system? Nonetheless we should add a debug message that a file was not found.

codechecker/CodeChecker.py
Xazax-hun: Indeed, I did not notice that, thanks.

codechecker_lib/analyzers/result_handler_plist_to_db.py
```diff
@@ -33,6 +34,133 @@ def __init__(self, buildaction, workspace, run_id):
...
+                                              bug.category,
+                                              bug.type,
+                                              severity,
+                                              suppress)
+ 
+            report_ids.append(report_id)
+
+    def handle_plist(self, plist):
+        with client.get_connection() as connection:
+            analisys_id = connection.add_build_action(self.__run_id,
```
Xazax-hun: Once the update mode fixes are landed this code needs to be updated.

codechecker_lib/analyzers/result_handler_plist_to_db.py
```diff
@@ -33,6 +34,133 @@ def __init__(self, buildaction, workspace, run_id):
...
+                                                      plist,
+                                                      'Build action from plist')
+
+            try:
+                files, bugs = plist_parser.parse_plist(plist)
+            except Exception as ex:
+                msg = 'Parsing the generated result file failed'
+                LOG.error(msg + ' ' + plist)
+                connection.finish_build_action(analysis_id, msg)
+                return 1
```
Xazax-hun: Why do you return a value here?

codechecker_lib/analyzers/result_handler_plist_to_stdout.py
```diff
@@ -62,6 +62,38 @@ def __format_bug_event(self, event):
...
+                    self.__output.write('\n')
+            self.__output.write('\n')
+
+    def handle_plist(self, plist):
+        try:
+            _, bugs = plist_parser.parse_plist(plist)
+        except Exception as ex:
+            LOG.error('The generated plist is not valid!')
+            LOG.error(ex)
+            return 1
```
Xazax-hun: why do you return a value here?

codechecker_lib/arg_handler.py
```diff
@@ -334,6 +336,70 @@ def handle_quickcheck(args):
...
+        check_env = analyzer_env.get_check_env(context.path_env_extra,
+                                               context.ld_lib_path_extra)
+    
+        sql_server = SQLServer.from_cmdline_args(args,
+                                                 context.codechecker_workspace,
+                                                 context.migration_root,
+                                                 check_env)
+    
+        conn_mgr = client.ConnectionManager(sql_server,
+                                            'localhost',
```
Xazax-hun: Is this always localhost? Isn't it configurable?

codechecker_lib/analyzers/result_handler_plist_to_db.py
bruntib: This intended to cover the case when the source files from which the plists were generated are not in the file system. I don't know if there is any use case when the plists are separated from the code base. But what else would "parse from plist" functionality be used for? Although debug messages for non-existing source files is a good idea.

Xazax-hun: It can be used for debug purpuses or as an interoperability option for scan-build scripts.

codechecker_lib/analyzers/result_handler_plist_to_db.py
bruntib: Because as far as I have seen, this is the general way of failure reporting, like in the `except` block of `handle_results()` function.

codechecker_lib/analyzers/result_handler_plist_to_db.py
gyorb: Update mode fix was merged see #330

codechecker_lib/arg_handler.py
gyorb: localhost is hard coded because the report storage server is started always where the analysis is done on the same host.
Currently it does not support the use case where the storage server is started on a separate machine. It reduces configuration complexity because the storage server is started automatically.

codechecker_lib/analyzers/result_handler_plist_to_db.py
gyorb: Could you update this api call?

PR #325, opened 2016-06-14
use analysis name unique temporary directory for reports

### LINKED ISSUE #321: unique temporary reports directory
The currently used reports directory name should be unique for each analysis. Currently it is set in [config.json](https://github.com/Ericsson/codechecker/blob/master/config/config.json#L19).
If multiple analysis is started with the same workspace they use the same temporary `reports` directory so the results might be stored to the wrong run.
This should be unique for each analysis run (analysis run name could be included), it can be removed from the config.json file.

gyorb:
Resolves #321

### DIFF
--- a/codechecker_lib/analyzers/result_handler_base.py
+++ b/codechecker_lib/analyzers/result_handler_base.py
@@ -54,7 +54,7 @@ class ResultHandler(object):
         self.__analyzer_returncode = 1
         self.__buildaction = action
 
-        self.__result_files = {}
+        self.__res_file = None
 
     @property
     def buildaction(self):
@@ -82,12 +82,6 @@ class ResultHandler(object):
         return the analyzer result files
         """
         return self.__result_files
-
-    def add_result_file(self, source_file_name,  result_file):
-        """
-        result file which should be cleaned
-        """
-        self.__result_files[source_file_name] = result_file
 
     @property
     def skiplist_handler(self):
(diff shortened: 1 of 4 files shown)


codechecker_lib/analyzers/result_handler_base.py
```diff
@@ -180,37 +175,36 @@ def analyzed_source_file(self, file_path):
...
 
-        out_file_name = str(self.buildaction.id) + '_' + \
-                         str(self.buildaction.analyzer_type) \
-                        + '_' + analyzed_file_name + '.out'
+            uid = str(uuid.uuid1()).split('-')[0]
 
-        out_file = os.path.join(self.__workspace, out_file_name)
-        self.add_result_file(analyzed_file_name, out_file)
+            out_file_name = str(self.buildaction.analyzer_type) + \
+                '_' + analyzed_file_name + '_' + uid + '.out'
```
Xazax-hun: Is .out a good extension? In case these files are plist, I would prefer to keep the plist extension.

gyorb: A more general extension was used because it just generates an output file name which can be used to store the result from the analyzers temporary.
Currently only the clang static analyzer provides plist output, clang tidy does not.
Where we use output files the clang tidy output is converted to plist too before writing it out.
I do not know if there are other places where non plist formatted information is stored in these files.

Xazax-hun:
LGTM!

PR #330, opened 2016-07-04
Update mode improvement

gyorb:
fix update mode, use buildaction and source file name to check if results should be cleaned up


alembic.ini
```diff
@@ -29,7 +29,7 @@ script_location = db_migrate
 # are written from script.py.mako
 # output_encoding = utf-8
 
-sqlalchemy.url = postgres://codechecker@localhost:8764/codechecker
+sqlalchemy.url = postgres://codechecker@localhost:5432/codechecker
```
Xazax-hun: Is this the default port? Should this work, when the used port is not the defult?

db_model/orm_model.py
```diff
@@ -204,6 +208,12 @@ class Report(Base):
     end_bugevent = Column(Integer, ForeignKey('bug_path_events.id', deferrable = True, initially = "DEFERRED", ondelete='CASCADE'), index = True)
     suppressed = Column(Boolean)
 
+    # cascade delete might remove rows SQLAlchemy warns about this
```
Xazax-hun: Start comment with upper case and end it with a period.

alembic.ini
gyorb: The default Postgresql 5432 port is the new default port for CodeChecker db port.

storage_server/report_server.py
```diff
@@ -184,30 +208,49 @@ def replaceConfigInfo(self, run_id, config_values):
...
         try:
+
             build_actions = \
                 self.session.query(BuildAction) \
-                            .filter(and_(BuildAction.run_id == run_id,
-                                         BuildAction.build_cmd == build_cmd,
-                                         BuildAction.check_cmd == check_cmd))\
-                            .all()
+                    .filter(and_(BuildAction.run_id == run_id,
+                                BuildAction.build_cmd == build_cmd,
```
dkrupp: It may cause problems if we include build command in this filter. What if, between two runs I change the compilation parameters for a file and modify that file. In that case, the old results will remain for that file. I think, we should delete all reports to a source file if we reanalize that source file.

PR #332, opened 2016-07-05
increase wait time in the tests

gyorb:
For some slower machines more time is needed for PostgreSQL to stop and restart.


tests/test_packages/functional/package_test/__init__.py
```diff
@@ -112,6 +112,8 @@ def setup_package():
     _run_check(shared_test_params, skip_list_file, test_project_build_cmd,
                test_project_1_name, test_project_path)
 
+    time.sleep(10)
```
dkrupp: why 10?

PR #335, opened 2016-07-08
fix the suppressiong handling for the same bug

### LINKED ISSUE #334: suppress the same bug twice
It is possible that the web UI does not refresh and the user tries to suppress the same bug twice.
The viewer server produces an error in that case and the browser needs to be refreshed.

gyorb:
Suppressing the same bug twice should not fail.
For the second suppression nothing should change.
The suppress comment from the first suppression is kept.

Resolves  #334
Added test for double suppression


Xazax-hun:
LGTM!

PR #349, opened 2016-08-15
Buildcmd escaping test and move test files

gyorb:
- Move test files one directory up.
- Add new build command escaping tests and fix build command escaping for the analyzer subprocess call.


codechecker_lib/analyzers/analyzer_base.py
```diff
@@ -118,3 +111,22 @@ def get_analyzer_checkers(self, config_handler, env):
...
         pass
+
+    @staticmethod
+    def run_proc(command, env=None, cwd=None):
+        """
+        Just run the given command and return the returncode
+        and the stdout and stderr outputs of the process.
+        """
+
+        cmd = shlex.split(command, posix=False)
```
Xazax-hun: Is tunring posix mode off the only functional change? What does this do?

Xazax-hun: After some offline discussion: posix=False is required to avoid removing some double quotes from the command. I think this worth a comment. Otherwise LGTM.

Xazax-hun:
I don't know whether this change introduced the following error:
[ERROR] - Analyzing player.cpp failed.
[ERROR] - clang-3.8: error: no such file or directory: '/home/xazax/Random/Xazax'
clang-3.8: error: no such file or directory: 'SVN/Zatacka2/player.cpp'
clang-3.8: error: no input files

I got this error when there is a space in the path to the cpp. Note that, the clang tidy did not fail, only the static analyzer.

gyorb:
clang-tidy handles errors differently. It will not fail during the analysis, but an error report will be generated. Can you see an error report in the results list for clang tidy? Can you test it with a previous release?

Xazax-hun:
I think we should have a test case with space in the path of the file that is being analyzed.

PR #350, opened 2016-08-16
Improve performance test framework

gyorb:
- improve performance test framework
- add new profiler options to CodeChecker
- do some source cleanup

### DIFF
--- a/codechecker_lib/profiler.py
+++ b/codechecker_lib/profiler.py
@@ -1,0 +1,72 @@
+# -------------------------------------------------------------------------
+#                     The CodeChecker Infrastructure
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+
+from datetime import datetime
+import cProfile
+import pstats
+import StringIO
+
+from codechecker_lib import logger
+
+LOG = logger.get_new_logger('PROFILER')
+
+
+def timeit(function):
+    '''
+    Decorator to measure function call time.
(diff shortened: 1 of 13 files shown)


bruntib:
The `number_of_runs` option in the config file is not taken into account. The performance test generates as many runs as the number of cores used (given by `-j` flag). Moreover it seems that around 10 runs set, only runs/cores are created, but none of them contain any reports.

Xazax-hun:
There is a conflict.

codechecker_lib/profiler.py
```diff
@@ -0,0 +1,72 @@
...
+import pstats
+import StringIO
+
+from codechecker_lib import logger
+
+LOG = logger.get_new_logger('PROFILER')
+
+
+def timeit(function):
+    '''
```
Xazax-hun: Use """ for docstrings.

tests/performance_test/readme.md
```diff
@@ -0,0 +1,56 @@
...
+`bug_length`: bug path length for each bug  
+`clean_after_fill`: remove all run results after each measurement  
+`wait_for_store`: if multiple processes are used for measurement wait until the database is filled up with dummy results before starting the measurements  
+`check_port`: the check port used in the CodeChecker server command  
+`view_port`: the view port used in the CodeChecker server command  
+
+A test configuration cloud look like this:  
+```
+{
+"number_of_runs" : 1,
```
Xazax-hun: What about adding some indentation?

PR #353, opened 2016-08-22
GUI refactoring

bruntib:


bruntib:
This refactored GUI lacks the browser compatibility check, and only one filter condition can be provided. Moreover there is a known bug which I plan to correct as soon as possible, but one can try if it works anyway.

bruntib:
Before this pull request first #367 should be accepted!

bruntib:
Now the width of the field is the currently selected option. This is the easy solution of the problem. This causes of course that in case of several fields their widths may differ.

PR #367, opened 2016-09-02
BuildAction thrift API

bruntib:
The getBuildActions now returns all information about the build actions,
not only the build command.


gyorb:
API version should be increased, this is a major change in case a client already used the API call, otherwise looks good to me.

bruntib:
I don't exactly know the versioning policy. Should it be 4.4?

gyorb:
I think it should be 5.0 because it is a major change which could break other clients. All of the current clients (command line and web) should be updated to check for the new major API version for compatibility.

PR #378, opened 2016-09-15
Print out the used analyzer and version

### LINKED ISSUE #372: Print out which analyzer is used
Print out at the beginning of the analysis which clang and clang-tidy (and what version) is
used for the analysis.

gyorb:
Before starting the analysis it is useful to see which analyzer and
what version of it is used.
Resolves #372


codechecker_lib/analyzer.py
```diff
@@ -36,6 +39,24 @@ def prepare_actions(actions, enabled_analyzers):
...
+    """
+    Print the path and the version of the analyzer binary.
+    """
+    check_env = analyzer_env.get_check_env(context.path_env_extra,
+                                           context.ld_lib_path_extra)
+
+    # get the analyzer binaries from the config_map which
+    # contains only the checked and available analyzers
+    for analyzer_name, analyzer_cfg in analyzer_config_map.iteritems():
+        LOG.info("Using analyzer: ")
```
gyorb: With the upcoming --quiet-build option this will be changed.

codechecker_lib/analyzer.py
```diff
@@ -36,6 +39,24 @@ def prepare_actions(actions, enabled_analyzers):
...
+        analyzer_bin = analyzer_cfg.analyzer_binary
+        print(analyzer_bin)
```
gyorb: The logging module formats the output differently which I think in this case is less readable.

tests/functional/quickcheck_test/test_quickcheck.py
```diff
@@ -49,7 +49,21 @@ def __check_one_file(self, path):
         output = subprocess.check_output(
             ['bash', '-c', command], env=self.env, cwd=self.test_dir)
 
-        self.assertEqual(output, correct_output)
+        # skip the analyzer version info between these two lines
+        # it might be different in the test running environments
+        skip_version_begin = "[INFO] - Using analyzer"
```
gyorb: The --quiet-build option will solve this too.

codechecker_lib/analyzer.py
```diff
@@ -36,6 +39,24 @@ def prepare_actions(actions, enabled_analyzers):
...
+        version = [analyzer_bin, u' --version']
+        proc = subprocess.Popen(shlex.split(' '.join(version)), env=check_env)
+        proc.wait()
```
gyorb: I will improve this.

gyorb:
With the new features in #402 this pull request can be simplified.

codechecker_lib/analyzer.py
gyorb: I made some exception handling improvements here (if calling the analyzer fails for some reason) but I do not think we need the return value here. The analysis will not stop here if getting the version command returns some non 0 value. There are other error handling parts during the analysis if the analyzer can not run for some reason.

Xazax-hun:
There is a conflict that needs to be resolved.

PR #393, opened 2016-10-03
Privileged access subsystem

### LINKED ISSUE #339: LDAP based authentication
Add optional LDAP based authentication to view the results.

whisperity:
This merge requests contains the privileged access infrastructure &mdash; closes #339.

If a user **is** authenticated (via any means described below), full access to the server is given: they can view results and destroy them. If the user **is not** authenticated, the entire access is denied &ndash; there are no further fine-graining, as per discussed with @dkrupp and @gyorb.

# The changes &mdash; summed up

 * A new *Thrift* endpoint at `/Authentication` was implemented, this handles authentication requests
 * Authentication can happen against a username-password storage, an LDAP server or the local PAM chain.
 * Existing interface was **not** changed, though an invalid session, on a **privileged** server, returns a HTTP error code which can make the `CodeChecker cmd` client fail on an exception &mdash; clients are expected to handle authentication before live request.
 * Two new configuration files, `config/session_config.json` and `config/session_client.json`. Both are created as working and descriptive templates.
    * The first one is the server's configuration. It is copied to `WORKSPACE/session_config.json` for every server started and contains the authentication setup.
    * The latter one gets copied for every user into their `~/.codechecker_passwords.json` as it contains per-user credentials.
 * Web-browser client authenticates via basic pop-ups (`HTTP WWW-Authenticate` header)
 * `CodeChecker cmd` got a new subcommand: `login`. With this, users can authenticate on a server and kill their session.
    * The `~/.codechecker_passwords.json` file can contain preconfigured credentials &mdash; if they are valid, executing a live query without a session will automatically log the user in for screen security and convenience.
    * Session information for a user is stored in a temporary file, identified on the system with the running user's name.

Full documentation for public use is available in `docs/authentication.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,3 +176,5 @@ See user guide for further configuration and check options.
 [Test documentation](tests/functional/package_test.md)
 
 [Database schema migration](docs/db_schema_guide.md)
+
+[Privileged server and authentication in client](docs/authentication.md)
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Tested on Ubuntu LTS 14.04.2
 
 # get ubuntu packages
 # clang-3.6 can be replaced by any later versions of clang
-sudo apt-get install clang-3.6 doxygen build-essential thrift-compiler python-virtualenv gcc-multilib git wget libldap2-dev libsasl2-dev libssl-dev
+sudo apt-get install clang-3.6 doxygen build-essential thrift-compiler python-virtualenv gcc-multilib git wget
 
 # create new python virtualenv
 virtualenv -p /usr/bin/python2.7 ~/checker_env
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Tested on Ubuntu LTS 14.04.2
 
 # get ubuntu packages
 # clang-3.6 can be replaced by any later versions of clang
-sudo apt-get install clang-3.6 doxygen build-essential thrift-compiler python-virtualenv gcc-multilib git wget
+sudo apt-get install clang-3.6 doxygen build-essential thrift-compiler python-virtualenv gcc-multilib git wget libldap2-dev libsasl2-dev libssl-dev
 
 # create new python virtualenv
 virtualenv -p /usr/bin/python2.7 ~/checker_env
--- a/build_package.py
+++ b/build_package.py
@@ -111,6 +111,15 @@ def generate_thrift_files(thrift_files_dir, env, silent=True):
         LOG.error('Failed to generate viewer server files')
         return ret
 
+    auth_thrift = os.path.join(thrift_files_dir, 'authentication.thrift')
+    auth_thrift = 'authentication.thrift'
+    auth_cmd = ['thrift', '-r', '-I', '.',
+               '--gen', 'py', auth_thrift]
+    ret = run_cmd(auth_cmd, thrift_files_dir, env, silent=silent)
+    if ret:
+        LOG.error('Failed to generate authentication interface files')
+        return ret
+
 
 # -------------------------------------------------------------------
 def generate_documentation(doc_root, env, silent=True):
--- a/build_package.py
+++ b/build_package.py
@@ -114,7 +114,7 @@ def generate_thrift_files(thrift_files_dir, env, silent=True):
     auth_thrift = os.path.join(thrift_files_dir, 'authentication.thrift')
     auth_thrift = 'authentication.thrift'
     auth_cmd = ['thrift', '-r', '-I', '.',
-               '--gen', 'py', auth_thrift]
+                '--gen', 'py', auth_thrift]
     ret = run_cmd(auth_cmd, thrift_files_dir, env, silent=silent)
     if ret:
         LOG.error('Failed to generate authentication interface files')
--- a/codechecker_lib/session_manager.py
+++ b/codechecker_lib/session_manager.py
@@ -31,6 +31,13 @@ try:
 except ImportError:
     unsupported_methods.append("ldap")
 
+try:
+    import pam
+    import grp
+    import pwd
+except ImportError:
+    unsupported_methods.append("pam")
+
 LOG = logger.get_new_logger("SESSION MANAGER")
 SESSION_COOKIE_NAME = "__ccPrivilegedAccessToken"
 session_lifetimes = {}
@@ -136,7 +143,6 @@ class SessionManager:
 
             if "method_ldap" in self.__auth_config and \
                     self.__auth_config["method_ldap"].get("enabled"):
-
                 if "ldap" not in unsupported_methods:
                     found_auth_method = True
                 else:
@@ -144,6 +150,16 @@ class SessionManager:
                                 "prerequisites are NOT installed on the system"
                                 "... Disabling LDAP authentication")
                     self.__auth_config["method_ldap"]["enabled"] = False
+
+            if "method_pam" in self.__auth_config and \
+                    self.__auth_config["method_pam"].get("enabled"):
+                if "pam" not in unsupported_methods:
+                    found_auth_method = True
+                else:
+                    LOG.warning("PAM authentication was enabled but "
+                                "prerequisites are NOT installed on the system"
+                                "... Disabling PAM authentication")
+                    self.__auth_config["method_pam"]["enabled"] = False
 
             #
             if not found_auth_method:
@@ -170,6 +186,7 @@ class SessionManager:
         """Validate an oncoming authorization request
         against some authority controller."""
         return self.__try_auth_dictionary(auth_string) \
+            or self.__try_auth_pam(auth_string) \
             or self.__try_auth_ldap(auth_string)
 
     def __is_method_enabled(self, method):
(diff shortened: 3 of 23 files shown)


Xazax-hun:
Did you consider storing client side information in the designated codechecker workspace? Does it make sense to have a global one?

Xazax-hun:
Is there a way to set timeout for a session?

gyorb:
If a user wants only to check the results and use only the command line no workspace is required. It is possible to store multiple server authentication credentials in one file.
By default a global one is created but the user should be able to use authentication files from different places if needed.

codechecker_lib/session_manager.py
```diff
@@ -0,0 +1,299 @@
...
+            ret = self.__save["credentials"].get(host)
+            if not ret:
+                ret = self.__save["credentials"].get("*:" + port)
+                if not ret:
+                    ret = self.__save["credentials"].get("*")
+
+        return ret
+
+    def saveToken(self, host, port, token, destroy=False):
+        if not destroy:
```
Xazax-hun: Reverse the branches and remove the not.

thrift_api/authentication.thrift
```diff
@@ -0,0 +1,31 @@
...
+service codeCheckerAuthentication {
+  // get basic authentication information from the server
+  HandshakeInformation getAuthParameters(),
+
+  // retrieves a list of accepted authentication methods from the server
+  list<string> getAcceptedAuthMethods(),
+
+  // handles creating a session token for the user
+  string performLogin(1: string auth_method,
+                      2: string auth_string)
```
Xazax-hun: Is it good practice to have one auth string instead of separate user/password etc?

viewer_clients/cmdline_client/authentication_helper.py
```diff
@@ -0,0 +1,100 @@
...
+
+        if session_token:
+            headers = {'Cookie': session_manager.SESSION_COOKIE_NAME +
+                       "=" + session_token}
+            self.transport.setCustomHeaders(headers)
+
+            # ------------------------------------------------------------
+
+    def ThriftClientCall(function):
+        # print type(function)
```
Xazax-hun: Commented code.

tests/test_utils/thrift_client_to_db.py
```diff
@@ -97,15 +97,21 @@ def __init__(self, host, port, auto_handle_connection=True):
...
-    def __init__(self, host, port, uri, auto_handle_connection=True):
+    def __init__(self, host, port, uri, auto_handle_connection=True,
+                 session_token=None):
         # import only if necessary; some tests may not add this to PYTHONPATH
+        from codechecker_lib import session_manager
         from codeCheckerDBAccess import codeCheckerDBAccess
         from codeCheckerDBAccess.constants import MAX_QUERY_SIZE
 
         self.max_query_size = MAX_QUERY_SIZE
         transport = THttpClient.THttpClient(host, port, uri)
```
Xazax-hun: Is the login done through https?

whisperity:
# To-Do:

 * Session tokens in separate files, proper locking
 * Server configuration into workspace
 * Warn that credentials travel without any encryption

thrift_api/authentication.thrift
whisperity: It was done this way to make sure that further extensions in auth methods (such as, e.g. SSH key based one) is most easily done without further breaking the Authentication API.

tests/test_utils/thrift_client_to_db.py
whisperity: **No.** (Docs will be updated.)

README.md
```diff
@@ -40,7 +40,7 @@ Tested on Ubuntu LTS 14.04.2
 
 # get ubuntu packages
 # clang-3.6 can be replaced by any later versions of clang
-sudo apt-get install clang-3.6 doxygen build-essential thrift-compiler python-virtualenv gcc-multilib git wget
+sudo apt-get install clang-3.6 doxygen build-essential thrift-compiler python-virtualenv gcc-multilib git wget libldap2-dev libsasl2-dev libssl-dev
```
whisperity: Agreed. CQ @gyorb @dkrupp on what the best way to support this through the automatic build chain (maybe a `--with-ldap` option in `build_package.py`?) would be.

gyorb: I think installing these extra requirements could go into the separate authentication documentation and start that if the user requires authentication he should install these extra packages and python modules.

build_package.py
```diff
@@ -107,6 +107,15 @@ def generate_thrift_files(thrift_files_dir, env, silent=True):
...
         return ret
 
+    auth_thrift = os.path.join(thrift_files_dir, 'authentication.thrift')
+    auth_thrift = 'authentication.thrift'
+    auth_cmd = ['thrift', '-r', '-I', '.',
+                '--gen', 'py', auth_thrift]
+    ret = run_cmd(auth_cmd, thrift_files_dir, env, silent=silent)
+    if ret:
+        LOG.error('Failed to generate authentication interface files')
+        return ret
```
gyorb: So far we generated the source files with Thrift files differently for the report server (python only) and the report viewer (python and javascript). The authentication is python only which could be refactored into a separate function with the report server source file generation. I'm not sure it will be so much shorter,  right now the generate_thrift_files function is not that long. If you think it is better you can refactor it into a separate function.

codechecker_lib/session_manager.py
```diff
@@ -0,0 +1,299 @@
...
+    def __init__(self):
+        LOG.debug('Loading session config')
+
+        session_cfg_file = os.path.join(os.environ['CC_PACKAGE_ROOT'],
+                                        "config", "session_config.json")
+        LOG.debug(session_cfg_file)
+        with open(session_cfg_file, 'r') as scfg:
+            scfg_dict = json.loads(scfg.read())
+
+        if not scfg_dict["authentication"]:
```
gyorb: Actually the next line ```scfg_dict["authentication"] = {'enabled': False}``` could be moved up before loading the config json to create a dict where by default the authentication is disabled. In that case we do not have to check it later. A simple dict update```scfg_dict.update(scfg_from_json)``` could be used to overwrite the default values with the values from the config file .

codechecker_lib/session_manager.py
```diff
@@ -0,0 +1,299 @@
...
+            return True
+        else:
+            return any(_sess.client == client
+                       and _sess.token == token
+                       and _sess.still_valid(access)
+                       for _sess in sessionManager.__valid_sessions)
+
+    def invalidate(self, client, token):
+        '''Remove a user's previous session from the store'''
+        for session in sessionManager.__valid_sessions[:]:
```
gyorb: Similar question why do we have to copy the whole __valid_sessions list if one of them is invalidated ore the sessions are cleaned up? Do we lose some session tokens if a cleanup is in progress and a new is added or checked for validity?

tests/functional/package_test/__init__.py
```diff
@@ -146,7 +152,30 @@ def setup_package():
 
     # start the CodeChecker server
     print("Starting server to get results")
-    _start_server(shared_test_params, test_config)
+    _start_server(shared_test_params, test_config, False)
+
+    #
+    # Create a dummy authentication-enabled configuration and an auth-enabled server
+    session_cfg_file = os.path.join(pkg_root, "config", "session_config.json")
+    with open(session_cfg_file, 'r') as scfg:
```
gyorb: Could we open the file only once? (read up the config modify it and write out)

viewer_server/client_db_access_server.py
```diff
@@ -20,39 +20,127 @@
...
 
         BaseHTTPRequestHandler.__init__(self,
                                         request,
                                         client_address,
                                         server)
 
-    def log_message(self, msg_format, *args):
-        ''' silenting http server '''
-        return
+    #def log_message(self, msg_format, *args):
```
gyorb: This was not commented out to silence the webserver. Do we need to see all the requests?

codechecker_lib/session_manager.py
```diff
@@ -0,0 +1,299 @@
...
+with a particular CodeChecker server.
+'''
+
+import os
+import shutil
+import uuid
+import json
+import hashlib
+import time
+import ldap
```
gyorb: importing ldap modules could be done optionally and if importing fails only the ldap authentication method will be disabled.
```
try:
   import ldap
   import ldap.sasl
except ImportError:
    # no ldap authentication is available
```

viewer_server/client_db_access_server.py
whisperity: This was a debug change that was accidentally added in development. It has been reverted ever since.

> GitHub does not hide the comments for hunks that had been modified since...

docs/authentication.md
```diff
@@ -0,0 +1,151 @@
...
+
+Authentication in the web browser is handled via standard *HTTP Authenticate* headers, the browser will prompt the user to supply their crendentials.
+
+For browser authentication to work, cookies must be enabled!
+
+### Command-line client
+
+The `CodeChecker cmd` client needs to be authenticated for a server before any data communication could take place.
+
+The client's configuration file is expected to be at `~/.codechecker_passwords.json`, which is created at the first command executed
```
gyorb: Please extend the documentation that the user should modify the access rights to the password file so only he can access it: ```chmod 0600 ~/.codechecker_password.json```

whisperity: :smirk: Better idea is to create the file initially as `0600`. Though I added it to the documentation and a warning (in `ssh` style) for the user if the file is with the wrong permissions.

gyorb:
Please update your branch, so all the tests can run on Linux and on OSX too. The CI is fixed now. Thanks!

whisperity:
Should I cherry-pick the test fixing changes or should I do a full rebase? With this many commits in the history &ndash; and the reviews &ndash; I would vote for cherry picking.

Xazax-hun:
I think rebasing is usually a cleaner solution for the history's sake.

codechecker_lib/session_manager.py
```diff
@@ -288,11 +313,12 @@ def getAuthString(self, host, port):
...
         else:
-            self.__save["tokens"][host + ":" + port] = token
+            self.__tokens[host + ":" + port] = token
 
-        session_cfg_file = os.path.join(os.path.expanduser("~"),
-                                        ".codechecker_passwords.json")
-        with open(session_cfg_file, 'w') as scfg:
-            json.dump(self.__save, scfg, indent=2, sort_keys=True)
+        with open(self.token_file, 'w') as scfg:
+            fcntl.lockf(scfg, fcntl.LOCK_EX)
```
gyorb: Is this way of locking unix specific? Will it work on OSX and maybe later on Windows?

whisperity: It is. It will work on OS X but will definitely *not* work on Windows, the filesystem concept of the two system families are entirely different.

Maybe we could use the package [portalocker](https://pypi.python.org/pypi/portalocker)? It is essentially just a branching wrapper, but it looks like it could work good enough.

whisperity:
As for squashing: not really without messing up order (I have done some reordering in this last rebase.)

I could squash the first two commits as they are one unit, I could also reorder the "update due to reviews" stuff and put it with the "misc." changes as one commit at the end, though. But putting any more commits together would create a big list of unrelated stuff in one commit...

gyorb:
Let's squash that few commits you mentioned and keep the rest.

PR #394, opened 2016-10-03
Bugfix: Bug view didn't open in diff view

bruntib:


sylvestre:
Fails with:
```TypeError: this.runData is undefined```

bruntib:
In which line do you get this error? What scenario leads to this? I can't reproduce this error.

sylvestre:
I confirm that it is fixing the issue for me.
Thanks

PR #397, opened 2016-10-04
Order runs by date descending.

### LINKED ISSUE #366: Default order of analyzer runs
The default order of the showed results should be ordered by descending date.

bruntib:
Fixes #366.


Xazax-hun:
I do not really like that 1 second delay stuff. Maybe it would be better to populate the grid with the good order to start with. So I would prefer a server side solution.

bruntib:
I totally agree. I replaced this ugly solution by a sorting, but on client side, not on server side.

PR #403, opened 2016-10-10
Update test dependency documentation

gyorb:


tests/functional/package_test.md
```diff
@@ -1,7 +1,11 @@
 # How to run the tests
 
+Additionally the [nose](https://nose.readthedocs.io/en/latest/) test framework is required to run the tests.
+~~~~~~{.sh}
+pip install nose
```
whisperity: Despite it being kind of trivial, I think we should still point out (just as how we did in the full README and how I did in the Authentication README) that this command should only be ran after source-ing into the virtual environment.

PR #407, opened 2016-10-16
Handle alembic command error.

### LINKED ISSUE #384: handle alembic exception
Alembic might throw an CommandError exception if an older CodeChecker tries to connect to a database with a newer db schema in it.
Handle alembic errors in the [database_handler.py] (https://github.com/Ericsson/codechecker/blob/03fd2a04fe8bfad4c56d586130efcbcc6955a440/codechecker_lib/database_handler.py#L64) while trying to create or migrate the database schema.

Xazax-hun:
Handle alembic command error. Fixes #384.

### DIFF
--- a/codechecker_lib/database_handler.py
+++ b/codechecker_lib/database_handler.py
@@ -15,6 +15,7 @@ import time
 from abc import ABCMeta, abstractmethod
 
 import sqlalchemy
+from alembic.util import CommandError
 from alembic import command, config
 from sqlalchemy import event
 from sqlalchemy.engine import Engine
@@ -95,6 +96,11 @@ class SQLServer(object):
 
         except sqlalchemy.exc.SQLAlchemyError as alch_err:
             LOG.error(str(alch_err))
+            sys.exit(1)
+        except CommandError as cerr:
+            LOG.error("Database schema and CodeChecker is incompatible."
+                      "Please update CodeChecker.")
+            LOG.debug(str(cerr))
             sys.exit(1)
 
     @abstractmethod


codechecker_lib/database_handler.py
```diff
@@ -96,6 +97,9 @@ def _create_or_update_schema(self, use_migration=True):
         except sqlalchemy.exc.SQLAlchemyError as alch_err:
             LOG.error(str(alch_err))
             sys.exit(1)
+        except CommandError as cerr:
+            LOG.error(str(cerr))
```
gyorb: Based on the code review feedback we should provide more user friendly error messages not just printing out the message from the exception. These error messages will be read by our users who might not be familiar with SQLAlchemy or Alembic errors.
I try to improve it in any new improvement and update our existing source too.

Xazax-hun: Done.

codechecker_lib/database_handler.py
```diff
@@ -96,6 +97,10 @@ def _create_or_update_schema(self, use_migration=True):
         except sqlalchemy.exc.SQLAlchemyError as alch_err:
             LOG.error(str(alch_err))
             sys.exit(1)
+        except CommandError as cerr:
+            LOG.error("Failed to update or create the database schema.")
```
gyorb: This message can be a bit misleading, because in the issue description the problem is if there is a newer schema in the database and we want to connect to it with an older CodeChecker.
In that case no schema update or creation is done. Might be useful to to print out here that there is a database incompatibility between the used CodeChecker version and the database, and updating your CodeChecker release to a newer should resolve it.

PR #408, opened 2016-10-16
Analysis progress reporting.

### LINKED ISSUE #370: Add some progress report to analyzer info messages
Big projects can take hours to finish, and it would be nice to know at least approximately how much stuff there is left to analyse.

As a first step I was thinking about a simple (ninja style) indicator on how many steps are left from the analysis process.

Something like:

```
[...]
[INFO] - [34/705] clangsa analyzed A.cpp successfully.
[INFO] - [35/705] clangsa analyzed B.cpp successfully.
[INFO] - [36/705] clangsa analyzed C.cpp successfully.
[INFO] - [37/705] clangsa analyzed D.cpp successfully.
[INFO] - [38/705] clangsa analyzed E.cpp successfully.
[INFO] - [39/705] clangsa analyzed F.cpp successfully.
[...]
```

wdyt?

Xazax-hun:
Fixes  #370.

Some notes to the changes:
- Shared memory values can not be passed as function arguments, that is the reason for the globals.
    - They could be moved into an empty imported module
- Result handler no longer prints successful analysis message. It was asymmetric. Now successful and unsuccessful results are printed from the same location.
- Result handler only invoked to process results, when the analysis succeeded. Updated the code to reflect this.
- Invoke the process pool implementation from the quickcheck command. This way multiprocess checking with quickcheck can be implemented in the future. (We might need to lock the standard output for concurrent writes.)

### DIFF
--- a/codechecker_lib/analysis_manager.py
+++ b/codechecker_lib/analysis_manager.py
@@ -58,6 +58,17 @@ def worker_result_handler(results):
     LOG.info("----=================----")
 
 
+# Progress reporting.
+progress_checked_num = None
+progress_actions = None
+
+
+def init_worker(x, y):
+    global progress_checked_num, progress_actions
+    progress_checked_num = x
+    progress_actions = y
+
+
 def check(check_data):
     """
     Invoke clang with an action which called by processes.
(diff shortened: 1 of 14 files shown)


codechecker_lib/analysis_manager.py
```diff
@@ -155,7 +174,7 @@ def signal_handler(*arg, **kwarg):
     signal.signal(signal.SIGINT, signal_handler)
 
     # Remove characters which could cause directory creation problems.
-    no_spec_char_name = re.sub(r'[^\w\-_\. ]', '_', args.name)
+    no_spec_char_name = re.sub(r'[^\w\-_. ]', '_', args.name)
```
gyorb: Why is this change needed? Just a temporary directory is created. The ```.``` char is replaced with ```_```.

Xazax-hun: It is just a cleanup, in that context no escaping of `.` required.

Xazax-hun: This is not a functional change just a cleanup. Dot was *not* replaced befor this modification and it is replaced afterwards.

gyorb: Ok, my bad I misinterpreted the regex.

PR #411, opened 2016-10-18
Improve option parser

gyorb:
Fixed issues:
* ```sysroot``` and ```isysroot``` compiler options were skipped and not forwarded to the analyzer.
* compilation language was not set properly (if detectable from the compilation command)
* compilation architecture was not set properly (if detectable from the compilation command)
* compilation action type was not set properly (compile, preprocess, link ...)

Some code refactoring and new documentation was created.
New unit tests are available for the option parser module.


codechecker_lib/option_parser.py
```diff
@@ -3,70 +3,49 @@
...
 #   License. See LICENSE.TXT for details.
 # -------------------------------------------------------------------------
+"""
+An extended lookup table for compiler arguments based on ScanBuild
+clang-analyzer.llvm.org/scan-build.html
 
+Used to filter out or change compiler argument not supported by clang
+or clang-tidy.
+
+Keys are the option name, value is the number of options to skip
```
Xazax-hun: End comments with a comma

codechecker_lib/option_parser.py
```diff
@@ -385,40 +412,65 @@ def skip_n(size):
...
-    """Requires a full compile command with the compiler, not only arguments."""
+    """ Requires a full compile command with the compiler,
+    not only arguments."""
 
     # Keep " characters.
     args = args.replace('"', '\\"')
 
     result_map = OptionParserResult()
+
+    # the first is the compiler skip it from parsing
```
Xazax-hun: Make the comments full sentences.

codechecker_lib/option_parser.py
```diff
@@ -171,22 +168,51 @@
 }
 
 
-# -----------------------------------------------------------------------------
 class ActionType(object):
-    LINK, COMPILE, PREPROCESS, INFO = range(4)
+    LINK = 0
+    COMPILE = 1
+    PREPROCESS = 3
+    INFO = 3
```
bruntib: Should be 4.

gyorb: Actually the preprocess should be 2, but thanks!

tests/unit/test_option_parser.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        Test the build command of a simple file.
+        """
+        source_files = ["main.cpp"]
+        build_cmd = "g++ -o main " + ' '.join(source_files)
+
+        res = option_parser.parse_options(build_cmd)
+        print(res)
+        self.assertTrue(len(source_files), len(res.files))
+        self.assertTrue(set(source_files) & set(res.files))
```
bruntib: These two assertions are true if the two sets have the same same size and they have at least one common element. Is there any reason why we don't check equality between the two sets?

tests/unit/test_option_parser.py
```diff
@@ -0,0 +1,187 @@
...
+
+""" Test the option parsing and filtering form the compilation commands. """
+
+import unittest
+
+from codechecker_lib import option_parser
+from codechecker_lib.option_parser import ActionType
+
+
+class OptionParserTest(unittest.TestCase):
```
bruntib: Would it make sense to always check every attributes of the returned `OptionParserResult` object? Or is this intentional that only a few of them is observed?

codechecker_lib/option_parser.py
```diff
@@ -385,40 +412,65 @@ def skip_n(size):
...
+        lang = get_language(os.path.splitext(source_file)[1])
+        if lang:
+            is_source = True
+            # If lang is not set already during the argument parsing
+            # set it based on the source file extension.
+            if result_map.lang is None:
+                result_map.lang = lang
+            break
+
+    # It there are no source files in the compilation argument
```
bruntib: Small typo: "It" -> "If".

tests/unit/test_option_parser.py
gyorb: Right now the tests focus on the options which are critical to the analysis and needed to be parsed correctly. Further tests can be added later.
gcc/g++ and clang have many options we do not target all of them now. These options are continuously extended if needed.

PR #412, opened 2016-10-18
Extend QuickCheck with suppress files, skiplists and parallelism.

### LINKED ISSUE #383: extend quickcheck
Quickcheck should support:
- [ ] skiplist
- [ ] suppress file

Xazax-hun:
Fixes #383.

Depends on: https://github.com/Ericsson/codechecker/pull/408


gyorb:
The parallelism will not mess up the output?
I think it was single threaded so far because the log output was mixed on multiple threads.
Could you add some simple test for the multithreaded version too?

Xazax-hun:
Locking added. Rebased to master as the dependency already merged.
To add a test with parallelism, there should be a project with multiple files, and there is no guarantee to the order in which the files are analyzed. I think I could add a separate test, (that is not using the current quickcheck test infrastructure completely, or extend it) that only compares input as a set of lines. Do you think that would be enough? Or is it ok to not have a test for this functionality?

codechecker_lib/analysis_manager.py
```diff
@@ -61,12 +61,14 @@ def worker_result_handler(results):
 # Progress reporting.
 progress_checked_num = None
 progress_actions = None
+progress_lock = None
 
 
-def init_worker(x, y):
-    global progress_checked_num, progress_actions
+def init_worker(x, y, l):
```
gyorb: I'm not a big fan of the one character variable names.

PR #419, opened 2016-10-20
PEP8 compliance second part

gyorb:
Refactoring the code is done during the cleanup,
mainly the suppress handling parts


codechecker_lib/option_parser.py
```diff
@@ -406,14 +408,16 @@ def skip_n(size):
...
+    Requires a full compile command with the compiler, not only arguments.
+    """
 
     # Keep " characters.
     args = args.replace('"', '\\"')
 
     result_map = OptionParserResult()
     for it in OptionIterator(shlex.split(args)[1:]):
-        arg_check(it, result_map) # TODO: do sth at False result, actually skip.
+        arg_check(it, result_map)  # TODO: do sth at False result, skip it
```
Xazax-hun: Trailing period removed.

codechecker_lib/client.py
```diff
@@ -41,10 +41,8 @@ def send_suppress(run_id, connection, file_name):
     Collect suppress information from the suppress file to be stored
     in the database.
     """
-    suppress_data = []
-    if os.path.exists(file_name):
-        with codecs.open(file_name, 'r', 'UTF-8') as s_file:
-            suppress_data = suppress_file_handler.get_suppress_data(s_file)
+    suppress_data = suppress_file_handler.SuppressFileHandler. \
```
bruntib: Is it on purpose not to use this UTF-8 thing? If it is, then `import codecs` is also unnecessary.

gyorb: It should be opened like this because we store user comment here which might contain non-ascii characters.

gyorb:
New pull request was created for pep8 cleanup.

PR #437, opened 2016-11-04
Display server start hint at the end of checking.

### LINKED ISSUE #413: plist importer command line
- extend help message that the results will be stored into a database
- print out how to start codechecker sever after the parsing of the plist files is done

bruntib:
At the end of checking and parsing plist files a hint is shown which
helps the user how to start the CodeChecker server.
Fixes #413.

### DIFF
--- a/codechecker_lib/arg_handler.py
+++ b/codechecker_lib/arg_handler.py
@@ -33,6 +33,18 @@ from viewer_server import client_db_access_server
 LOG = LoggerFactory.get_new_logger('ARG_HANDLER')
 
 
+def log_startserver_hint(args):
+    db_data = ""
+    if args.postgresql:
+        db_data += " --postgresql" \
+                   + " --dbname " + args.dbname \
+                   + " --dbport " + str(args.dbport) \
+                   + " --dbusername " + args.dbusername
+
+    LOG.info("To view results run:\nCodeChecker server -w " +
+             args.workspace + db_data)
+
+
 def handle_list_checkers(args):
     """
     List the supported checkers by the analyzers.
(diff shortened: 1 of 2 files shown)


codechecker_lib/arg_handler.py
```diff
@@ -231,19 +241,12 @@ def handle_check(args):
...
-            workspace = args.workspace
-        except AttributeError:
-            # If no workspace value was set for some reason
-            # in args set the default value.
-            workspace = util.get_default_workspace()
-
-        workspace = os.path.realpath(workspace)
-        if not os.path.isdir(workspace):
-            os.mkdir(workspace)
+        args.workspace = os.path.realpath(args.workspace)
```
gyorb: How about using absolute path?

codechecker_lib/arg_handler.py
```diff
@@ -433,14 +428,17 @@ def handle_plist(args):
 
     try:
         items = [(plist, args, context) for plist in os.listdir(args.directory)]
-        pool.map_async(consume_plist, items, 1).get(float('inf'))
+        pool.map_async(consume_plist, items, 1)
```
gyorb: I do not think this should be removed (timeout is set to "infinite") see out other [usage of map_async](https://github.com/Ericsson/codechecker/blob/eeac2d6a1b1188dce6ef65bcc4a82faa545a5f7e/codechecker_lib/analysis_manager.py#L198-L209).

Xazax-hun: Maybe not giving any argument would be even better? 
https://docs.python.org/2/library/multiprocessing.html#multiprocessing.pool.AsyncResult.get

bruntib: I can't really see the aim of that `get()` function. This gives me an error:
```py
import multiprocessing
pool = multiprocessing.Pool(2)
pool.map_async(lambda x: 2 * x, [1, 2, 3, 4, 5], 1).get(float('inf'))
```

Xazax-hun: You get an error because lambda funtions can not be pickled (serialized). If you replace the lambda with an out of line function you will not get an error. The get is necessary to actually run the tasks and wait for them being finished.

Xazax-hun:
This functionaly has no tests at all. Could you add one in this merge request?

Xazax-hun:
I think using existing plist files would be a great idea. Even if we do not check whether they were imported properly we could at least make sure that there are no exceptions on that code path. That would be a great first step.

bruntib:
I agree we need tests for plist parsing but this pull request is not about plist parsing. It's about simply displaying a hint to the user how to start the server when parsing is done.

Issue #361, opened 2016-08-29
Please provide a checker view


sylvestre:
For big projects like Firefox (more than 180 000 warnings), it would be nice to have the list of all defects found by a checker (either in the webview or the cli).

gyorb:
Do you want to get some summary view about the checkers?
Right now you have two options:
1. you can filter the results by checker name on the GUI
2. use the command line ```CodeChecker cmd sum --help``` to get a summary

Do you have another type of view in your mind?

sylvestre:
Are you sure? This looks incorrect:
```
$ CodeChecker cmd sum -p 8001 -n clang-analyzer-deadcode.DeadStores 
No check name found: clang-analyzer-deadcode.DeadStores
Possible check names are:
firefox_201609010755
firefox_201609040755
firefox_201608232251
firefox_201608310756
firefox_201609030755
firefox_201608241541
firefox_201608240045
firefox_201608271710
firefox_201608241057
firefox_201608300756
firefox_201608280753
firefox_201608250753
firefox_201609050753
firefox_201608290753
firefox_201609020755
firefox_201608242327
```

sylvestre:
> 1. you can filter the results by checker name on the GUI
In a specific run? What is I want to get all of them for all runs?

gyorb:
Please try this command ```CodeChecker cmd sum -p 8001 -a``` or ```CodeChecker cmd sum -p 8001 -a -o json``` for json format. This will give a summary list for all the runs in the database this is not a merged summary.
Maybe we could add some ```CodeChecker sum global ...``` command to get a summary about all the checker results found in multiple runs.
Until there is no support for this I think it can be easily done by a simple script to go through the json formatted output and sum all the checkers.

gyorb:
The new feature in #361 is related to this but will not fully resolve the new feature request. The new UI improvement will show a summary for each run based on the severity levels.
I still think a command-line summary command would be nice where the summary based on checker names or severity levels could be collected.

dkrupp:
CodeChecker v6.0 implements this view
Checkers Summary Table: Provides a summary statistics of reports found by checkers summarized for all runs. (#826)

PR #438, opened 2016-11-07
Show number of bugs by severity

bruntib:
This commit may be related to #361.

### DIFF
--- a/viewer_clients/web-client/scripts/codecheckerviewer/ListOfRuns.js
+++ b/viewer_clients/web-client/scripts/codecheckerviewer/ListOfRuns.js
@@ -6,16 +6,20 @@
 
 define([
   'dojo/_base/declare',
+  'dojo/dom',
   'dojo/dom-construct',
+  'dojo/dom-style',
   'dojo/data/ItemFileWriteStore',
   'dojo/topic',
+  'dojo/when',
   'dijit/Dialog',
   'dijit/form/Button',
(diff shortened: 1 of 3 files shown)


Xazax-hun:
Isn't it a performance bottleneck to do this query on the client side instead of making the server to send this data to the client? In case this is cheap, LGTM!

bruntib:
The computation goes on server side by calling `getRunResultCount()` API function, in addition it is called asynchronously. I'm rather worried about server side, because this tends to be a slow function. I haven't tested it on a big database, but for the same reason is the "checker id" filter field asynchronous. I think it would even be worth to store these sums in the database or at least in a server-side cache as these are queried all the time when the GUI page is loaded.

gyorb:
The server side cache for the summaries is a nice idea.

Xazax-hun:
Is the cache ever cleared? Is this a good idea? 
Maybe a new API to query this info efficiently would be better.

Xazax-hun:
I think we should not introduce caching but a new API call for now for the following results:
* What is the cache invalidation logic? And currently the cache can only grow, it is a leak. It can be a problem on long running CI servers eventually.
* Should we protect the cache with a lock? Interprocess or interthread lock? 
* Should we introduce separate caching mechanisms for all of the features?

Using a new API call we do not need to solve those problems and can be still performant.

Xazax-hun:
This will be subsumed by the new filtering GUI.

PR #442, opened 2016-11-10
Use POSIX mode to lex shell command line arguments.

### LINKED ISSUE #439: Set back the shlex splitting method to posix
Modifying the build arguments in the test files is needed too.

Xazax-hun:
Fixes #439.


gyorb:
Shouldn't be the [printout](https://github.com/Ericsson/codechecker/blob/master/tests/unit/test_buildcmd_escaping.py#L53) changed in the generated test cpp file?

PR #443, opened 2016-11-10
Command line cleanup

Xazax-hun:
Cleanups to the invocation of clang-tidy and the static analyzer:
* There can be only one resource dir.
* Do not add resource dir as include path.
* Made the code more compact.


Xazax-hun:
This change is mainly for better signal to noise ratio, so the code is easier to read, the arguments are grouped nicer. I would be surprised if this part of codechecker would be a performance bottleneck. But I did a small benchmark, the difference between extending with a 2 argument list or calling append twice is about 10%.

PR #448, opened 2016-11-15
Various cleanups to the codebase.

Xazax-hun:
See the commit messages for details.


codechecker_lib/analyzers/analyzer_types.py
```diff
@@ -299,14 +289,11 @@ def build_config_handlers(args, context, enabled_analyzers, connection=None):
...
             config_handler = __build_clangsa_config_handler(args, context)
-            analyzer_config_map[ea] = config_handler
-
         elif ea == CLANG_TIDY:
             config_handler = __build_clang_tidy_config_handler(args, context)
-            analyzer_config_map[ea] = config_handler
         else:
-            LOG.debug_analyzer('Not supported analyzer type. '
-                               'No configuration handler will be created.')
+            assert False, 'Analyzer types should have been checked already.'
```
gyorb: Should be the default value of the config_handler None?
Do we need to fail with an assert if it is misconfigured? The basic idea was that if a non supported analyzer is configured just skip it and run the supported and available analyzers.

codechecker_lib/analyzers/analyzer_types.py
```diff
@@ -330,52 +317,37 @@ def construct_result_handler(args,
     """
     Construct a result handler.
     """
+    assert buildaction.analyzer_type in supported_analyzers, \
```
gyorb: Should we fail here with an assert?

codechecker_lib/analyzers/config_handler_clang_tidy.py
```diff
@@ -91,16 +80,18 @@ def get_checker_configs(self):
...
+            # Match for clang tidy analyzer names and attributes.
+            clang_tidy_checker_rex = r'^(?P<checker_name>([^.]+))' \
+                                     r'\.(?P<checker_attribute>([^.]+))$'
+
+            tidy_pattern = re.compile(clang_tidy_checker_rex)
             tidy_config = json.loads(args.tidy_config)
             for checker_option in tidy_config.get('CheckOptions', []):
                 value = checker_option['value']
-                key_values = re.match(clangsa_pattern, checker_option['key'])
+                # We only store configs related to tidy checks. We run static
```
gyorb: Why do we store only clang tidy configs?

codechecker_lib/analyzers/result_handler_base.py
```diff
@@ -183,17 +183,24 @@ def get_analyzer_result_file(self):
                 '_' + analyzed_file_name + '_' + uid + '.plist'
 
             out_file = os.path.join(self.__workspace, out_file_name)
-            self.__res_file = out_file
+            self.__result_file = out_file
+
+        return self.__result_file
 
-        return self.__res_file
+    @analyzer_result_file.setter
```
gyorb: There was no setter for the result file so it should not be possible to accidentally overwrite during the analysis process the result file. The cleanup might fail to remove temporary result files.

codechecker_lib/analyzers/result_handler_base.py
```diff
@@ -183,17 +183,24 @@ def get_analyzer_result_file(self):
...
+    def result_file(self, file_path):
+        """
+        The source file which is analyzed.
```
gyorb: Is this a source file or a result file, the naming and the comment is misleading.

codechecker_lib/analyzers/result_handler_plist_to_db.py
```diff
@@ -190,3 +165,9 @@ def handle_results(self):
             self.__store_bugs(files, bugs, connection, analysis_id)
 
             connection.finish_build_action(analysis_id, self.analyzer_stderr)
+
+    def postprocess_result(self):
```
gyorb: I'm not sure why is this here better than in a base class with a default behaviour doing nothing?
It is just duplicated in the result_handler_plist_to_stdout.py module.

codechecker_lib/arg_handler.py
```diff
@@ -39,27 +39,21 @@ def handle_list_checkers(args):
     List the default enabled and disabled checkers in the config.
     """
     context = generic_package_context.get_context()
-    enabled_analyzers = args.analyzers
+    # If nothing is set, list checkers for all supported analyzers.
+    enabled_analyzers = args.analyzers or analyzer_types.supported_analyzers
```
gyorb: supported_analyzers is not a list type, please recheck if it will not cause problems in other places.

codechecker_lib/arg_handler.py
```diff
@@ -90,30 +84,21 @@ def handle_server(args):
     """
     Starts the report viewer server.
     """
-    if not host_check.check_zlib():
-        LOG.error("zlib error")
-        sys.exit(1)
+    host_check.check_zlib()
```
gyorb: We should fail at the beginning if we can not compress results.

codechecker_lib/arg_handler.py
```diff
@@ -186,9 +160,9 @@ def handle_log(args):
     args.logfile = os.path.realpath(args.logfile)
     if os.path.exists(args.logfile):
         os.remove(args.logfile)
+    open(args.logfile, 'a').close()  # Same as linux's touch.
```
gyorb: Could we remove this and update the later parts to create the file if needed?

codechecker_lib/arg_handler.py
```diff
@@ -439,7 +383,7 @@ def handle_plist(args):
     try:
         items = [(plist, args, context)
                  for plist in os.listdir(args.directory)]
-        pool.map_async(consume_plist, items, 1).get(float('inf'))
+        pool.map_async(consume_plist, items, 1).get()
```
gyorb: Are you sure we should remove this?

codechecker_lib/analysis_manager.py
```diff
@@ -212,7 +212,7 @@ def signal_handler(*arg, **kwarg):
         pool.map_async(check,
                        analyzed_actions,
                        1,
-                       callback=worker_result_handler).get(float('inf'))
+                       callback=worker_result_handler).get()
```
Xazax-hun: According to the docs, not passing an argument is the way to have no timeout. See https://docs.python.org/2/library/multiprocessing.html#multiprocessing.pool.AsyncResult.get

codechecker_lib/analyzers/analyzer_types.py
```diff
@@ -20,7 +20,8 @@
 from codechecker_lib.analyzers import config_handler_clang_tidy
 from codechecker_lib.analyzers import config_handler_clangsa
 from codechecker_lib.analyzers import result_handler_clang_tidy
```
Xazax-hun: No. In line 20 its the config handler, this is the result handler.

codechecker_lib/analyzers/analyzer_types.py
Xazax-hun: It does not work this way. See check_supported_analyzers which terminates the check when one of the analyzer is unsupported. And since the check right now is terminated there, in case this code is invoked with an unsupported analyzer, it is a programming error, so we should not continue running.

codechecker_lib/analyzers/analyzer_types.py
Xazax-hun: See my answer above. Right now the other parts of the code terminates when one of the analyzers is unsupported.

codechecker_lib/analyzers/config_handler_clang_tidy.py
Xazax-hun: Because we run the static analyzer separately, and the SA related configs in the tidy config file will not have any effect.

codechecker_lib/analyzers/result_handler_base.py
Xazax-hun: Ok, will fix.

codechecker_lib/analyzers/result_handler_base.py
Xazax-hun: It was introduced in order to be able to reduce the code duplication between plist processing and checking. It might be moved to a constructor argument, but in that case several methods and functions will need an extra argument.

codechecker_lib/arg_handler.py
Xazax-hun: We do not need this type to be a list type. We only use this in iterations, so having it as iterable is good.

codechecker_lib/arg_handler.py
Xazax-hun: The check_zlib method will terminate the execution if zlib is not found. Having additional termination here is redundant.

codechecker_lib/arg_handler.py
Xazax-hun: Ok.

codechecker_lib/arg_handler.py
Xazax-hun: Yes, see my answer above.

codechecker_lib/analysis_manager.py
gyorb: I was thinking about this [signal handling comment](https://github.com/Ericsson/codechecker/blob/15d6ae051bc1d3770800ac80193fcde996d60415/codechecker_lib/analysis_manager.py#L198-L202) why it should be kept as it was.

Xazax-hun: Done.

codechecker_lib/arg_handler.py
```diff
@@ -448,9 +448,8 @@ def handle_plist(args):
     pool = multiprocessing.Pool(args.jobs)
 
     try:
-        items = [(plist, args, context)
-                 for plist in os.listdir(args.directory)]
-        pool.map_async(consume_plist, items, 1).get(float('inf'))
+        items = [(plist, args, context) for plist in os.listdir(args.directory)]
+        pool.map_async(consume_plist, items, 1).get()
```
Xazax-hun: That is already fixed.

gyorb: Ok, there are many changes and I checked an older version.

PR #449, opened 2016-11-15
Cleanup pep8 warnings

gyorb:
This is a new pull request because there were many refactorings since #419 was started.

### DIFF
--- a/codechecker_lib/checkers.py
+++ b/codechecker_lib/checkers.py
@@ -13,7 +13,8 @@ def get_env_checkers_list(env_name):
     env_set_checkers = util.get_env_var(env_name)
 
     LOG = logger.get_new_logger("CHECKERS")
-    LOG.debug_analyzer('Getting checkers list from environment variable %s' % (env_name))
+    LOG.debug_analyzer('Getting checkers list from environment variable %s'
+                       % env_name)
 
     if env_set_checkers:
         checkers_list = env_set_checkers.split(':')
--- a/codechecker_lib/plist_helper.py
+++ b/codechecker_lib/plist_helper.py
@@ -72,42 +72,103 @@ def get_check_name(current_msg):
 
 # This map needs extending.
 checker_message_map = \
-                      {
-    "": "NOT FOUND",
-    "Access out-of-bound array element (buffer overflow)": "alpha.security.ArrayBound",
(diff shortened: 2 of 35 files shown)


Xazax-hun:
This pull request is very big. Maybe we should split it into two separate one for tests and one for the lib?

gyorb:
Because this pull request contains mostly formatting I do not think it is hard to review. I planed to put the missing modules here too but if you think this will get too big I will create separate pull requests for them. I wanted to do this as fast as possible, because if we start a new feature there can be a lot of conflicts which is not always easy to resolve especially if there is some refactoring too.

codechecker_lib/checkers.py
```diff
@@ -13,7 +13,8 @@ def get_env_checkers_list(env_name):
     env_set_checkers = util.get_env_var(env_name)
 
     LOG = logger.get_new_logger("CHECKERS")
-    LOG.debug_analyzer('Getting checkers list from environment variable %s' % (env_name))
+    LOG.debug_analyzer('Getting checkers list from environment variable %s'
+                       % (env_name))
```
Xazax-hun: Redundant paren around env_name

codechecker_lib/plist_helper.py
```diff
@@ -72,42 +72,103 @@ def get_check_name(current_msg):
...
-    "Undefined or garbage value returned to caller": "core.uninitialized.UndefReturn",
-    "Use of memory after it is freed": "alpha.unix.MallocWithAnnotations",
-    "Value stored to  during its initialization is never read": "deadcode.DeadStores",
-    "Value stored to  is never read": "deadcode.DeadStores"
-}
+    {
+        "": "NOT FOUND",
+
+        "Access out-of-bound array element (buffer overflow)":
+
```
Xazax-hun: Is this a good idea to have this much new lines? Does this make it more readable? Maybe exporting this to a JSON is better?

gyorb: Good idea but I would do this as a separate issue. Many things could be refactored but just focus on pep8 cleanup now.

codechecker_lib/checkers.py
gyorb: Done.

PR #459, opened 2016-11-22
Add options to delete multiple runs.

### LINKED ISSUE #453: Extend delete options
To support better automatic cleanup the command line delete function could be extended where the user can give a time stamp or run name and every runs before that will be removed.

* ```-keep-after myrunname``` keep runs stored after this run (myrunname will not be deleted)
* ```-keep-before myrunname``` keep runs stored before this run (myrunname will not be deleted)
* ```-keep-after hour:day:month:year``` keep runs stored after the given date
* ```-keep-before hour:day:month:year``` keep runs stored before the given date

bruntib:
The user can delete multiple runs by name, by date (before/after), and
by date based on a name (all runs checked after/before a given run.)
Fixes #453.

### DIFF
--- a/codechecker_lib/arg_handler.py
+++ b/codechecker_lib/arg_handler.py
@@ -141,9 +141,9 @@ def handle_server(args):
             # A STOP only stops the server associated with the given workspace
             # and view-port.
             if i['hostname'] != socket.gethostname() or (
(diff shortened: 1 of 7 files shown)


Xazax-hun:
There are no tests for this functionality.

gyorb:
A separate issue was created to test this #571

PR #464, opened 2016-12-06
Use absolute workspace paths everywhere

gyorb:


whisperity:
This is no longer relevant, as `arg_handler.py` has been annihilated. Individual subcommands handle workspace resolution well. _`workspace`_, as a concept, is to be eliminated, as per `TODO`-handling in #745.

PR #473, opened 2016-12-23
Fix llvm,clang download url for OSX in travis

gyorb:
The compressed file was moved to another place.


Xazax-hun:
LGTM! What was the problem with the original one? I could download from the original link.

PR #476, opened 2017-01-04
Log verbosity -v (cmdline argument change)

gyorb:
With the new command line option logging verbosity levels
can be increased with the given number of 'v'-s in the
command line.
To use a common log verbosity configuration for all
the commands the server view port argument needed to be changed.

### DIFF
--- a/codechecker/CodeChecker.py
+++ b/codechecker/CodeChecker.py
@@ -154,10 +154,10 @@ def add_verbose_arguments(parser):
     """
     Verbosity level arguments.
     """
-    parser.add_argument('--verbose', type=str, dest='verbose',
-                        choices=['info', 'debug', 'debug_analyzer'],
-                        default='info',
-                        help='Set verbosity level.')
+    parser.add_argument('-v', '--verbose', action='count',
+                        help="Set verbosity level. "
+                        "Increase the number of 'v'-s "
+                        "for more verbose output.")
 
 
 # ------------------------------------------------------------------------------
@@ -397,7 +397,7 @@ Build command which is used to build the project.''')
                                    default=util.get_default_workspace(),
                                    help=workspace_help_msg)
 
-        server_parser.add_argument('-v', '--view-port', type=int,
+        server_parser.add_argument('-p', '--view-port', type=int,
                                    dest="view_port",
                                    default=8001, required=False,
                                    help='Port used for viewing.')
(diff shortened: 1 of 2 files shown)


gyorb:
I'm not sure if anybody will use the `--verbose --verbose` when `-vv` is much shorter.
It is done this way because with the increased number of v-s more log output will be generated which means in this case that the `DEBUG_ANALYZER` log level gives less information (only related to the analyzers) compared to `DEBUG`.

whisperity:
`--verbose --verbose` is ugly and generall everyone will use `-vv` instead (even in third-party and automation scripts). It's a _potentially_ unsupported thing in `argparse`, the nice way would be using `--verbose 4` (where `4` is the number of `v`s) instead... but this is railing to off-topic.

Though I agree with @bruntib, generally one expects less critical "messages" the deeper verbosity gets for the program `*`, compare [`syslog` RFC3164](https://en.wikipedia.org/wiki/Syslog#Severity_level).

`*`: Ask ourselves the question: _Are analyzer logs in the same plane in terms of verbosity with CodeChecker output?_ Maybe it's an entirely orthogonal message feed and thus cannot be applied into "verbosity levels" neatly. Maybe a separate `--debug-analysers` (`-da`) flag &ndash; or something like that &ndash; would do the trick?

(Also, I'd move for renaming `DEBUG_ANALYZER` to simply `ANALYZER`.)

bruntib:
I just thought that one may type `--verbose` by routine instead of `-v` or `-vv` and that will not contain the CodeChecker debug messages which we usually need for debugging.

gyorb:
@whisperity I'm not sure what you meant by
> generally one expects less critical "messages" the deeper verbosity gets for the program

with deeper verbosity I expect more messages (critical or not).

By default we use `INFO` [loglevel](https://docs.python.org/2/library/logging.html#logging-levels).
We introduced `DEBUG_ANALYZERS` which is more than info but less than debug (mainly analysis related information)
Right now it looks like this.

| Level 	| Numeric value| 
-------------|-------------|
|CRITICAL 	|50|
|ERROR 	|40|
|WARNING 	|30|
|INFO 	|20|
|DEBUG_ANALYZER 	|15|
|DEBUG 	|10|
|NOTSET 	|0|

The general idea is to have a separate debug level for the analyzers is that if you develop checkers and you are interested mainly in the output of the analyzers and not how CodeChecker works.

We named it this way because it is meant to be a debug level, using `ANALYZER` only might not be obvious when this level should be used, but if everybody thinks `ANALYZER` is better we can rename it.

whisperity:
@gyorb Did we drop this idea, or will it happen in a future refactor?

PR #477, opened 2017-01-05
Use timestamps in the logs

### LINKED ISSUE #465: add timestamp to logs
with the timestamps it will be easier to detect when the analysis was started

gyorb:
Timestamps give more information about the analysis and build
process than the current log levels. Resolves #465

### DIFF
--- a/codechecker_lib/logger.py
+++ b/codechecker_lib/logger.py
@@ -9,6 +9,7 @@
 import logging
 import os
 import sys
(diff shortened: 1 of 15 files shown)


Xazax-hun:
I think in case of an error the log level should still be displayed.

gyorb:
I've updated and log levels are kept for error, warning and critical messages.

PR #478, opened 2017-01-05
Extend version numbering with revision number

gyorb:


gyorb:
Yes, thanks! I've made some cleanups too, to load the version info from the json file at one place.

whisperity:
For further reference, it would also be good to document where the line is drawn between a _minor_ change and a _revision_ change.

gyorb:
With the versioning we try to follow this:
- major - there is some backward incompatible change (db scheme, api ...)
- minor - several new features bugfixes no backward incompatible changes
- revision - mainly bugfixes

I'm thinking to change the next release version to 5.7 not 5.6.1 because there are some new features and not only bugfixes.

codechecker_lib/generic_package_context.py
```diff
@@ -64,24 +66,40 @@ def set_env(self, env_vars):
...
         try:
-            with open(version_file, 'r') as vfile:
+            with open(self.version_file, 'r') as vfile:
                 vfile_data = json.loads(vfile.read())
 
             package_version = vfile_data['version']
-            db_version = vfile_data['db_version']
+            package_build_date = vfile_data['package_build_date']
+            package_git_hash = vfile_data['git_hash']
+            database_version = vfile_data['db_version']
```
bruntib: As far as I can see the `version.json` file contains only the `version` and `db_version` attributes. How do `package_build_date` and `git_hash` get there?

codechecker_lib/analyzer.py
```diff
@@ -97,7 +97,7 @@ def run_check(args, actions, context):
 
     actions = prepare_actions(actions, enabled_analyzers)
 
-    package_version = context.version['major'] + '.' + context.version['minor']
+    package_version = context.version
```
bruntib: This variable is not necessary, because it is used only once.

codechecker_lib/arg_handler.py
```diff
@@ -367,9 +367,8 @@ def handle_plist(args):
 
         conn_mgr.start_report_server()
 
+        package_version = context.version
```
bruntib: This variable is not necessary, because it is used only once.

codechecker_lib/arg_handler.py
```diff
@@ -396,33 +395,14 @@ def handle_version_info(args):
...
-        print(('Package build date: \t' +
-              version_data['package_build_date']).expandtabs(30))
-        print(('Git hash: \t' + version_data['git_hash']).expandtabs(30))
-        print(('DB schema version: \t' + db_schema_version).expandtabs(30))
-
-    except ValueError as verr:
-        LOG.error('Failed to decode version information from the config file.')
-        LOG.error(verr)
-        sys.exit(1)
+    db_version = str(context.db_version_info)
```
bruntib: This variable is not necessary, because it is used only once.

codechecker_lib/generic_package_context.py
gyorb: The package builder script extends it.

codechecker_lib/analyzer.py
gyorb: Done.

codechecker_lib/arg_handler.py
gyorb: Done.

Issue #343, opened 2016-08-05
cleanup and refactor database schema


gyorb:
- [ ] Some schema cleanup is needed because of performance issues during the remove of the results.
- [ ] There are some tables which can be merged (BugPathEvent, BugReportPoint)
- [ ] BuildAction #329

Xazax-hun:
I am not sure that merging BugPathEvent and  BugReportPoint is a good idea. They are separate concepts in the plist generated by clang, and I doubt that this would provide us with any serious performance gain.

gyorb:
I think it is worth to try. The queries could be simplified a lot. I know they are separate concepts but the only difference between them storage wise is that the BugPathEvent contains a message not just a position. We definitely will measure if there is any performance gain.

whisperity:
Note: We should consider #565 for this!

dkrupp:
I guess this is solved by detection status related refactoring.

PR #480, opened 2017-01-05
Build command storage changed to build cmd hash

gyorb:
In the previous releases the whole original build command from the compilation
command json file was stored. It could take too much space in the database
if the build commands were long (no compression was used).
To address this issue in the commit ffef1ebfb1a831d84d925bef8df9dd78b8ae2ed7
storing the build command was only done in debug mode.
With that solution the update mode did not work correctly because it
depends on the build command right now.

With this commit the hashed build command from the compilation database
is stored to address the storage size problem and the update mode can
use the stored hash values.

### DIFF
--- a/codechecker_lib/analyzers/result_handler_plist_to_db.py
+++ b/codechecker_lib/analyzers/result_handler_plist_to_db.py
@@ -14,6 +14,7 @@ import shared
 from codechecker_lib import client
 from codechecker_lib import plist_parser
 from codechecker_lib import suppress_handler
+from codechecker_lib import logger
 from codechecker_lib.logger import LoggerFactory
 from codechecker_lib.analyzers.result_handler_base import ResultHandler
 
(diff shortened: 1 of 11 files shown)


codechecker_lib/analyzers/result_handler_plist_to_db.py
```diff
@@ -164,16 +165,19 @@ def handle_results(self):
 
             _, source_file_name = ntpath.split(self.analyzed_source_file)
 
+            if LoggerFactory.get_log_level() == logger.DEBUG:
+                analyzer_cmd = ''
```
Xazax-hun: We set analyzer_cmd to empty string only when the logger is in debug mode? Did you mean the opposite way?

gyorb: Thanks for pointing this out! I fixed it.

db_model/orm_model.py
```diff
@@ -116,6 +116,7 @@ class BuildAction(Base):
     run_id = Column(Integer,
                     ForeignKey('runs.id', deferrable=True, initially="DEFERRED",
                                ondelete='CASCADE'))
+    # hash of the build command
```
Xazax-hun: I would prefer renaming this field instead of a comment.

gyorb: Yes I know, I was thinking about it, but that means db schema change which I do not wanted to do right now that is why I put the comment there.
With v6.0 there will be more schema changes  #343 and we definitely should change it.
Is it enough if we change it for 6.0?

Xazax-hun: I think if you pass a string to the first argument of Column, than you can set the name of the column in the database. So you can change the name of the field without changing the database layout.

gyorb: If I change the just the column name this way `build_cmd = Column("build_cmd_hash", String)
` is not enough.
I got these error messages:
`[ERROR] - (psycopg2.ProgrammingError) column build_actions.build_cmd_hash does not exist`

The proper way to do it is to create an alembic schema upgrade script, which will rename the column.

gyorb: I can generate the schema upgrade script, but if we go with that all of the users need to change their CodeChecker versions to the latest, which I do not wanted to do in 5.6.1.
If we go with the db upgrade release version 5.7 might be better, because this is minor change in the database.

Xazax-hun: What is the problem when the name of the variable and the name of the column differs temporarily?

PR #486, opened 2017-01-10
Do not delete compile cmd json if set in command line

gyorb:


Xazax-hun:
There is no test for this functionality.

PR #492, opened 2017-01-11
Clean pep8 warnings from viewer_server

gyorb:


viewer_server/client_db_access_handler.py
```diff
@@ -65,9 +68,9 @@ def construct_report_filter(report_filters):
             # severity value can be 0
             AND.append(Report.severity == report_filter.severity)
         if report_filter.suppressed:
-            AND.append(Report.suppressed == True)
+            AND.append(Report.suppressed == true())
```
Xazax-hun: What is the reason behind this change?

viewer_server/client_db_access_handler.py
```diff
@@ -487,20 +491,21 @@ def __update_suppress_storage_data(self,
         source_file = session.query(File).get(report.file_id)
         source_file_path, source_file_name = ntpath.split(source_file.filepath)
 
-        LOG.debug('Updating suppress data for: ' + str(report_id) + ' bug id ' +
-                  bug_id_hash + ' file name ' + source_file_name +
+        LOG.debug('Updating suppress data for: ' + str(report_id) +
```
Xazax-hun: Maybe using a format string would be better here?

viewer_server/client_db_access_handler.py
```diff
@@ -487,20 +491,21 @@ def __update_suppress_storage_data(self,
...
-                 SuppressBug.run_id.in_(run_ids)),
-            and_(SuppressBug.hash == bug_id_hash,
-                 SuppressBug.file_name == '',
-                 SuppressBug.run_id.in_(run_ids))
-        )) \
+            .filter(or_(
+                and_(SuppressBug.hash == bug_id_hash,
+                     SuppressBug.file_name == source_file_name,
+                     SuppressBug.run_id.in_(run_ids)),
+                and_(SuppressBug.hash == bug_id_hash,
```
Xazax-hun: Do we plan to support the case when the suppress file name is empty? If we break the schema in 6.0 anyways, maybe we should delete that part of the code. (We could open a ticket for that.)

viewer_server/client_db_access_handler.py
```diff
@@ -661,10 +666,13 @@ def getCheckerDoc(self, checkerId):
 
         text = "No documentation found for checker: " + checkerId + \
                "\n\nPlease refer to the documentation at the "
+        sa_link = "http://clang-analyzer.llvm.org/available_checks.html"
+        tidy_link = "http://clang.llvm.org/extra/clang-tidy/checks/list.html"
+
         if "." in checkerId:
-            text += "[ClangSA](http://clang-analyzer.llvm.org/available_checks.html)"
+            text += "[ClangSA]("+sa_checkers_link+")"
```
Xazax-hun: Here and on the next link, it might be better to have space around the `operator+`.

viewer_server/client_db_access_handler.py
gyorb: The reason for this is because of pep8 warning `E712: Comparison to False should be "if cond is False:" or "if not cond:"`
This can be fixed if we use [this style](http://stackoverflow.com/questions/18998010/flake8-complains-on-boolean-comparison-in-filter-clause)

viewer_server/client_db_access_handler.py
gyorb: Suppression changes should be handled with caution. What if there are multiple suppressions generated by an older release, you might not want to loose all the suppression information and start again to go over the results which can be quite time consuming.
Suppression information might be stored in external files too, not just in the database.

viewer_server/client_db_access_handler.py
gyorb: It is interesting that pep8 did not complain here.

viewer_server/client_db_access_handler.py
Xazax-hun: Sounds reasonable. Do we plan to support the old format for eternity?

viewer_server/client_db_access_handler.py
Xazax-hun: Ok.

viewer_server/client_db_access_handler.py
gyorb: If we change the suppress file format I think some converter script should be provided to the users which converts to the new file format.

PR #520, opened 2017-01-19
Getpwd

gyorb:
Contains bug fixes and cleanups for authentication.

There is a command line argument change (getting the user password in the command line).

### DIFF
--- a/codechecker_lib/session_manager.py
+++ b/codechecker_lib/session_manager.py
@@ -104,6 +104,28 @@ class _Session():
             self.last_access = datetime.now()
 
 
+def check_file_owner_rw(file_to_check):
+    """
+    Check the file permissions.
+    Return:
+        True if only the owner can read or write the file.
+        False if other users or groups can read or write the file.
+    """
+
+    mode = os.stat(file_to_check)[stat.ST_MODE]
+    if mode & stat.S_IRGRP \
+            or mode & stat.S_IWGRP \
+            or mode & stat.S_IROTH \
+            or mode & stat.S_IWOTH:
+        LOG.warning(file_to_check + " is readable by users other than you!"
+                    " This poses a risk of leaking senitive"
+                    "information passwords, session tokens ...!\n"
+                    "Please 'chmod 0600 " + file_to_check + "' so only you can"
+                    " read and write it.")
+        return False
+    return True
+
+
 def load_session_cfg(session_cfg_file):
     """
     Tries to load the session config file which should be a
@@ -111,12 +133,18 @@ def load_session_cfg(session_cfg_file):
     """
 
     scfg_dict = {}
-    with open(session_cfg_file, 'r') as scfg:
-        try:
+    try:
+        with open(session_cfg_file, 'r') as scfg:
             scfg_dict = json.loads(scfg.read())
-        except ValueError as verr:
-            LOG.warning(verr)
-            LOG.warning('Not valid session config file.')
+        check_file_owner_rw(session_cfg_file)
+
+    except IOError:
+        LOG.debug('Failed to open user authentication file: ' +
+                  session_cfg_file)
+    except ValueError as verr:
+        LOG.warning(verr)
+        LOG.warning('Not valid user authentication file: ' +
+                    session_cfg_file)
 
     return scfg_dict
 
@@ -133,6 +161,7 @@ class SessionManager:
         # Check whether workspace's configuration exists.
         session_cfg_file = os.path.join(SessionManager.CodeChecker_Workspace,
                                         "session_config.json")
+
         if not os.path.exists(session_cfg_file):
             LOG.info("CodeChecker server's authentication example "
                      "configuration file created at " + session_cfg_file)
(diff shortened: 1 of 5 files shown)


codechecker_lib/session_manager.py
```diff
@@ -305,29 +310,22 @@ def __init__(self):
...
+        user_home = os.path.expanduser("~")
+        session_cfg_file = os.path.join(user_home,
                                         ".codechecker_passwords.json")
-        if not os.path.exists(session_cfg_file):
-            LOG.info("CodeChecker authentication client's example "
-                     "configuration file created at " + session_cfg_file)
-            shutil.copyfile(os.path.join(os.environ['CC_PACKAGE_ROOT'],
-                                         "config", "session_client.json"),
-                            session_cfg_file)
-            os.chmod(session_cfg_file, stat.S_IRUSR | stat.S_IWUSR)
```
Xazax-hun: Is it okay to not set the permissions so this file is only readable by the owner? Is that the default permission on every system we support?

docs/authentication.md
```diff
@@ -255,7 +252,10 @@ Privileged session expire after a set amount of time. To log out manually, issue
 
 To alleviate the need for supplying authentication in the command-line every time a server is connected to, users can pre-configure their credentials to be used in authentication.
 
-To do so, open `~/.codechecker_passwords.json`. The `credentials` section is used by the client to read pre-saved authentication data in `username:password` format.
+To do so first copy the `config/session_client.json` file from the CodeChecker package to your home directory and rename it to `codechecker_passwords.json`
+After creating the new file open `~/.codechecker_passwords.json`.
```
Xazax-hun: Shouldn't we warn the users about setting the permissions properly in case we do not set it?

viewer_clients/cmdline_client/cmd_line_client.py
```diff
@@ -41,63 +44,74 @@ def default(self, obj):
...
-                                                         session.getToken(
-                                                             args.host,
-                                                             args.port))
-
+                                                         auth_token)
     try:
         handshake = auth_client.getAuthParameters()
+
+        if not handshake.requiresAuthentication:
+            print("This server does not require privileged access.")
```
Xazax-hun: Here and some other places: is print appropriate? When to use print and when to use LOG?

viewer_clients/cmdline_client/cmd_line_client.py
```diff
@@ -466,51 +486,55 @@ def register_client_command_line(argument_parser):
...
+    listruns_parser = subparsers.add_parser('runs',
+                                            formatter_class=ADHFormatter,
+                                            help='Get the run data.')
     listruns_parser.add_argument('--host', type=str, dest="host",
                                  default='localhost',
                                  help='Server host.')
     listruns_parser.add_argument('-p', '--port', type=str, dest="port",
-                                 default=11444,
-                                 required=True, help='HTTP Server port.')
+                                 default="8001", help='HTTP Server port.')
```
Xazax-hun: Is changing the port intentional here and some other places? Why did you make it a string? Maybe factoring these ports out to somewhere is a good idea.

Xazax-hun:
Not requesting password as a command line argument is always a good idea.

docs/authentication.md
gyorb: It is in the [documentation](https://github.com/Ericsson/codechecker/blame/971aec0f2641bf0667b79aec119fd4ec1923b106/docs/authentication.md#L229) already.

codechecker_lib/session_manager.py
gyorb: I removed this part so the user have to do this manually if this way of authentication is preferred.

viewer_clients/cmdline_client/cmd_line_client.py
gyorb: The idea here is that this was started as a "separate tool" so it should work without most of the modules in the codechecker_lib modules that is why not the logger was used. To fix the verbosity levels the logger module needed to be imported, we might change the logging here too.
Printing the tables for results/runs should be done using the print not the LOG because of the formatting.

viewer_clients/cmdline_client/cmd_line_client.py
gyorb: I made this a string because the type of it is `type=str`. This was a bug before because if the user did not provide a port as an argument the type of default value would be integer, which would cause errors in other modules. Most of the code depends on that the type of the port is string.
Further refactoring is needed in multiple modules to handle the integer type properly.

codechecker_lib/session_manager.py
```diff
@@ -305,29 +310,22 @@ def __init__(self):
...
         self.__save = scfg_dict
         self.__autologin = scfg_dict.get("client_autologin") \
             if "client_autologin" in scfg_dict else True
 
         # Check and load token storage for user
-        self.token_file = os.path.join(tempfile.gettempdir(), ".codechecker_" +
+        self.token_file = os.path.join(user_home, ".codechecker_" +
                                        getpass.getuser() + ".session.json")
         LOG.debug(self.token_file)
         if os.path.exists(self.token_file):
```
whisperity: Below this line, are we sure about what the output is talking about? This part of the code is supposed to load the token storage (which is, just like session cookies, expected to be volatile), but the permission setting code warns about the `session_cfg_file` instead.

gyorb: I moved out to a separate function the permission checking so its better now than before. We can check the permissions for both files. Mainly we should warn about the config file because it needs to be created manually by the user. The token file is handled automatically with the right permissions, but I've put a warning there too, just in case.

viewer_clients/cmdline_client/cmd_line_client.py
Xazax-hun: I see, that makes sense.

codechecker_lib/session_manager.py
gyorb: @whisperity any other comments on the changes?

PR #523, opened 2017-01-20
Proper error message in case of used server address.

### LINKED ISSUE #510: handle viewer 'Address already in use' exception
Handle viewer  `Address already in use` exception, provide a proper error message to the users.

bruntib:
Fixes #510.

### DIFF
--- a/codechecker_lib/arg_handler.py
+++ b/codechecker_lib/arg_handler.py
@@ -6,10 +6,12 @@
 """
 Handle command line arguments.
 """
+import errno
 import json
 import multiprocessing
 import os
 import shutil
+import socket
 import sys
 import tempfile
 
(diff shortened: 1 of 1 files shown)


codechecker_lib/arg_handler.py
```diff
@@ -157,12 +159,19 @@ def handle_server(args):
...
+    try:
+        client_db_access_server.start_server(package_data,
+                                             args.view_port,
+                                             db_connection_string,
+                                             suppress_handler,
+                                             args.not_host_only,
+                                             context.db_version_info)
+    except socket.error as err:
+        if err.errno == errno.EADDRINUSE:
+            LOG.error("Server can't be started, maybe the given port number "
```
gyorb: Could we print out the already used port number here?

Xazax-hun: Also don't we usually stop the execution of codechecker after an error? Shouldn't there be a sys.exit here?

whisperity: There should be, to indicate the error to automation scripts. (However, technically, execution is halted here, as this is an arg-handler function.)

Issue #509, opened 2017-01-17
fix verbose logging for cmd subcommands


gyorb:
Verbose log level can not be turned on for `CodeChecker cmd` commands.

PR #529, opened 2017-01-24
Fix `CodeChecker cmd sum` unrunnable due to mistakenly left out call after #511

whisperity:


gyorb:
Thanks. I missed one of our subcommands in  #509

PR #530, opened 2017-01-24
Unduplicate code in cmd sum and results

whisperity:
Also, renamed a variable in viewer_cmd/thrift_helper so name matches "serverside" name.


whisperity:
No, we don't.

Tests only exists for the Thrift API endpoints, not for the "human" interface, i.e. the command-line calls.

gyorb:
Travis runs pep8 too, not just the tests, so it should not be skipped!

PR #531, opened 2017-01-24
Attempt to fix a concurrency error.

Xazax-hun:
If multiple processes want to insert the same file at the same time all of them might succeed. This pull request attempts to fix that. No test case, since it is hard to reproduce.


whisperity:
Version 6 will bring upon an entirely new database schema.
This fix is irrelevant to v6, and won't be a target for triage in the upcoming 5.9 release.

PR #532, opened 2017-01-24
Change --not-host-only to --listen-address

### LINKED ISSUE #331: add a configuration option listen address
Add a configuration option for viewer listen address:

The
CodeChecker server
sub option should be:
-l listen address

the default should be localhost.

Then the not so practical
--not-host-only
option can be removed.

bruntib:
Fixes #331.


gyorb:
The new command line changes in #655 will resolve this issue. The new command line arguments will be available in v6.0.

Issue #500, opened 2017-01-16
check db version instead of listing databases


gyorb:
Currently the databases are listed to check if postgresql is [available](https://github.com/Ericsson/codechecker/blob/9707c6d823e26d8966a31948901988259ac7ebf8/codechecker_lib/database_handler.py#L377). 
Checking the database version in this case should be better:
`check_db = ['psql', '-U', self.user, '-d', self.database, '-c', '"SELECT version();"', '-p', str(self.port), '-h', self.host]`

PR #535, opened 2017-01-25
Check db version instead of table list for checking database existence.

bruntib:


gyorb:
This pull request resolves the issue #500

Issue #487, opened 2017-01-10
refactor test infrastucture and layout


gyorb:
To enable more and better functional tests refactor the current layout.
Create more separated functional test packages
- [ ] update mode
- [ ] delete mode
- [ ] command line tests
- [ ] plist importer tests #485
- [ ] double suppress in suppress file and in the source code
...

gyorb:
Separate issues were created for the function tests. This issue can be closed if refactoring the layout is ready.

PR #538, opened 2017-01-25
New simple make file

gyorb:
Introduced a new Makefile for a more standard way to:
  - create a package
  - run different types of tests

Done some refactoring in the test running scripts.

This is the first step for #487.


Makefile
```diff
@@ -0,0 +1,85 @@
...
+	# virtual environment for development
+	virtualenv -p /usr/bin/python2.7 venv && . venv/bin/activate && pip install -r .ci/python_requirements && pip install nose
+
+clean_venv:
+	rm -rf venv
+
+clean:
+	rm -rf build
+	rm -rf gen-docs
+
```
whisperity: The clean command does not clean the thrift-generated files (not sure if this is intended or an oversight.)

PR #540, opened 2017-01-26
Fix bug classification in the plist importer

### LINKED ISSUE #499: severity levels are missing
all severity levels are categorized as unspecified

detected in codechecker 5.6

bruntib:
The context reads the severity map file so the clients don't need to
reproduce this code.
The plist importer now also can use this severity map.

Fixes #499

### DIFF
--- a/codechecker_lib/context_base.py
+++ b/codechecker_lib/context_base.py
@@ -5,7 +5,11 @@
 # -------------------------------------------------------------------------
 
 import abc
+import json
 import os
+from codechecker_lib.logger import LoggerFactory
+
+LOG = LoggerFactory.get_new_logger("CONTEXT BASE")
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


codechecker_lib/context_base.py
```diff
@@ -60,6 +61,9 @@ def set_env(self, env_vars):
         if codechecker_workspace:
             self._codechecker_workspace = codechecker_workspace
 
+        with open(self.checkers_severity_map_file) as severity_file:
```
gyorb: I think some error handling if the file is missing or it fails to load the json would be nice here.

codechecker_lib/context_base.py
```diff
@@ -60,6 +64,16 @@ def set_env(self, env_vars):
...
 
+        try:
+            with open(self.checkers_severity_map_file) as severity_file:
+                self._severity_map = json.load(severity_file)
+        except IOError:
+            LOG.warning('No such file: {}'.format(
+                self.checkers_severity_map_file))
+        except ValueError:
+            LOG.warning('File not in JSON format: {}'.format(
+                self.checkers_severity_map_file))
```
gyorb: I think we should print here to the users that the severity levels will not be available if the file is missing or not in a valid json format.

PR #553, opened 2017-02-02
HTML escaping in big viewer

### LINKED ISSUE #541: Incorrect display of a description with <>
one of the warning display the following in the code:
```
casting (double + 0.5) to integer leads to incorrect rounding; consider using lround (#include <cmath>) instead
```

<cmath> is interpreted, it should be using &lt; &gt; instead

bruntib:
Fixes #541

### DIFF
--- a/viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
+++ b/viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
@@ -7,6 +7,7 @@ define([
   'dojo/store/Memory',
   'dojo/store/Observable',
   'dojo/topic',
+  'dojox/html/entities',
   'dijit/Dialog',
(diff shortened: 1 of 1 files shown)


viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -439,19 +440,22 @@ function (declare, dom, style, on, query, Memory, Observable, topic,
         var column = item.report.lastBugPosition.startCol;
       }
 
-      if (fileId !== this.editor.get('sourceFileData').fileId)
+      if (fileId !== this.editor.get('sourceFileData').fileId) {
```
Xazax-hun: Is this a related change? Maybe you mixed two pull requests?

PR #556, opened 2017-02-03
--export-plist command in CodeChecker quickcheck

whisperity:
Specifying a path to this **extra** argument will output the generated plist files into the given folder.

    CodeChecker quickcheck --build "the usual" [...] --export-plist ~/myplists

### DIFF
--- a/codechecker/CodeChecker.py
+++ b/codechecker/CodeChecker.py
@@ -340,6 +340,13 @@ Build command which is used to build the project.''')
                                         ' and defines and pass them to Clang.'
                                         'This is useful when you'
                                         'do cross-compilation.')
+        qcheck_parser.add_argument('--export-plist',
+                                   type=str,
+                                   required=False,
+                                   help='Instead of printing analysis '
+                                        'results to the standard output, '
+                                        'export the generated plist files '
+                                        'to the specified directory.')
         add_analyzer_arguments(qcheck_parser)
         logger.add_verbose_arguments(qcheck_parser)
         qcheck_parser.set_defaults(func=arg_handler.handle_quickcheck)
--- a/codechecker_lib/analysis_manager.py
+++ b/codechecker_lib/analysis_manager.py
@@ -79,7 +79,7 @@ def check(check_data):
     skiplist handler is None if no skip file was configured.
     """
     args, action, context, analyzer_config_map, skp_handler, \
-        report_output_dir, use_db = check_data
+        report_output_dir, use_db, plist_export_path = check_data
 
     skipped = False
     try:
@@ -111,7 +111,8 @@ def check(check_data):
                                                          context.severity_map,
                                                          skp_handler,
                                                          progress_lock,
-                                                         use_db)
+                                                         use_db,
+                                                         plist_export_path)
 
             # Create a source analyzer.
             source_analyzer = \
@@ -161,7 +162,7 @@ def check(check_data):
 
 
 def start_workers(args, actions, context, analyzer_config_map, skp_handler,
-                  use_db=True):
+                  use_db=True, export_plist_path=True):
     """
     Start the workers in the process pool
     for every buildaction there is worker which makes the analysis.
@@ -207,7 +208,8 @@ def start_workers(args, actions, context, analyzer_config_map, skp_handler,
                              analyzer_config_map,
                              skp_handler,
                              report_output,
-                             use_db) for build_action in actions]
+                             use_db,
+                             export_plist_path) for build_action in actions]
 
         pool.map_async(check,
                        analyzed_actions,
--- a/codechecker_lib/analysis_manager.py
+++ b/codechecker_lib/analysis_manager.py
@@ -162,7 +162,7 @@ def check(check_data):
 
 
 def start_workers(args, actions, context, analyzer_config_map, skp_handler,
-                  use_db=True, export_plist_path=True):
+                  use_db=True, export_plist_path=None):
     """
     Start the workers in the process pool
     for every buildaction there is worker which makes the analysis.
--- a/codechecker_lib/analyzers/analyzer_types.py
+++ b/codechecker_lib/analyzers/analyzer_types.py
@@ -354,7 +354,6 @@ def construct_result_handler(args,
             res_handler = result_handler_plist_to_file.PlistToFile(
                 buildaction,
                 report_output,
-                lock,
                 export_plist_path)
             res_handler.print_steps = args.print_steps
 
@@ -362,7 +361,6 @@ def construct_result_handler(args,
             res_handler = result_handler_clang_tidy.ClangTidyPlistToFile(
                 buildaction,
                 report_output,
-                lock,
                 export_plist_path)
 
     res_handler.severity_map = severity_map
--- a/codechecker_lib/analyzers/analyzer_types.py
+++ b/codechecker_lib/analyzers/analyzer_types.py
@@ -21,6 +21,7 @@ from codechecker_lib.analyzers import config_handler_clang_tidy
 from codechecker_lib.analyzers import config_handler_clangsa
 from codechecker_lib.analyzers import result_handler_clang_tidy
 from codechecker_lib.analyzers import result_handler_plist_to_db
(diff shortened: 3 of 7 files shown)


codechecker_lib/analysis_manager.py
```diff
@@ -161,7 +162,7 @@ def check(check_data):
 
 
 def start_workers(args, actions, context, analyzer_config_map, skp_handler,
-                  use_db=True):
+                  use_db=True, export_plist_path=True):
```
Xazax-hun: Is True a good default here?

codechecker_lib/analyzers/analyzer_types.py
```diff
@@ -347,6 +349,21 @@ def construct_result_handler(args,
                 buildaction,
                 report_output,
                 lock)
+    elif export_plist_path:
+        if buildaction.analyzer_type == CLANG_SA:
```
Xazax-hun: Note to a future refactoring: it might scale better to have a seperate post processing step to plist and the next stage would consume the plist regardless what happened before, so no branching on the analysis type is required here.

codechecker_lib/analyzers/result_handler_plist_to_file.py
```diff
@@ -0,0 +1,82 @@
...
+class PlistToFile(ResultHandler):
+    """
+    Result handler for copying a plist file to a different location.
+    """
+
+    __metaclass__ = ABCMeta
+
+    def __init__(self, buildaction, workspace, lock, export_plist_path):
+        super(PlistToFile, self).__init__(buildaction, workspace)
+        self.__lock = lock
```
Xazax-hun: Are we sure that we need locking here?

codechecker_lib/analyzers/result_handler_plist_to_file.py
```diff
@@ -0,0 +1,82 @@
...
+
+LOG = LoggerFactory.get_new_logger('PLIST TO FILE')
+
+
+class PlistToFile(ResultHandler):
+    """
+    Result handler for copying a plist file to a different location.
+    """
+
+    __metaclass__ = ABCMeta
```
Xazax-hun: What does this do?

codechecker_lib/analyzers/analyzer_types.py
whisperity: This is most likely expected to happen in the future.

codechecker_lib/analyzers/result_handler_plist_to_file.py
whisperity: Subclassing support. The other _result handler_ classes are written the same way.

codechecker_lib/analyzers/result_handler_plist_to_file.py
whisperity: Can we be sure that every analysis for every action produces a unique plist file (name)?

Xazax-hun: There is a uniuque plist filename created for the SA,
see: https://github.com/Ericsson/codechecker/blob/master/codechecker_lib/analyzers/result_handler_base.py#L171

I did not find suddenly what happens to clang tidy's output (which is passed to PListConverter). But in case that is unique as well, we should not do any locking. The locking is there for the standard output.

whisperity: Understood, I will remove the locking call.

codechecker/CodeChecker.py
```diff
@@ -340,6 +340,13 @@ def signal_handler(sig, frame):
                                         ' and defines and pass them to Clang.'
                                         'This is useful when you'
                                         'do cross-compilation.')
+        qcheck_parser.add_argument('--export-plist',
```
gyorb: Could we use other command line argument here like `-o, --output` and some separate output `format/type` argument (`--output-type` maybe).
With that we get more flexibility in the future if we want to provide not only `plist` output.

codechecker_lib/analyzers/analyzer_types.py
```diff
@@ -313,7 +314,8 @@ def construct_result_handler(args,
                              severity_map,
                              skiplist_handler,
                              lock,
-                             store_to_db=False):
+                             store_to_db=False,
+                             export_plist_path=None):
```
gyorb: Couldn't we solve this somehow else? There are already too many arguments for this function.
Maybe splitting up result handler construction would be better.
`run_id, severity_map, store_to_db` are not required for a file output only result handler.

codechecker_lib/analyzers/result_handler_clang_tidy.py
```diff
@@ -58,3 +59,19 @@ def postprocess_result(self):
         output_file = self.analyzer_result_file
         tidy_stdout = self.analyzer_stdout.splitlines()
         generate_plist_from_tidy_result(output_file, tidy_stdout)
+
+
+class ClangTidyPlistToFile(PlistToFile):
+    """
+    Print the clang tidy results to the standard output.
```
gyorb: Why does the comment say print to standard output?

codechecker_lib/analyzers/result_handler_plist_to_file.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, buildaction, workspace, export_plist_path):
+        super(PlistToFile, self).__init__(buildaction, workspace)
+        self.__folder = export_plist_path
+
+    @property
+    def print_steps(self):
```
gyorb: Do we use this property if file is the output?

codechecker_lib/analyzers/result_handler_plist_to_file.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        Print the multiple steps for a bug if there is any.
+        """
+        return self.__print_steps
+
+    @print_steps.setter
+    def print_steps(self, value):
```
gyorb: Do we use this property if file is the output?

codechecker_lib/analyzers/result_handler_plist_to_file.py
```diff
@@ -0,0 +1,76 @@
...
+        Print the multiple steps for a bug if there is any.
+        """
+        self.__print_steps = value
+
+    def handle_results(self):
+        """This handler copies the plist file into the jailed_root."""
+        plist = self.analyzer_result_file
+
+        try:
+            files, _ = plist_parser.parse_plist(plist)
```
gyorb: I do not think we should check the plist validity here. It is checked if we want to import it.

codechecker_lib/arg_handler.py
```diff
@@ -316,10 +317,17 @@ def handle_quickcheck(args):
     """
 
     args.workspace = tempfile.mkdtemp(prefix='codechecker-qc')
+
+    if args.export_plist:
+        if not os.path.exists(args.export_plist):
+            os.mkdir(args.export_plist)
```
gyorb: I would recommend `makedirs` instead of mkdir.

codechecker_lib/analyzers/result_handler_plist_to_file.py
```diff
@@ -0,0 +1,76 @@
...
+        except Exception as ex:
+            LOG.error('The generated plist is not valid!')
+            LOG.error(ex)
+            return 1
+
+        err_code = self.analyzer_returncode
+
+        if len(files) == 0:
+            LOG.debug("Will not export '" +
```
gyorb: I think it is enough to return if the analyzer returns some error code. I that case no plist is generated.
I'm not sure we should log that.

codechecker_lib/analyzers/result_handler_plist_to_file.py
whisperity: Is this variable used _anywhere_ in the codebase at all? I don't seem to find any actual usage apart from propery accessors.

whisperity:
A better set of changes covering this functionality will be introduced after #584 concludes. @gyorb we should abandon this PR, right?

gyorb:
Yes @whisperity, I've created a new issue for this feature  #588.

PR #557, opened 2017-02-06
Lock when printing plists to stdout.

Xazax-hun:


Xazax-hun:
The new version of code checker will have a separate parse command to print the contents of a plist and that will run on one core, so no locking required.

PR #558, opened 2017-02-06
Fix duplicate suppression insertions

### LINKED ISSUE #504: suppression should be stored only once
Suppression should be run unique.
Currently the same bug hash can be stored multiple times for suppression. 
This can happen if the bug was suppressed in the suppress file before and suppressed later in the source file too.
After the suppress information from the suppress file is stored to the database and suppress information is available in the source file too, the source file suppress message should overwrite the already stored suppress information in the database.

bruntib:
When the suppress file and the source code both contain a suppression
for the same bug then both are inserted to the database resulting a
duplicate.
Fixes #504

### DIFF
--- a/storage_server/report_server.py
+++ b/storage_server/report_server.py
@@ -526,13 +526,19 @@ class CheckerReportHandler(object):
         try:
             suppressList = []
             for bug_to_suppress in bugs_to_suppress:
-                suppress_bug = SuppressBug(run_id,
-                                           bug_to_suppress.bug_hash,
-                                           bug_to_suppress.file_name,
-                                           bug_to_suppress.comment)
-                suppressList.append(suppress_bug)
+                res = self.session.query(SuppressBug) \
+                    .filter(SuppressBug.hash == bug_to_suppress.bug_hash,
+                            SuppressBug.run_id == run_id,
+                            SuppressBug.file_name == bug_to_suppress.file_name)
 
-            self.session.bulk_save_objects(suppressList)
+                if res.one_or_none():
+                    res.update({'comment': bug_to_suppress.comment})
+                else:
+                    self.session.add(SuppressBug(run_id,
+                                                 bug_to_suppress.bug_hash,
+                                                 bug_to_suppress.file_name,
+                                                 bug_to_suppress.comment))
+
             self.session.commit()
 
         except Exception as ex:


codechecker_lib/analyzer.py
```diff
@@ -134,6 +131,9 @@ def run_check(args, actions, context):
     end_time = time.time()
 
     with client.get_connection() as connection:
+        if os.path.exists(suppress_file):
+            client.send_suppress(context.run_id, connection, suppress_file)
+
```
Xazax-hun: Is there any reason you moved this? I see no problem with the moving itself, but was wondering whether this is a fix or just a refactoring with no functional change.

storage_server/report_server.py
```diff
@@ -526,11 +526,16 @@ def addSuppressBug(self, run_id, bugs_to_suppress):
         try:
             suppressList = []
             for bug_to_suppress in bugs_to_suppress:
-                suppress_bug = SuppressBug(run_id,
-                                           bug_to_suppress.bug_hash,
-                                           bug_to_suppress.file_name,
-                                           bug_to_suppress.comment)
-                suppressList.append(suppress_bug)
+                res = self.session.query(SuppressBug) \
+                    .filter(SuppressBug.hash == bug_to_suppress.bug_hash) \
```
Xazax-hun: I think we should also check for the file name when looking for duplicates.

codechecker_lib/analyzer.py
bruntib: The idea here was that when a bug suppression is stored then we check the bug hash if it exists in the database so we don't store those twice. But since the intention is to keep the suppression from the source code then the suppressions from suppress file have to come later.

storage_server/report_server.py
bruntib: Ok, I'll do it, but by the way what is the scenario when the same hash is generated in two different files?

Xazax-hun: Hash collision or copy pasted code, or accidentally similar line.

bruntib: So copy-paste codes may still cause hash collision if they are in the same file, right? Or is line number part of the hash? Then why file name isn't? Anyway I should once check how a hash is generated :wink:

Xazax-hun: Yes, copy paste in the same file is still a problem, we could not come up with a good enough hash to deal with that.

storage_server/report_server.py
```diff
@@ -526,13 +526,18 @@ def addSuppressBug(self, run_id, bugs_to_suppress):
...
-                suppress_bug = SuppressBug(run_id,
-                                           bug_to_suppress.bug_hash,
-                                           bug_to_suppress.file_name,
-                                           bug_to_suppress.comment)
-                suppressList.append(suppress_bug)
+                res = self.session.query(SuppressBug) \
+                    .filter(SuppressBug.hash == bug_to_suppress.bug_hash,
+                            SuppressBug.file_name == bug_to_suppress.file_name)
+
+                if res.first():
```
Xazax-hun: Do we want to update for all of the runs? Does this piece of code does that?

Xazax-hun: Also, do we want to add a new entry for the current run?

storage_server/report_server.py
```diff
@@ -526,13 +526,19 @@ def addSuppressBug(self, run_id, bugs_to_suppress):
...
             for bug_to_suppress in bugs_to_suppress:
-                suppress_bug = SuppressBug(run_id,
-                                           bug_to_suppress.bug_hash,
-                                           bug_to_suppress.file_name,
-                                           bug_to_suppress.comment)
-                suppressList.append(suppress_bug)
+                res = self.session.query(SuppressBug) \
+                    .filter(SuppressBug.hash == bug_to_suppress.bug_hash,
+                            SuppressBug.run_id == run_id,
+                            SuppressBug.file_name == bug_to_suppress.file_name)
```
Xazax-hun: We are expected to have exactly one result from this query. Maybe you could use one_or_none or an assert about the length of the results to ensure the invariant.

bruntib: Done, but the change is a few lines lower.

PR #559, opened 2017-02-07
Fix environment not getting cleaned up

whisperity:
In most of the errorneous cases where CodeChecker exits via `sys.exit()`, the bootstrap does not remove saved environments from `/tmp`, which, over long-running machines, created a real clutter.

After this patch, every user-graceful exit of CodeChecker will properly remove the pickle dir.

### DIFF
--- a/codechecker/CodeChecker
+++ b/codechecker/CodeChecker
@@ -11,6 +11,7 @@ Used to run the logging in the same env.
 """
 from __future__ import print_function
 
+import errno
 import ntpath
 import os
 import pickle
@@ -69,6 +70,16 @@ def main():
 
     original_env_file = os.path.join(tmp_dir, 'original_env.pickle')
 
+    def _remove_tmp():
+        # Remove temporary directory.
+        try:
+            shutil.rmtree(tmp_dir)
+        except Exception as ex:
+            if type(ex) != OSError or ex.errno != errno.ENOENT:
+                print('Failed to remove temporary directory: ' + tmp_dir)
+                print('Manual cleanup is required.')
+                print(ex)
+
     try:
         with open(original_env_file, 'wb') as env_save:
             pickle.dump(original_env, env_save)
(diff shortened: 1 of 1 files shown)


codechecker/CodeChecker
```diff
@@ -66,9 +66,23 @@ def main():
     checker_env = original_env
 
     tmp_dir = tempfile.mkdtemp()
+    global is_tmp_removed
```
gyorb: Why do we need a global variable? Shouldn't be enough to check if `tmp_dir` still exists and remove it if it is there in `__remove_tmp()`?
I'm not a big fan of global variables.

codechecker/CodeChecker
```diff
@@ -66,9 +66,23 @@ def main():
...
 
+    def _remove_tmp():
+        # Remove temporary directory.
+        try:
+            shutil.rmtree(tmp_dir)
+
+            global is_tmp_removed
+            is_tmp_removed = True
+        except Exception as ex:
+            print('Failed to remove temporary directory: ' + tmp_dir)
```
Xazax-hun: Do we want to use a warning log instead?

whisperity: This is the bootstrapper script, the concept of `log` does not exist here.

codechecker/CodeChecker
```diff
@@ -69,6 +69,15 @@ def main():
 
     original_env_file = os.path.join(tmp_dir, 'original_env.pickle')
 
+    def _remove_tmp():
+        # Remove temporary directory.
+        try:
+            shutil.rmtree(tmp_dir)
+        except Exception as ex:
```
gyorb: It would be a nice simplification if we would handle `OSError` exceptions separately like `No such file or directory` in that case we do not have to check if the directory exists at line 104.

whisperity: Done! :slightly_smiling_face:

PR #564, opened 2017-02-10
Highlights and icons in BugPathView

### LINKED ISSUE #526: Differentiation of BugPath elements (e.g. 'entered call' and 'calling') in the bugpath overview list
Come up with an idea to understand and categorise certain important and usual bugpath elements (like entered call, calling, "Assuming ...", etc.) and show them in a user-friendly way (icons? background color? etc.) would make a "glimpse skim" of the overview much easier.

### LINKED ISSUE #560: UI coloring improvements
With coloring the different bug path elements (event, position) it is easier recognize the most important information, how a bug happened.
 - bug events
 - last bug path position)
 - bug path changes between files
 - bug path parts inside one function?

whisperity:
- Checker messages resulting from functions deeper down the call stack will have different background to easily distinguish function blocks.
 - Certain common checker messages are given custom icons (`Calling`, `Entered call`, _Loop_ actions, assumes)
 - Better generic iconization for bug categories (compliant to usual severity colouring schemes) and messages (instead of plain :file_folder: and :page_facing_up:)
 - If the BugPath only contains one line (such as the case with `deadcode.DeadStores` and most of clang-tidy), don't show the result two times)
 - Filenames are not shown if the BugPath is restricted to a single file only; and if not, the too long filenames are cropped.

Colouring and icons
Motherlode of conditional assumes
No multiple showing for DeadStores
Filename cropping

> **Note!** The severity categories not having any bugs as children are only shown on the images for demonstration purposes, the code still hides empty categories.

The icons used in this change are subset that of [FontAwesome](http://fontawesome.io/).

(closes #560, closes #526)


viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,49 +462,201 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
+        };
+        var extractFuncName = function(prefix) {
+          if (contains(prefix)) {
+            return msg.replace(prefix, "").replace("'", "");
+          }
+          return false;
+        };
+
+        var func;
+        if (func = extractFuncName("Calling ")) {
```
Xazax-hun: Maybe these string could be separated into a JSON instead of being hardcoded?

whisperity:
> @Xazax-hun wrote:
>
> Maybe these string could be separated into a JSON instead of being hardcoded?

I disagree. Depending on where we read the JSON file, we will always end up having an extra HTTP request for the file itself, which could, at the wrong place, tremendously hinder the experience...

... unless we put the JSON into one of the JS files itself.

But with this route, we also have a problem: the match rules are not simple, and because this is an initial implementation, I am not expecting them to be simple as we discover further "common" _bugevent_s. As you can see in this code, we have matches where (to prevent multiple string matching), we need to extract a token, sometimes we match a substring (where the behaviour of `else if` is neccessarsy, as `Assuming` is a substring of `Assuming the condition`), and at some places because the check message is not a template, we use full string equality.

Xazax-hun:
Oh, I see. These messages are the subject to change, such changes are not frequent though. Do you have an idea how to handle compatibility with multiple clang versions in case such changes are introduced? What about regexes to deal with the different kinds of matching strategies?

whisperity:
**DO NOT MERGE UNTIL I FIX THIS**

I have came to the realisation that we could at some places better use a `startsWith` instead of an `indexOf`. 🙂 

Also we need to update the README with this feature's changes.

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,49 +462,201 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
+          stack.funcStack.push(func);
+          stack.background = that._highlight_colours[stack.funcStack.length % that._highlight_colours.length];
+
+          highlight.iconOverride = "calling";
+        } else if (contains("Entered call from ")) {
```
whisperity: `startsWith` here

Xazax-hun: Your last commit did not change this line. Is that intentional?

whisperity: No, it was an accidental oversight due to having multiple redesigns in the code. It will be fixed in the next commit, along with a slight README update.

whisperity:
> @Xazax-hun wrote:
>
> [...] Do you have an idea how to handle compatibility with multiple clang versions in case such changes are introduced? [...]

See #566, such information stored for the run itself could, at a later point, also be used here for such distinction.

Though I don't think this is an issue at this point, as the highlights were only made for the most common and most important checker messages (I went through the code of the implemented checkers in current `master` LLVM, looking for bugstep prints), and everything _else_ is getting (though, unfortunately, not visible on the screenshots I attached) a "command-line" icon (`>_`) which also sort of highlights the message, showing that it is important because it is not a "general" thing usually seen in BugPaths.

(However, the level of how general a message is were my/our arbitrary decision (such is the case with _Severity_ levels also!)

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,49 +462,204 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
+      var msg = step.msg;
+
+      // The background must be saved BEFORE stack transition. Calling is in the caller, return is in the called func.
+      var highlight = {
+        background : stack.background,
+        iconOverride : undefined
+      };
+
+      var stackTransition = function() {
+        var contains = function(substr) {
```
bruntib: This function is not used anywhere.

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,49 +462,204 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
+          return msg.indexOf(substr) !== -1;
+        };
+        var startsWith = function(substr) {
```
bruntib: Don't introduce this function. Its body can be used at the place of usage.

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,49 +462,204 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
+      var that = this;
+      var msg = step.msg;
+
+      // The background must be saved BEFORE stack transition. Calling is in the caller, return is in the called func.
+      var highlight = {
+        background : stack.background,
+        iconOverride : undefined
+      };
+
+      var stackTransition = function() {
```
bruntib: Why do we need a function for this? It is called once after its definition.

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,49 +462,204 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
+        var contains = function(substr) {
+          return msg.indexOf(substr) !== -1;
+        };
+        var startsWith = function(substr) {
+          return msg.startsWith(substr);
+        };
+        var extractFuncName = function(prefix) {
```
bruntib: You can use the traditional function definition syntax instead of storing a lambda in a variable.

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,49 +462,204 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
+          if (startsWith(prefix)) {
+            return msg.replace(prefix, "").replace("'", "");
+          }
+          return false;
```
bruntib: A function without `return` statement returns `undefined` by default which is considered a falsey value.

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,49 +462,204 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
+      var stackTransition = function() {
+        var contains = function(substr) {
+          return msg.indexOf(substr) !== -1;
+        };
+        var startsWith = function(substr) {
+          return msg.startsWith(substr);
+        };
+        var extractFuncName = function(prefix) {
+          if (startsWith(prefix)) {
+            return msg.replace(prefix, "").replace("'", "");
```
bruntib: What is this apostrophe which is eliminated here? Is it conscious that only one occurrence is removed?

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,49 +462,204 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
         parent : report.reportId + '',
         report : report,
-        isLeaf : true
+        isLeaf : true,
+        kind : 'msg'
       });
 
+      var filePaths = [];
+      var areThereMultipleFiles = false;
+      var highlightStack = {
```
bruntib: Reduce the scope of these variables. These are used only in the callback of `getReportDetails`.

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,49 +462,204 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
       CC_SERVICE.getReportDetails(report.reportId, function (reportDetails) {
-        reportDetails.pathEvents.forEach(function (step, index) {
-          that.bugStore.put({
-            id : report.reportId + '_' + (index + 1),
-            name : 'Line ' + step.startLine + ': ' + entities.encode(step.msg),
-            parent : report.reportId,
-            bugPathEvent : step,
-            isLeaf : true,
-            report : report
+        if (reportDetails.pathEvents.length > 1) {
```
bruntib: You can reduce an indentation level if you negate this condition and simply return from this callback.

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
whisperity: (Good catch, I can't believe JavaScript works this way...)

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,46 +462,204 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
+        name : '<b><u>' + entities.encode(report.checkerMsg) + '</u></b>',
         parent : report.reportId + '',
         report : report,
-        isLeaf : true
+        isLeaf : true,
+        kind : 'msg'
       });
 
       CC_SERVICE.getReportDetails(report.reportId, function (reportDetails) {
+        console.log(reportDetails.pathEvents.length);
```
bruntib: A log message was left in the source accidentally.

viewer_clients/web-client/scripts/codecheckerviewer/BugViewer.js
```diff
@@ -462,46 +462,204 @@ function (declare, dom, style, on, query, Memory, Observable, topic, entities,
...
-             + '</b>',
+        name : '<b><u>' + entities.encode(report.checkerMsg) + '</u></b>',
         parent : report.reportId + '',
         report : report,
-        isLeaf : true
+        isLeaf : true,
+        kind : 'msg'
       });
 
       CC_SERVICE.getReportDetails(report.reportId, function (reportDetails) {
```
bruntib: Could you please refactor this lambda to a "private" report detail formatter function? Don't use too big lambdas.

PR #577, opened 2017-02-15
Show version info in Viewer

whisperity:
`CodeChecker server`:
image

`CodeChecker server --verbose debug`
image

The same output is added to `CodeChecker version`. In debug mode, extra annotation is showed, while "normal" mode only shows the release tag.

### DIFF
--- a/codechecker_lib/logger.py
+++ b/codechecker_lib/logger.py
@@ -40,9 +40,8 @@ def add_verbose_arguments(parser):
                         help='Set verbosity level.')
 
 # ------------------------------------------------------------------------------
-logging.DEBUG_ANALYZER = 15
-DEBUG_ANALYZER = logging.DEBUG_ANALYZER
-logging.addLevelName(logging.DEBUG_ANALYZER, 'DEBUG_ANALYZER')
+DEBUG_ANALYZER = logging.DEBUG_ANALYZER = 15
+logging.addLevelName(DEBUG_ANALYZER, 'DEBUG_ANALYZER')
 
 
 class CCLogger(logging.Logger):
--- a/codechecker_lib/logger.py
+++ b/codechecker_lib/logger.py
@@ -41,6 +41,7 @@ def add_verbose_arguments(parser):
 
 # ------------------------------------------------------------------------------
 logging.DEBUG_ANALYZER = 15
+DEBUG_ANALYZER = logging.DEBUG_ANALYZER
 logging.addLevelName(logging.DEBUG_ANALYZER, 'DEBUG_ANALYZER')
 
 
--- a/thrift_api/report_viewer_server.thrift
+++ b/thrift_api/report_viewer_server.thrift
@@ -250,6 +250,9 @@ service codeCheckerDBAccess {
   // returns the database access handler api version
   string getAPIVersion();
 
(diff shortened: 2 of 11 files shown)


codechecker_lib/logger.py
```diff
@@ -41,6 +41,7 @@ def add_verbose_arguments(parser):
 
 # ------------------------------------------------------------------------------
 logging.DEBUG_ANALYZER = 15
+DEBUG_ANALYZER = logging.DEBUG_ANALYZER
 logging.addLevelName(logging.DEBUG_ANALYZER, 'DEBUG_ANALYZER')
```
gyorb: I think we could use the `DEBUG_ANALYZER` instead of `logging.DEBUG_ANALYZER` when we add the new level name.

thrift_api/report_viewer_server.thrift
```diff
@@ -250,6 +250,9 @@ service codeCheckerDBAccess {
   // returns the database access handler api version
   string getAPIVersion();
 
+  // returns the CodeChecker version that is running on the server
+  string getPackageVersion();
```
gyorb: Shouldn't we increase the `API_VERSION = '5.0'` to `5.1`?  
We are adding a new API method, this is just an extension and won't break backward compatibility.

viewer_clients/web-client/style/codecheckerviewer.css
```diff
@@ -1,3 +1,5 @@
+@import url('https://fonts.googleapis.com/css?family=Raleway|Source+Code+Pro');
```
gyorb: I like new fonts. But after building a package we should not depend on external web resources. I see there is a fall back font, but I'm not a big fan of this.

viewer_server/client_db_access_handler.py
```diff
@@ -89,11 +89,13 @@ def __init__(self,
                  checker_md_docs,
                  checker_md_docs_map,
                  suppress_handler,
-                 db_version_info):
+                 db_version_info,
+                 version):
```
gyorb: maybe naming `version` to `package_version` or something similar can be more descriptive

viewer_clients/web-client/style/codecheckerviewer.css
whisperity: How should we handle it then? Should the `build_package.py` also download and "copy" the font into the webserver's folder?

Fonts available on _Google Fonts_ are entirely open-source and free to use even in commercial projects. (Though they advise _against_ downloading and serving it locally, it is not _legally_ prohibited.)

gyorb:
> We need to also find a way for this to be working with releases downloaded from GitHub. Currently a
> downloaded release source code does not include a Git structure, so git rev-parse HEAD and various
> other commands fail, leaving version information empty. @whisperity 

I think GitHub by design does not pack git information into the compressed files in the release.
We still can show the user the major version and if we need to support a release we know from which tag it was generated

whisperity:
@gyorb Oh, right, I forgot to mention, but that "comment" of mine was fixed in commit cd40aa2. The "describe" output defaults to the hardcoded Git information and is only rewritten to _actual_ git tag if it is possible.

PR #597, opened 2017-02-27
Use the python2 from the environment (more flexible)

gyorb:
Do not use the hardcoded full path.
In some machines you might not want to use that python version.


gyorb:
OSX Travis build is down because http://releases.llvm.org/3.8.0/ is not available now.

PR #584, opened 2017-02-21
Restructure the project

whisperity:
In this PR, we introduce a new internal structure for CodeChecker in anticipation of further modularisation.

 * The many different libraries, such as `viewer_clients` and `report_server` are condensed in a **single** `lib/` structure.
    * The `libcodechecker` structure is as follows:
       * Every "module" of _CodeChecker_ (logger, analyzer, storage, cmdline-client) is now given its own subfolder/sublibrary.
       * Shared libraries are in the root of `libcodechecker`
       * (Some cross-references between libraries remain which should be ironed out in the future, marked with a `TODO` in the code.)
    * This rename makes it easier for IDEs such as PyCharm to properly resolve references between parts of the program.
 * The build and install script has been modified accordingly.
 * The behaviour of _CodeChecker_ is **not** modified at all: the same entry-points and invocation rules apply.
 * Generated Thrift Python codes are stored in `gencodechecker` which is included as a top-level directory in the `PYTHONPATH`.
 * The developer folder structure now follows conventions such as `bin`, `libcodechecker`, `vendor` instead of names unique to our project.

Resolves #590

Issue #576, opened 2017-02-15
Autocomplete support


whisperity:
There is a way to enable shell `[TAB]` autocompletion for `argparse`-driven entry points: [`https://argcomplete.readthedocs.io/en/latest/`](https://argcomplete.readthedocs.io/en/latest/).

This should be laid out in the build script if we are to give a binary package (see #528) as this introduces a **tremendous** help in usage.

Issue #580, opened 2017-02-16
intercept build error handling


gyorb:
improve intercept build error handling and detection

dkrupp:
intercept-build created build json is empty on ubuntu 14, clang 4.0-master

dkrupp:
intercept-build logging does not work on Ubuntu 16 either

whisperity:
How are you invoking `intercept-build`?

whisperity:
My investigation reveals that this is an issue when `intercept-build` is invoked with a direct compiler invocation. It has been reported to @rizsotto. More on the thing here: rizsotto/scan-build#79.

whisperity:
Side question @gyorb, @ednikru: Should we just _always_ build ldlogger, no matter what? It could be that at `make package` intercpt-build is available, but it moves out of your path when you are using CodeChecker. In this case, the logger automatically tries to fall back to ldlogger but fails horribly.

If we were to build ld-logger even if intercept is availabe, we could make this transition a lot more easier. And building ld-logger doesn't increase CC's build time that much at all.

whisperity:
@gyorb What did you mean by "improve" in the original summary?
The issue itself with intercept-build has been resolved by #610.

gyorb:
Until we support out own ldlogger solution we can build it always into the package (use intercept-build at runtime if available).
By improvement I meant here that we should provide better error/debug messages to the users (if it is possible) when logging the compile commands fail for some reason (not just the message that the compilation command db is empty).

bruntib:
I close this ticket as a duplicate of https://github.com/Ericsson/codechecker/issues/4678

Issue #505, opened 2017-01-17
Parsing string values in -D compiler option


goshabog:
Hi
I have compilation option -DPRODUCT_NAME='"SCTP IETF"' for my project. But when I’m using CodeChecker for code analysis, the following  error is raised:
sctp.c:715:1: error: use of undeclared identifier 'SCTP'
REGVERSION_DEF(SCTP)
version.h:66:36: note: expanded from macro 'REGVERSION_DEF'
                                   PRODUCT_NAME,                        \
                                   ^
<command line>:12:22: note: expanded from here
#define PRODUCT_NAME SCTP
Also I see that the following option is used in command line for clang
 -DPRODUCT_NAME="SCTP
Does CodeChecker support such parameter – string with space? 
I am using 5.6 version of CodeChecker
Best regards, Igor

whisperity:
At first glance, this looks like an escaping issue. Do you use a build environment or a script file, or are you calling the `--build` option and supply the compile command bare?

As a workaround, try making a shell script which runs your compiler and run CodeChecker as:

```bash
CodeChecker check --name foobar --build "./myscript.sh"
```

gyorb:
@goshabog did you manage to analyze your project with the suggestions by @whisperity ?

PR #610, opened 2017-03-10
New subcommand infrastructure + splitting out log command

### LINKED ISSUE #593: new log command
create a separate log command to log compilation actions

The CodeChecker log should use this in the background.

whisperity:
This PR introduces a new command infrastructure to help in further compartmentalising CodeChecker which begun in #584. Closes #593.

# Overview 

## New entrypoints
New CodeChecker entrypoints are introduced, following the "usual scheme" set out by the giants ahead of us. In line with, e.g. how Git works (`git log` and `git-log` are both commands), we will now have each "part"/"subcommand"/"feature" of CodeChecker as a direct entry point, such as `codechecker-log`. This, in the future, will help us create more dynamic helps, such as _man pages_ for a Debian package, or a TAB autocompletion support (#576).

In reality, `codechecker-log` and `CodeChecker log` has the same behaviour, `codechecker-log` is just a directly callable wrapper script that is also put in the `PATH`.

## Putting the argument list and the executed function together

> AKA: The decline of `arg_handler.py`. :wink:

`libcodechecker/arg_handler.py` is a voloptous patchwork, which imports **everything** from all parts of CodeChecker. While the executed code are in this file, their help and argument definitions are in `bin/CodeChecker.py`...

From now on, subcommands of _CodeChecker_ will be located in their own file, `libcodechecker/log.py`, which has an easy interface.

 * `get_argparser_ctor_args` returns a dict which specifies how the subcommand argument parser should be constructed
 * `add_arguments_to_parser(parser)` is the function which puts the arguments into the constructed argument parser.
 * `main(args)` executes the method of the subcommand. (Here `args` is a parsed `argparse.Namespace`.)

## Dynamic package building and loading

The `CodeChecker.py` script will now dynamically load the available subcommands' help and argument list. This dynamic list is created at **package build** via the install script. However, if the user supplies a _valid_ subcommand (either calling `CodeChecker log` or `codechecker-log`, the two are equal), the helpers, **and** executed actions (and thus, every library module that executed function needs) for all the _other_ commands will not be loaded.

While CodeChecker is already really fast on most machines, I hope this will help to increase speed even more, and this will definitely help to further compartmentalise CodeChecker.

----

All the above, as of this PR, only applies to `CodeChecker log`, the remaining commands are untouched, for now.

# Helpers

Just like how the test framework is given a "bootstrap script" to create new test cases, I'm introducing `scripts/create_new_subcommand.py` which, when called with a single `subcommand-name` argument, will create, in the development directory, the neccessary skeleton for a new command, `CodeChecker subcommand-name` (or `codechecker-subcommand-name`).

----

Also, as a minor change, I've updated the `CodeChecker log`'s help message a bit.

### DIFF
--- a/bin/codechecker-log
+++ b/bin/codechecker-log
@@ -1,0 +1,21 @@
+#!/usr/bin/env python
+# -------------------------------------------------------------------------
+#                     The CodeChecker Infrastructure
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+"""
+Entry point for the build logger command.
+"""
+
+import imp
+import os
+
+THIS_PATH = os.path.dirname(os.path.abspath(__file__))
+CC = os.path.join(THIS_PATH, "CodeChecker")
+
+# Load CodeChecker from the current folder (the wrapper script (without .py))
+CodeChecker = imp.load_source('CodeChecker', CC)
+
+# Execute CC's main script with the current subcommand.
+CodeChecker.main("log")
--- a/libcodechecker/analyze/log_parser.py
+++ b/libcodechecker/analyze/log_parser.py
@@ -93,6 +93,13 @@ def parse_compile_commands_json(logfile, add_compiler_defaults=False):
     counter = 0
     for entry in data:
         sourcefile = entry['file']
+
+        if not os.path.isabs(sourcefile):
+            # Newest versions of intercept-build can create the 'file' in the
+            # JSON Compilation Database as a relative path.
+            sourcefile = os.path.join(os.path.abspath(entry['directory']),
+                                      sourcefile)
+
         lang = option_parser.get_language(sourcefile[sourcefile.rfind('.'):])
 
         if not lang:
@@ -100,7 +107,26 @@ def parse_compile_commands_json(logfile, add_compiler_defaults=False):
 
         action = build_action.BuildAction(counter)
 
-        command = entry['command']
+        if 'command' in entry:
+            command = entry['command']
+        elif 'arguments' in entry:
+            # Newest versions of intercept-build create an argument vector
+            # instead of a command string.
+            for i in range(0, len(entry['arguments'])):
+                arg = entry['arguments'][i]
+                if ' ' in arg:
+                    # If there is an argument with a space in it, the join
+                    # below will mess the invocation up. (-DVAR=va lue will be
+                    # passed as -DVAR=va)
+                    #
+                    # Build.json created by ld-logger escapes these strings
+                    # and they never reach the analyser later on.
+                    #
+                    # TODO: Better handle this. (See issue #505.)
+                    entry['arguments'][i] = '\"' + arg + '\"'
+            command = ' '.join(entry['arguments'])
+        else:
+            raise KeyError("No valid 'command' or 'arguments' entry found!")
         results = option_parser.parse_options(command)
 
         action.original_command = command
--- a/libcodechecker/log/host_check.py
+++ b/libcodechecker/log/host_check.py
@@ -1,0 +1,43 @@
+# -------------------------------------------------------------------------
+#                     The CodeChecker Infrastructure
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+
+import errno
+import os
+import subprocess
+
+from libcodechecker.logger import LoggerFactory
+
+LOG = LoggerFactory.get_new_logger('HOST CHECK')
+
+
+def check_intercept(env):
+    """
+    Simple check if intercept (scan-build-py) is available.
+    """
+    intercept_cmd = ['intercept-build']
+    try:
+        with open(os.devnull, 'wb') as null:
+            res = subprocess.check_call(intercept_cmd,
+                                        env=env,
+                                        stdout=null,
+                                        stderr=null)
+
+        if not res:
+            return True
+        else:
+            LOG.debug('Failed to run: "' + ' '.join(intercept_cmd) + '"')
+            return False
+    except subprocess.CalledProcessError:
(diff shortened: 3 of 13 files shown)


whisperity:
Due to the log command already being refurbished and thus we nevertheless would need to rebase this, I am introducing the fix for #580 here. Once again, thank you @rizsotto for the prompt reply at rizsotto/scan-build#79.

libcodechecker/log/host_check.py
```diff
@@ -0,0 +1,38 @@
...
+
+def check_intercept(env):
+    """
+    Simple check if intercept (scan-build-py) is available.
+    """
+    intercept_cmd = ['intercept-build', '--help']
+    try:
+        res = subprocess.call(intercept_cmd,
+                              env=env,
+                              stdout=subprocess.PIPE,
```
gyorb: There is a note not to use `subprocess.PIPE` in the `call` because it can deadlock. `popen` and `communicate` is the [recommended solution](https://docs.python.org/2/library/subprocess.html).

scripts/build_package.py
```diff
@@ -639,9 +639,35 @@ def build_package(repository_root, build_package_config, env=None):
     time_now = time.strftime("%Y-%m-%dT%H:%M")
     version_json_data['package_build_date'] = time_now
 
+    version_json_data['available_commands'] = []
```
gyorb: Why do we put the subcommands into the version config file?

libcodechecker/log/host_check.py
whisperity: We don't even need to properly communicate with `intercept-build` here, we only need to see if the command returns properly.

scripts/build_package.py
whisperity: The version file is the one which contains the _most_ build-time created information, such at git describe and the like.

I felt it is appropriate to put the information here as the list of available commands is a property of a particular build taken, ie. users could arbitrarily add or remove commands before building their package.

libcodechecker/analyze/log_parser.py
```diff
@@ -93,14 +93,28 @@ def parse_compile_commands_json(logfile, add_compiler_defaults=False):
...
 
         action = build_action.BuildAction(counter)
 
-        command = entry['command']
+        if 'command' in entry:
+            command = entry['command']
+        elif 'arguments' in entry:
+            # Newest versions of intercept-build create an argument vector
+            # instead of a command string.
+            command = ' '.join(entry['arguments'])
```
rizsotto: When the argument list is `["one", "two three"] then you concatenate it as you do it here, you won't be able to get the original list back with split. You need to annotate the list items better. (Or if you won't do the split by yourself, the called shell will do it. So you need shell escaping. How portable is that?) The main driver to change this in `intercept-build` was to get rid of these concatenation and splitting errors. The current compilation database parser in Clang understands the `arguments` filed... So the question, why don't you pass the command around as a list instead of a string?

whisperity: The changes in this patch are specifically around creating a new architecture for the commands. There are already at least one issue about argument list escaping, #505 comes to mind. The moment we start to refurbish the `check` command, we'll look into these issues.

Right now, `ld-logger` puts custom arguments from

```bash
CodeChecker check -n run -b "g++ -DVAR='va lu e' main.cpp -o a.out"
```

to the build.json as: `\"-DVAR=va lu e\"` but these values &ndash; and I've checked it just now &ndash; do not reach the analyzer command.

In case of `intercept-build`, however, `-DVAR=va ` reaches the analyzer. I will now make sure the same escape sequence is applied to the command here, and we will discuss proper argument forwarding later.

bin/codechecker-log
```diff
@@ -0,0 +1,21 @@
...
+"""
+
+import imp
+import os
+
+THIS_PATH = os.path.dirname(os.path.abspath(__file__))
+CC = os.path.join(THIS_PATH, "CodeChecker")
+
+# Load CodeChecker from the current folder (the wrapper script (without .py))
+CodeChecker = imp.load_source('CodeChecker', CC)
```
Xazax-hun: What is the reason for using imp here? Can't we load it with the regular import syntax?

whisperity: No. The folder where the environment bootstrapper resides is not in the `PYTHONPATH`. Also it's not a real Python package either, and importing from relative paths outside a package is an error, and inside a package is still discouraged.

See, for example, [this question on SO](http://stackoverflow.com/a/7506029). We couldn't use `from ..bin` because we can't make assumptions on where the CC scripts that go into the users' `PATH` are installed.

gyorb:
Could we squash some of the commits?
* some text formating only (could be part of another)
* the main part of his pull request the log command changes... (maybe squash some of them to be more in one group)
* intercept-build changes (I think this should be at the end/beginning of the commits and not mixed together with the log command changes)

whisperity:
Sure, done. :wink:

libcodechecker/analyze/log_parser.py
whisperity: @rizsotto How would you pass argument lists to clang in a shell invocation? The analyzer command is built by hand, the build.json does not reach the individual clang calls.

rizsotto: Not sure that i get the question... the point i was trying to rise was that when you read the compilation database, it's better to keep `arguments` as list (don't concatenate it) and if it was a `command` make it as a list (with `shlex.split` or something clever)... So either way, you get a list of arguments... So, you can call `subprocess.call` or `subprocess.check_call` without `shell=True`. (Which will make your code more portable and get rid of the problem i've mentioned before.)

whisperity: For the record --- I've already removed the question because I found the solution meanwhile, the question was:

> How would you pass an argument list to clang, as right now the `build.json` never directly reaches the analyzer.

The build database is parsed and filtered, and from that do we create the analyzer invocation.

I've investigated the issue and came up with the solution that works both for _intercept-build_ and _ld-logger_, while also esentially fixing the aforementioned issue #505. Please see #631.
#631

PR #611, opened 2017-03-14
Default checks

Xazax-hun:
This pull request enables some additional clang tidy checks by default. Some of those checks did not exist when the initial list of default checks was created and some of them was improved since then. This pull request also adds missing severity levels to some checks.


whisperity:
See #612, maybe we should also discuss that before merging.

Xazax-hun:
I think we should fix that in a subsequent pull request.

PR #613, opened 2017-03-17
Refurbish 2d-array to (table|json|csv) converters

whisperity:
Introduce a more reusable module for formatting 2d-array data matrices to various formats, such as:

 * `rows`: minimal formatting, just rows printed after each other (currently analogue to how `CodeChecker checkers`) output
 * `table`: `libcodechecker.util.print_table`
 * `dictlist`: _zip_ (the list-to-tuple, not the archiver) the "header" row and each "data" rows into a list of dict objects
 * `json`: the `dictlist` in JSON format

`dictlist_to_twodim` is the inverse of `dictlist`.

----

Some of these functions might appear unused as of now, they are getting used in features that I'm introducing later. The reason for this PR is to tighten the scope of features added.

### DIFF
--- a/libcodechecker/arg_handler.py
+++ b/libcodechecker/arg_handler.py
@@ -134,7 +134,7 @@ def handle_server(args):
                              str(instance['port'])))
 
         print("Your running CodeChecker servers:")
-        util.print_table(rows)
+        print(util.twodim_to_table(rows))
         sys.exit(0)
     elif args.stop or args.stop_all:
         for i in instance_manager.list():
--- a/libcodechecker/cmd/cmd_line_client.py
+++ b/libcodechecker/cmd/cmd_line_client.py
@@ -20,7 +20,7 @@ from Authentication import ttypes as AuthTypes
 
 from libcodechecker import session_manager
 from libcodechecker import logger
-from libcodechecker.util import print_table
+from libcodechecker.output_formatters import twodim_to_str
 
 from . import thrift_helper
 from . import authentication_helper
@@ -258,15 +258,12 @@ def handle_list_runs(args):
         print(CmdLineOutputEncoder().encode(results))
 
     else:  # plaintext, csv
-        rows = [('Name', 'ResultCount', 'RunDate')]
+        header = ['Name', 'ResultCount', 'RunDate']
+        rows = []
         for run in runs:
             rows.append((run.name, str(run.resultCount), run.runDate))
 
-        if args.output_format == 'csv':
-            writer = csv.writer(sys.stdout)
-            writer.writerows(rows)
-        else:
-            print_table(rows)
+        print(twodim_to_str(args.output_format, header, rows))
 
 
 def handle_list_results(args):
@@ -302,13 +299,13 @@ def handle_list_results(args):
     if args.output_format == 'json':
         print(CmdLineOutputEncoder().encode(all_results))
     else:
-        rows = []
-        if args.suppressed:
-            rows.append(
-                ('File', 'Checker', 'Severity', 'Msg', 'Suppress comment'))
-        else:
-            rows.append(('File', 'Checker', 'Severity', 'Msg'))
 
+        if args.suppressed:
+            header = ['File', 'Checker', 'Severity', 'Msg', 'Suppress comment']
+        else:
+            header = ['File', 'Checker', 'Severity', 'Msg']
+
+        rows = []
         for res in all_results:
             bug_line = res.lastBugPosition.startLine
             checked_file = res.checkedFile + ' @ ' + str(bug_line)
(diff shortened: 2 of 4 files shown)


libcodechecker/output_formatters.py
```diff
@@ -0,0 +1,186 @@
...
+# -------------------------------------------------------------------------
+"""
+Contains functions to format and pretty-print data from two-dimensional arrays.
+"""
+
+import json
+
+# The supported formats the users should specify. (This is not an exhaustive
+# list of ALL formats available.)
+USER_FORMATS = ['rows', 'table', 'csv', 'json']
```
gyorb: Could we use this list in the argument handler to show the user the selectable options
[here](https://github.com/Ericsson/codechecker/blob/master/libcodechecker/cmd/cmd_line_client.py#L497)?

How about checking the valid formats based on this list in the `twodim_to_str` function?
I did not found any usage of this list in the code.

libcodechecker/cmd/cmd_line_client.py
```diff
@@ -321,11 +318,7 @@ def handle_list_results(args):
                 rows.append(
                     (checked_file, res.checkerId, sev, res.checkerMsg))
 
-        if args.output_format == 'csv':
-            writer = csv.writer(sys.stdout)
-            writer.writerows(rows)
-        else:
-            print_table(rows)
+        print(twodim_to_str(args.output_format, header, rows))
```
gyorb: `ValueError` exceptions should be handled if `twodim_to_str` is called, what if the user makes a typo in the format?

libcodechecker/output_formatters.py
```diff
@@ -0,0 +1,186 @@
...
+    Converts the given two-dimensional array (with the specified keys)
+    to the given format.
+    """
+
+    all_rows = rows
+    if keys is not None and len(keys) > 0:
+        all_rows = [keys] + rows
+
+    if format == 'rows':
+        return twodim_to_rows(rows)
```
gyorb: Possible `TypeError` exceptions are not handled.

libcodechecker/arg_handler.py
```diff
@@ -134,7 +134,7 @@ def handle_server(args):
                              str(instance['port'])))
 
         print("Your running CodeChecker servers:")
-        util.print_table(rows)
+        print(util.twodim_to_table(rows))
```
gyorb: Possible `TypeError` exceptions are not handled.

libcodechecker/output_formatters.py
whisperity: Right now it could be that there isn't.

Once this PR is merged, the upcoming `checkers` and `analyzers` command will use this as their output format `choices`.

libcodechecker/output_formatters.py
whisperity: Expected to throw. It is the client code's responsibility to prepare a well-formatted twodim.

libcodechecker/cmd/cmd_line_client.py
whisperity: What do you mean as _format_ here? The `args.output_format`'s value?

libcodechecker/arg_handler.py
whisperity: Not possible. The twodim is well-formatted, with each row being a `2` or `3`-tuple. (Based on `instances_on_multiple_hosts`' value.)

libcodechecker/output_formatters.py
whisperity: But good catch for the `listrun` in `cmd`, will fix.

libcodechecker/cmd/cmd_line_client.py
gyorb: Yes, by accidentally making a typo like `jsson` it will be not a valid format and the exception will go back to the user, or is it handled somewhere else and I missed it?

gyorb:
The general idea is that the user should not get any exceptions just some error message which is easy to to understand.

libcodechecker/cmd/cmd_line_client.py
whisperity: `output_format` is a [_choices_](https://github.com/whisperity/CodeChecker/blob/10c8bb37b8c9a04b4fd1c0456c45a3efb15f7ad0/libcodechecker/cmd/cmd_line_client.py#L507) argument, which makes _argparse_ error if the choice is invalid.

whisperity:
The _end user_ (who is using CodeChecker) will not get any error messages because of this module.

The _user_ (as in, the _user_ code, client code, who uses `output_formatters`) will not get any error messages **if** the the person who writes the client code writes it in a way that it calls the methods in a way that they should be called.

Also, an error message which was borderline possible if calling to `print_table` (the predecessor of `twodim_to_table`) in a bad way (let's say the header is 3 column but every row is only 2 column) was `index out of range` which was updated to `one of the rows contain different columns than the rest`.

gyorb:
Ok, I see. I just did not remember that we check the choices given by the user in the [argument handler](https://github.com/whisperity/CodeChecker/blob/10c8bb37b8c9a04b4fd1c0456c45a3efb15f7ad0/libcodechecker/cmd/cmd_line_client.py#L482).

PR #614, opened 2017-03-20
Travis and auth cfg cleanup

gyorb:
* remove dummy authentication credentials in it enough it they are in the documentation
* cleanup configuration files in each travis run

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,5 +57,7 @@ addons:
     postgresql: "9.3"
 
 script:
+    - make clean_travis
     - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make travis; fi
     - make test
+    - make clean_travis
--- a/Makefile
+++ b/Makefile
@@ -38,14 +38,18 @@ package: build_dir gen-docs thrift
 	fi
 
 travis:
-	# travis specific changes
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -57,5 +57,6 @@ addons:
     postgresql: "9.3"
 
 script:
+    - make clean_travis
     - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make travis; fi
     - make test
```
Xazax-hun: I think we could also add a make clean at the end just to make sure that none of the scripts/command exit unexpectedly.

Makefile
```diff
@@ -41,6 +41,10 @@ travis:
 	# travis specific changes
 	scripts/change_clang_version.py $(CURRENT_DIR)
 
+clean_travis:
+	# cleanup codechecker config files
```
Xazax-hun: We should have full sentence comments in the makefile too.

.travis.yml
gyorb: I'm not sure what you mean here by the scripts/commands exit unexpectedly, but we can make some cleanup after the tests too.

Makefile
gyorb: Done.

PR #616, opened 2017-03-20
Find clang automatically if configured binary does not match

### LINKED ISSUE #578: autodetect installed clang version
autodetect installed clang version so the users do not have to manually edit the package layout config

whisperity:
If the user's `package_layout.json` configuration is missing, or it contains relative only binary names, so the binary is executed from the environmental `PATH`, but such binary is not found, an error **was** reported.

The rules to this new detection is:

 1. If the path is an absolute path, or a package-relative path, use that. If the binary doesn't exist, configuration error.
 2. If not walk the `PATH` and attempt to find the binaries:
    * for _clang-sa_: `clang`, `clang-3.8`, `clang++`, `clang++-3.8`, `clang-5.0`, etc.
    * for _clang-tidy_: `clang-tidy` and `clang-tidy-3.8`, `clang-tidy-5.0`, etc.
 3. Iteration is done in the `PATH` order, just as how the OSes do it.
 4. If multiple binaries are found in the PATH that match, the "newest", lexicographically (`clang` < `clang-3.8` < `clang-5.0`) is used.

Closes #578.


libcodechecker/analyze/analyzers/analyzer_clang_tidy.py
```diff
@@ -116,3 +117,41 @@ def construct_analyzer_cmd(self, res_handler):
         except Exception as ex:
             LOG.error(ex)
             return []
+
+    @classmethod
+    def resolve_missing_binary(cls, configured_binary, env):
```
Xazax-hun: This method is very similar for both analyzers. Maybe a good candidate to abstract out into a base class or utility function? There is no magic function in the python standard library to do this?

whisperity: No, there isn't. We attempted the `distutils.spawn.find_executable` but it can only match exact binary names.

Also right now it seems that the code is a duplicate, it _is_, but I would vouch **against** abstracting it out: it would create an extra layer of abstraction which doesn't really fit into the model we are using.

`BaseAnalyzer`
`RegexMatchableInstallNamedAnalyzer : public BaseAnalyzer`
`ClangSA : public RegexMatchableInstallNamedAnalyzer`

Any time there comes a new analyzer (we are still advertising CodeChecker to be compatible with many analyzers even though we only officially support clangsa and -tidy) which cannot be matched on any of the regex patters and we would break this chain. In my view, a new layer of abstraction here would only cause more clutter than how much imminent duplication it could solve.

whisperity: Though a utility could work. But not the base class thing. I'll do it into `util`.

gyorb: +1 for moving out the common parts to `util`.

README.md
```diff
@@ -140,32 +140,6 @@ cd ..
 Check your first project
 ------------------------
```
gyorb: I think we should write down that the static analyzers are auto detected, and the latest available version is used for the analysis and if somebody wants to use a specific analyzer (custom build), how to configure CodeChecker.

whisperity: 👍 I'll write up an understandable paragraph about this.

whisperity: I've reverted the removal and touched-up the wording of the sentences that already existed here. Also, rebased to newest master.

libcodechecker/analyze/analyzers/analyzer_types.py
```diff
@@ -86,9 +93,37 @@ def check_supported_analyzers(analyzers, context):
         if not analyzer_bin:
             LOG.warning('Failed to detect analyzer binary ' + analyzer_name)
             available_analyzer = False
-        elif not host_check.check_clang(analyzer_bin, check_env):
+        elif not os.path.isabs(analyzer_bin):
+            # If the analyzer is not in an absolute path, try to find it...
+            if analyzer_name == CLANG_SA:
+                found_bin = analyzer_clangsa.ClangSA.\
```
Xazax-hun: Can't we use the "current analyzer" somehow?

It is pretty strange that we have a common base class but instead of virtual calls we are dispatching manually almost always.

whisperity:
Definitely in a separate one. That line there is in such a way so that the style of the code keeps the style that is used elsewhere in the codebase.

PR #624, opened 2017-04-03
Update log command in the user guide

whisperity:
Follow-up for #610, I've extended the user guide with the new log command's documentation.


whisperity:
Yes, the "help" has been copied from the output verbatim.

Issue #445, opened 2016-11-11
Update plist test files for newer clang versions


gyorb:
The latest generated plist test files for the plist parser is from clang 3.8.
New plist test files should be generated for the newer clang versions:

* clang 3.9
* clang 4.0
* clang 5.0 trunk

Especially if the new notes feature will be available in clang #381.

gyorb:
Test file for `clang 3.9` is missing.

gyorb:
New test files were added, with notes and macro expansions too (available since clang v8).

Issue #20, opened 2015-07-28
Better skip file handling


igalex:
A "bug" should be skipped if its path's last point/step should be skipped according to a skip file (given by --skip command line option).

PR #628, opened 2017-04-10
Extend/refactor plist parser module

gyorb:
* Intoduce new types for Reports, Ranges, Positions, DiagSections

* The main event is handled separately from the other events in a
  bug path. Main event should be always available.

* DiagSections store diagnostic related informations from the
  analyzers (can be part of a path or a standalone information).
  Usually a position/range in the source file and optional message.

* Multiple hash generation options introduced for later versions.
  Old hash generation is still supported for backward compatibility.

* New unit test files for Clang 4.0 and 5.0 plist formats Resolves: #445


whisperity:
Is this for `5.8` or `6.0`? I wholeheartedly welcome the `bug` ↪️ `report` transition in our wording in the code, myself trying to keep to this mindset in the CLI refurbishing too.

libcodechecker/report.py
```diff
@@ -0,0 +1,445 @@
...
+        WARNING uses a different algorithm compared to obsolate hash.
+        """
+        if not self.__hash_value:
+            self.__hash_value = self.__gen_report_hash()
+
+        return self.__hash_value
+
+    def events(self):
+        """
+        FOR STORAGE BACKWARD COMPATIBILITY CLEANUP AFTER v6.0
```
whisperity: If _cleanup_ here means `please remove` then we should at least introduce a `,` in the text.
Also, I'd move for marking this as a `TODO` for more general IDE support on this comment.

libcodechecker/analyze/plist_parser.py
```diff
@@ -3,170 +3,165 @@
...
+                if item['kind'] == 'event':
+                    event = __handle_event(item, files)
+                    bug_path.add_diag(event)
+
+                elif item['kind'] == 'control':
+                    ctrl_sections = __handle_control(item, files)
+                    for cs in ctrl_sections:
+                        bug_path.add_diag(cs)
+                else:
+                    LOG.warning('Not supported item kind found: ' +
```
whisperity: This messages feels clunky on the output. Maybe `Unsupported item kind (VALUE) found during plist parsing`?

libcodechecker/report.py
```diff
@@ -0,0 +1,445 @@
...
+        string_to_hash = '|||'.join(hash_content)
+
+        return hashlib.md5(string_to_hash.encode()).hexdigest()
+
+    def __gen_report_hash(self):
+        """
+        Generate a report hash if the analyzer failed to generate
+        or not supported by the analyzer.
+        """
+        # FIXME NEEDS REVIEW WHAT TO INCLUDE IN THE HASH
```
whisperity: `:` after `FIXME`

libcodechecker/report.py
```diff
@@ -0,0 +1,445 @@
...
+        hash_content = [file_name,
+                        self.checker_name,
+                        msg,
+                        line_content,
+                        str(from_col),
+                        str(until_col)]
+
+        diag_sections = self.__bug_path.get_diag_sections()
+
+        # use only the contol diag sections and
```
whisperity: Not a full sentence.

libcodechecker/report.py
```diff
@@ -0,0 +1,445 @@
...
+        Contains all the control and event diag sections!
+        """
+        return self.__path_length
+
+    @property
+    def path_hash(self):
+        """
+        Generate a unique id for a bug path.
+        """
+        # TODO review hash generation
```
whisperity: Not a full sentence either.

gyorb:
This is for 5.8 and have some preparation for v6.0. This improvement is backward compatible with the current behavior.

libcodechecker/analyze/analyzers/result_handler_plist_to_db.py
```diff
@@ -58,48 +58,49 @@ def __store_bugs(self, files, bugs, connection, analisys_id):
...
+                    # Issue #20: this report is in a file which
+                    # should be skipped
+                    LOG.debug(report.hash_value + ' is skipped (in ' +
+                              events[-1].location.file_path + ")")
                     continue
 
             # Create remaining data for bugs and send them to the server.
             bug_paths = []
-            for path in bug.paths():
+            for path in report.paths():
```
Xazax-hun: I think there should be only one path for a bug. Paths probably contains different pieces of the path. Maybe this should be renamed as well to reflect this.

libcodechecker/report.py
```diff
@@ -0,0 +1,444 @@
...
+        self.__bug_path = bug_path
+
+    def get_path(self):
+        return self.__bug_path
+
+    @property
+    def path_hash(self):
+        if not self.__path_hash:
+            self.__path_hash = self.__bug_path.path_hash
+        return self.__path_hash
```
Xazax-hun: Why not just always return self.__bug_path.path_hash?

libcodechecker/report.py
```diff
@@ -0,0 +1,444 @@
...
+            self.__hash_value = self.__gen_report_hash()
+
+        return self.__hash_value
+
+    def events(self):
+        # TODO: FOR STORAGE BACKWARD COMPATIBILITY REMOVE AFTER v6.0
+        return [x for x in self.__bug_path.get_diag_sections() if
+                x.kind == "event"]
+
+    def paths(self):
```
Xazax-hun: I do not like the term paths. This is the pieces if one path.

libcodechecker/report.py
```diff
@@ -0,0 +1,444 @@
...
+class BugPath(object):
+    """
+    Represents a bug path which contains multiple DiagSections.
+    The order of the diagnostic sections in the bug path matters, and
+    needed for the storage in the database and for the UI to
+    render arrows properly.
+    """
+    def __init__(self):
+        self.__diag_sections = []
+        self.__path_length = 0
```
Xazax-hun: Why do we have a separate length? Wouldn't it be better to just use the length of the list?

libcodechecker/report.py
```diff
@@ -0,0 +1,444 @@
...
+    @property
+    def begin(self):
+        return self.__begin
+
+    @property
+    def end(self):
+        return self.__end
+
+
+class BugPath(object):
```
Xazax-hun: Does it woth to have a large wrapper over a list basically?

libcodechecker/report.py
```diff
@@ -0,0 +1,444 @@
...
+    If position_in_path is set this section belongs to a BugPath.
+    It is possible that a DiagSection is a standalone "message" or
+    "warning" by the analyzer and does not belong to BugPath.
+
+    See storage thrift api desctiption for kinds.
+    """
+
+    def __init__(self, kind='', msg=''):
+
+        self.__start_range = None
```
Xazax-hun: Why do we have a start range and an end range?
Shouldn't a diag section only contain one range? (start and end position)

libcodechecker/report.py
```diff
@@ -0,0 +1,444 @@
...
+        self.__end_range = None
+        self.__msg = msg
+        self.__extended_msg = ''
+        self.__kind = kind
+
+        # One Position.
+        self.__location = None  # Optional in some cases only range is
```
Xazax-hun: Can the location be separate from the beginning of the range? Isn't this redundant?

libcodechecker/analyze/analyzers/result_handler_plist_to_db.py
```diff
@@ -58,48 +58,49 @@ def __store_bugs(self, files, bugs, connection, analisys_id):
...
             # Skip list handler can be None if no config file is set.
             if self.skiplist_handler:
                 if events and self.skiplist_handler.should_skip(
-                        events[-1].start_pos.file_path):
-                    # Issue #20: this bug is in a file which should be skipped
-                    LOG.debug(bug.hash_value + ' is skipped (in ' +
-                              events[-1].start_pos.file_path + ")")
+                        events[-1].location.file_path):
+                    # Issue #20: this report is in a file which
+                    # should be skipped
```
whisperity: Are we sure we need to refer issues that has already been closed here? I'm all in for keeping a little explanation, but no `Issue #20`. (Or if it remains, at least put a proper link there.)

libcodechecker/analyze/analyzers/result_handler_plist_to_db.py
gyorb: This is kept this way to be backward compatible with the current way of storing path information. It will be changed in the upcoming schema rewriting.

libcodechecker/report.py
gyorb: This will be removed in the upcoming refactoring, kept here for backward compatibility with the current way of storing the path elements.

libcodechecker/report.py
gyorb: The control (and some event) diag sections have a start and an end range not only a position. A position is only one point in the source and a range is made of two positions. With the extra information we can provide better and more precise report visualization (better highlighting for the important parts).

libcodechecker/report.py
gyorb: It is possible that the location is not the same as the range. For events only location is available in most cases only for some specific events is a range available.

libcodechecker/report.py
gyorb: I wanted to create some separation from the path independent diagnostic sections ...
I will refactor this, so the path can be built only once and no element can be added later which might change the path hash and move the path hash generation here.

libcodechecker/analyze/analyzers/result_handler_plist_to_db.py
gyorb: This was not my comment I just moved it. I will change it and add some TODO because in v6.0 this checking should be done based on the main diagnostic section.

libcodechecker/analyze/analyzers/result_handler_plist_to_db.py
Xazax-hun: I don't see why renaming this would be backward incompatible.

libcodechecker/report.py
Xazax-hun: Could you point me to an example? It is really hard to imagine to have two ranges in a control/event.

libcodechecker/report.py
Xazax-hun: Maybe it would be worth to have a catalog of these diagnostic pieces, like a list of example plist fragments about the cases that we would like to cover.

libcodechecker/report.py
```diff
@@ -0,0 +1,434 @@
...
+    @property
+    def start_range(self):
+        return self.__start_range
+
+    @start_range.setter
+    def start_range(self, range):
+        self.__start_range = range
+
+    @property
+    def end_range(self):
```
Xazax-hun: do we want to have getters/setter for these trivial members?

gyorb:
I will create a new pull request.

Issue #579, opened 2017-02-16
better error handling at listing checkers


gyorb:
if one of the analyzer binaries are not found exception is thrown which should be handled

PR #581, opened 2017-02-16
Better error handling at checker listing

gyorb:
handle exceptions if analyzers are missing

PR #629, opened 2017-04-11
Add analyzer listing and detail view

whisperity:
Added a new command `codechecker-analyzers` to list the available analyzers and their version strings in a machine or human-readable format.

`CodeChecker analyzers` will simply list the **available** analyzer names (in the format as they are arguments to `--analyzers` in `check`, `quickcheck`, etc.)

Extra information is queryable via `--details`, all supported analyzers (those which are supported by CodeChecker but not found on the system) can be get by `--all`. Using the improvements from #613, the result is queryable in every "twodim" format introduced.

 * Closes: #579 (as `CodeChecker check` will no longer attempt to invoke nonexistent binaries)
 * Supersedes: #581


libcodechecker/analyzers.py
```diff
@@ -0,0 +1,124 @@
...
+    if args.output_format not in ['csv', 'json']:
+        if not args.details:
+            header = ['Name']
+        else:
+            header = ['Name', 'Path', 'Version']
+    else:
+        if not args.details:
+            header = ['name']
+        else:
+            header = ['name', 'path', 'version_string']
```
Xazax-hun: Why the distinction in header names?

whisperity: If `--details` is not specified, only the analyzer names are listed. If specified, we add some extra fields.

Xazax-hun: I mean the other distinction (lower vs upper case)?

whisperity: Oh, you meant the other conditional. Basically it is there so output would be nice for the user, while keeping a machine-readable format in csv and json (so keys are lowercase, no space but `_`, etc.) I think it's better if we make the distinction here instead of using some obscure `util` call which automatically transforms (later on, maybe in an uninteded way?) human-readable header names to machine-readable keys.

libcodechecker/analyzers.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+
+    context = generic_package_context.get_context()
+    working, errored = \
+        analyzer_types.check_supported_analyzers(
+            analyzer_types.supported_analyzers,
+            context)
+
+    if args.output_format not in ['csv', 'json']:
+        if not args.details:
```
Xazax-hun: Could your remove the `not` and swap the branches? At the other places too.

whisperity: I'd offer a counterargument here: imagine reading the code. `--details` being specified is the _special_ and not the _general_ case &ndash; if you are searching in the code and skimming it with your eyes, having to find the _general_ case would, after a swap, require potentially a lot (here only a few lines, but otherwise potentially a lot) of LOC to skip over.

Same with the outside `if` (about `output_format`). `json` and `csv` is the more special cases here. I think it's more logical to have the _generic_ cases evaluated first.

PR #631, opened 2017-04-13
Fix escaping multiword string literals, once again

### LINKED ISSUE #505: Parsing string values in -D compiler option

whisperity:
Consider the following file:

```c++
int main()
{
  if ( VARIABLE == "a b" )
  {
    int *p = 0;
    *p = 42;
  }
}
```

built with `g++ -DVARIABLE='"a b"' main.cpp`.

# Old behaviour

## LD-LOGGER
If we use _LD_LOGGER_, the resulting output will show no errors, as the string in the `build.json` is `"-DVARIABLE="a b""` (one escape level higher, so after stripping the escapes put in by the JSON serializer!) and the options parser will understand this entry as a _file_ (instead of a _compiler option_) in [`log_parser.py#130`](https://github.com/Ericsson/codechecker/blob/d3e53ff1254509770a7cc535a4a58b53bd9be31c/libcodechecker/analyze/log_parser.py#L130). This way, the define never reaches into the analyzer invocation &mdash; thus the macro expansion is never made.

## Newest intercept-build
The same issue arises with _intercept-build_ **newest version**, as in #610, a makeshift fix was introduced to make the parsed `argument` vector appear as a `command` string in the same way. Removing this makeshift fix would result in the `option_parser` making `-DVARIABLE="a` a command-line argument and `b" main.cpp` the file, making the analysers uncallable.

## intercept-build in clang-5.0
The old _intercept-build_, currently shipping with **upstream** clang also created a `command` format, seen below, which botched analyzer calls, as the part of the define after a space was considered a file input:

```
"command": "c++ -c -DVARIABLE=\\\"a b\\\" -o main.out main.cpp",
```

```
compiler options: -DVARIABLE=\a b\
linker options: 
files: variable.cpp
```

resulting in invocation `clang [...] -x c++  -DVARIABLE=\a b\ /home/user/main.cpp` to which clang said `no such file or directory: 'b /home/user/main.cpp'`, and clang-tidy said `No escaped character`.

This behaviour was applicable even on `master` before #610 had been introduced.

# New behaviour

This pull-request fixes the escape issues once more, agnostic to the logger tool used. The command-line defines (or defines in the `argument` vector) are now properly passed to the analyzers and `clang-tidy` happily reports

```
main.cpp:4:17: both side of operator are equivalent [misc-redundant-expression]
  if ( VARIABLE == "a b" )
                ^

clang-tidy found 1 defect(s) while analyzing main.cpp

[15:44] - [2/2] clang-tidy analyzed main.cpp successfully.
```

showing that the macro is defined properly now.

This change now properly handles arguments with quotes in them, as seen above.

 * If the log has been created by the old _LD-LOGGER_, no functional change, the define does not reach the call.
 * _LD-LOGGER_ introduced in the current patch (which esentially reverts relevant parts of #109) will now properly passes the defines into the logfile and to the analyzers.
 * Using upstream [_intercept-build_](http://github.com/rizsotto/scan-build-py), the defines are also properly passed.
 * Using _intercept-build_ currently shipping with clang also works properly now.

This MR should close #505.

### DIFF
--- a/libcodechecker/analyze/log_parser.py
+++ b/libcodechecker/analyze/log_parser.py
@@ -106,24 +106,20 @@ def parse_compile_commands_json(logfile, add_compiler_defaults=False):
             continue
 
         action = build_action.BuildAction(counter)
-
         if 'command' in entry:
             command = entry['command']
+
+            # Old versions of intercept-build (confirmed to those shipping
+            # with upstream clang-5.0) do escapes in another way:
+            # -DVARIABLE="a b" becomes -DVARIABLE=\"a b\" in the output.
+            # This would be messed up later on by options_parser, so need a
+            # fix here. (Should be removed once we are sure noone uses this
+            # intercept-build anymore!)
+            if r'\"' in command:
+                command = command.replace(r'\"', '"')
         elif 'arguments' in entry:
             # Newest versions of intercept-build create an argument vector
             # instead of a command string.
-            for i in range(0, len(entry['arguments'])):
-                arg = entry['arguments'][i]
-                if ' ' in arg:
-                    # If there is an argument with a space in it, the join
-                    # below will mess the invocation up. (-DVAR=va lue will be
-                    # passed as -DVAR=va)
-                    #
-                    # Build.json created by ld-logger escapes these strings
-                    # and they never reach the analyser later on.
-                    #
-                    # TODO: Better handle this. (See issue #505.)
-                    entry['arguments'][i] = '\"' + arg + '\"'
             command = ' '.join(entry['arguments'])
         else:
             raise KeyError("No valid 'command' or 'arguments' entry found!")
@@ -131,6 +127,7 @@ def parse_compile_commands_json(logfile, add_compiler_defaults=False):
 
         action.original_command = command
         action.analyzer_options = results.compile_opts
+
         action.lang = results.lang
         action.target = results.arch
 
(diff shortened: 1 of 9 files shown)


whisperity:
Where or how should I add the test? The problem with automatic testing for this is that, as discussed in the PR, different behaviours require very specific combinations in the environment.

whisperity:
DO **NOT** MERGE, I'm updating the tests. The current test cases found in [`tests/unit/test_buildcmd_escaping.py`](https://github.com/whisperity/CodeChecker/blob/cd8c249b84f438812665bc7de818e7a5cc861a87/tests/unit/test_buildcmd_escaping.py) cannot reproduce this problem.

whisperity:
Tests have been added to all four cases described in the original summary.

PR #632, opened 2017-04-13
Fix various --add-compiler-defaults errors

whisperity:
* From `#define __VERSION__ "5.4.0 20160609"`, `20160609` would be understood as a file or folder name, resulting in clang erroring away.

* Newer versions of GCC/G++ define `__has_include_next` to `__has_include_next__` which is not understood by Clang, as it supplies its **own** `__has_include_next` internally and does **not** define the postfix `__` version.
Thus, such macro definition retrieved from GCC &mdash; which would inevitably lead to a `undefined macro` error &mdash; should not be passed to clang.

### DIFF
--- a/libcodechecker/analyze/log_parser.py
+++ b/libcodechecker/analyze/log_parser.py
@@ -67,9 +67,27 @@ def get_compiler_defines(compiler):
                                     stderr=subprocess.PIPE)
             out, err = proc.communicate("")
             for line in out.splitlines(True):
-                LOG.debug("define:"+line)
+                LOG.debug("define: " + line)
                 define = line.strip().split(" ")[1:]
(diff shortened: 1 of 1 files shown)


libcodechecker/analyze/log_parser.py
```diff
@@ -67,9 +80,30 @@ def get_compiler_defines(compiler):
...
             for line in out.splitlines(True):
-                LOG.debug("define:"+line)
+                LOG.debug("define: " + line)
                 define = line.strip().split(" ")[1:]
-                d = "-D"+define[0] + '=' + '"' + ' '.join(define[1:]) + '"'
+
+                variable = define[0]
+                value = ' '.join(define[1:])
+
+                if variable in IGNORED_DEFINES:
```
whisperity: This should be the other way around, as `__has_include_next` must match `__has_include_next(STR)`, not swapped.

PR #634, opened 2017-04-19
Refurbished checker listing and detail view 

whisperity:
New, extended and refurbished command for available checker listing

Splitting "old-style" and  "new-style" invocation
==================================

`CodeChecker checkers` as of now is most likely used by many of our users &ndash; yours truly included! &ndash; in wrapper scripts to quickly generate specific analyzer enable and disable args for invocation. Due to this, I did not wish to break this interface with the upcoming releases.

Starting from the acceptance of this MR up until a later release, the principle introduced in #610 is somewhat stabbed and broken: in the context of this new command, `CodeChecker checkers` **will NOT have the same behaviour as** `codechecker-checkers`. Code parts related to this behaviour (essentially just *HACK*s) are marked for removal once `later release` rolls.

`CodeChecker checkers` retains its old behaviour.
Issue #579 will be solved by #629.
A warning is implemented as unexecuted code, so users can later be notified about the changes upcoming in a later release.

New `codechecker-checkers` command
=============================

For easier use and better pipeability, the new `checkers` command works in the following way:

 * Initially, all checkers are listed, one by one.
 * The user can request to `--only-disabled` and `--only-enabled` for filtering.

 * By default, only names are printed.
 * Similar to `CodeChecker analyzers`, a `--details` view is available, complete with name, analyzer, description, enabled status
   * similarly, `--output json, csv, ...` is present here too.
 * Just like the "old" `CodeChecker checkers` command, checkers can be filtered by-analyzer.

----
> Depends: #629


gyorb:
I would recommend to change this behavior
> Initially, only the enabled (based on the config file) checkers are listed.

to that all the available checkers are listed first, so the users can see the full list and choose what to enable or disable. It can be misleading if we print out only a limited set of the available checkers, users might think that the enabled list is the full list.

bin/CodeChecker.py
```diff
@@ -193,6 +193,21 @@ def signal_handler(sig, frame):
...
 
+        # TODO: Delete these once v6.0 rolls.
+        old_subcommands = []
+
+        def _warn_deprecated_command(cmd_name):
+            # Write to stderr so the output is not captured by pipes, e.g.
+            # with the "checkers" command, the "-" in the new command's name
+            # would mess up pipe usage.
+            sys.stderr.write(
+                "[WARNING] The called command 'CodeChecker {0}' is DEPRECATED "
```
gyorb: I do not think we should warn the users before v6.0 is released. After the v6.0 release we should inform them that a new option is available and the old might be removed in one of the later releases.

docs/user_guide.md
```diff
@@ -477,16 +477,68 @@ optional arguments:
...
+## 4. `checkers` mode
 
-List all available checkers.
+List the checkers available in the installed analyzers which can be used when
+performing an analysis.
 
-The ```+``` (or ```-```) sign before a name of a checker shows whether the checker is enabled (or disabled) by default.
+### Old `checkers` command (before `v6.0` release)
+
+**WARNING!** Usage of this command is deprecated and with the release of `v6.0`
```
gyorb: The same warning after v6.0 is out.

docs/user_guide.md
```diff
@@ -477,16 +477,68 @@ optional arguments:
...
+The ```+``` (or ```-```) sign before a name of a checker shows whether the
+checker is enabled (or disabled) by default.
 
 ~~~~~~~~~~~~~~~~~~~~~
 CodeChecker checkers
 ~~~~~~~~~~~~~~~~~~~~~
 
+### New `checkers` command (since `v5.8` release)
+
+By default, `codechecker-checkers` will list only the *enabled* checkers, one
```
gyorb: As I've already mentioned by in one of my comments we should list all the checkers by default, not just the enabled ones.

docs/user_guide.md
```diff
@@ -477,16 +477,68 @@ optional arguments:
...
+  --show-disabled       Do show the disabled checkers. (default: False)
+  -o {rows,table,csv,json}, --output {rows,table,csv,json}
+                        The format to list the applicable checkers as.
+                        (default: rows)
+  --verbose {info,debug,debug_analyzer}
+                        Set verbosity level. (default: info)
+
+The list of checkers that are enabled of disabled by default can be edited by
+editing the file
+'(install-dir)/CodeChecker/config/config.json'.
```
gyorb: We should mention it might be overwritten if the package is reinstalled!

docs/user_guide.md
whisperity: Please clarify, I don't understand what you mean here.

whisperity:
I've done the asked changes, updated the documentation, the MR's code, and also fixed merge conflicts introduced.

docs/user_guide.md
gyorb: I meant to remove this warning from the user guide too before v6.0 is released. We will add it later when the new v6.0 is out and we see how we roll out the old command line or we keep compatibility.

bin/CodeChecker
```diff
@@ -61,6 +61,14 @@ def run_codechecker(checker_env, subcommand=None):
         # wrapper entry point, add it to the command list).
         # E.g. 'codechecker-log ...' is the same as 'CodeChecker log ...'.
         checker_cmd.append(subcommand)
+    else:
+        # For some commands the "legacy" old implementation is different,
+        # such as the case with 'CodeChecker checkers' instead of
+        # 'codechecker-checkers'. For this, we use a little HACK here through
+        # environmental variables.
+        #
+        # TODO: Delete this once "old" commands are eradicated in v6.0 release.
```
gyorb: I would remove the exact version number from the comment because the version where we clean this up might change.

whisperity: Very small oversight. :)

(I wish GitHub would not mark only the PR as "changes requested" but the commits which introduced the changes that are reviewed and commented on too, so such oversights would be pointed out easier for the patch submitter.)

PR #635, opened 2017-04-21
Refactor plist parser module

gyorb:
The plist parser does only some minimal parsing and result processing:
  * every data in the plist files is kept in dictionaries
  * split the big dictionary into two smaller in the Report
   - main (main dictionary with checker name, report hash ...)
   - bug path (only bug path related items)
  * if checker name is missing try to find out based on the checker
    description
  * if the report hash is missing try to generate based on the
    available information in the report

Add new tests and test files for newer clang versions.


libcodechecker/analyze/plist_parser.py
```diff
@@ -3,170 +3,185 @@
...
-import plistlib
-from xml.parsers.expat import ExpatError
-
-from libcodechecker.analyze import plist_helper
-from libcodechecker.logger import LoggerFactory
-
-LOG = LoggerFactory.get_new_logger('PLIST_PARSER')
+With the newer clang releases more information is available in the plist files.
 
+* Before Clang v3.7:
```
Xazax-hun: Clang 3.7 was released in 2015 Sept. Do we want to support the earlier versions? (Clang 3.6 was released in 2015 feb.), 3.8 is available in ubuntu 16.04 LTS.  And 3.8 is also available in the 14.04 backports repository.

libcodechecker/report.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # Dictionary containing bug path related data
+        # with control, event ... sections.
+        self.__bug_path = bugpath
+
+    @property
+    def main(self):
+        return self.__main
+
+    def print_main(self):
```
Xazax-hun: Do we need both `print_main` and `__str__`?
I'd rather remove the former, since it is not flexible (can not be used with log levels).

gyorb: I used the `print_main` only for debugging and I left it there, but I've removed it now.

libcodechecker/analyze/plist_parser.py
gyorb: I think right now it is not too much of an overhead to support plists generated by older clang versions (name detection code and hash generation). The hash generation part should be kept later for backward compatibility. The name detection is just a few simple functions so I think we should keep it, at least for the next few releases. In the upcoming v6 there will be multiple changes and we can review if we should support plists generated by older analyzer versions.

PR #636, opened 2017-04-24
Rename authentication config files to .codechecker.foobar.json format

whisperity:
Instance configuration is saved already as `~/.codechecker.instances.json`. Due to having the session token and credential file under `~`, there is no use to have the username in the file. This way, all CodeChecker files are in line with each other, in terms of the name format in the users' home.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -223,7 +223,7 @@ For browser authentication to work, cookies must be enabled!
 
 The `CodeChecker cmd` client needs to be authenticated for a server before any data communication could take place.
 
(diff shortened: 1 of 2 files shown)


gyorb:
* What if, the user wants to log in with multiple user names?
* The configuration file names are changed, so the users need to login/authenticate again to generate the new config files. What happens to the old files? Manual cleanup is required?
* We might consider to automatically rename the old file names to the new.

whisperity:
If the user would log in with another username, the files would already be broken. `username` in this context means the username on the operating system, **not** the username used while logging in to CodeChecker.

The old files at this point are left behind, yes. I'll add a "rename" step aptly marked as a migration between versions.

PR #639, opened 2017-04-25
New user commands for analysis

### LINKED ISSUE #307: --enable/--disable all option
To be able to quickly run the tests that you want this would be nice. Right now there is no way to see a nice list of which checkers you have enabled. You have to count them yourself.

### LINKED ISSUE #436: Command line argument to disable or enable all checkers
Add an command line argument which implies that all checkers are disabled but explicitly enabled ones.

This would help to see which checkers are actually used, just by looking at the command issued.

This argument in combination of the issue 427 https://github.com/Ericsson/codechecker/issues/427  would make the configuration file easy to read.

Enabling all checkers is probably not that useful, as there are always experimental checkers in the build and those will in many cases cause problems.

### LINKED ISSUE #586: More general import results command
Right now the import/export functionality is limited to plist (command line argument).
We could introduce a more general import command line with sub arguments for format, which can be extended more easily later.

Something like this:
`CodeChecker import --format plist -w ~/workspace ...`

### LINKED ISSUE #588: Analysis results without database storage
Currently the results are continuously processed and automatically stored to the database.
We should support that the user do not want to store the analysis results and we just produce plist or other result output files. (Later can be imported/stored into a database.)

### LINKED ISSUE #591: new store results command
separate command and file under bin to store results (import results feature)

The current CodeChecker plist import command should use this.

### LINKED ISSUE #592: new analyze project command
Create new analyze command under bin.
It should just run the analysis and do not store the results.

The current CodeChecker check command should use this.

whisperity:
This MR introduces the new, more split analysis infrastructure commands `analyze`, `parse` and `store`.

# New commands
## `codechecker-analyze`
Using one or more `build.json`s created in the previous step (#610, `codechecker-log`), create output plists to an output folder (or `~/.codechecker/reports`) as default. This command takes all the arguments related to analysis: analyzer configuration, skipfile, checker list.

> Supersedes #556.
> Closes #588, closes #592

`--enable-all` has been introduced: specifying `--enable-all` will automatically enable every checker, except `alpha.` and `debug.`. The list of checkers can still be fine-tuned later on, i.e. `--enable-all -e alpha -d core` does what you would expect.

> Closes #307, closes #436

## `codechecker-parse`
Taking one or more folders containing input `plist`s, `parse` prints analysis results to the standard output.

## `codechecker-store`
Taking one or more folders containing input `plist`s, `store` stores analysis data into the database specified. 

> Closes #586, closes #591

## `metadata.json` files
Metadata JSONs are created by `analyze` in the output folder. These files contain crucial persistent information about an analysis (as from now on, the analyse and the storage parts are **entirely** disconnected!), such as the original analysis command, the duration, the number of succeeded and failed files. In addition to this, the list of checkers that were actually used in the analysis are also saved here. Using these new files, a `store` (formerly `CodeChecker plist`) can properly show the original `analyze` invocation in the web browser, and we can properly show in `parse` which file was analyzed that resulted in the plist 

Metadata JSONs are expected to be an internal thing, not meant to be read or edited by users. They are used to carry information from `analyze` to `store`/`parse`.

# Changes
`codechecker-log --quiet-build` has been introduced.

## `check` and `quickcheck`
 *  `quickcheck = log + analyze + parse`
 * `check = log + analyze + store`

The old `quickcheck` and `check` commands have been cleaned up and turned into wrappers of the aforementioned command chains. The user invocations of `quickcheck` and `check` **did not change at all** (excluding some better help messages), previous _deprecated_ arguments are still there. Arguments are transformed dynamically to corresponding arguments of the 3-long call chain.

Argument option token "discrepancies" between the wrapper's interface and the actual new commands' interfaces are marked accordingly, with a `TODO:`.

## `plist`

The old `plist` command has also been dropped in favour of a wrapper. The wrapper wraps over `parse` and `store`, depending on the "output" of `plist`, i.e. whether `--stdout` was specified, or not.

The invocation didn't change here either. This also means that "strange" restrictions about the invocation (such as the need for `--name` even if `--stdout` is given...) are present still.


docs/user_guide.md
```diff
@@ -336,58 +688,109 @@ void test() {
...
+                        files not affected by the analysis, and only
+                        incrementally update defect reports for source files
+                        that were analysed.) (default: False)
+  --verbose {info,debug,debug_analyzer}
+                        Set verbosity level. (default: info)
 
-For example:
+database arguments:
+  --sqlite SQLITE_FILE  Path of the SQLite database file to use. (default:
+                        {HOME_DIR}/.codechecker/codechecker.sqlite)
```
gyorb: Use `/home/<username>/` instead of `{HOME_DIR}` to be consistent.

libcodechecker/analyze.py
```diff
@@ -0,0 +1,290 @@
...
+                               action='store_true',
+                               default=False,
+                               required=False,
+                               help="Retrieve compiler-specific configuration "
+                                    "from the analyzers themselves, and use "
+                                    "them with Clang. This is used when the "
+                                    "compiler on the system is special, e.g. "
+                                    "when doing cross-compilation.")
+
+    # TODO: One day, get rid of these. See Issue #36, #427.
```
gyorb: I'm not a big fan of issue numbers in the comments.

libcodechecker/analyze.py
```diff
@@ -0,0 +1,290 @@
...
+    analyzer_opts.add_argument('--saargs',
+                               dest="clangsa_args_cfg_file",
+                               required=False,
+                               default=argparse.SUPPRESS,
+                               help="File containing argument which will be "
+                                    "forwarded verbatim for the Clang Static "
+                                    "analyzer.")
```
gyorb: analyzer with an upper case `a`

libcodechecker/analyze.py
```diff
@@ -0,0 +1,290 @@
...
+            LOG.error("The specified logfile '" + log_file + "' does not "
+                      "exist!")
+            continue
+
+        actions += log_parser.parse_log(log_file,
+                                        args.add_compiler_defaults)
+
+    if 'enable_all' in args:
+        # enable_all is handled in
+        # libcodechecker/analyze/analyzers/analyzer_types.py.
```
gyorb: Should we leave this type of comments, what if the filename changes or it will be moved into another module?

libcodechecker/analyze.py
```diff
@@ -0,0 +1,290 @@
...
+        LOG.info("'--enable-all' was supplied for this analysis.")
+
+    # Run the analysis.
+    args.output_path = os.path.abspath(args.output_path)
+    if os.path.isdir(args.output_path):
```
gyorb: We should print some info message to the user, that the previous analysis results will be removed.

libcodechecker/analyze.py
```diff
@@ -0,0 +1,290 @@
...
+
+    LOG.debug("Output will be stored to: '" + args.output_path + "'")
+
+    metadata = {'action_num': len(actions),
+                'command': sys.argv,
+                'versions': {
+                    'codechecker': "{0} ({1})".format(context.package_git_tag,
+                                                      context.package_git_hash)
+                },
+                'working_directory': os.getcwd(),
```
gyorb: Should we use the current directory here? Some temporary directory wouldn't be better with the `tempfile` module?

libcodechecker/analyze.py
```diff
@@ -0,0 +1,290 @@
...
+                'output_path': args.output_path}
+
+    if 'name' in args:
+        metadata['name'] = args.name
+
+    if 'skipfile' in args:
+        # Skip information needs to be saved because reports in a header
+        # can only be suppressed by the report-server used in 'store' later
```
gyorb: Here the skip and the suppress functionality is mixed in the comment. Skip will not store the results, with suppress the results will be stored but suppressed.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -176,23 +198,13 @@ def signal_handler(*arg, **kwarg):
 
     signal.signal(signal.SIGINT, signal_handler)
 
-    # Remove characters which could cause directory creation problems.
-    no_spec_char_name = re.sub(r'[^\w\-_. ]', '_', args.name)
```
gyorb: Do we handle this somewhere else?

libcodechecker/analyze/analyzer.py
```diff
@@ -69,97 +68,44 @@ def _get_skip_handler(args):
...
-                                             context,
-                                             enabled_analyzers)
-
-    __print_analyzer_version(context, analyzer_config_map)
-
-    LOG.info("Static analysis is starting ...")
-
-    analysis_manager.start_workers(args, actions, context, analyzer_config_map,
-                                   _get_skip_handler(args), False)
+    metadata['duration'] = end_time - start_time
```
gyorb: We should store the start and the end time separately. The new db schema will store them separately.

libcodechecker/check.py
```diff
@@ -0,0 +1,515 @@
...
+    parser.add_argument('--keep-tmp',
+                        action=DeprecatedOptionAction)
+
+    parser.add_argument('-c', '--clean',
+                        action=DeprecatedOptionAction)
+
+    parser.add_argument('--update',
+                        action=DeprecatedOptionAction)
+
+    # TODO: --name is not required in 'store'.
```
gyorb: I think we might want to keep the `--name`. The user might want to store the results with a different name which was used for `analysis` or the plist results are from a scan-build output where not metadata values are available.

libcodechecker/check.py
```diff
@@ -0,0 +1,515 @@
...
+                                    "are: " +
+                                    ', '.join(analyzer_types.
+                                              supported_analyzers) + ".")
+
+    analyzer_opts.add_argument('--add-compiler-defaults',
+                               action='store_true',
+                               default=False,
+                               required=False,
+                               help="Retrieve compiler-specific configuration "
+                                    "from the analyzers themselves, and use "
```
gyorb: "from the compilers" instead of "from the analyzers"

libcodechecker/check.py
```diff
@@ -0,0 +1,515 @@
...
+                                    "them with Clang. This is used when the "
+                                    "compiler on the system is special, e.g. "
+                                    "when doing cross-compilation.")
+
+    # TODO: One day, get rid of these. See Issue #36, #427.
```
gyorb: I'm not a big fan of issue numbers in the source code.

libcodechecker/parse.py
```diff
@@ -0,0 +1,175 @@
...
+    parser.set_defaults(func=main)
+
+
+def parse(f, context, metadata_dict, suppress_handler, steps):
+    """
+    Prints the results in the given file to the standard output in a human-
+    readable format.
+    """
+
+    if not f.endswith(".plist"):
```
gyorb: I think we should inform the user that only files with `.plist` extension will be parsed.

libcodechecker/plist.py
```diff
@@ -0,0 +1,332 @@
...
+                "store the results."
+    }
+
+
+def add_arguments_to_parser(parser):
+    """
+    Add the subcommand's arguments to the given argparse.ArgumentParser.
+    """
+
+    # TODO: --name is not required in 'store'.
```
gyorb: As I mentioned before we should keep the `--name` in store.

libcodechecker/store.py
```diff
@@ -0,0 +1,436 @@
...
+                              "arguments\" section on how to configure the "
+                              "database connection.")
+
+    pgsql = parser.add_argument_group("PostgreSQL arguments",
+                                      "Values of these arguments are ignored, "
+                                      "unless '--postgresql' is specified!")
+
+    # WARNING: '--dbaddress' default value influences workspace creation
+    # in SQLite.
+    pgsql.add_argument('--db-address',
```
gyorb: I was thinking about the `--db-host` argument name for the new `store`, instead of `--db-address`. What do you think?

libcodechecker/store.py
```diff
@@ -0,0 +1,436 @@
...
+
+    LOG.debug("Parsing input file '" + f + "'")
+
+    buildaction = build_action.BuildAction()
+    if os.path.basename(f).startswith("clangsa_"):
+        buildaction.analyzer_type = analyzer_types.CLANG_SA
+    elif os.path.basename(f).startswith("clang-tidy_"):
+        buildaction.analyzer_type = analyzer_types.CLANG_TIDY
+
+    buildaction.original_command = "IMPORTED"
```
gyorb: Do we store the original build command as a meta information?

libcodechecker/check.py
whisperity: The `--name` argument will be kept. I meant that it's not `required`. Previously, `check` had `required=True` for `--name`, but `store` does not, as `store` _can_ be invoked without specifying `--name`, as it will _try_ to figure a name out for the run... and if it can't, it shown an error and asks the user to explicitly specify `--name`.

libcodechecker/plist.py
whisperity: Same answer as before. `--name` **is** an argument for `store`, with `required=False`. But to keep previous invocation consistent, `--name` is `reuqired=True` in `plist` and `check`. (Though in case of `plist` it's a mistake, see last two paragraphs of the patch description.)

libcodechecker/store.py
whisperity: No, we don't. Only the `codechecker-analyzer` invocation command-line is kept... as the build-jsons could come from anywhere and in an arbitrary number.

libcodechecker/analyze.py
whisperity: This information saves in which folder the `CodeChecker analyze` was invoked. We discussed it earlier that this information is needed, but I don't recall the reason right now.

docs/user_guide.md
```diff
@@ -232,81 +451,214 @@ CodeChecker will get the hardcoded values for the compilers set in the
...
+~~~~
+CodeChecker analyze mylogfile.json --saargs extra_compile_flags.txt -n myProject
+~~~~
+
+Where the `extra_compile_flags.txt` file contains additional compilation
+options, for example:
 
-    CodeChecker check --saargs extra_compile_flags -n myProject -b "make -j4"
+~~~~~
+-I~/include/for/analysis -I$(MY_LIB)/include -DDEBUG
```
whisperity: I'm not sure what this part is about. I just aligned the documentation to have a better wording, but am using the same examples from the previous version.

libcodechecker/analyze/analysis_manager.py
whisperity: This was only used to create a temporary folder, identified by the run-name, under the workspace.
Because the analysis results, from now on, are directly saved into a folder given by the user, it is no longer needed. The default value is `~/.codechecker/reports`, which is safe, and if the user specifies a custom one, it will be earlier an error, when the output folder is attempted to be created.

libcodechecker/analyze.py
whisperity: Done.

libcodechecker/parse.py
whisperity: Done.

libcodechecker/analyze.py
gyorb: Where do we use this information in the new changes? It is exported out but do we use it?

libcodechecker/plist.py
gyorb: Ok, I understand, it was a bit misleading to me based comment what should be done here. Maybe some more descriptive TODO message would be better, that it should not be a required argument later.

libcodechecker/analyze.py
whisperity: Uhm... I'd like to apologise, the relevant code parts must have been annihilated by a bogus merge. I'm adding them right now.

libcodechecker/plist.py
whisperity: I have elaborated a bit on the `TODO` to make it more understandable.

docs/user_guide.md
```diff
@@ -232,81 +451,214 @@ CodeChecker will get the hardcoded values for the compilers set in the
...
+be edited manually, instead, a CodeChecker viewer server should manage them,
+but the suppress file could be committed into the source code repository for
+distribution to teammates.
+
+If `--suppress` is specified, `parse` will use the given suppress file and
+will not show reports on the standard output which are marked as suppressed.
+
+### Suppression in the source code
+
+Suppress comments can be used in the source files to suppress specific or all
```
whisperity: I'd rather implement this feature as part of a later pull request. This one is already huge enough.

gyorb: Yes, it can be a separate pull request, I've just created an issue so we won't forget it.

PR #648, opened 2017-05-03
configurable test projects directory

gyorb:
Test projects directory can be configured in the Makefile.local
for different test environments.

### DIFF
--- a/tests/libtest/env.py
+++ b/tests/libtest/env.py
@@ -91,6 +91,10 @@ def repository_root():
     return os.path.abspath(os.environ['REPO_ROOT'])
 
 
+def test_proj_root():
+    return os.path.abspath(os.environ['TEST_PROJ'])
+
+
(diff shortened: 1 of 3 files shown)


tests/libtest/env.py
```diff
@@ -91,6 +91,10 @@ def repository_root():
     return os.path.abspath(os.environ['REPO_ROOT'])
 
 
+def test_proj_root():
+    return os.path.abspath(os.environ['TEST_PROJ'])
```
bruntib: Didn't you mean `TEST_PROJECT` here?

gyorb: No. In the Makefile the `TEST_PROJ` environment variable will be set [here](https://github.com/Ericsson/codechecker/blob/c00b352804a99029d89c222634d4be7bba2db5ef/tests/Makefile#L13). `TEST_PROJECT` is a Makefile only variable name

PR #654, opened 2017-05-09
Speed up CI

whisperity:
Instead of having two instances on Travis to test Linux and OSX, we can make a bigger build matrix for all three database configurations, totalling 6 different instances.

Why is this good?
 * Better separation of errors that _might_ result only in a particular backend. Currently, if `sqlite` tests fail, `psql` tests are never touched.
 * Better resillience against OSX-test particularities: instead of one download failure from LLVM/Clang, we can now fail three separate times.

Unfortunately, there wasn't an easy way of generating the new build matrix, due to the peculiarities of cross-OS building &mdash; hence the apparently heavy duplication in `.travis.yml`.

----

I have also cleaned up the install script because it used to install, or at least _check if is installed_ dependencies multiple times.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -40,6 +40,7 @@ test_unit_novenv:
 	$(UNIT_TEST_CMD)
 
 test_functional: test_sqlite test_psql
+
 test_functional_novenv: test_sqlite_novenv test_psql_novenv
 
 test_sqlite: package venv_dev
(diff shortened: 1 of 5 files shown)


tests/Makefile
```diff
@@ -17,7 +17,9 @@ REPO_ROOT ?= REPO_ROOT=$(ROOT)
 # Nose test runner configuration options.
 NOSECFG = --config .noserc
 
-test: pep8 test_unit test_functional
+test_agnostic: pep8 test_unit
```
gyorb: Could we find out a different name for this target instead of `test_agnostic`?

whisperity: How about putting `pep8` as a dependency of `test_unit` and removing this target entirely?

whisperity: Or just repeating `pep8 test_unit` in the `test_matrix_*` targets.

gyorb: Repeating should be fine, there are not too many places, and that way it is more visible what is needed for each target.

PR #655, opened 2017-05-16
Move viewer server to new structure

whisperity:
The server command has been moved to the new subcommand structure, was given a better help and an updated documentation.

> Supersedes: #532.

# Changes to invocation

 * Database arguments have been made akin to `store`. This means `--sqlite` is no longer deprecated and takes the argument of the database file. `--db-something` is introduced as opposed to `--dbsomething`, but old arguments are kept for backwards compatibility.
 * In line with popular webservers taking an argument for configuration directory (such as `-f` in nginx and Apache HTTPD), `CodeChecker server` also has an `-f CONFIG_DIRECTORY` option. Server configuration, such as authentication data, will be read from this folder.
 * The database and configuration location has been entirely separated, as `--workspace`, a command that never made its way to `store`, is kept only as a *legacy* support. Specifying `--workspace` means `--sqlite <workspace>/codechecker.sqlite -f <workspace>` so no changes to the behaviour, but the argument itself is eliminated from usage.
 * `-p/--port` has been introduced for port number, to be in line with `CodeChecker cmd` and common sense. `-v/--view-port` is kept, `TODO` to deprecated them.
 * The behaviour of #532 is introduced: `--not-host-only` is now only a wrapper for `--host ""`, while `--host` takes priority in specifying which address to listen on. The default behaviour remains as `--host localhost`, so a `server` started without arguments only listens on the local machine.

----

This pull request also fixes an unreported bug of `CodeChecker server --list` failing due to an earlier code restructuring hiccup result in a missing function.

Various code formatting changes has been introduced to the _actual_ server.


docs/user_guide.md
```diff
@@ -933,6 +933,152 @@ A detailed view of the available analyzers is available via `--details`. In the
...
+                        result viewing interface. NOTE: The suppress file
+                        relies on the "bug identifier" generated by the
+                        analyzers which is experimental, take care when
+                        relying on it.
+  --verbose {info,debug,debug_analyzer}
+                        Set verbosity level. (default: info)
+
+database arguments:
+  --sqlite SQLITE_FILE  Path of the SQLite database file to use. (default:
+                        /home/ericsza/.codechecker/codechecker.sqlite)
```
gyorb: There is a typo in the doc use `/home/<username>/` instead.

libcodechecker/server.py
```diff
@@ -0,0 +1,490 @@
...
+                        '-p', '--port',
+                        type=int,
+                        dest="view_port",
+                        metavar='PORT',
+                        default=8001,
+                        required=False,
+                        help="The port which will be used as listen port for "
+                             "the server.")
+
+    # TODO: This should be removed later on, in favour of --listen-address.
```
gyorb: I'm not sure we want to keep the `--listen-address` I think `--host 0.0.0.0` will mean the same configuration as the `--not-host-only` right now.

whisperity: `--listen-address` here was meant to mean `--host`. As you can see, there is no _actual_ `--listen-address` optstring, it's only the variable name of `--host`.

docs/user_guide.md
whisperity: I shall write a pre-commit hook which prevents such mistakes from ever happen again.
Will fix. 🙂

PR #658, opened 2017-05-20
Address various suggestions by static analysis tools.

Xazax-hun:


scripts/build_package.py
```diff
@@ -251,7 +251,7 @@ def handle_external_repository(dep, clean, env, verbose):
         if source:
             tmp = tempfile.mkdtemp()
             LOG.debug("Downloading font files " + source + " to " + tmp)
-            for fformat, agent in font_user_agents.iteritems():
+            for fformat, agent in font_user_agents.items():
```
whisperity: `items()` creates a copy in the iteration, which is, at this point, worthless memory overhead.

Take note that we are using **Python 2.7**, and **not** Python 3. The removal of `iteritems()` and the behaviour change of `items()` to only be a view is a Python 3 thing.

scripts/build_package.py
```diff
@@ -688,7 +687,7 @@ def build_package(repository_root, build_package_config, env=None):
     copy_tree(source, target)
 
     # Copy font files.
-    for _, dep in vendor_projects.iteritems():
+    for _, dep in vendor_projects.items():
```
whisperity: Same as above.

libcodechecker/session_manager.py
```diff
@@ -8,16 +8,17 @@
...
 import hashlib
 import json
 import os
-import portalocker
-import stat
 import shutil
+import stat
 import time
 import uuid
+from datetime import datetime
```
whisperity: This should stay at the front line. System imports are ordered by their package name.

libcodechecker/server/client_auth_handler.py
```diff
@@ -75,5 +68,5 @@ def performLogin(self, auth_method, auth_string):
 
     @timeit
     def destroySession(self):
-        return self.__manager.invalidate(self.__client_host,
+        return SessionManager.invalidate(self.__client_host,
```
whisperity: Same comment as the changes in `libcodechecker/instance_manager.py`. The static method here breaks symmetry. What if a module overrides `self.__manager` in this class?

libcodechecker/server/client_auth_handler.py
```diff
@@ -52,8 +46,7 @@ def getAuthParameters(self):
 
     @timeit
     def getAcceptedAuthMethods(self):
-        result = []
-        result.append("Username:Password")
+        result = ["Username:Password"]
         return result
```
whisperity: `return ["Username:Password"]`? It's even more compact that way.

libcodechecker/report.py
```diff
@@ -136,7 +134,6 @@ def compare_ctrl_sections(curr, prev):
 
             for i, section in enumerate(event_sections):
                 loc = section['location']
-                line_num = loc['line']
```
whisperity: This change here is @gyorb's jurisdiction. Is this a good catch on a left-behind unused variable, or did we miss something last time this file was touched?

scripts/build_package.py
Xazax-hun: I think since this collection should not be large, it is not a performance critical part of the code. I value forward compatibility in this case over performance.

libcodechecker/report.py
gyorb: `line_num` is not used for hash generation. It can be removed.

libcodechecker/session_manager.py
```diff
@@ -38,7 +39,7 @@
 session_lifetimes = {}
 
 
-class _Session():
+class _Session:
```
gyorb: I would modify it to this `class _Session(object)`

scripts/build_package.py
gyorb: We made similar changes other places in the code previously. The other thing that this is only in the packaging script.

libcodechecker/session_manager.py
Xazax-hun: Done.

PR #662, opened 2017-05-30
Fix suppress feature in parse broken by plist_parser refactoring

whisperity:
After the refactor in #635, the suppression feature in `CodeChecker parse` broke, as the new `report` object didn't contain the data that was needed to check if the file is suppressed according to the suppress file.

### DIFF
--- a/libcodechecker/analyze/analyzers/result_handler_plist_to_stdout.py
+++ b/libcodechecker/analyze/analyzers/result_handler_plist_to_stdout.py
@@ -91,9 +91,13 @@ class PlistToStdout(ResultHandler):
                 LOG.debug(report + ' is skipped (in ' + f_path + ")")
                 continue
 
+            bug = {'hash_value':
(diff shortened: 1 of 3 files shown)


libcodechecker/analyze/analyzers/result_handler_plist_to_stdout.py
```diff
@@ -91,9 +91,14 @@ def __print_bugs(self, reports, files):
...
 
+            bug = {'hash_value':
+                   report.main['issue_hash_content_of_line_in_context'],
+                   'file_path': f_path
+                   }
             if self.suppress_handler and \
-                    self.suppress_handler.get_suppressed(report):
-                LOG.debug(report + " is suppressed by suppress file.")
+                    self.suppress_handler.get_suppressed(bug):
+                LOG.debug(report.__str__() +
```
Xazax-hun: Isn't `str(report)` more idiomatic?

PR #664, opened 2017-05-30
Don't remove results folder if no analysis could be ran

whisperity:
Running `CodeChecker analyze nonexistent.file` removed potentially valid results while not replacing them with new ones.


whisperity:
Closed until I can figure out why this makes the OSX tests fail.

PR #675, opened 2017-06-26
Turn off two ClangTidy checkers

bruntib:
These two checkers give too many false-positives

### DIFF
--- a/config/config.json
+++ b/config/config.json
@@ -41,7 +41,6 @@
           {"misc-dangling-handle" : true },
           {"misc-inaccurate-erase" : true },
           {"misc-incorrect-roundings" : true },
-          {"misc-misplaced-widening-cast" : true },
           {"misc-move-forwarding-reference" : true },
           {"misc-move-constructor-init" : true },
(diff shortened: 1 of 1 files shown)


Xazax-hun:
I do not see why should those give too many false positives. Could you report the bugs to the clang tidy devs? I'm ok with turning them off by default in the meantime.

bruntib:
Actually I ran these on an opensource project and most of the findings were:
- According to the checker the thrown object should be constructed in place as a temporary object instead of constructing a named one and throw that. Most of the findings threw an enum constant.
- The other checker reported problems related to the known signed/unsigned - int/long int conversions. Of course these can be problems but in practice the findings were not bugs since the values were not close to INT_MAX where an error may occur.

gyorb:
I think the in the newest clang release (v4.0) there were some updates to these checker which caused to report more results.
We do not want flood the users with too much results. Even if we remove them from the default enabled list the users can re-enable them in the command line.

I think these type of changes should go into the release notes so the users are informed.

PR #696, opened 2017-06-29
extend analyze command output

gyorb:
print more information to the user that the analysis is finished and what to do to view the results


libcodechecker/analyze.py
```diff
@@ -291,3 +291,11 @@ def main(args):
     LOG.debug("Analysis metadata write to '" + metadata_path + "'")
     with open(metadata_path, 'w') as metafile:
         json.dump(metadata, metafile)
+
+    LOG.info("Anlysis finished.")
```
whisperity: There is a typo here.

gyorb: Thanks, I've fixed it.

PR #698, opened 2017-06-30
increase current work in progress version to 5.9

gyorb:

### DIFF
--- a/Doxyfile.in
+++ b/Doxyfile.in
@@ -34,7 +34,7 @@ PROJECT_NAME           = "Code Checker"
 # This could be handy for archiving the generated documentation or
 # if some version control system is used.
 
-PROJECT_NUMBER         = 5.8.0
+PROJECT_NUMBER         = 5.9.0
 
 # Using the PROJECT_BRIEF tag one can provide an optional one line description
 # for a project that appears at the top of each page and should give viewer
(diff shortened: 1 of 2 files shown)


whisperity:
Starting the client with verbose mode will make `CodeChecker version` output a tainted git tag `v5.8.0-123-g1234567` showing that the current build was created with changes above `5.8.0`.

Doesn't introducing this change create a confusion, as in this state, `5.9.0` is used for all version strings, including the verbose "taint" string too?

gyorb:
With this release we will change when the version is updated. 
In the current case we released v5.8 and we are working on v5.9 the version info will reflect that. Until v5.9 is not released we assume this is not a stable release.
If the users build a package from the master branch it will indicate that this version is ahead of the latest released version.
I do not think it will be too much confusing. The taint tag will show what is the latest commit and v5.9 will show that this is the version we are working now.

whisperity:
Actually, upon further thinking into it, the taint tag will still show `v5.8.0-NUMBER-gSHA`, because the latest git tag available in the repository is still `v5.8`. 🙂

gyorb:
Ok, than this needs to be changed later. So not the latest git tag is used as the version.

whisperity:
It only shows in `--verbose debug` mode. In "normal" modes, the API-given version is used.

gyorb:
I've created a separate issue to change the printed version so it will be consistent everywhere.

PR #704, opened 2017-07-05
update compilation command handling

gyorb:
In the current solution the update mode depends on the compilation
commands. If not given to the store command cleaning up the
previous reports will not work properly.

Introduce a new argument for the store command where the
compilation database (json) generated by the log can be used to
get the compilation commands for each source file.

### DIFF
--- a/libcodechecker/analyze.py
+++ b/libcodechecker/analyze.py
@@ -238,6 +238,10 @@ def main(args):
     # Parse the JSON CCDBs and retrieve the compile commands.
     actions = []
 
+    if len(args.logfile) != 1:
+        LOG.warning("Only one log file can be processed right now!")
+        sys.exit(1)
+
     for log_file in args.logfile:
         if not os.path.exists(log_file):
             LOG.error("The specified logfile '" + log_file + "' does not "
@@ -291,6 +295,17 @@ def main(args):
     LOG.debug("Analysis metadata write to '" + metadata_path + "'")
     with open(metadata_path, 'w') as metafile:
         json.dump(metadata, metafile)
+
+    # WARN: store command will search for this file!!!!
+    compile_cmd_json = os.path.join(args.output_path, 'compile_cmd.json')
+    try:
+        source = os.path.abspath(args.logfile[0])
+        target = os.path.abspath(compile_cmd_json)
+        shutil.copyfile(source, target)
+    except shutil.Error as serr:
+        LOG.debug("Compile command json file is the same")
+    except Exception as ex:
+        LOG.debug("Copying compile command json file failed.")
 
     LOG.info("Analysis finished.")
     LOG.info("To view results in the terminal use the "
--- a/libcodechecker/store.py
+++ b/libcodechecker/store.py
@@ -247,7 +247,7 @@ def add_arguments_to_parser(parser):
 
 
 def consume_plist(item):
-    f, context, metadata_dict = item
+    f, context, metadata_dict, compile_cmds = item
 
     LOG.debug("Parsing input file '" + f + "'")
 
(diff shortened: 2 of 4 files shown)


libcodechecker/analyze.py
```diff
@@ -292,6 +296,11 @@ def main(args):
     with open(metadata_path, 'w') as metafile:
         json.dump(metadata, metafile)
 
+    # WARN: store command will search for this file!!!!
+    compile_cmd_json = os.path.join(args.output_path, 'compile_cmd.json')
+    if not args.logfile[0] == compile_cmd_json:
```
bruntib: - The two paths should be converted to absolute path, because if relative path is given then the two strings may not be equal even for the same paths.
- `a != b` should be used instead of `not a == b`, because this way negation applies to `a`.
- What if the target directory already contains a `compile_cmd.json` file? It won't be overwritten and it can be confusing that the existing JSON file will be used.

libcodechecker/analyze.py
```diff
@@ -238,6 +238,10 @@ def main(args):
     # Parse the JSON CCDBs and retrieve the compile commands.
     actions = []
 
+    if len(args.logfile) != 1:
```
bruntib: `logfile` flag is declared to be a string in the command line arguments. Isn't it going to check the length of that string? This can be an array if this `analyze` module is loaded from a wrapper module (e.g. `CodeChecker check`) which gives it an array. By the way `CodeChecker check` gives it an array of length 1. Can this array be longer?

whisperity: `logfile` has `nargs = '+'` declared, so it is not a string, but an array of strings.
It _could_ be longer. Analyze was supposed to use an arbitrary number of log files (but of course at least one), and from _all_ these build actions, it would generate the output.

(Akin to how `store` and `parse` takes an arbitrary number of plists or plist-folders.)

`check` gives an array of one, because previously, check only operated on a single build command (causing **1** output json) or a single json.

libcodechecker/store.py
```diff
@@ -308,6 +314,25 @@ def __get_run_name(input_list):
...
+    compile_db = os.path.abspath(compile_db)
+
+    comp_cmds = {}
+    with open(compile_db, 'r') as ccdb:
+        comp_cmds = json.load(ccdb)
+
+    processed = {}
+    for cc in comp_cmds:
+        processed[str(os.path.join(cc['directory'], cc['file']))] = \
+                str(' '.join(cc['arguments']))
```
bruntib: Is this compilation database the one which is created by `CodeChecker log`? If I run `CodeChecker log` then the resulting JSON file has no "arguments" attribute but "command". And the "file" attribute contains the full path of the file, not just the file name. However I have seen a compilation database which contained a relative path to the file. It was generated by scan-build if I remember well.

libcodechecker/store.py
```diff
@@ -323,6 +348,11 @@ def main(args):
     if isinstance(args.input, str):
         args.input = [args.input]
 
+    compile_commands = {}
+    if len(args.logfile) != 1:
+        LOG.warning("Only one log file can be processed right now!")
+        compile_commands = process_compile_db(args.logfile[0])
```
bruntib: `analyze` command exited when not 1 log file was given. Shouldn't it behave the same way?

libcodechecker/store.py
whisperity: Newer `scan-build-py` creates such compilation command databases. The build log parser was patched to support both file formats, see the last commit 4137434e96c36557fb4612bd32ca175fe59d3267 in #610.

libcodechecker/store.py
gyorb: I've cleaned this up. No new argument will be available so no check is needed. The reports directory is checked for a `compile_cmd.json` file.

libcodechecker/store.py
```diff
@@ -367,13 +397,20 @@ def main(args):
     items = []
     for input_path in args.input:
         input_path = os.path.abspath(input_path)
+
+        # WARN: analyze command should create this file!
+        cmp_db = os.path.join(input_path, 'compile_cmd.json')
```
gyorb: Cleaned up this too.

libcodechecker/store.py
gyorb: I've updated the processing part to handle both formats.

PR #708, opened 2017-07-06
Removing BuildAction table and related elements.

bruntib:
In this commit the BuildAction is removed as a concept. This also means
that update mode doesn't work any more, because a report is connected to
a run instead of the build action. If a report is already found in the
run then it won't be updated, but skipped. Later in this case the
detection status of the report will be updated.


bruntib:
It went to the wrong branch.

PR #709, opened 2017-07-06
Removing BuildAction table and related elements.

bruntib:
In this commit the BuildAction is removed as a concept. This also means
that update mode doesn't work any more, because a report is connected to
a run instead of the build action. If a report is already found in the
run then it won't be updated, but skipped. Later in this case the
detection status of the report will be updated.

### DIFF
--- a/libcodechecker/analyze/analyzers/result_handler_plist_to_db.py
+++ b/libcodechecker/analyze/analyzers/result_handler_plist_to_db.py
@@ -34,7 +34,7 @@ class PlistToDB(ResultHandler):
         super(PlistToDB, self).__init__(buildaction, workspace)
         self.__run_id = run_id
 
-    def __store_bugs(self, files, reports, client, analisys_id):
+    def __store_bugs(self, files, reports, client):
         file_ids = {}
         # Send content of file to the server if needed.
         for file_name in files:
@@ -170,7 +170,7 @@ class PlistToDB(ResultHandler):
             type = report.main['type']
 
             LOG.debug("Storing report")
-            report_id = client.addReport(analisys_id,
+            report_id = client.addReport(self.__run_id,
                                          file_ids[fpath],
                                          bug_hash,
                                          msg,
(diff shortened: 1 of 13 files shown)


Xazax-hun:
I was wondering why are we removing the BuildAction?
Ok, I do remember, we will have a different design for the update mode, which makes the schema simpler (at the cost of managing the state on the disk in form of plists.)

But the BuildAction (which should be named AnalysisAction in the first place) contains other useful information like how long did a certain action take? What was the command? We can use this information for diagnosing performance problems or rerunning actions in debug mode for additional diagnostics. 

So my question is, do we loose these functionalities with this change, or are they going to work after this change just using information from other sources?

gyorb:
With the current database change/cleanup we will loose some of the functionality or at least will not be stored in the database.
We are splitting up the analysis process to log/analyze/store.
With this setup the failed analysis and build information should be handled at the analyze step and store will handle only the storage of the produced reports.

libcodechecker/analyze/analyzers/result_handler_plist_to_db.py
```diff
@@ -176,35 +176,13 @@ def handle_results(self):
...
-
-            build_cmd_hash = self.buildaction.original_command_hash
-            analysis_id = \
-                connection.add_build_action(self.__run_id,
-                                            build_cmd_hash,
-                                            analyzer_cmd,
-                                            self.buildaction.analyzer_type,
-                                            source_file_name)
-
-            assert self.analyzer_returncode == 0
```
Xazax-hun: Are you sure that this piece of error checking is related to build action handling?

libcodechecker/storage/report_server.py
```diff
@@ -45,104 +45,6 @@ class CheckerReportHandler(object):
     information to the database.
     """
 
-    def __sequence_deleter(self, table, first_id):
```
Xazax-hun: Are you sure you do not need this? To delete the path of bug reports.

bruntib: The bug path representation also changes: instead of linked list we are ordering by an enumeration column, so no sequence deleter is needed.

api/report_storage_server.thrift
```diff
@@ -52,16 +52,8 @@ service CheckerReport {
 
                 // The next few following functions must be called via the same connection.
```
dkrupp: Is this still valid?

I mean do we really want to handle storage transactions (storage of reports belonging to the same run instance) based on connections? What if a connection is breaks down due to network error? Will the run under being updated be broken too?

Wouldn't it be more safe to handle transactions independently of connections?
Something like:

i64 **openStorageTransaction**(run_id) //returns a run storage transaction id
addReport(...)
addReport(...)
**closeStorageTransaction**(i64 storage_transaction_id)

when the closeTransaction is called, we set the status of the reports that were not present in the session would be set to resolved.

bruntib: This will be the issue of #724.

PR #715, opened 2017-07-10
Remove debug command

gyorb:
Direct database access will not be supported only through the
server api.
Analysis failures should be handled at analysis not from the
database.


gyorb:
The analyze command could be extended with this functionality without the database see #716

PR #723, opened 2017-07-12
Local compare mode added to CodeChecker cmd diff mode.

dkrupp:
fixes #692

PR #717, opened 2017-07-10
new storage

gyorb:
Resolves #683

PR #730, opened 2017-07-14
Update documentation

whisperity:
There were minor issues in the previous release that the documentation has been updated to reflect usage changes of #717, but the code was not.

Also improved the help just a little bit with an example for #723.


dkrupp:
LGTM

PR #752, opened 2017-07-24
Minor fix for server --stop-all

whisperity:
The refactoring in #655 introduced a crash if `codechecker-server --stop-all` was called due to `--stop` and `--stop-all` being an `argparse.SUPPRESS` argument and thus raising a `KeyError` if `args.stop` would be used when `--stop` was **not** given.

I've also refactored the ugly long conditional to two early returns.


Xazax-hun:
Shouldn't we add a test case for that?

whisperity:
@Xazax-hun We should have added when this feature was originally implemented. :frowning_face: Done now.

PR #754, opened 2017-07-24
Adds username to the authenticated session. Shows logged in user in t…

### LINKED ISSUE #731: Show logged in user [24h]
When a user logs, the currently logged in user should be shown in the CodeChecker web UI.

Implementation:

- [x] Store the user name and groups along with the session data on the server. [8h]
- [x] Implement an internal API to query the currently logged in user [4h]
- [x] Implement an internal API to query the groups of the user [12h]

dkrupp:
…he webgui. Fixes #731

### DIFF
--- a/api/authentication.thrift
+++ b/api/authentication.thrift
@@ -7,6 +7,7 @@
 include "shared.thrift"
 
 namespace py Authentication
+namespace js codeCheckerAuthentication
 
 struct HandshakeInformation {
   1: bool requiresAuthentication,       // true if the server has a privileged zone --- the state of having a valid access is not considered here
@@ -27,5 +28,11 @@ service codeCheckerAuthentication {
 
   // performs logout action for the user (must be called from the corresponding valid session)
   bool destroySession()
+             throws (1: shared.RequestFailed requestError),
+
+  // returns currently logged in user within the active session
+  // returns empty string if the session is not active
+  string getLoggedInUser()
              throws (1: shared.RequestFailed requestError)
+
 }
--- a/libcodechecker/server/client_auth_handler.py
+++ b/libcodechecker/server/client_auth_handler.py
@@ -30,16 +30,26 @@ class ThriftAuthHandler(object):
     Handle Thrift authentication requests.
     """
 
-    def __init__(self, manager, session_token=None):
+    def __init__(self, manager, auth_session):
         self.__manager = manager
-        self.__session_token = session_token
+        self.__auth_session = auth_session
 
     @timeit
     def getAuthParameters(self):
+        token = None
+        if self.__auth_session:
+            token = self.__auth_session.token
         return HandshakeInformation(self.__manager.isEnabled(),
                                     self.__manager.is_valid(
-                                        self.__session_token,
+                                        token,
                                         True))
+
+    @timeit
+    def getLoggedInUser(self):
+        if self.__auth_session:
+            return self.__auth_session.user
+        else:
+            return ""
 
     @timeit
     def getAcceptedAuthMethods(self):
@@ -48,9 +58,10 @@ class ThriftAuthHandler(object):
     @timeit
     def performLogin(self, auth_method, auth_string):
         if auth_method == "Username:Password":
-            authToken = self.__manager.create_or_get_session(auth_string)
-            if authToken:
-                return authToken
+            session = self.__manager.create_or_get_session(auth_string)
+
+            if session:
+                return session.token
             else:
                 raise shared.ttypes.RequestFailed(
                     shared.ttypes.ErrorCode.PRIVILEGE,
@@ -62,4 +73,7 @@ class ThriftAuthHandler(object):
 
     @timeit
     def destroySession(self):
-        return self.__manager.invalidate(self.__session_token)
+        token = None
+        if self.__auth_session:
+            token = self.__auth_session.token
+        return self.__manager.invalidate(token)
--- a/libcodechecker/server/client_db_access_server.py
+++ b/libcodechecker/server/client_db_access_server.py
@@ -68,12 +68,15 @@ class RequestHandler(SimpleHTTPRequestHandler):
     def check_auth_in_request(self):
         """
         Wrapper to handle authentication needs from both GET and POST requests.
+        Returns a session object if correct cookie is presented or creates a
+        new session if the Authorization header and the correct credentials are
+        present.
         """
 
         if not self.manager.isEnabled():
-            return True
+            return None
 
-        success = False
+        success = None
 
         # Authentication can happen in two possible ways:
         #
@@ -90,9 +93,9 @@ class RequestHandler(SimpleHTTPRequestHandler):
                         values[0] == session_manager.SESSION_COOKIE_NAME:
                     if self.manager.is_valid(values[1], True):
                         # The session cookie contains valid data.
-                        success = values[1]
+                        success = self.manager.get_session(values[1], True)
 
-        if not success:
+        if success is None:
             # Session cookie was invalid (or not found...)
             # Attempt to see if the browser has sent us
             # an authentication request.
@@ -104,14 +107,16 @@ class RequestHandler(SimpleHTTPRequestHandler):
                     self.headers.getheader("Authorization").
                     replace("Basic ", ""))
 
-                token = self.manager.create_or_get_session(authString)
-                if token:
+                session = self.manager.create_or_get_session(authString)
+                if session:
                     LOG.info("Client from " + client_host + ":" +
-                             str(client_port) + " successfully logged in")
-                    return token
+                             str(client_port) +
+                             " successfully logged in as user " +
+                             session.user)
+                    return session
 
         # Else, access is still not granted.
-        if not success:
+        if success is None:
             LOG.debug(client_host + ":" + str(client_port) +
                       " Invalid access, credentials not found " +
                       "- session refused.")
@@ -120,13 +125,13 @@ class RequestHandler(SimpleHTTPRequestHandler):
         return success
 
     def do_GET(self):
-        authToken = self.check_auth_in_request()
-        if authToken:
+        auth_session = self.check_auth_in_request()
+        if not self.manager.isEnabled() or auth_session:
             self.send_response(200)
-            if isinstance(authToken, str):
+            if auth_session:
                 self.send_header("Set-Cookie",
                                  session_manager.SESSION_COOKIE_NAME + "=" +
-                                 authToken + "; Path=/")
+                                 auth_session.token + "; Path=/")
             SimpleHTTPRequestHandler.do_GET(self)
         else:
             self.send_response(401)
@@ -165,8 +170,9 @@ class RequestHandler(SimpleHTTPRequestHandler):
         iprot = input_protocol_factory.getProtocol(itrans)
         oprot = output_protocol_factory.getProtocol(otrans)
 
-        sess_token = self.check_auth_in_request()
-        if self.path != '/Authentication' and not sess_token:
+        auth_session = self.check_auth_in_request()
+        if self.manager.isEnabled() and self.path != '/Authentication' \
+                and not auth_session:
             # Bail out if the user is not authenticated...
             # This response has the possibility of melting down Thrift clients,
             # but the user is expected to properly authenticate first.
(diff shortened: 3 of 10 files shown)


api/authentication.thrift
```diff
@@ -28,4 +29,10 @@ service codeCheckerAuthentication {
   // performs logout action for the user (must be called from the corresponding valid session)
   bool destroySession()
              throws (1: shared.RequestFailed requestError)
```
whisperity: Though it seems Thrift doesn't seem to complain about this, just for the sake of code appearance, put a `,` at the end of this function declaration.

api/authentication.thrift
```diff
@@ -28,4 +29,10 @@ service codeCheckerAuthentication {
...
+
+  // returns currently logged in user within the active session
+  // returns empty string if the session is not active
```
whisperity: Maybe for better type safety, this should return `None` instead of `""` if no user is present.

libcodechecker/server/client_auth_handler.py
```diff
@@ -42,13 +43,23 @@ def getAuthParameters(self):
                                         True))
 
     @timeit
+    def getLoggedInUser(self):
+        if self.__auth_session:
+            return self.__auth_session.user
+        else:
+            return ""
```
whisperity: Akin to the comment in the Thrift definition,
```py
return None
```

libcodechecker/server/client_auth_handler.py
```diff
@@ -42,13 +43,23 @@ def getAuthParameters(self):
...
+    @timeit
     def getAcceptedAuthMethods(self):
         return ["Username:Password"]
 
     @timeit
     def performLogin(self, auth_method, auth_string):
         if auth_method == "Username:Password":
-            authToken = self.__manager.create_or_get_session(auth_string)
+            session = self.__manager.create_or_get_session(auth_string)
+            authToken = None
```
whisperity: This variable is unused, and the lack of newline between the conditionals make the code hard to read. Could be simplified as:

```py
session = ...

if session:
    return session.token
else:
    raise ...
```

libcodechecker/server/client_db_access_server.py
```diff
@@ -120,10 +125,13 @@ def check_auth_in_request(self):
         return success
 
     def do_GET(self):
-        authToken = self.check_auth_in_request()
-        if authToken:
+        authSession = self.check_auth_in_request()
+        authToken = None
+        if authSession:
```
whisperity: Same comment as in `client_auth_handler`, this extra conditional should be simplified out, or better to remove the `authToken` variable and use `authSession.token` instead. A session object, if valid, always has a token, anyways.

libcodechecker/server/client_db_access_server.py
```diff
@@ -120,10 +125,13 @@ def check_auth_in_request(self):
         return success
 
     def do_GET(self):
-        authToken = self.check_auth_in_request()
-        if authToken:
+        authSession = self.check_auth_in_request()
```
whisperity: Also a variable named `authSession` is not very Pythonic. This is my bad when I wrote the original version of this code I didn't really know the naming conventions. :innocent: Python uses `auth_session`.

libcodechecker/server/client_db_access_server.py
```diff
@@ -165,8 +173,12 @@ def do_POST(self):
         iprot = input_protocol_factory.getProtocol(itrans)
         oprot = output_protocol_factory.getProtocol(otrans)
 
-        sess_token = self.check_auth_in_request()
-        if self.path != '/Authentication' and not sess_token:
+        auth_session = self.check_auth_in_request()
+        sess_token = None
```
whisperity: Same thing about the conditional here too.

tests/functional/authentication/test_authentication.py
```diff
@@ -54,6 +54,10 @@ def test_privileged_access(self):
                                                 "invalid:invalid")
         self.assertIsNone(sessionToken, "Invalid credentials gave us a token!")
 
+        # A non-authenticated session should return an empty user
```
whisperity: Comment formatting.

libcodechecker/session_manager.py
```diff
@@ -274,6 +285,9 @@ def create_or_get_session(self, auth_string):
             self.__cleanup_sessions()
 
         if self.__handle_validation(auth_string):
+            # If the session is still valid
```
whisperity: Could we constrict this comment to be longer and less line-breaks, if it still conforms to pep8?

tests/functional/authentication/test_authentication.py
```diff
@@ -54,6 +54,10 @@ def test_privileged_access(self):
...
+        user = auth_client.getLoggedInUser()
+        self.assertEqual(user, "")
```
whisperity: If `return None` this test case should be updated. There is an `assertIsNone` method.

www/scripts/codecheckerviewer/codecheckerviewer.js
```diff
@@ -29,6 +29,9 @@ function (declare, topic, domConstruct, Dialog, DropDownMenu, MenuItem,
 
     CC_SERVICE = new codeCheckerDBAccess.codeCheckerDBAccessClient(
       new Thrift.Protocol(new Thrift.Transport("CodeCheckerService")));
+    var transport = new Thrift.TXHRTransport("/Authentication");
+    var protocol  = new Thrift.TJSONProtocol(transport);
+    var CC_AUTH_SERVICE    = new codeCheckerAuthentication.codeCheckerAuthenticationClient(protocol);
```
whisperity: Does `TXHRTransport` differ from `Thrift.Transport`? What's `Thrift.Protocol` versus `Thrift.TJSONProtocol`? (I think stating simply protocol means default, which is either hard-coded to be JSON in terms of the browser (as CodeChecker works :grinning: ), or Thrift actually negotiates with the server on which protocol to use.

Nonetheless, if these variables are not used later on, I think we should simplify this initialisation into one line, as is the case with `CC_SERVICE`.

whisperity: Why is `CC_SERVICE` not a `var` and this one is a `var`?

www/scripts/codecheckerviewer/codecheckerviewer.js
```diff
@@ -60,7 +63,16 @@ function (declare, topic, domConstruct, Dialog, DropDownMenu, MenuItem,
 
     domConstruct.place(logo, headerPane.domNode);
 
-    //--- Menu button ---//
+    var user = CC_AUTH_SERVICE.getLoggedInUser();
+    var loginUserSpan = null
```
whisperity: `;` missing

www/scripts/codecheckerviewer/codecheckerviewer.js
```diff
@@ -60,7 +63,16 @@ function (declare, topic, domConstruct, Dialog, DropDownMenu, MenuItem,
...
+    if (user.length > 0) {
```
whisperity: If behaviour changes to `return None`, this should be a `!= null` check.

www/scripts/codecheckerviewer/codecheckerviewer.js
```diff
@@ -103,7 +115,17 @@ function (declare, topic, domConstruct, Dialog, DropDownMenu, MenuItem,
       dropDown : menuItems
     });
 
-    headerPane.addChild(menuButton);
+    var headerMenu = domConstruct.create('div', {
+        id : 'header-menu'
+      });
+
+    if (loginUserSpan != null)
+        domConstruct.place(loginUserSpan,headerMenu);
```
whisperity: Space is missing after `,`.

www/scripts/codecheckerviewer/codecheckerviewer.js
```diff
@@ -103,7 +115,17 @@ function (declare, topic, domConstruct, Dialog, DropDownMenu, MenuItem,
...
+
+    domConstruct.place(menuButton.domNode,headerMenu);
```
whisperity: Space is missing after `,`.

www/style/codecheckerviewer.css
```diff
@@ -38,7 +38,14 @@ html, body {
   text-align: center;
 }
 
-.mainMenuButton {
+#loggedin {
+  padding-right: 10px;
+}
+
+#header-menu {
+  font-family: 'Source Code Pro', monospace;
```
whisperity: I've been tweaking a bit with the appearance of this text here. I would like to suggest a better idea.

First of all, the font specifications should go into `#loggedin`'s selector, not `#header-menu`'s.

And the settings I think that look much nicer than a monospace font:

```css
  font-family: 'Raleway', sans-serif;
  font-size: 11pt;
  color: white;
```

Compare:

image

to

image

www/scripts/codecheckerviewer/codecheckerviewer.js
```diff
@@ -60,7 +63,16 @@ function (declare, topic, domConstruct, Dialog, DropDownMenu, MenuItem,
 
     domConstruct.place(logo, headerPane.domNode);
 
-    //--- Menu button ---//
+    var user = CC_AUTH_SERVICE.getLoggedInUser();
+    var loginUserSpan = null
+    if (user.length > 0) {
+      loginUserSpan = domConstruct.create('span', {
+        id: 'loggedin',
+        innerHTML: "Logged in as " + user
```
whisperity: Closing `.` or a `:` somehow is missing, this sentence feels orphaned without it.

`Logged in as admin.`
`Logged in: admin.`

Something like that.

libcodechecker/server/client_auth_handler.py
Xazax-hun: An empty string is a deliberate choice or would None be more idiomatic?

libcodechecker/server/client_db_access_server.py
Xazax-hun: Almost each time when we query a session we also extract the token. Maybe it would be cleaner to return both of these?

libcodechecker/server/client_db_access_server.py
```diff
@@ -189,10 +201,15 @@ def do_POST(self):
...
+                                                 auth_session, sess_token)
                 processor = codeCheckerAuthentication.Processor(auth_handler)
             else:
-                acc_handler = ThriftRequestHandler(session,
+                if auth_session:
+                    LOG.debug("Accessing as user " + auth_session.user)
+                else:
+                    LOG.debug("Unauthenticated access.")
+
+                acc_handler = ThriftRequestHandler(session, auth_session,
```
whisperity: If this variable list is so long already, I'd put `auth_session` into a new line.

Xazax-hun: What is the difference between session and auth_session? Maybe better naming would be good for these variables?

libcodechecker/session_manager.py
```diff
@@ -57,18 +57,22 @@ def calc_persistency_hash(auth_string):
...
                               _Session.__initial_salt).hexdigest()
 
-    def __init__(self, token, phash):
+    def __init__(self, token, phash, user, authenticated):
+        self.authenticated = authenticated
+        self.user = user
         self.token = token
         self.persistent_hash = phash
         self.last_access = datetime.now()
+        self.destroyed = False
```
Xazax-hun: Why do we have both a list of valid sessions and an instance variable if a session is destroyed?

whisperity: We were trying with explicit logouts. Could have been a left-behind. We weren't sure with @dkrupp if the logout feature not working is a result of multithreading errors and sync issues or something else.

The answer was simple, mainly that `POST` queries don't send a new cookie, only `GET`, and that `HTTP-Authenticate` wasn't planned to support a "Logout" anyways. (The browser **always** sends the username-password header in **every** request once it caches it, so even if a session serverside is invalidated explicitly, the browser always tries to reauthorize.)

libcodechecker/server/client_db_access_server.py
whisperity: `session` is the database session.

libcodechecker/server/client_auth_handler.py
dkrupp: Unfortunately the javascript thrift parser throws an exception when we return None from python for function with string return type. So I will leave it like this.

libcodechecker/server/client_db_access_server.py
dkrupp: token is part of the session object. se we return both at the same time.

libcodechecker/server/client_auth_handler.py
dkrupp: Unfortunately the javascript thrift parser throws an exception when we return None from python for function with string return type. So I will leave it like this.

Issue #745, opened 2017-07-20
[Generic] Wade through all TODOs and handle them


whisperity:
Quite many TO-DOs have been introduced for breaking changes that are expected to happen with version 6 to make the command-line experience more streamlined, and inline with the newly introduced CLI.

This is a *generic issue*. It is expected to be changed on the fly as more action points are added.

# Important stuff

 - [ ] Remove `plist` command. It has been superseded by `store` and `parse`. It's already not documented.
 - [ ] Remove database arguments from `check`, completely.
 - [ ] Handle the removal of `TODO`s in the `cmd_line_client`, such as making run name(s) a positional argument, and **not** an opt-string anymore.

# Not so important stuff

 - [ ] Refactor cross-sublibrary imports.

PR #756, opened 2017-07-25
Deduplicate files between runs.

### LINKED ISSUE #728: Optimize file storage
If a file is stored several places with different paths then these shouldn't be stored several times. Storing the files uniquely can be accomplished by splitting `File` table to `File` and `FileContent` tables which are joined by file content hash.

Xazax-hun:
If two file has the same SHA256, store the content only once. The chance of hash collision is very small, so we do not handle that case.

Since several runs might share files with the same content this update should reduce the database size significantly.

Closes #728.


libcodechecker/server/client_db_access_handler.py
```diff
@@ -1546,50 +1546,45 @@ def replaceConfigInfo(self, run_id, config_values):
...
+            except sqlalchemy.exc.IntegrityError:
+                # Other transaction might added the same file in the meantime.
+                self.__session.rollback()
+
+        needed = self.__session.query(FileContent).get(content_hash) is None
         return NeedFileResult(needed, f.id)
 
     @timeit
-    def addFileContent(self, fid, content, encoding):
+    def addFileContent(self, content_hash, content, encoding):
```
dkrupp: Do not trust the client that it calculates the content_hash correctly. A malicious client could corrupt the whole file storage. Verify the content hash here and throw an exception if invalid hash was sent.

api/report_server.thrift
```diff
@@ -363,12 +363,12 @@ service codeCheckerDBAccess {
                  throws (1: shared.RequestFailed requestError),
 
   NeedFileResult needFileContent(
-                                 1: i64 run_id,
-                                 2: string filepath)
+                                 1: string filepath,
```
dkrupp: Add documentation to the API function.

If matching filepath, content_hash is found in the DB return the correspondig,existing NeedFileResult.fileId,  NeedFileResult.needed=false
* If only the content_hash matches, a new a new NeedFileResult.fileId is returned,  NeedFileResult.needed=false
* If the filepath matches, but content_hash not, a new file_id is generated and NeedFileResult.needed=true.
* If none of the conent_hash and the filepath matches a new file_id is generated and NeedFileResult.needed is false in the response.

libcodechecker/server/client_db_access_handler.py
Xazax-hun: I think it is redundant. We already trust the client in other places, and hashing is not cheap.

libcodechecker/orm_model.py
```diff
@@ -89,25 +87,31 @@ def __init__(self, run_id, checker_name, attribute, value):
...
+    content_hash = Column(String,
+                          ForeignKey('file_content.content_hash',
+                                     deferrable=True,
+                                     initially="DEFERRED", ondelete='CASCADE'))
 
-    def __init__(self, run_id, filepath):
-        self.run_id, self.filepath = run_id, filepath
-        self.inc_count = 0
+    __table_args__ = (UniqueConstraint('filepath', 'content_hash',
+                                       name='_filpath_content_hash_uc'),)
```
whisperity: The name of the `UNIQUE KEY` should be created by SQLAlchemy per the `CC_META` naming convention. There isn't a need to specify a name here.

Make sure to update the migration script to the observed name when you change this!

libcodechecker/orm_model.py
```diff
@@ -89,25 +87,31 @@ def __init__(self, run_id, checker_name, attribute, value):
         self.checker_name, self.run_id = checker_name, run_id
 
 
+class FileContent(Base):
+    __tablename__ = 'file_content'
```
whisperity: The usual convention is singular class name and plural table name. (Yes I know this is already violated at some places, but no need to violate it further, I have added an action point to my meta-issue #745.)

db_migrate/versions/1dd7cd2978aa_deduplicate_files_with_same_content.py
```diff
@@ -0,0 +1,46 @@
...
+depends_on = None
+
+from alembic import op
+import sqlalchemy as sa
+from hashlib import sha256
+import zlib
+
+
+def upgrade():
+    content = 'dummy content, backward incompatible update'
```
whisperity: Let's enclose non-machine strings in `"` instead of `'`.

db_migrate/versions/1dd7cd2978aa_deduplicate_files_with_same_content.py
```diff
@@ -0,0 +1,46 @@
...
+    hasher = sha256()
+    hasher.update(content)
+    content_hash = hasher.hexdigest()
+    content = zlib.compress(content,
+                            zlib.Z_BEST_COMPRESSION)
+    op.drop_column('files', 'content')
+    op.drop_column('files', 'inc_count')
+    op.drop_column('files', 'run_id')
+    op.add_column('files',
+                  sa.Column('content_hash', sa.String(),
```
whisperity: The `FOREIGN KEY` relation is missing from this table, and thus the resulting schema doesn't match the ORM.

libcodechecker/server/client_db_access_handler.py
```diff
@@ -1245,9 +1219,19 @@ def removeRunResults(self, run_ids):
             runs_to_delete.append(run_to_delete)
 
         for run_to_delete in runs_to_delete:
+            # FIXME: clean up bugpaths. Once run_id is a foreign key there,
+            # it should be automatic.
             session.delete(run_to_delete)
             session.commit()
 
+        s1 = select([BugPathEvent.file_id])
```
whisperity: Let's make this code more verbose. It's really clunky to read, especially the `~something` instead of `not something`.

But if this code doesn't do **anything** (seemingly it does not...), we shouldn't introduce it in the first place.

At this point, when I try to delete a run from the database, I get the following error in the server console:

    [ERROR] [12:05] - 'DeclarativeMeta' object is not iterable

And while run is deleted, the files and file_contents table is left intact.

db_migrate/versions/1dd7cd2978aa_deduplicate_files_with_same_content.py
```diff
@@ -0,0 +1,52 @@
...
+    hasher.update(content)
+    content_hash = hasher.hexdigest()
+    content = zlib.compress(content,
+                            zlib.Z_BEST_COMPRESSION)
+    content_table = op.create_table('file_contents',
+                                    sa.Column('content_hash', sa.String(),
+                                              primary_key=True),
+                                    sa.Column('content', sa.Binary(),
+                                              nullable=False))
+    op.bulk_insert(content_table,
```
whisperity: This `bulk_insert` should check if there are a file in the database or not. For freshly created databases, the migration script automatically insert this dummy dead weight into the database, which is not good.

PR #761, opened 2017-07-28
CTU Work-In-Progress

gerazo:
Please do not merge as it still misses functional tests.
CTU is working (and detects clang CTU capabilities) and functionally complete. Please review!

Problem: I cannot run the functional tests without errors (not even on master). Some tests fail (like functional/skip). I am following the README and get this: 
  File "/home/gerazo/repos/codechecker_dev_env/samsungXTU/codechecker/tests/functional/skip/test_skip.py", line 92, in test_skip
    self.assertEqual(bug['file'], skipped_file)
AssertionError: u'call_and_message.cpp' != 'file_to_be_skipped.cpp'

### DIFF
--- a/libcodechecker/analyze/analysis_manager.py
+++ b/libcodechecker/analyze/analysis_manager.py
@@ -8,7 +8,6 @@
 
 from collections import defaultdict
 import glob
-import json
 import multiprocessing
 import os
 import signal
@@ -18,7 +17,6 @@ import zipfile
 
 from libcodechecker import util
 from libcodechecker.analyze import analyzer_env
-from libcodechecker.analyze import ctu_manager
 from libcodechecker.analyze.analyzers import analyzer_types
 from libcodechecker.logger import LoggerFactory
 
@@ -275,8 +273,7 @@ def check(check_data):
 
 
 def start_workers(actions, context, analyzer_config_map,
-                  jobs, output_path, skip_handler, metadata,
-                  ctu_collect, ctu_analyze, ctu_dir, ctu_func_map_cmd):
+                  jobs, output_path, skip_handler, metadata):
     """
     Start the workers in the process pool.
     For every build action there is worker which makes the analysis.
@@ -288,13 +285,6 @@ def start_workers(actions, context, analyzer_config_map,
             pool.terminate()
         finally:
             sys.exit(1)
-
-    if ctu_collect:
-        ctu_manager.do_ctu_collect(actions, context, analyzer_config_map,
-                                   jobs, skip_handler,
-                                   ctu_dir, ctu_func_map_cmd)
-        if not ctu_analyze:
-            return
 
     signal.signal(signal.SIGINT, signal_handler)
 
--- a/libcodechecker/analyze/analysis_manager.py
+++ b/libcodechecker/analyze/analysis_manager.py
@@ -195,7 +195,8 @@ def check(check_data):
 
 
 def start_workers(actions, context, analyzer_config_map,
-                  jobs, output_path, skip_handler, metadata):
+                  jobs, output_path, skip_handler, metadata,
+                  ctu_collect, ctu_analyze, ctu_dir, ctu_func_map_cmd):
     """
     Start the workers in the process pool.
     For every build action there is worker which makes the analysis.
--- a/libcodechecker/analyze/analysis_manager.py
+++ b/libcodechecker/analyze/analysis_manager.py
@@ -15,6 +15,7 @@ import sys
 import traceback
 
 from libcodechecker.analyze import analyzer_env
+from libcodechecker.analyze import ctu_manager
 from libcodechecker.analyze.analyzers import analyzer_types
 from libcodechecker.logger import LoggerFactory
 
@@ -208,6 +209,13 @@ def start_workers(actions, context, analyzer_config_map,
             pool.terminate()
         finally:
             sys.exit(1)
+
+    if ctu_collect:
+        ctu_manager.do_ctu_collect(actions, context, analyzer_config_map,
+                                   jobs, skip_handler,
+                                   ctu_dir, ctu_func_map_cmd)
+        if not ctu_analyze:
+            return
 
     signal.signal(signal.SIGINT, signal_handler)
 
--- a/libcodechecker/analyze/analyzer.py
+++ b/libcodechecker/analyze/analyzer.py
@@ -89,7 +89,8 @@ def perform_analysis(args, context, actions, metadata):
     if 'ctu_phases' in args:
         ctu_collect = args.ctu_phases[0]
         ctu_analyze = args.ctu_phases[1]
-        ctu_dir = os.path.join(args.output_path, context.ctu_dir)
+        ctu_dir = os.path.join(args.output_path, 'ctu-dir')
+        args.ctu_dir = ctu_dir
         if analyzer_types.CLANG_SA not in analyzers:
             LOG.error("CTU can only be used with the clang static analyzer.")
             return
@@ -125,7 +126,7 @@ def perform_analysis(args, context, actions, metadata):
         ctu_manager.do_ctu_collect(actions, context, config_map, args.jobs,
                                    __get_skip_handler(args), ctu_dir)
 
-    if ctu_analyze or not ctu_analyze and not ctu_collect:
+    if ctu_analyze or (not ctu_analyze and not ctu_collect):
         analysis_manager.start_workers(actions, context, config_map,
                                        args.jobs, args.output_path,
                                        __get_skip_handler(args), metadata)
--- a/libcodechecker/analyze/analyzer.py
+++ b/libcodechecker/analyze/analyzer.py
@@ -7,6 +7,7 @@
 Prepare and start different analysis types
 """
 import copy
+import os
 import shlex
 import subprocess
 import time
@@ -75,6 +76,18 @@ def perform_analysis(args, context, actions, metadata):
     Additionally, insert statistical information into the metadata dict.
     """
 
+    ctu_collect = False
+    ctu_analyze = False
+    ctu_dir = ''
+    ctu_func_map_cmd = ''
+    if hasattr(args, 'ctu_phases'):
+        args.ctu_dir = os.path.abspath(args.ctu_dir)
+        args.logfile = os.path.abspath(args.logfile)
+        ctu_collect = args.ctu_phases[0]
+        ctu_analyze = args.ctu_phases[1]
+        ctu_dir = args.ctu_dir
+        ctu_func_map_cmd = args.ctu_func_map_cmd
+
     analyzers = args.analyzers if 'analyzers' in args \
         else analyzer_types.supported_analyzers
     analyzers, _ = analyzer_types.check_supported_analyzers(
@@ -103,7 +116,9 @@ def perform_analysis(args, context, actions, metadata):
 
     analysis_manager.start_workers(actions, context, config_map,
                                    args.jobs, args.output_path,
-                                   __get_skip_handler(args), metadata)
+                                   __get_skip_handler(args), metadata,
+                                   ctu_collect, ctu_analyze,
+                                   ctu_dir, ctu_func_map_cmd)
 
     end_time = time.time()
     LOG.info("Analysis length: " + str(end_time - start_time) + " sec.")
--- a/libcodechecker/analyze/analyzer.py
+++ b/libcodechecker/analyze/analyzer.py
@@ -9,6 +9,7 @@ Prepare and start different analysis types
 import copy
 import os
 import shlex
+import shutil
 import subprocess
 import time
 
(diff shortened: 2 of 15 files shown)


libcodechecker/analyze/analysis_manager.py
```diff
@@ -274,7 +275,8 @@ def check(check_data):
 
 
 def start_workers(actions, context, analyzer_config_map,
-                  jobs, output_path, skip_handler, metadata):
+                  jobs, output_path, skip_handler, metadata,
+                  ctu_collect, ctu_analyze, ctu_dir, ctu_func_map_cmd):
```
whisperity: Adding this many new parameters to the function doesn't feel the best. Perhaps we could somehow use the extended `config_handler` classes in `libcodechecker.analyze` instead?

libcodechecker/analyze/analyzer.py
```diff
@@ -80,6 +82,22 @@ def perform_analysis(args, context, actions, metadata):
     analyzers, _ = analyzer_types.check_supported_analyzers(
         analyzers, context)
 
+    ctu_collect = False
+    ctu_analyze = False
+    ctu_dir = ''
+    ctu_func_map_cmd = ''
+    # ctu_in_memory = False
+    if hasattr(args, 'ctu_phases'):
```
whisperity: While `hasattr` seems to be correct, we are using `'dest' in args` for these checks.

libcodechecker/libhandlers/analyze.py
```diff
@@ -89,6 +90,12 @@ def add_arguments_to_parser(parser):
     Add the subcommand's arguments to the given argparse.ArgumentParser.
     """
 
+    def is_ctu_capable():
```
whisperity: We have a detection function implemented for checking if clang exists or not. I don't recall the exact location, but you could find it by starting from `libcodechecker.libhandlers.analyzers` (or `checkers`, even).

This function should be introduced in a similar fashion, under `libcodechecker.analyze` package.

libcodechecker/libhandlers/analyze.py
```diff
@@ -22,6 +22,7 @@
 from libcodechecker.logger import LoggerFactory
 
 LOG = LoggerFactory.get_new_logger('ANALYZE')
+CTU_FUNC_MAP_CMD = 'clang-func-mapping'
```
whisperity: This default variable should be moved into a config variable, created by `package_context`, and read from `config\package_layout.json`.

libcodechecker/analyze/analyzers/analyzer_clangsa.py
```diff
@@ -137,6 +137,18 @@ def construct_analyzer_cmd(self, res_handler):
...
+                analyzer_cmd.extend(['-Xclang', '-analyzer-config',
+                                     '-Xclang', 'xtu-dir=' + config.ctu_dir,
+                                     '-Xclang', '-analyzer-config',
+                                     '-Xclang', 'reanalyze-xtu-visited=true'])
+                if config.ctu_in_memory:
+                    analyzer_cmd.extend(['-Xclang', '-analyzer-config',
+                                         '-Xclang',
+                                         'xtu-reparse=' +
+                                         os.path.abspath(config.log_file[0])])
+            # TODO Make it work on multiple logfiles
```
whisperity: (Note: The analyzer command is constructed for each and every build action separately. Maybe this fact could be used to handle this TODO.)

libcodechecker/libhandlers/analyze.py
```diff
@@ -177,6 +184,58 @@ def add_arguments_to_parser(parser):
...
+                                   "directory used between CTU phases.")
+        ctu_mutex_group.add_argument('--ctu-collect-only',
+                                     action='store_const',
+                                     const=[True, False],
+                                     dest='ctu_phases',
+                                     default=argparse.SUPPRESS,
+                                     help="Perform only the collect phase of "
+                                          "CTU. Keep <ctu-dir> for further "
+                                          "use.")
+        ctu_mutex_group.add_argument('--ctu-analyze-only',
```
whisperity: Instead of negative option strings (that turn out feature X and leave only Y on, **while** the default is both turned off...), we should use positive ones, such as `--ctu-run-analyze`.

libcodechecker/libhandlers/analyze.py
```diff
@@ -177,6 +184,58 @@ def add_arguments_to_parser(parser):
...
+                                     action='store_const',
+                                     const=[False, True],
+                                     dest='ctu_phases',
+                                     default=argparse.SUPPRESS,
+                                     help="Perform only the analyze phase of "
+                                          "CTU. <ctu-dir> should be present "
+                                          "and will not be removed after "
+                                          "analysis.")
+        ctu_opts.add_argument('--on-the-fly',
```
whisperity: This optstring lacks a `ctu` prefix.

libcodechecker/libhandlers/analyze.py
```diff
@@ -177,6 +184,58 @@ def add_arguments_to_parser(parser):
...
+                                     const=[True, True],
+                                     dest='ctu_phases',
+                                     default=argparse.SUPPRESS,
+                                     help="Perform cross translation unit "
+                                          "(CTU) analysis (both collect and "
+                                          "analyze phases) using default "
+                                          "<ctu-dir> for temporary output. "
+                                          "At the end of the analysis, the "
+                                          "temporary directory is removed.")
+        ctu_opts.add_argument('--ctu-dir',
```
whisperity: Ordering is clunky. `--ctu-dir` and the optstring for specifying the funcmap command should be the last in this argument group.

libcodechecker/libhandlers/analyze.py
```diff
@@ -177,6 +184,58 @@ def add_arguments_to_parser(parser):
...
+                                          "and will not be removed after "
+                                          "analysis.")
+        ctu_opts.add_argument('--on-the-fly',
+                              action='store_true',
+                              dest='ctu_in_memory',
+                              default=False,
+                              help="Do not create AST dumps in collect phase, "
+                                   "recompile external files on the fly "
+                                   "(memory intensive).")
+        ctu_opts.add_argument('--use-func-map-cmd',
```
whisperity: If the path of this external binary is moved to a config option, this optstring could just as well be eliminated. The command-line doesn't allow the user to specify which build tool or exactly which clang binary to use, the config JSONs are used to set up (parts of) the environment.

libcodechecker/libhandlers/analyze.py
```diff
@@ -177,6 +184,58 @@ def add_arguments_to_parser(parser):
...
+                                     action='store_const',
+                                     const=[True, True],
+                                     dest='ctu_phases',
+                                     default=argparse.SUPPRESS,
+                                     help="Perform cross translation unit "
+                                          "(CTU) analysis (both collect and "
+                                          "analyze phases) using default "
+                                          "<ctu-dir> for temporary output. "
+                                          "At the end of the analysis, the "
+                                          "temporary directory is removed.")
```
whisperity: The concept of using temporary directories, at least outside of `quickcheck` has been eliminated earlier. @gyorb What do you think of a subcommand creating heavy temporary dumps and then removing them later on?

If the analysis fails, these files could be useful. Do we really want to clean up state that could help debugging? We don't clean failed analysis outputs either.

The option commands infer a lot of different responsibilities, they trigger multiple phases, decide on temporary folders, etc.

libcodechecker/libhandlers/analyze.py
```diff
@@ -177,6 +184,58 @@ def add_arguments_to_parser(parser):
...
+        ctu_opts.add_argument('--ctu-dir',
+                              metavar='<ctu-dir>',
+                              default=os.path.join(
+                                  util.get_default_workspace(),
+                                  'ctu-dir'),
+                              help="Optional parameter defines the temporary "
+                                   "directory used between CTU phases.")
+        ctu_mutex_group.add_argument('--ctu-collect-only',
```
whisperity: I really have a deep concern about having this option here, in `analyze`. `analyze` is supposed to **run analysis**, and NOT prepare and/or create some dumps that are "to be used later". This option undermines the concept of this command.

I would be more happy if at least the "collect" option would be part of `CodeChecker log` instead. But introducing a separate sub-command would be the best course?

`analyze` has a precondition, that build logs are available. (and that `clang` is, etc.)
Thus, a CTU-version of `analyze` should have the precondition that alongside the build logs, the CTU dumps are also already available.

What happens if `--ctu-collect-only` (note my other comment about negative optstrings) is stated? The collection runs, an analysis (which is the cornerstone of the `analyze` command) does not?

If we are not to introduce `CodeChecker ctu` as a separate subcommand (we shouldn't), then at least how `check` was split into `log`+`analyze` (ignore `store` for now), so should an implicit `ctu` be split into `ctu-log` and `ctu-analyze`, with the subcommands handling parts of the full stack respectively.

In my opinion, `log` should use the `func-map` binary (@gerazo I take it right that that command is needed for the "1st phase" only?) from a config (see earlier comments). It should have a `--ctu-collect-output` option, that specifies the folder into which the dumps should be put. An optional argument, which **if** specified, automatically run the dumping.

And `analyze` should get a `--use-ctu-dumps` option (or something akin), that specifies in which folder (so it's a path argument, just like `--sqlite` in `server`.) the CTU dumps are available (akin to a positional argument specifying the `build.json`s to use), and if this is given, we infer that CTU-analysis must be ran (using the given logfiles).
The alternative approach would be, mutually excluding `--use-ctu-dumps` (the opstring is up for debate!), `--ctu-on-the-fly` (or somesuch), which would trigger CTU mechanics described below.

---

And of course, the unified set of arguments could be presented on the interface of `check` and `quickcheck`, which, with proper routing, would handle CTU usage. `check` would do `log --ctu` and then `--use-ctu-dumps`; whereas `quickcheck` should do

 * either a "simple" logging and then `--ctu-on-the-fly`
 * or a `log --ctu` and then `--use-ctu-dumps`, where the CTU files (just as how the `plist`s in `quickcheck`'s case) are cleaned up later on.

libcodechecker/analyze/analyzers/analyzer_clangsa.py
Xazax-hun: Also terminate sentences with period.

libcodechecker/libhandlers/analyze.py
gerazo: The original concept was is to merge this code with as little changes as possible to not diverge too much from the ctu code in scan-build. Options and everything also fits the scan-build implementation.

The tools create auxiliary data for CTU. This means that they do not really help debugging a clang crash as this information is kept elsewhere anyway. I would say, this data is not particularly useful unless we want to debug these tools (clang-func-mapping) themselves. Still the real and only reason removing the temporary folder is user's convenience, so nothing will happen if we don't do it.

whisperity: @gerazo Keeping in line with scan-build is understandable. However, there are some concerns we must take to keep the integrity of CodeChecker, and not turn it into a "scanbuild 2".

A concern about `--ctu-dir` being an argument, and `--on-the-fly` not having `--ctu` prefix (thus allowing a collision with other non-CTU option of the same optstring, if such arises), is a problem. I'll comment about `--ctu-dir` below.

libcodechecker/libhandlers/analyze.py
gerazo: Yes, it also raised my concerns after seeing that this concept doesn't really fit the current architecture. My problem was that while clangsa and clangtidy can be nicely separated along a multithreaded interface as these plugins can run separately, we cannot do this with CTU 1st and 2nd phase as the order is important (You are right in that clang-func-mapping is only for the 1st phase). Currently no 2-phased process is supported without creating a separate subcommand for it.

So about the subcommands: The original reason for putting these phases together was user convenience. It was an expectation from multiple sources who want to use this.
But: although putting 1st phase into log seems to be a solution, it is also raises confusion. Running the log is only necessary when you change your project structure like adding new sources. CTU collect (1st) phase on the other hand has to be run on any code change as function names and the ast can change anytime during development. This drastic workflow change would confuse the user I think. Based on this, a new, separate subcommand would be the cleanest, but again, it would be clangsa only, would rely on other subcommand's parametrization (for example if you use reparse here, you should use reparse there...). To be honest, none of the proposed solutions seem perfect so far.

Anyway, thanks for the insight, I will fix all your comments.

libcodechecker/libhandlers/analyze.py
gerazo: Yes, I will fix these. I am wondering that putting ctu-dir under the report folder would do any harm... so this option is removable then.

whisperity: @gerazo Technically it SHOULD NOT. Unless the user's folder structure is special enough that they have enough free space for the reports, but the CTU intermediate files can only be stored on an entire separate path (e.g. a separate network storage or hard drive.)

In that case, I'd tell the user to set `--output-dir` somewhere else, free up space, or use OTF mode.

libcodechecker/libhandlers/analyze.py
whisperity: Yeah, there seems not to exist a pristine, ideal solution, as CTU in itself greatly violates (with the two-phase invocation) the usual usage of clang(sa) too.

I could agree with leaving the collection and the analysis here, as they are _somewhat_ tied together, but I am still pro eliminating the `-only` options in favour of a "positive" optstring (as opposed to a "negatively phrased", see somewhere in the review.)

libcodechecker/libhandlers/analyze.py
gerazo: Agreed on that.

libcodechecker/libhandlers/analyze.py
gerazo: I like this. Let's wait for the other's opinions (see mail).

libcodechecker/libhandlers/analyze.py
whisperity: :innocent: Discontinuing the concept of "workspace" (which was the... duh... work space for both servers and analyses) in favour of using a more common `--config-dir` or `--output-dir` argument was one thing, but making 8234 separate """"output""" dirs is overkill.

PR #764, opened 2017-08-01
Bug path ordering by enumeration instead of linked list

bruntib:
Bug path events and report points are now ordered by enumeration instead
of storing them as a linked list in the database.

### DIFF
--- a/libcodechecker/orm_model.py
+++ b/libcodechecker/orm_model.py
@@ -116,63 +116,59 @@ class File(Base):
 class BugPathEvent(Base):
     __tablename__ = 'bug_path_events'
 
-    id = Column(Integer, autoincrement=True, primary_key=True)
     line_begin = Column(Integer)
     col_begin = Column(Integer)
     line_end = Column(Integer)
     col_end = Column(Integer)
+
+    order = Column(Integer, primary_key=True)
+
     msg = Column(String)
     file_id = Column(Integer, ForeignKey('files.id', deferrable=True,
                                          initially="DEFERRED",
                                          ondelete='CASCADE'), index=True)
+    report_id = Column(Integer, ForeignKey('reports.id', deferrable=True,
+                                           initially="DEFERRED",
+                                           ondelete='CASCADE'),
+                       primary_key=True)
 
-    next = Column(Integer)
-    prev = Column(Integer)
-
-    def __init__(self, line_begin, col_begin, line_end, col_end, msg, file_id):
+    def __init__(self, line_begin, col_begin, line_end, col_end,
+                 order, msg, file_id, report_id):
         self.line_begin, self.col_begin, self.line_end, self.col_end = \
             line_begin, col_begin, line_end, col_end
-        self.file_id, self.msg = file_id, msg
 
-    def addPrev(self, prev):
-        self.prev = prev
-
-    def addNext(self, next):
-        self.next = next
-
-    def isLast(self):
-        return self.next is None
-
-    def isFirst(self):
-        return self.prev is None
+        self.order = order
+        self.msg = msg
+        self.file_id = file_id
+        self.report_id = report_id
 
 
 class BugReportPoint(Base):
     __tablename__ = 'bug_report_points'
 
-    id = Column(Integer, autoincrement=True, primary_key=True)
     line_begin = Column(Integer)
     col_begin = Column(Integer)
     line_end = Column(Integer)
     col_end = Column(Integer)
+
+    order = Column(Integer, primary_key=True)
+
     file_id = Column(Integer, ForeignKey('files.id', deferrable=True,
                                          initially="DEFERRED",
                                          ondelete='CASCADE'), index=True)
+    report_id = Column(Integer, ForeignKey('reports.id', deferrable=True,
+                                           initially="DEFERRED",
+                                           ondelete='CASCADE'),
+                       primary_key=True)
 
-    # TODO: Add check, the value must be an existing id or null.
-    # Be careful when inserting.
-    next = Column(Integer)
-
-    def __init__(self, line_begin, col_begin, line_end, col_end, file_id):
+    def __init__(self, line_begin, col_begin, line_end, col_end,
+                 order, file_id, report_id):
         self.line_begin, self.col_begin, self.line_end, self.col_end = \
             line_begin, col_begin, line_end, col_end
+
+        self.order = order
         self.file_id = file_id
-
-    def addNext(self, next):
-        self.next = next
-
-    def isLast(self):
-        return self.next is None
+        self.report_id = report_id
 
 
 class Report(Base):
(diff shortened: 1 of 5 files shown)


libcodechecker/orm_model.py
```diff
@@ -117,62 +117,60 @@ class BugPathEvent(Base):
...
     line_end = Column(Integer)
     col_end = Column(Integer)
+
+    order = Column(Integer)
+
     msg = Column(String)
     file_id = Column(Integer, ForeignKey('files.id', deferrable=True,
                                          initially="DEFERRED",
                                          ondelete='CASCADE'), index=True)
+    report_id = Column(Integer, ForeignKey('reports.id', deferrable=True,
```
Xazax-hun: Do we need an id for this table? Maybe report_id, order pair could be together a key.

libcodechecker/orm_model.py
```diff
@@ -117,62 +117,60 @@ class BugPathEvent(Base):
...
+        self.order = order
+        self.msg = msg
+        self.file_id = file_id
+        self.report_id = report_id
 
 
 class BugReportPoint(Base):
     __tablename__ = 'bug_report_points'
 
     id = Column(Integer, autoincrement=True, primary_key=True)
```
Xazax-hun: Same comment for the id column as above.

libcodechecker/server/client_db_access_handler.py
```diff
@@ -101,43 +120,113 @@ def __init__(self,
...
+    def __lastBugEventPos(self, report_id):
+        """
+        This function returns the last BugPathEvent object position which
+        belongs to the given report. If no such event is found then None
+        returns.
+        """
+        last = self.__session.query(BugPathEvent) \
+            .filter(BugPathEvent.report_id == report_id) \
+            .order_by(BugPathEvent.order.desc()) \
+            .limit(1).all()
```
Xazax-hun: I would use one_or_none() or something similar.

libcodechecker/server/client_db_access_handler.py
```diff
@@ -101,43 +120,113 @@ def __init__(self,
...
+
+        session = self.__session
+        results = []
+        try:
+            # Count the reports subquery.
+            stmt = session.query(Report.run_id,
+                                 func.count(literal_column('*')).label(
+                                     'report_count')) \
+                .filter(Report.suppressed == false()) \
+                .group_by(Report.run_id) \
```
Xazax-hun: Can't you achieve the same using the count method instead of using literal_column and func.count()?

libcodechecker/server/client_db_access_handler.py
```diff
@@ -101,43 +120,113 @@ def __init__(self,
...
+                                       str(instance.date),
+                                       instance.name,
+                                       instance.duration,
+                                       reportCount,
+                                       instance.command
+                                       ))
+            return results
+
+        except sqlalchemy.exc.SQLAlchemyError as alchemy_ex:
+            msg = str(alchemy_ex)
```
Xazax-hun: I would eliminate the msg variable here.

libcodechecker/server/client_db_access_handler.py
```diff
@@ -101,43 +120,113 @@ def __init__(self,
...
-                .outerjoin(File,
-                           Report.file_id == File.id) \
-                .outerjoin(BugPathEvent,
-                           Report.end_bugevent == BugPathEvent.id) \
-                .outerjoin(SuppressBug,
-                           SuppressBug.hash == Report.bug_id)
+                .outerjoin(File, Report.file_id == File.id) \
+                .outerjoin(SuppressBug, SuppressBug.hash == Report.bug_id)
 
             results = q.limit(1).all()
```
Xazax-hun: I would use one or none here as well. It would make the code cleaner.

libcodechecker/server/client_db_access_handler.py
```diff
@@ -1613,28 +1540,29 @@ def addFileContent(self, content_hash, content, encoding):
...
         """
         try:
-            # There should be at least one bug event.
-            point2 = self.__session.query(BugPathEvent).get(start_bugevent_id)
+            q = self.__session.query(BugPathEvent) \
+                .filter(BugPathEvent.report_id == report_id) \
+                .order_by(BugPathEvent.order)
+
+            for i, point2 in enumerate(q):
+                if i == len(events):
```
Xazax-hun: Do we want to compute the len of events in every iteration?
Are you sure this condition is correct? Especially in the last iteration?

tests/functional/report_viewer_api/test_hash_clash.py
```diff
@@ -160,8 +160,8 @@ def test_hash_clash(self):
                                                  run_id1,
                                                  'XXX',
                                                  ((2, 1), (2, 2)))
-            # Same file, same hash and different position
-            self.assertNotEqual(rep_id1, rep_id3)
+            # Same file, same hash and different position in line
```
whisperity: Comment formatting.
Why exactly did we need this change?

libcodechecker/server/client_db_access_handler.py
whisperity: The rest of the handler follows a schema of

```py
msg = "" # some error string, either from database, or manually constructed
LOG.error(msg)
raise ...(msg)
```

whisperity: The rest of the handler follows a schema of

```py
msg = "" # some error string, either from database, or manually constructed
LOG.error(msg)
raise ...(msg)
```

tests/functional/report_viewer_api/test_hash_clash.py
bruntib: `assertNotEqual` has been changed to `assertEqual` because as of this commit we consider two bug paths the same if only the line numbers of the path event change (e.g. the definition of a function got few lines down). See the changes here: https://github.com/Ericsson/codechecker/pull/764/files#diff-b2e3db7161aa24606040a31602cb60a5L1626. By the way this is considered a heuristic, since we don't check if the line numbers in a bug path move equally.

libcodechecker/server/client_db_access_handler.py
bruntib: I don't really understand this question. Doesn't this code generate a count() function in SQL? By the way this is not my code. The diff indicates this function new because it has been moved in the source code from another place.

Xazax-hun: I was thinking about`query(...).count()` instead, but I think that might not be equivalent in this case, so this is ok.

libcodechecker/server/client_db_access_handler.py
```diff
@@ -344,64 +335,25 @@ def getRunResultCount(self, run_id, report_filters):
...
-        while event.next is not None:
-
-            event = session.query(BugPathEvent).get(event.next)
+        file_path_cache = {}
+        bug_items = []
 
+        for event in q:
             file_path = file_path_cache.get(event.file_id)
             if not file_path:
                 f = session.query(File).get(event.file_id)
```
Xazax-hun: I wonder, do we need a cache here? Wouldn't SQLAlchemy's ORM already do caching?
See http://docs.sqlalchemy.org/en/latest/orm/session_basics.html#is-the-session-a-cache

libcodechecker/orm_model.py
```diff
@@ -116,63 +116,59 @@ def __init__(self, filepath, content_hash):
...
-    id = Column(Integer, autoincrement=True, primary_key=True)
     line_begin = Column(Integer)
     col_begin = Column(Integer)
     line_end = Column(Integer)
     col_end = Column(Integer)
+
+    order = Column(Integer, primary_key=True)
+
     msg = Column(String)
     file_id = Column(Integer, ForeignKey('files.id', deferrable=True,
```
gyorb: We should decide when there events will be automatically deleted. Right now cascade delete tries to remove this if the corresponding file is removed or the corresponding report is removed. Is it possible that there is a conflict at deletion because of the foreign keys?

Xazax-hun: I think right now it should be deleted when the report is removed.

PR #790, opened 2017-08-10
Improve dependency generation at failed analyses

### LINKED ISSUE #793: failure collection error
with Codechecker version 5.9.1 collection fails

pointer.c:

```
#include <stdint.h>

void main(){
    int *p;
    int32_t i=(int32_t) p;     
}
```

if we compile this in C++ mode:

```
(venv)ednikru@md-mtas2:/ssd/ednikru/ctu/tests$ CodeChecker quickcheck --analyzers clangsa -b "gcc -Wall -std=c++11 -m32 ./pointer.c"
[10:23] - Starting build ...
cc1: warning: command line option \u2018-std=c++11\u2019 is valid for C++/ObjC++ but not for C [enabled by default]
./pointer.c: In function \u2018main\u2019:
./pointer.c:5:13: warning: \u2018p\u2019 is used uninitialized in this function [-Wuninitialized]
     int32_t i=(int32_t) p;
             ^
[10:23] - Build finished successfully.
[10:23] - Previous analysis results in '/tmp/codechecker-qcpW2YyO' have been removed, overwriting with current result
[10:23] - Starting static analysis ...
  File "/home/ednikru/work/codechecker/codechecker_git/codechecker/build/CodeChecker/lib/python2.7/libcodechecker/analyze/analysis_manager.py", line 244, in check
    zipfile.ZIP_DEFLATED)
  File "/usr/lib/python2.7/zipfile.py", line 1118, in write
    st = os.stat(filename)
OSError: [Errno 2] No such file or directory: 'is'
Exception in thread Thread-3:
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 396, in _handle_results
    cache[job]._set(i, obj)
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 599, in _set
    self._callback(self._value)
  File "/home/ednikru/work/codechecker/codechecker_git/codechecker/build/CodeChecker/lib/python2.7/libcodechecker/analyze/analysis_manager.py", line 317, in <lambda>
    results, metadata, output_path)
  File "/home/ednikru/work/codechecker/codechecker_git/codechecker/build/CodeChecker/lib/python2.7/libcodechecker/analyze/analysis_manager.py", line 47, in worker_result_handler
    failed_analysis[buildaction.analyzer_type] += 1
AttributeError: 'str' object has no attribute 'analyzer_type'
```

whisperity:
> Closes #793.

# Enhancements
 - Certain build systems, such as that of Linux and Apache Xerces, make `make` make dependency files on the fly. These directives are recorded in the build action, and they derail our dependency list generation which would create the source files in the `failed.zip`.
 - If the crash happens in CTU, the dependency list only contains the files originally found by G++/Clang to be a dependency for the build action... Additional to this, we now parse the analyzer output with a regular expression matching lines that mention files, and add these files to the ZIP too.

# Fixes
 - A nonexistent dependency file reported by the compiler would crash the current analyzer call. This has been fixed in a way that we explicitly append to the archive, under `failed-sources-root/<filename>` the reason why it couldn't be added to the archive.
 - Python Exceptions in the analyzer invoker code (such as the one above) would result in the process pool crashing and the `analyze` call hanging indefinitely. This has been fixed too.

### DIFF
--- a/libcodechecker/analyze/analyzers/analyzer_clangsa.py
+++ b/libcodechecker/analyze/analyzers/analyzer_clangsa.py
@@ -172,6 +172,38 @@ class ClangSA(analyzer_base.SourceAnalyzer):
             LOG.error(ex)
             return []
 
+    def get_analyzer_mentioned_files(self, output):
+        """
+        Parse ClangSA's output to generate a list of files that were mentioned
+        in the standard output or standard error.
+        """
+
+        # A line mentioning a file in ClangSA's output looks like this:
+        # /home/.../.cpp:L:C: warning: foobar.
+        regex = re.compile(
+            # File path followed by a ':'.
+            '^(?P<path>[\S ]+?):'
+            # Line number followed by a ':'.
+            '(?P<line>\d+?):'
+            # Column number followed by a ':' and a space.
+            # This is optional, as lines beginning with
+            # "In file included from" don't have a column number.
+            '((?P<column>\d+?):\ )?')
+
+        paths = []
+
+        for line in output.splitlines():
+            # Files can also be referenced in a line that
+            # begins as "In file included from /home/...".
+            if line[0] == 'I':
+                line = line.replace("In file included from ", "", 1)
+
+            match = re.match(regex, line)
+            if match:
+                paths.append(match.group('path'))
+
+        return set(paths)
+
     @classmethod
     def resolve_missing_binary(cls, configured_binary, env):
         """
--- a/libcodechecker/log/option_parser.py
+++ b/libcodechecker/log/option_parser.py
@@ -107,7 +107,14 @@ REPLACE_OPTIONS_MAP = {
 
 IGNORED_OPTION_MAP = {
     '-MT': 1,
+    '-MQ': 1,
     '-MF': 1,
+    '-MJ': 1,
+    '-MM': 0,
+    '-MP': 0,
+    '-MD': 0,
+    '-MV': 0,
+    '-MMD': 0,
     '-fsyntax-only': 0,
     '-save-temps': 0,
     '-install_name': 1,
@@ -123,6 +130,10 @@ IGNORED_OPTION_MAP = {
     '--param': 1,
     '-u': 1,
     '--serialize-diagnostics': 1
+}
+
+IGNORED_OPTION_MAP_REGEX = {
+    '^-g(.+)?$': 0,
 }
 
 # Unknown options by clang, will be skipped.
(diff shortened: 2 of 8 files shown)


libcodechecker/log/option_parser.py
```diff
@@ -107,7 +107,14 @@
 
 IGNORED_OPTION_MAP = {
     '-MT': 1,
+    '-MQ': 1,
     '-MF': 1,
+    '-MJ': 1,
```
Xazax-hun: Why do we have both this map and some of these hard coded above?

whisperity: This map is the `option_parser` which filters the arguments needed from a build command to an analyzer command. It was lacking some options to ignore, which resulted in a lot of `Unhandled argument` output if `analyze` was ran with `--verbose debug`. This removes arguments that are not needed or should not be passed or should be passed in a different way to Clang.

The other, in `analysis_manager.py` uses the **original** build command verbatim, which is **NOT** translated by this parser, and thus, must be cleaned separately. This is, because the dependency listing invokes the same compiler that was used, and thus should get as much arguments as possible. But a `-MF` dependency output file specified in the build command would derail the generation to `STDOUT`.

whisperity:
@dkrupp One regression to note for further CTU usage:

> Additional to this, we now parse the analyzer output with a regular expression matching lines that mention files, and add these files to the ZIP too.

Assume the following project: `X.cpp` includes `H.hpp` which we analyze, and there is a separate TU `A.cpp` (which includes `W.hpp`) where CTU finds an error. In this case, **only** `X.cpp`, `H.hpp` (original build action's dependencies) and `A.cpp` (because it was mentioned in the output) will be zipped. That is, `W.hpp` **is not ZIPped**.

I'd rather not add dependency generation for arbitrary files, because we don't know (at the place of the analysis failure) what build command was used to compile the other TUs. (Of course, the whole process _usually_ knows, but the thread invoking the current analysis does not.)

libcodechecker/analyze/analyzers/analyzer_clangsa.py
```diff
@@ -172,6 +172,38 @@ def construct_analyzer_cmd(self, res_handler):
             LOG.error(ex)
             return []
 
+    def get_analyzer_mentioned_files(self, output):
```
gyorb: Shouldn't we depend on the files listed in the plist output instead of parsing the clang sa output?

whisperity: This function is called when the analyzer crashes. There is no plist output in this case.

whisperity: I've updated the DocStrings of these functions to better convey what `output` means in this context (and also added a small optimisation for `In file included from` stripping)

PR #796, opened 2017-08-11
Fix check and quickcheck not forwarding analyzer args

### LINKED ISSUE #795: quickcheck mode ignores --tidyargs and --saargs parameters
if CodeChecker is run like this 
 CodeChecker quickcheck  --saargs ./saarg.txt --tidyargs tidyargs.txt --logfile ./b.log
the content of saargs and tidy args are ignored and not passed to the respective analyzers.
in "analyze" mode it is working fine.

whisperity:
Closes #795. It didn't work in `check` either.


libcodechecker/libhandlers/check.py
```diff
@@ -452,12 +452,12 @@ def __update_if_key_exists(source, target, key):
                 quiet_build=args.quiet_build,
```
dkrupp: any testcase for saargs, and tidyargs argument processing?

whisperity: I don't know if such exist. Most likely they don't, or if they were, they were moved to using `analyze` directly.

I don't know what or how could be added as an SA or Tidy args, so I'm unsure what could reliably made into a test case.

PR #801, opened 2017-08-14
Fix exception

gyorb:
An integer was returned instead of the proper exception type in case of an error.


gyorb:
Thanks @whisperity I've fixed it at the other place too.

PR #802, opened 2017-08-14
A new analysis argument --capture-analysis-output is introduced to st…

### LINKED ISSUE #800: Add analysis option to catch analyzer stdout
Some debug checkers, and checkers that create statistics (see https://github.com/baloghadamsoftware/clang/tree/ReturnCheckers ) emit analysis output to standard out.
Therefore it would be useful to capture and store the standard output of the analysis.

`--capture-analysis-output`  flag to analyze command to optionally capture analysis output.
The option will be off by default.
The output will be stored in `report_dir/success` directory. Note: the naming is the opposite of the "failed" directory.

There will be one file per analysis action, following the naming convention of the failure collection.

dkrupp:
…ore stdout/stderr of analysis invocation into the success subdirectory

Closes #800.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -343,6 +343,7 @@ usage: CodeChecker analyze [-h] [-j JOBS] [-i SKIPFILE] [-o OUTPUT_PATH]
                            [-t {plist}] [-n NAME]
                            [--analyzers ANALYZER [ANALYZER ...]]
                            [--add-compiler-defaults]
+                           [--capture-analysis-output]
                            [--saargs CLANGSA_ARGS_CFG_FILE]
                            [--tidyargs TIDY_ARGS_CFG_FILE]
                            [--ctu | --ctu-collect | --ctu-analyze]
@@ -422,6 +423,10 @@ analyzer arguments:
                         compilers themselves, and use them with Clang. This is
                         used when the compiler on the system is special, e.g.
                         when doing cross-compilation. (default: False)
+  --capture-analysis-output
+                        Store standard output and standard error of
+                        successful analyzer invocations into the
+                        '<OUTPUT_DIR>/success' directory.
   --saargs CLANGSA_ARGS_CFG_FILE
                         File containing argument which will be forwarded
                         verbatim for the Clang Static Analyzer.
--- a/libcodechecker/analyze/analysis_manager.py
+++ b/libcodechecker/analyze/analysis_manager.py
@@ -173,7 +173,7 @@ def check(check_data):
     """
 
     action, context, analyzer_config_map, \
-        output_dir, skip_handler = check_data
+        output_dir, skip_handler, capture_analysis_output = check_data
 
     skipped = False
     try:
@@ -237,13 +237,27 @@ def check(check_data):
             # sequence, whereas clang-tidy does not. We rewrite the file
             # names to contain no escape sequences for every analyzer.
             result_file = rh.analyzer_result_file.replace(r'\ ', ' ')
+            result_base = os.path.basename(result_file)
 
             if rh.analyzer_returncode == 0:
                 # Analysis was successful processing results.
-                if rh.analyzer_stdout != '':
+                if (capture_analysis_output):
+                    success_dir = os.path.join(output_dir, "success")
+                    if not os.path.exists(success_dir):
+                        os.makedirs(success_dir)
+                if len(rh.analyzer_stdout) > 0:
                     LOG.debug_analyzer('\n' + rh.analyzer_stdout)
-                if rh.analyzer_stderr != '':
+                    if (capture_analysis_output):
+                        with open(os.path.join(success_dir, result_base) +
+                                  ".stdout.txt", 'w') as outf:
+                            outf.write(rh.analyzer_stdout)
+                if len(rh.analyzer_stderr) > 0:
                     LOG.debug_analyzer('\n' + rh.analyzer_stderr)
+                    if (capture_analysis_output):
+                        with open(os.path.join(success_dir, result_base) +
+                                  ".stderr.txt", 'w') as outf:
+                            outf.write(rh.analyzer_stderr)
+
                 rh.postprocess_result()
                 # Generated reports will be handled separately at store.
 
@@ -267,7 +281,7 @@ def check(check_data):
                     os.makedirs(failed_dir)
                 LOG.debug("Writing error debugging to '" + failed_dir + "'")
 
-                zip_file = os.path.basename(result_file) + '.zip'
+                zip_file = result_base + '.zip'
                 with zipfile.ZipFile(os.path.join(failed_dir, zip_file),
                                      'w') as archive:
                     if len(rh.analyzer_stdout) > 0:
@@ -363,7 +377,8 @@ def check(check_data):
 
 
 def start_workers(actions, context, analyzer_config_map,
-                  jobs, output_path, skip_handler, metadata):
+                  jobs, output_path, skip_handler, metadata,
+                  capture_analysis_output):
     """
     Start the workers in the process pool.
     For every build action there is worker which makes the analysis.
@@ -397,7 +412,8 @@ def start_workers(actions, context, analyzer_config_map,
                              context,
                              analyzer_config_map,
                              output_path,
-                             skip_handler)
+                             skip_handler,
+                             capture_analysis_output)
                             for build_action in actions]
 
         pool.map_async(check,
--- a/libcodechecker/analyze/analyzer.py
+++ b/libcodechecker/analyze/analyzer.py
@@ -129,7 +129,9 @@ def perform_analysis(args, context, actions, metadata):
     if ctu_analyze or (not ctu_analyze and not ctu_collect):
         analysis_manager.start_workers(actions, context, config_map,
                                        args.jobs, args.output_path,
-                                       __get_skip_handler(args), metadata)
+                                       __get_skip_handler(args),
+                                       metadata,
+                                       'capture_analysis_output' in args)
 
     end_time = time.time()
     LOG.info("Analysis length: " + str(end_time - start_time) + " sec.")
--- a/libcodechecker/libhandlers/analyze.py
+++ b/libcodechecker/libhandlers/analyze.py
@@ -174,6 +174,16 @@ def add_arguments_to_parser(parser):
                                     "compiler on the system is special, e.g. "
                                     "when doing cross-compilation.")
 
+    analyzer_opts.add_argument('--capture-analysis-output',
+                               dest='capture_analysis_output',
+                               action='store_true',
+                               default=argparse.SUPPRESS,
+                               required=False,
+                               help="Store standard output and standard error "
+                                    "of successful analyzer invocations "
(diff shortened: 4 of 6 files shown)


libcodechecker/analyze/analysis_manager.py
```diff
@@ -237,13 +237,27 @@ def check(check_data):
...
+                if (args.capture_analysis_output):
+                    success_dir = os.path.join(output_dir, "success")
+                    if not os.path.exists(success_dir):
+                        os.makedirs(success_dir)
+                if len(rh.analyzer_stdout) > 0:
                     LOG.debug_analyzer('\n' + rh.analyzer_stdout)
-                if rh.analyzer_stderr != '':
+                    if (args.capture_analysis_output):
+                        with open(os.path.join(success_dir, result_base) +
+                                  ".stdout", 'w') as outf:
```
whisperity: I would use `.stdout.txt` as the postfix. Using a graphical browser on the result folder makes it more apparent what these files are.

libcodechecker/analyze/analyzer.py
```diff
@@ -129,7 +129,8 @@ def perform_analysis(args, context, actions, metadata):
     if ctu_analyze or (not ctu_analyze and not ctu_collect):
         analysis_manager.start_workers(actions, context, config_map,
                                        args.jobs, args.output_path,
-                                       __get_skip_handler(args), metadata)
+                                       __get_skip_handler(args),
+                                       metadata, args)
```
whisperity: Propagating the `args` here makes it really redundant, as `args.jobs` and `args.output_path` is also propagated.

It is also isn't the best approach to propagate a magic dynamic configuration container (such as `args` in this case) so deep and so much across thread and subprocess boundaries in the code.
It makes for a much better documented behaviour if we only passed the actual boolean here, and used it later on.

libcodechecker/libhandlers/analyze.py
```diff
@@ -174,6 +174,14 @@ def is_ctu_capable():
                                     "compiler on the system is special, e.g. "
                                     "when doing cross-compilation.")
 
+    analyzer_opts.add_argument('--capture-analysis-output',
+                               action='store_true',
```
whisperity: Please introduce an explicit `dest` for this argument, e.g. `capture_analysis_output`.

libcodechecker/libhandlers/analyze.py
```diff
@@ -174,6 +174,14 @@ def is_ctu_capable():
...
+                               default=False,
```
whisperity: For extension arguments like these, `argparse.SUPPRESS` is used as the default, so `(Default: False)` is not printed to the help.

Be careful, with this change, you need to change `args.dest` to `'dest' in args` everywhere, as a non-specified optstring results in a key error.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -237,13 +237,30 @@ def check(check_data):
             # sequence, whereas clang-tidy does not. We rewrite the file
             # names to contain no escape sequences for every analyzer.
             result_file = rh.analyzer_result_file.replace(r'\ ', ' ')
+            result_base = os.path.basename(result_file)
+            capture_analysis_output = False
+            if hasattr(args, 'capture_analysis_output'):
```
whisperity: This is woefully redundant here and we don't use this construct anywhere in CodeChecker.

If `default=False` is used, `args.capture_analysis_output` is always defined.

I suggested `default=argparse.SUPPRESS`, for which `'capture_analysis_output' in args` is equivalent to "it is passed and it is `True`".

libcodechecker/libhandlers/analyze.py
```diff
@@ -174,6 +174,14 @@ def is_ctu_capable():
...
+                               required=False,
+                               help="Store the (standard and error) output "
```
whisperity: Perhaps "standard output" and "standard error" would convey better, and also keep us in line with the UNIX Principles.

libcodechecker/libhandlers/analyze.py
```diff
@@ -174,6 +174,14 @@ def is_ctu_capable():
...
+                                    "of the successful analyzer invocations "
+                                    "into the <report-dir>/success directory.")
```
whisperity: Path should be formatted as `'<OUTPUT_DIR>/success'` (note the apostrophes around the path, this is how relative PATHs are mentioned in help). The `metavar` for `-o` is `OUTPUT_DIR`.

tests/functional/analyze/test_analyze.py
```diff
@@ -42,6 +42,41 @@ def tearDown(self):
...
+        print(out+err)
+        errcode = process.returncode
+        self.assertEquals(errcode, 0)
+
+        # We expect the sucess stderr file in the success directory.
+        success_files = os.listdir(success_dir)
+        print(success_files)
+        self.assertEquals(len(success_files), 1)
+        self.assertIn("success.c", success_files[0])
+
```
whisperity: Please add an

```py
os.remove(os.path.join(success_dir, success_files[0]))
```

at the end. These tests should always go into an empty report dir.

tests/functional/analyze/test_analyze.py
```diff
@@ -42,6 +42,41 @@ def tearDown(self):
         """Restore environment after tests have ran."""
         os.chdir(self.__old_pwd)
 
+    def test_capture_analysis_output(self):
+        """
+        Test if reports/success/<output_file>.[stdout,stderr] files are created
+        """
+        build_json = os.path.join(self.test_workspace, "build_success.json")
+        reports_dir = os.path.join(self.test_workspace, "reports")
```
whisperity: This variable is redundant. `self.report_dir` already contains this value.

libcodechecker/libhandlers/analyze.py
```diff
@@ -174,6 +174,14 @@ def is_ctu_capable():
                                     "compiler on the system is special, e.g. "
                                     "when doing cross-compilation.")
 
+    analyzer_opts.add_argument('--capture-analysis-output',
```
whisperity: I wonder, mayhaps _capture-**successful**-output_ could be a better optstring? Considering that failure output is also captured (about which I wonder perhaps we should introduce an optstring that _disables_ that notion, as there could be cases someone does not truly care about the analysis failures as deeply as the code currently does.)

libcodechecker/analyze/analysis_manager.py
```diff
@@ -237,13 +237,30 @@ def check(check_data):
...
+                if (capture_analysis_output):
+                    success_dir = os.path.join(output_dir, "success")
+                    if not os.path.exists(success_dir):
+                        os.makedirs(success_dir)
+                if len(rh.analyzer_stdout) > 0:
                     LOG.debug_analyzer('\n' + rh.analyzer_stdout)
-                if rh.analyzer_stderr != '':
+                    if (capture_analysis_output):
+                        with open(os.path.join(success_dir, result_base) +
+                                  ".stdout", 'w') as outf:
```
whisperity: I'd use `.stdout.txt` and `.stderr.txt` (or `.log`!) so that file types are made apparent in a graphical browser.

libcodechecker/analyze/analysis_manager.py
dkrupp: It is not defined in case analysis is executed in quickcheck mode.  I needed to introduce this check because of that. In quickheck mode, this parameter is non-sense.

libcodechecker/libhandlers/analyze.py
dkrupp: I will keep capture-analysis-output

whisperity: A'ight, as discussed, every "ideal" optstring would be too long. And just for the record, I'm noting that the suggestion about turning fail-zipping off was dropped.

libcodechecker/libhandlers/analyze.py
```diff
@@ -174,6 +174,15 @@ def is_ctu_capable():
...
+                               dest='capture_analysis_output',
+                               action='store_true',
+                               default=argparse.SUPPRESS,
+                               required=False,
+                               help="Store standard output and standard error"
+                                    "of the successful analyzer invocations "
```
whisperity: `of successful analyzer invocations`. A _the_ should not be here in this case.

libcodechecker/libhandlers/analyze.py
```diff
@@ -174,6 +174,15 @@ def is_ctu_capable():
...
+                                    "into the <report-dir>/success directory.")
```
whisperity: `into the '<OUTPUT_DIR>/success' directory.`

docs/user_guide.md
```diff
@@ -422,6 +423,10 @@ analyzer arguments:
                         compilers themselves, and use them with Clang. This is
                         used when the compiler on the system is special, e.g.
                         when doing cross-compilation. (default: False)
+  --capture-analysis-output
+                        Store the (standard and error) output of the
+                        successful analyzer invocations into the
+                        <report-dir>/success directory.
```
whisperity: Update the help here accordingly to the changes in the `help=` of this argument.

docs/user_guide.md
```diff
@@ -252,6 +252,7 @@ analyzer arguments:
   -i SKIPFILE, --skip SKIPFILE
   --analyzers ANALYZER [ANALYZER ...]
   --add-compiler-defaults
+  --capture-analysis-output
```
whisperity: This should **NOT** be added here, as this argument is not part of the argument list of `check`.

docs/user_guide.md
```diff
@@ -251,7 +251,7 @@ analyzer arguments:
   -j JOBS, --jobs JOBS
   -i SKIPFILE, --skip SKIPFILE
   --analyzers ANALYZER [ANALYZER ...]
-  --add-compiler-defaults
+  --add-compiler-defaults
```
whisperity: Extra whitespace added.

libcodechecker/libhandlers/analyze.py
```diff
@@ -174,6 +174,15 @@ def is_ctu_capable():
...
                                     "when doing cross-compilation.")
 
+    analyzer_opts.add_argument('--capture-analysis-output',
+                               dest='capture_analysis_output',
+                               action='store_true',
+                               default=argparse.SUPPRESS,
+                               required=False,
+                               help="Store standard output and standard error "
+                                    "of the successful analyzer invocations "
+                                    "into the <OUTPUT_DIR>/success directory.")
```
whisperity: Put an apostrophe (`'`) around the path.

libcodechecker/libhandlers/analyze.py
```diff
@@ -174,6 +174,15 @@ def is_ctu_capable():
                                     "compiler on the system is special, e.g. "
                                     "when doing cross-compilation.")
 
+    analyzer_opts.add_argument('--capture-analysis-output',
+                               dest='capture_analysis_output',
+                               action='store_true',
+                               default=argparse.SUPPRESS,
+                               required=False,
+                               help="Store standard output and standard error "
+                                    "of the successful analyzer invocations "
```
whisperity: `the` is not needed here.

PR #803, opened 2017-08-15
fix if workspace is removed before server stop

gyorb:
It is possible that the workspace was removed before the
server was stopped in that case an IOError was thrown
by the instance manager during the rewriting the configuration
files.

If stopping the server takes more time in the comment tests
the cleanup fails beacause there are files used by the server.

### DIFF
--- a/libcodechecker/server/client_db_access_server.py
+++ b/libcodechecker/server/client_db_access_server.py
@@ -340,13 +340,28 @@ def start_server(package_data, port, db_conn_string, suppress_handler,
                                      context.db_version_info,
                                      session_manager.SessionManager())
 
-    instance_manager.register(os.getpid(),
-                              os.path.abspath(context.codechecker_workspace),
-                              port)
+    try:
+        instance_manager.register(os.getpid(),
+                                  os.path.abspath(
+                                      context.codechecker_workspace),
+                                  port)
+    except IOError as ex:
+        LOG.debug(ex.strerror)
 
     LOG.info("Server waiting for client requests on [{0}:{1}]"
              .format(listen_address, str(port)))
 
-    atexit.register(instance_manager.unregister, os.getpid())
+    def unregister_handler(pid):
+        """
+        Handle errors during instance unregistration.
+        The workspace might be removed so updating the
+        config content might fail.
+        """
+        try:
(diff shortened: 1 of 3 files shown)


libcodechecker/server/client_db_access_server.py
```diff
@@ -340,13 +340,28 @@ def start_server(package_data, port, db_conn_string, suppress_handler,
...
 
     LOG.info("Server waiting for client requests on [{0}:{1}]"
              .format(listen_address, str(port)))
 
-    atexit.register(instance_manager.unregister, os.getpid())
+    def unregister_handler(pid):
+        """
+        Handle errors during instance unregistration.
+        The workspace might be removed so updating the
+        config content migth fail.
```
whisperity: s/migth/might

libcodechecker/server/client_db_access_server.py
```diff
@@ -340,13 +340,28 @@ def start_server(package_data, port, db_conn_string, suppress_handler,
...
+        """
+        try:
+            instance_manager.unregister()
```
whisperity: `pid` needs to be passed here.

libcodechecker/server/client_db_access_server.py
```diff
@@ -340,13 +340,28 @@ def start_server(package_data, port, db_conn_string, suppress_handler,
...
+        except IOError as ex:
+            LOG.debug(ex.strerror)
+
+    atexit.register(unregister_handler(os.getpid()))
```
whisperity: Is this invocation valid here? I think [the syntax is different](https://docs.python.org/2/library/atexit.html#atexit-example):

```py
atexit.register(unregister_handler, os.getpid())
```

libcodechecker/server/client_db_access_server.py
gyorb: Fixed.

libcodechecker/server/client_db_access_server.py
gyorb: Fixed.

libcodechecker/server/client_db_access_server.py
whisperity: This typo is left unfixed.

PR #809, opened 2017-08-16
Howto instructions extended for the case when the buildlog is empty.

dkrupp:
Howto is extended to describe frequently occurring usage error.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -23,6 +23,20 @@ CodeChecker log -b "make" -o compilation.json
 ```
 cat ./compilation.json
 ```
+
+**What to do if the `compilation.json` is empty?**
+* Make sure that your build system actually invoked the compiler (e.g. `gcc`,`g++`). 
+  In case your software was built once (and the binaries are already generated),
+  the compiler will not be invoked. In this case do a build cleanup (e.g. `make clean`) and 
+  retry to log your build.
+  
+* Make sure that the `CC_LOGGER_GCC_LIKE` environment variable is set correctly and contains your compilers. 
+  For detailed description see the [user guide](/docs/user_guide.md#1-log-mode).
+
+* MacOS users need `intercept-build` to be available on the system, 
+  and in most cases, _System Integrity Protection_ needs to be turned off. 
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -23,6 +23,16 @@ CodeChecker log -b "make" -o compilation.json
...
+
+**What to do if the `compilation.json` is empty?**
+* Make sure that your build system actually invoked the compiler (e.g. `gcc`,`g++`). 
+  In case your software was built once (and the binaries are already generated),
+  the compiler will not be invoked. In this case do a build cleanup (e.g. `make clean`) and 
+  retry to log your build.
+  
+* Make sure that the `CC_LOGGER_GCC_LIKE` environment variable is set correctly and contains your compilers. 
+  For detailed description see the [user guide](user_guide.md#1-log-mode).
+
```
whisperity: I would add a link back to the [README's _Mac OS X_](https://github.com/Ericsson/codechecker#mac-os-x) section as macOS requires an extra step to allow for build logging.

Something like this:

```markdown
macOS users need `intercept-build` to be available on the system, and in most cases, _System Integrity Protection_ needs to be turned off. See the [README](/README.md#mac-os-x) for details.
```

docs/usage.md
```diff
@@ -23,6 +23,16 @@ CodeChecker log -b "make" -o compilation.json
...
 ```
+
+**What to do if the `compilation.json` is empty?**
+* Make sure that your build system actually invoked the compiler (e.g. `gcc`,`g++`). 
+  In case your software was built once (and the binaries are already generated),
+  the compiler will not be invoked. In this case do a build cleanup (e.g. `make clean`) and 
+  retry to log your build.
+  
+* Make sure that the `CC_LOGGER_GCC_LIKE` environment variable is set correctly and contains your compilers. 
+  For detailed description see the [user guide](user_guide.md#1-log-mode).
```
whisperity: It would be better if documentation used absolute links: `[user guide](/docs/user_guide.md#1-log-mode)`, we use such crossreferences in other places too.

PR #827, opened 2017-08-22
Skipping and replacing non-utf8 characters from source files.

### LINKED ISSUE #820: CodeChecker storage fails for certain files
We get the following fault on the server side when storing the result of xerces.
This makes the storage of the whole run to fail:

17:05:30] [29020] <139951674660608> - client_db_access_handler.py:1444 storeZip() - Parsing input file 'clangsa_TraverseSchema.cpp_9c8d802e.plist'
[17:05:30] [29020] <139951674660608> - plist_parser.py:132 parse_plist() - Parsing plist: /tmp/tmpVmkRWB/reports/clangsa_TraverseSchema.cpp_9c8d802e.plist
[ERROR] [17:05:30] - 'utf8' codec can't decode byte 0xb7 in position 301483: invalid start byte

dkrupp:
Fixing utf8 parsing errors at store (which makes the storage fail) and showing Exception stack trace when processing thrift POST calls. Will be useful for debugging production environment... 

Fixes #820

### DIFF
--- a/libcodechecker/server/client_db_access_server.py
+++ b/libcodechecker/server/client_db_access_server.py
@@ -222,6 +222,8 @@ class RequestHandler(SimpleHTTPRequestHandler):
             return
 
         except Exception as exn:
+            import traceback
+            traceback.print_exc()
             LOG.error(str(exn))
             self.send_error(404, "Request failed.")
             return
--- a/libcodechecker/server/client_db_access_handler.py
+++ b/libcodechecker/server/client_db_access_handler.py
@@ -1487,7 +1487,8 @@ class ThriftRequestHandler(object):
                     LOG.debug(file_name + ' not found or already stored.')
                     continue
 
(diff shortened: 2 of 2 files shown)


libcodechecker/server/client_db_access_server.py
```diff
@@ -222,6 +224,8 @@ def do_POST(self):
             return
 
         except Exception as exn:
+            tb = traceback.format_exc()
```
whisperity: The usual format for exception printing is:

```py
except Type as var:
    import traceback
    traceback.print_exc()
```

libcodechecker/server/client_db_access_server.py
```diff
@@ -15,6 +15,8 @@
 import posixpath
 import socket
 import urllib
+import traceback
```
whisperity: This is not needed if the other comment is handled.

Issue #830, opened 2017-08-22
Sorage of runs to remote server fails


dkrupp:
If the contents are not stored onto localhost (so that source files are accessible from their original location), the storage fails.

PR #834, opened 2017-08-23
Storage on remote server (non localhost) is fixed. Source files are t…

dkrupp:
…aken

from the ZIP file instead of the directories on the server.
content_hashes.json file is now generated and sent over by the store command.
This file contains the content hashes of all source files that are referred to from the plist.
So the server can now find the correct file versions for the source files when storing the reports.

Fixed a bug which caused source files not being put into the ZIP file when their content hashes were the same.

Renamed necessaryFileContents to getMissingContentHashes to better express its functionality.

### DIFF
--- a/libcodechecker/libhandlers/store.py
+++ b/libcodechecker/libhandlers/store.py
@@ -199,7 +199,10 @@ def assemble_zip(inputs, zip_file, client):
     source_dir = os.path.join(temp_dir, 'root')
     os.makedirs(source_dir)
 
-    file_dict = {}
+    hash_to_file = {}
+    # There can be files with same hash,
+    # but different path.
+    file_to_hash = {}
 
     def collect_file_hashes_from_plist(plist_file):
         try:
@@ -210,7 +213,8 @@ def assemble_zip(inputs, zip_file, client):
                     hasher = sha256()
                     hasher.update(content.read())
                     content_hash = hasher.hexdigest()
-                    file_dict[content_hash] = f
+                    hash_to_file[content_hash] = f
+                    file_to_hash[f] = content_hash
         except Exception as ex:
             LOG.error('Parsing the plist failed: ' + str(ex))
 
@@ -239,8 +243,8 @@ def assemble_zip(inputs, zip_file, client):
                     if f.endswith(".plist"):
                         collect_file_hashes_from_plist(plist_file)
 
-        necessary_hashes = client.necessaryFileContents(file_dict.keys())
-        for h, f in file_dict.items():
+        necessary_hashes = client.getMissingContentHashes(hash_to_file.keys())
+        for f, h in file_to_hash.items():
             if h in necessary_hashes:
                 target_file = os.path.join(source_dir, f.lstrip('/'))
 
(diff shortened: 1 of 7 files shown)


dkrupp:
This fixes #830

libcodechecker/libhandlers/store.py
```diff
@@ -254,10 +256,15 @@ def collect_file_hashes_from_plist(plist_file):
...
+        hashes_fname = "content_hashes.json"
+        hashes_fpath = os.path.join(temp_dir, hashes_fname)
+        with open(hashes_fpath, "w") as f:
+            json.dump(file_to_hash, f)
+
         with zipfile.ZipFile(zip_file, 'w') as zipf:
             for root, _, files in os.walk(temp_dir):
-                for file in files:
-                    f = os.path.join(root, file)
+                for sf in files:
```
gyorb: Why was a rename `file` to `sf` needed?

libcodechecker/server/client_db_access_handler.py
```diff
@@ -1481,13 +1489,25 @@ def massStoreRun(self, name, version, b64zip, force):
                 continue
 
             for file_name in files:
-                zip_file_name = os.path.join(source_root, file_name)
+                source_file_name = os.path.join(source_root, file_name[1:])
```
gyorb: What do we split here? `file_name.strip("/")` wouldn't be better?

libcodechecker/server/client_db_access_handler.py
```diff
@@ -1481,13 +1489,25 @@ def massStoreRun(self, name, version, b64zip, force):
...
+                LOG.debug("Storing source file:"+source_file_name)
 
-                if not os.path.isfile(zip_file_name):
+                if not os.path.isfile(source_file_name):
                     LOG.debug(file_name + ' not found or already stored.')
+                    fid = store_handler.getFileIdByContentHash(
+                        self.__Session(),
```
gyorb: Do we need to create a new session for each `getFileIdByContentHash` query?

libcodechecker/libhandlers/store.py
dkrupp: file is a builtin and its not nice to use builtins for var names.

libcodechecker/libhandlers/store.py
```diff
@@ -200,6 +200,7 @@ def assemble_zip(inputs, zip_file, client):
     os.makedirs(source_dir)
 
     file_dict = {}
```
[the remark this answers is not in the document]
whisperity: Yes, I agree, we should rename this dict.

tests/Makefile
```diff
@@ -5,8 +5,8 @@
 PSQL ?= TEST_USE_POSTGRESQL=true
 PG800 ?= CODECHECKER_DB_DRIVER=pg8000
 PSYCOPG2 ?= CODECHECKER_DB_DRIVER=psycopg2
-DBPORT ?= TEST_DBPORT=5432
-DBUNAME ?= TEST_DBUSERNAME=postgres
+DBPORT ?= TEST_DBPORT=4995
```
gyorb: I think these should be set back to the original values.

tests/libtest/codechecker.py
```diff
@@ -199,6 +199,7 @@ def serv_cmd(codechecker_cfg, test_config):
                        '--port',
                        str(codechecker_cfg['viewer_port'])
                        ])
+    # server_cmd.extend(['--verbose', 'debug'])
```
gyorb: This can be removed too.

tests/Makefile
whisperity: @dkrupp Always use `git add --patch`. :wink:

PR #845, opened 2017-08-29
Remove hardcoded checker messages

gyorb:
Checker messages can change quite rapidly with new clang releases.
Tests can fail quite easily if we depend on them.


whisperity:
Are we sure no coverage is lost with this change that we don't want to lose?

gyorb:
I do not think we loose coverage. The api is called in multiple ways (one run, all runs). What we loose is maybe some precision because we do not  compare the results for a fix result set which should be returned by the query.

PR #856, opened 2017-08-31
Nomenclature fixes and some TODO cleanup

whisperity:
> Implements parts of #745 and related issues.

For now, I've added many comments so we can individually see the changes. Of course this can be squashed later.


api/report_server.thrift
```diff
@@ -21,58 +21,55 @@ const string API_VERSION = '6.0'
 const i64 MAX_QUERY_SIZE = 500
 //=================================================
 
-//-----------------------------------------------------------------------------
-struct RunData{
+struct RunData {
   1: i64            runId,        // unique id of the run
```
Xazax-hun: Maybe we could adopt the convention of having full sentences in comments to comply with the rest of the codebase as the part of this revamp?

api/report_server.thrift
```diff
@@ -21,58 +21,55 @@ const string API_VERSION = '6.0'
...
-  3: string              checkedFile,     // this is a filepath
-  4: string              checkerMsg,      // description of the bug report
-  5: i64                 reportId,        // id of the report in the current run in the db
-  6: i64                 fileId,          // unique id of the file the report refers to
-  7: i64                 line,            // line number or the reports main section (not part of the path)
-  8: i64                 column,          // column number of the report main section (not part of the path)
-  9: shared.Severity     severity         // checker severity
-  10: ReviewData          review           // bug review status informations.
+struct ReportData {
+  1: string                  checkerId,       // the qualified id of the checker that reported this
```
Xazax-hun: Maybe this should be renamed as this is the name of the check.

api/report_server.thrift
```diff
@@ -21,58 +21,55 @@ const string API_VERSION = '6.0'
...
+  2: string                  bugHash,         // This is unique id of the concrete report.
+  3: string                  checkedFile,     // this is a filepath
+  4: string                  checkerMsg,      // description of the bug report
+  5: i64                     reportId,        // id of the report in the current run in the db
+  6: i64                     fileId,          // unique id of the file the report refers to
+  7: i64                     line,            // line number or the reports main section (not part of the path)
+  8: i64                     column,          // column number of the report main section (not part of the path)
+  9: shared.Severity         severity         // checker severity
+  10: ReviewData             review           // bug review status informations.
```
Xazax-hun: I would prefer to call this `reviewStatus`.

api/report_server.thrift
```diff
@@ -21,58 +21,55 @@ const string API_VERSION = '6.0'
...
-  5: i64                 reportId,        // id of the report in the current run in the db
-  6: i64                 fileId,          // unique id of the file the report refers to
-  7: i64                 line,            // line number or the reports main section (not part of the path)
-  8: i64                 column,          // column number of the report main section (not part of the path)
-  9: shared.Severity     severity         // checker severity
-  10: ReviewData          review           // bug review status informations.
+struct ReportData {
+  1: string                  checkerId,       // the qualified id of the checker that reported this
+  2: string                  bugHash,         // This is unique id of the concrete report.
+  3: string                  checkedFile,     // this is a filepath
```
Xazax-hun: In case this is the `main file` which was given for the analyzer, we should use the same nomenclature.

api/report_server.thrift
```diff
@@ -93,7 +90,7 @@ typedef list<ReportFilter> ReportFilterList
  * Members of this struct are interpreted in "OR" relation with each other.
```
Xazax-hun: I think this comment is very confusing. Both sentences referring to the members of the struct. Is this correct? 
Why do we need the `v1`? Can we remove that now?

Xazax-hun: As a side note, I am not sure this has the same expressivity as the old version. 

E.g. how do you represent:
```
(filepath == A AND checkerName == B) OR (severity == C and reviewStatus == D)
```

Note that the old format was basically a disjunctive normal form, so we could be sure that every predicate could be translated to that.

libcodechecker/libhandlers/analyze.py
```diff
@@ -176,8 +176,8 @@ def is_ctu_capable():
 
     analyzer_opts.add_argument('--add-compiler-defaults',
                                action='store_true',
-                               default=False,
                                required=False,
+                               default=argparse.SUPPRESS,
```
Xazax-hun: Why do we need this change?

libcodechecker/libhandlers/check.py
```diff
@@ -453,21 +382,21 @@ def __update_if_key_exists(source, target, key):
...
-            jobs=args.jobs,
-            add_compiler_defaults=args.add_compiler_defaults
+            jobs=args.jobs
         )
         # Some arguments don't have default values.
         # We can't set these keys to None because it would result in an error
         # after the call.
         args_to_update = ['skipfile',
                           'analyzers',
+                          'add_compiler_defaults',
```
Xazax-hun: Why can't this have a default value?

libcodechecker/libhandlers/quickcheck.py
```diff
@@ -322,14 +318,14 @@ def __update_if_key_exists(source, target, key):
...
             output_format='plist',
             clean=True,
-            jobs=args.jobs,
-            add_compiler_defaults=args.add_compiler_defaults
+            jobs=args.jobs
         )
         # Some arguments don't have default values.
         # We can't set these keys to None because it would result in an error
         # after the call.
         args_to_update = ['skipfile',
```
Xazax-hun: Looks like this list is replicated. In the future, it might be better to refactor it in a way that this list is created together with the arg parsers.

api/report_server.thrift
whisperity: v1 can only be removed if the command-line client is upgraded to use the same, new, "v2" filtering semantics.

`(filepath == A AND checkerName == B) OR (severity == C and reviewStatus == D)`
This kind of filter was not explicable by any of the clients in the v1 version either, even if the API -- if directly called -- seemed to had supported that.

@gyorb, @csordasmarton rewrote the filtering system.

libcodechecker/libhandlers/analyze.py
whisperity: To not have `(default: False)` printed for the help message in every argument if the argument is already a boolean, and its help already makes it apparent that "it does _something_ if given, otherwise it does not do _something_ ".

libcodechecker/libhandlers/check.py
whisperity: As above. Boolean options' help messages usually look better if `default: False` is not given. The code behaves in a way that if it is not passed, it is considered `False`.

api/report_server.thrift
```diff
@@ -21,58 +21,55 @@ const string API_VERSION = '6.0'
...
-struct RunData{
+struct RunData {
   1: i64            runId,        // unique id of the run
   2: string         runDate,      // date of the run
   3: string         name,         // human readable name of the run
   4: i64            duration,     // duration of the run; -1 if not finished
   5: i64            resultCount,  // number of results in the run
   6: string         runCmd,       // the used check command
-  7: optional bool  can_delete    // true if codeCheckerDBAccess::removeRunResults()
+  7: optional bool  canDelete,    // true if codeCheckerDBAccess::removeRunResults()
```
gyorb: Do we still need the `canDelete`? I'm not sure we use it in the database with the new schema.

api/report_server.thrift
gyorb: The old `ReportFilter` will be removed.

whisperity:
If everything is fine, I'll squash these commits! **Don't merge beforehand.**

PR #858, opened 2017-08-31
Local compare mode takes in-code-suppressions into consideration.

### LINKED ISSUE #851: source-code suppression does not work in local-compare mode
https://github.com/Ericsson/codechecker/blob/master/docs/user_guide.md#suppression-in-the-source-code

        int *something=0;
        // codechecker_suppress [all] suppress all checker results
        *something=5;
and 
CodeChecker cmd diff -n ./myreports -b tinyxml_master -p 4999 --new
/home/ednikru/work/codechecker/TinyXML/tinyxml.cpp:1884:12: Dereference of null pointer (loaded from variable 'something') [core.NullDereference]
	*something=5;

still lists it as a new bug, however it should not.

dkrupp:
This fixes #851

### DIFF
--- a/libcodechecker/cmd/cmd_line_client.py
+++ b/libcodechecker/cmd/cmd_line_client.py
@@ -12,6 +12,7 @@ import sys
 
 from codeCheckerDBAccess_v6 import constants, ttypes
 
+from libcodechecker import suppress_handler
 from libcodechecker import suppress_file_handler
 from libcodechecker.analyze import plist_parser
 from libcodechecker.libclient.client import handle_auth
@@ -20,6 +21,7 @@ from libcodechecker.logger import LoggerFactory
 from libcodechecker.output_formatters import twodim_to_str
 from libcodechecker.report import Report
 from libcodechecker.util import split_server_url
+
 
 LOG = LoggerFactory.get_new_logger('CMD')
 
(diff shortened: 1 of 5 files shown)


libcodechecker/cmd/cmd_line_client.py
```diff
@@ -21,6 +21,8 @@
 from libcodechecker.logger import LoggerFactory
 from libcodechecker.output_formatters import twodim_to_str
 from libcodechecker.report import Report
+from libcodechecker import suppress_handler
```
whisperity: Import order is not proper here. `from libcodechecker import X` must precede `from libcodecher.a import b`.

tests/functional/diff/__init__.py
```diff
@@ -16,6 +16,7 @@
 import sys
 import time
 import uuid
+from distutils import dir_util
```
whisperity: Import order is not proper here. Not in this line, and not the line directly below.

tests/functional/diff/__init__.py
```diff
@@ -45,6 +55,10 @@ def setup_package():
     test_project = 'cpp'
 
     test_project_path = project.path(test_project)
+    test_project_path_altered = os.path.join(TEST_WORKSPACE, "cpp_copy")
+    # we create a copy of the test project which we will change
```
whisperity: This is not a full sentence.

tests/projects/cpp/call_and_message.cpp
```diff
@@ -16,6 +16,7 @@ void f(struct S s);
 
 void test1() {
   struct S s;
+  //insert_suppress_here
```
whisperity: Can we put a space between `//` and the keyword?

PR #859, opened 2017-08-31
Storage should work even if no metadata.json exists.

bruntib:


Xazax-hun:
I think it should be easy to test this. You could delete the metadata json after the analyze and try store again.

whisperity:
Well, I mean, it's okay that this is easy to test, but for resillience, I'd rather add a test case for it directly into the code.

Xazax-hun:
I agree with @whisperity

tests/functional/detection_status/test_detection_status.py
```diff
@@ -100,9 +100,21 @@ def _create_source_file(self, version):
         with open(os.path.join(self._test_dir, self._source_file), 'w') as f:
             f.write(self.sources[version])
 
-        codechecker.check(self._codechecker_cfg,
+        codechecker.analyze(self._codechecker_cfg,
+                            'hello',
+                            self._test_dir)
+
+        try:
```
gyorb: Couldn't we create a separate test for the metadata.json removal? At least a separate test case if not a separate test package.

PR #860, opened 2017-08-31
fix skip for header files

### LINKED ISSUE #812: skip file does not apply for headers
`CodeChecker analyze` command can be given a skip file to ignore results from certain files (such as 3rd party libraries).

If a skip file contains a directory such as 
-/skip/all/files/in/directory/*
and this directory contains a header file for example
/skip/all/files/in/directory/lib.h

bug reports from this header file should not be stored by the `store` sub-command and should not be shown by the `parse` sub-command.

gyorb:
Skipping reports in header files included by a c/cpp
file can be done only after the analysis is done and the
plist output is generated.

With this modification if there is a skip file with headers
in it, the reports from the header will be removed from the
generated plist file.

resolves #812

### DIFF
--- a/libcodechecker/analyze/plist_parser.py
+++ b/libcodechecker/analyze/plist_parser.py
@@ -184,3 +184,130 @@ def parse_plist(path):
         LOG.error(ex)
     finally:
         return files, reports
+
+
+def fids_in_range(rng):
+    """
+    Get the file ids from a range.
+    """
+    fids = []
+    for r in rng:
+        for l in r:
+            fids.append(l['file'])
+    return fids
+
+
+def fids_in_edge(edges):
+    """
+    Get the file ids from an edge.
+    """
+    fids = []
+    for e in edges:
+        start = e['start']
+        end = e['end']
+        for l in start:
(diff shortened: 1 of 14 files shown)


libcodechecker/analyze/plist_parser.py
```diff
@@ -184,3 +184,117 @@ def parse_plist(path):
...
+    file_ids_to_remove = []
+
+    for i, f in enumerate(report_data['files']):
+        if skip_handler.should_skip(f):
+            file_ids_to_remove.append(i)
+
+    fids, kept_diagnostics = fids_in_path(report_data, file_ids_to_remove)
+    report_data['diagnostics'] = kept_diagnostics
+
+    remove_file = True
```
whisperity: What does this loop filter exactly? If we mark a file to be removed, but it is still used somewhere else?

gyorb: In the plist there is a `files` array with all the files where a report was found. We can safely remove a file from this list only if that file is not found in any of the other report paths.

PR #862, opened 2017-09-01
Reopened status should turn unresolved instead of staying reopened.

bruntib:


libcodechecker/analyze/store_handler.py
```diff
@@ -338,15 +338,15 @@ def addReport(storage_session,
 
                 if report.detection_status == 'new' and not \
```
whisperity: I know it's well-defined, but I believe it's more pleasing to the eye if statements where `and` and `or` are combined are given some extra parentheses around the `and`. Just to avoid having to think about preference if someone is reading the code.

PR #866, opened 2017-09-02
fix log at analyzer detection

gyorb:
`get_binary_in_path` can return `False` in that case the debug log fails


gyorb:
We should not throw an exception because python could not concatenate a boolean value with a string in the debug log. (printing out false does not give us more information)

The case if there was no analyzer binary was found is handled in the analyzer_types, but the execution does not get there because of the exception before.
https://github.com/Ericsson/codechecker/blob/8d5089709133e59e7ca035e07bc7d2b79b53c330/libcodechecker/analyze/analyzers/analyzer_types.py#L99-L122

whisperity:
I knew one day we should turn every log print into `.format` instead of simple concatenation...

PR #872, opened 2017-09-04
extend profiling options

gyorb:
* new Timer class which can be used to measure code blocks
* print time in seconds for all of the measurements

### DIFF
--- a/libcodechecker/profiler.py
+++ b/libcodechecker/profiler.py
@@ -19,6 +19,23 @@ from libcodechecker.logger import LoggerFactory
 LOG = LoggerFactory.get_new_logger('PROFILER')
 
 
+class Timer(object):
+    """
+    Simple timer context manager
+    to measure code block execution time.
+    """
+    def __init__(self, block_name=''):
+        self.block_name = block_name
+
+    def __enter__(self):
+        self.before = datetime.datetime.now()
+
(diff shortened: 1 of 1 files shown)


libcodechecker/profiler.py
```diff
@@ -19,6 +19,25 @@
...
+    def __init__(self, block_name=''):
+        self.before = None
+        self.after = None
+        self.block_name = block_name
+
+    def __enter__(self):
+        self.before = datetime.datetime.now()
+
+    def __exit__(self, type, value, traceback):
+        after = datetime.datetime.now()
```
Xazax-hun: I think either this should be `self.after` or `self.after` should not be initialized.

libcodechecker/profiler.py
```diff
@@ -19,6 +19,25 @@
 LOG = LoggerFactory.get_new_logger('PROFILER')
 
 
+class Timer(object):
+    """
+    Simple timer context manager
+    to measure code block execution time.
+    """
+    def __init__(self, block_name=''):
+        self.before = None
```
Xazax-hun: Is it idiomatic to initialize these members in python?

libcodechecker/profiler.py
gyorb: I've cleaned it up.

PR #873, opened 2017-09-04
Use --url instead of --host/--port for login and product command-line too

whisperity:
> Merge #857 first!


whisperity:
I fixed the merge conflict in the documentation.

PR #879, opened 2017-09-05
Auto-detect compiler target and add it to the analyze command.

rnkovacs:
Inclusion of target-specific defines is disabled.

### DIFF
--- a/libcodechecker/analyze/analyzers/analyzer_clangsa.py
+++ b/libcodechecker/analyze/analyzers/analyzer_clangsa.py
@@ -88,9 +88,6 @@ class ClangSA(analyzer_base.SourceAnalyzer):
 
             analyzer_cmd = [config.analyzer_binary]
 
-            analyzer_cmd.extend(self.buildaction.compiler_defines)
-            analyzer_cmd.extend(self.buildaction.compiler_includes)
-
             if len(config.compiler_resource_dir) > 0:
                 analyzer_cmd.extend(['-resource-dir',
                                      config.compiler_resource_dir,
@@ -163,6 +160,8 @@ class ClangSA(analyzer_base.SourceAnalyzer):
             analyzer_cmd.append(config.analyzer_extra_arguments)
 
             analyzer_cmd.extend(self.buildaction.analyzer_options)
+
+            analyzer_cmd.extend(self.buildaction.compiler_includes)
 
             analyzer_cmd.append(self.source_file)
 
--- a/libcodechecker/analyze/log_parser.py
+++ b/libcodechecker/analyze/log_parser.py
@@ -55,78 +55,52 @@ def get_compiler_includes(compiler, extra_opts=None):
             if line.startswith(end_mark):
                 do_append = False
             if do_append:
-                include_paths.append("-I"+line)
+                include_paths.append("-I" + line)
             if line.startswith(start_mark):
                 do_append = True
 
     except OSError as oerr:
-        LOG.error("Cannot find include paths:" + oerr.strerror+"\n")
+        LOG.error("Cannot find include paths: " + oerr.strerror + "\n")
     return include_paths
 
 
-# Certain includes are defined by GCC to a GCC internal "macro" never exported
-# by the compiler, such as
-# #define __has_include(STR) __has_include__(STR)
-# which causes Clang to fail in certain cross-compiler and cross-architecture
-# setups due to blabla__ not being defined by Clang in a way GCC does it.
-#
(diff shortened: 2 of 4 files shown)


libcodechecker/analyze/analyzers/analyzer_clangsa.py
```diff
@@ -164,6 +161,9 @@ def construct_analyzer_cmd(self, res_handler):
 
             analyzer_cmd.extend(self.buildaction.analyzer_options)
 
+            # analyzer_cmd.extend(self.buildaction.compiler_defines)
```
Xazax-hun: I prefer to remove code rather than check in commented out code.

libcodechecker/analyze/log_parser.py
```diff
@@ -60,9 +60,34 @@ def get_compiler_includes(compiler, extra_opts=None):
...
+    LOG.debug("Retrieving target platform information via '" + cmd + "'")
+    target = ""
+
+    try:
+        proc = subprocess.Popen(shlex.split(cmd),
+                                stdin=subprocess.PIPE,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+
+        out, err = proc.communicate("")
```
whisperity: Is the output written to `stderr`? If `out` is not used anywhere, don't introduce a variable for it: `_, err = proc.communicate()`.

rnkovacs:
Removed code connected to `compiler_defines` and the unnecessary `out` variable.

libcodechecker/analyze/log_parser.py
```diff
@@ -196,12 +221,11 @@ def parse_compile_commands_json(logfile, add_compiler_defaults=False):
...
                             extra_opts.append(comp_opt)
 
-                compiler_defines[results.compiler] = \
-                    get_compiler_defines(results.compiler, extra_opts)
                 compiler_includes[results.compiler] = \
                     get_compiler_includes(results.compiler, extra_opts)
-            action.compiler_defines = compiler_defines[results.compiler]
+                compiler_target = get_compiler_target(results.compiler)
             action.compiler_includes = compiler_includes[results.compiler]
+            action.analyzer_options.append("--target=" + compiler_target)
```
whisperity: Let's introduce an empty newline before this function to make this code better readable.

libcodechecker/analyze/log_parser.py
```diff
@@ -60,9 +60,34 @@ def get_compiler_includes(compiler, extra_opts=None):
...
-        LOG.error("Cannot find include paths:" + oerr.strerror+"\n")
+        LOG.error("Cannot find include paths: " + oerr.strerror + "\n")
     return include_paths
 
+def get_compiler_target(compiler):
+    """
+    Returns the target triple of the given compiler as a string.
+    """
+    target_label = "Target:"
+    cmd = compiler + " " + "-v"
```
whisperity: No need to use two strings here, and this is a machine string and not a human string.

```py
    cmd = compiler + ' -v'
```

libcodechecker/analyze/log_parser.py
```diff
@@ -55,78 +55,40 @@ def get_compiler_includes(compiler, extra_opts=None):
...
-# Thus, we ignore these "defines" and let Clang fall back to its own
-# __has_include definition. List taken from the "extension" list of Clang and
-# the GCC extension list referenced therein:
-# https://clang.llvm.org/docs/LanguageExtensions.html
-IGNORED_DEFINES = frozenset(
-    ['__has_include'])  # Matches __has_include and __has_include_next.
-
-
-def get_compiler_defines(compiler, extra_opts=None):
+def get_compiler_target(compiler):
```
dkrupp: if the compiler is not gcc return an empty string.
For example for clang compilers this function will always return the host architecture as opposed to the build target.

libcodechecker/analyze/log_parser.py
```diff
@@ -196,13 +158,13 @@ def parse_compile_commands_json(logfile, add_compiler_defaults=False):
...
 
-                compiler_defines[results.compiler] = \
-                    get_compiler_defines(results.compiler, extra_opts)
                 compiler_includes[results.compiler] = \
                     get_compiler_includes(results.compiler, extra_opts)
-            action.compiler_defines = compiler_defines[results.compiler]
+                compiler_target = get_compiler_target(results.compiler)
             action.compiler_includes = compiler_includes[results.compiler]
 
+            action.analyzer_options.append("--target=" + compiler_target)
```
dkrupp: if compiler_target is empty do not append.

PR #880, opened 2017-09-05
Fix a storage bug

### LINKED ISSUE #878: CTU storage of Xerces fails
When analyzing Apache Xerces 3.1 with `--ctu`, for some reason, the created ZIP is bogus.
The client properly sends the data through, and on the server's log, the following data is printed:

```
[14:00] - 127.0.0.1:55233 -- [Anonymous] POST /Xerces/CodeCheckerService
  File "/home/compass/srv/CodeCheckerv6/workspace/install/CodeChecker/lib/python2.7/gencodechecker/codeCheckerDBAccess/codeCheckerDBAccess.py", line 2478, in 
process_massStoreRun
    result.success = self._handler.massStoreRun(args.run_name, args.version, args.zipfile, args.force)
  File "/home/compass/srv/CodeCheckerv6/workspace/install/CodeChecker/lib/python2.7/libcodechecker/profiler.py", line 45, in release_wrapper
    res = function(*args, **kwargs)
  File "/home/compass/srv/CodeCheckerv6/workspace/install/CodeChecker/lib/python2.7/libcodechecker/server/client_db_access_handler.py", line 2355, in massStor
eRun
    store_handler.collect_paths_events(report, file_ids)
  File "/home/compass/srv/CodeCheckerv6/workspace/install/CodeChecker/lib/python2.7/libcodechecker/analyze/store_handler.py", line 106, in collect_paths_event
s
    file_path = files[event['location']['file']]
IndexError: list index out of range
[14:02] - Session timeout for run 4
```

Apparently, something in the ZIP is missing, a file is not created or something.

> Note, that this analysis is done into a server where "normal" Xerces analysis is already done, so there are some files already present on the server.

----

Moreover, the client prints the following:

```
[14:00] - ----==== Summary ====----
[14:00] - Total compilation commands: 1262
[14:00] - Successfully analyzed
[14:00] -   clangsa: 631
[14:00] -   clang-tidy: 631
[14:00] - ----=================----
[14:00] - Analysis length: 796.202292919 sec.
[14:00] - Analysis finished.
[14:00] - To view results in the terminal use the "CodeChecker parse" command.
[14:00] - To store results use the "CodeChecker store" command.
[14:00] - See --help and the user guide for further options about parsing and storing the reports.
[14:00] - ----=================----
[14:00] - Storing analysis results for run 'Xerces_3.1.1_CTU'
[14:02] - Storage failed: Internal error
```

What is a big problem, however, that even though `Storage failed: Internal error` is printed, the **`store` command exits with `0`**, so scripting this tool at this point is problematic.

bruntib:
Fixes #878
In the CTU parsing there happened to be a path which goes threw a file
twice, including another file between these two. In the previous
implementation we had a dictionary which contained the files as keys,
but according to the precondition of `collect_path_events()` function
the keys of this dictionary have to be the file list in the same order
as in the plist file including multiple occurrences.

### DIFF
--- a/libcodechecker/analyze/store_handler.py
+++ b/libcodechecker/analyze/store_handler.py
@@ -53,9 +53,17 @@ def metadata_info(metadata_file):
     return check_commands, check_durations, skip_handlers
 
 
-def collect_paths_events(report, file_ids):
-    files = file_ids.keys()
+def collect_paths_events(report, file_ids, files):
+    """
+    This function creates the BugPathPos and BugPathEvent objects which belong
+    to a report.
 
+    report -- A report object from the parsed plist file.
+    file_ids -- A dictionary which maps the file paths to file IDs in the
+                database.
+    files -- A list containing the file paths from the parsed plist file. The
+             order of this list must be the same as in the plist file.
+    """
     bug_paths = []
     bug_events = []
 
--- a/libcodechecker/server/client_db_access_handler.py
+++ b/libcodechecker/server/client_db_access_handler.py
@@ -10,7 +10,6 @@ Handle Thrift requests.
 import base64
 import codecs
 from collections import defaultdict
-from collections import OrderedDict
 import datetime
 import json
 import os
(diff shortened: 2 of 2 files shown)


dkrupp:
LGTM

PR #896, opened 2017-09-11
Remove support for old clang versions.

### LINKED ISSUE #607: Drop support for guessing check name
For clang version older than 3.7 codechecker tries to guess the check name based on error message. I think we should not support old clang versions indefinitely. So the code for this workaround should be deleted in the next major release.

Xazax-hun:
Closes #607.


README.md
```diff
@@ -225,15 +225,10 @@ Important limitations with older Clang versions
...
+are not supported.
 
-If your installed Clang does not support these features you will see the
-following debug messages in your log:
-
-> Check name wasn't found in the plist file.
-
- * Use clang `>= 3.7` or trunk `r228624` &mdash; otherwise CodeChecker makes
-   a guess based on the report message.
+If you have clang 3.7 installed you might see the following warning message:
```
whisperity: Clang `3.7`

README.md
```diff
@@ -225,15 +225,10 @@ Important limitations with older Clang versions
 -----------------------------------------------
 
 CodeChecker requires some features from LLVM/Clang to work properly which are
-not available in the `3.6` or earlier releases.
+not available in the `3.6` or earlier releases. Thus those versions of clang
```
whisperity: The two sentences should be joined, or rewritten something like this:

> Clang `3.6` or earlier releases are **NOT** supported due to relying on features not available in these releases.

libcodechecker/analyze/plist_parser.py
```diff
@@ -46,67 +43,15 @@
...
-
-    return checker_message_map[closest_msg]
-
-
 def get_checker_name(diagnostic):
     """
     Check if checker name is available in the report.
     Checker name was not available in older clang versions before 3.7.
     """
     checker_name = diagnostic.get('check_name')
```
gyorb: Shouldn't we set some default value if there is no `check_name` in the plist?
We will return `None` if there is no `check_name`.

Xazax-hun: Done.

PR #898, opened 2017-09-11
Add-compiler-defaults improvements

rnkovacs:
Take language and sysroot into consideration when adding default
includes to the analyze command.

### DIFF
--- a/libcodechecker/analyze/log_parser.py
+++ b/libcodechecker/analyze/log_parser.py
@@ -27,7 +27,7 @@ COMPILE_OPTS_FWD_TO_DEFAULTS_GETTER = frozenset(
      '^-std=.*'])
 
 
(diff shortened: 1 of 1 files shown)


libcodechecker/analyze/log_parser.py
```diff
@@ -37,8 +37,11 @@ def get_compiler_includes(compiler, extra_opts=None):
     if extra_opts is None:
         extra_opts = []
 
-    # What if not c++?
-    cmd = compiler + " " + ' '.join(extra_opts) + " -E -x c++ - -v "
+    sysroots = [item for item in compile_opts if item.startswith("--sysroot=")]
```
Xazax-hun: Is it guaranteed to have sysroot? If no, we might index into an empty array.

PR #911, opened 2017-09-13
Git ignore Vim temporary .swp flies

bruntib:


whisperity:
Be careful though, `.swp` is not the only extension Vim swap files can have! It starts at `swp` and then goes all the way to `swa` if multiple exists under the same name (e.g. if the Vim process crashed).

Also: should these developer environment specific rules be in the `.gitignore` of *the repository*? If you are using Vim, you can perfectly configure `.git/info/exclude` (the **local** `gitignore` for your repository) to ignore these files.

PR #914, opened 2017-09-13
Fixing root.user file deletion exception

bruntib:
Initially `root.user` file doesn't exist. If the server is started with
`--reset-root` flag then this file is removed which produces an
exception if the file doesn't exist.

### DIFF
--- a/libcodechecker/libhandlers/server.py
+++ b/libcodechecker/libhandlers/server.py
@@ -426,9 +426,13 @@ def main(args):
         GenericSuppressHandler(None, False)
 
     if 'reset_root' in args:
-        os.remove(os.path.join(args.config_directory, 'root.user'))
-        LOG.info("Master superuser (root) credentials invalidated and "
-                 "deleted. New ones will be generated...")
+        try:
+            os.remove(os.path.join(args.config_directory, 'root.user'))
+            LOG.info("Master superuser (root) credentials invalidated and "
+                     "deleted. New ones will be generated...")
+        except OSError:
+            # File doesn't exist.
+            pass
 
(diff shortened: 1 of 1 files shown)


libcodechecker/libhandlers/server.py
```diff
@@ -426,7 +426,11 @@ def main(args):
         GenericSuppressHandler(None, False)
 
     if 'reset_root' in args:
-        os.remove(os.path.join(args.config_directory, 'root.user'))
+        try:
+            os.remove(os.path.join(args.config_directory, 'root.user'))
+        except OSError:
+            # File doesn't exist.
+            pass
         LOG.info("Master superuser (root) credentials invalidated and "
```
whisperity: If the root user doesn't exist, then nothing is "invalidated", because there is nothing to invalidate in the first place. I would go with moving the log line into the `try` statement too.

So if `server --reset-root` is the first start then the behaviour is the same as if `--reset-root` was not given at all.

libcodechecker/libhandlers/server.py
```diff
@@ -426,9 +426,12 @@ def main(args):
...
 
     if 'reset_root' in args:
-        os.remove(os.path.join(args.config_directory, 'root.user'))
-        LOG.info("Master superuser (root) credentials invalidated and "
-                 "deleted. New ones will be generated...")
+        try:
+            os.remove(os.path.join(args.config_directory, 'root.user'))
+        except OSError:
+            # File doesn't exist.
+            LOG.info("Master superuser (root) credentials invalidated and "
```
Xazax-hun: Do we want to write this message only if the file did not exist?

PR #916, opened 2017-09-13
Can query ldap groups can now be queried for a user to perform group …

### LINKED ISSUE #807: fetch ldap groups for authenticated users

dkrupp:
…based access control.

A bug was fixed:the authentication token cookie was not always set in the response header
which caused frequent re-authnentications.

Closes #807.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -191,19 +191,17 @@ servers as it can elongate the authentication process.
 
  * `groupPattern`
 
-   Group query pattern used. Must be a valid LDAP query expression.
-
- * `groupMemberPattern`
-
-  Group member pattern will be combined with the group patten to query user
-  for ldap group membership. `$USERDN$` will be automatically replaced by the
-  queried user account DN.
-
-  Example configuration: `(member=$USERDN$)`
+   Group query pattern used LDAP query expression to find the group objects
+   a user is a member of. It must contain a `$USERDN$` pattern. 
+   `$USERDN$` will be automatically replaced by the queried user account DN.
+   
+ * `groupNameAttr`
+ 	
+   The attribute of the group object which contains the name of the group. 
 
  * `groupScope`
 
-  Scope of the search performed. (Valid values are: `base`, `one`, `subtree`)
+   Scope of the search performed. (Valid values are: `base`, `one`, `subtree`)
 
 ~~~{.json}
 "method_ldap": {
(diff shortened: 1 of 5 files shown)


docs/authentication.md
```diff
@@ -191,15 +191,13 @@ servers as it can elongate the authentication process.
...
-  queried user account DN.
-
-  Example configuration: `(member=$USERDN$)`
+   Group query pattern used LDAP query expression to find the group objects
+   a user is a member of. It must contain a `$USERDN$` pattern. 
+   `$USERDN$` will be automatically replaced by the queried user account DN.
+   
+ * `groupNameAttr`
+ 	
+ 	The attribute of the group object which contains the name of the group.
```
whisperity: Keep the indentation in this line.

libcodechecker/libauth/cc_ldap.py
```diff
@@ -307,17 +341,36 @@ def auth_user(ldap_config, username=None, credentials=None):
...
+                      'Its value must be the name'
+                      'attribute of the group.')
+            return []
 
-        member_pattern = member_pattern.replace('$USERDN$', user_dn)
-        member_query = '(& ' + group_pattern + member_pattern + ')'
+        # Attribute name must be ascii encoded
+        group_name_attr = group_name_attr.encode('ascii',
+                                                 'ignore')
+        attr_list = []
```
whisperity: This list and the `append()` call can be simplified, as you only append once to the list:

```py
attr_list = [group_name_attr]
```

libcodechecker/libauth/cc_ldap.py
```diff
@@ -307,17 +341,36 @@ def auth_user(ldap_config, username=None, credentials=None):
...
+        LOG.debug("Performing LDAP search for group:" + group_pattern +
+                  "Group Name Attr:" + group_name_attr)
+
+        with ldap_error_handler():
+            group_result = connection.search_s(group_base,
+                                               group_scope,
+                                               group_pattern,
+                                               attr_list)
+        groups = []
+        for g in group_result:
```
gyorb: `group_result` will be undefined if there is some error in the `search_s`

libcodechecker/libauth/cc_ldap.py
```diff
@@ -307,17 +341,36 @@ def auth_user(ldap_config, username=None, credentials=None):
...
+            groups.append(g[1][group_name_attr][0])
```
gyorb: Could you add some example what is the expected group format which is processed here?

libcodechecker/server/client_db_access_server.py
```diff
@@ -127,6 +128,17 @@ def __check_auth_in_request(self):
 
         return success
 
+    def end_headers(self):
+        # Sending the authentication cookie
+        # in every response if any.
```
gyorb: I would add some extra comment here too, why this is needed.

bruntib:
Python dependencies for LDAP are listed in `.ci/auth_requirements`. No makefile installs these dependencies and no documentation mentions that it is necessary.

PR #922, opened 2017-09-14
Compiler detection improvements. Fixes #921

### LINKED ISSUE #921: add-compiler-defaults problems
- [ ]  --add-compiler-defaults option should cut out the "sysroot" arg from the original build action and call the cross compiler detection in (log_parser.py)

- [ ]  auto-detected include paths should be added by -isystem to clang (instead of -I) to get at the end of the inclusion list

- [ ] clang tidy does not get the auto-detected target and include pahts, however it should

Xazax-hun:
Fixes #921


Xazax-hun:
I think we should not do it now. Right now the release is quite close, so we should only do minimal changes to the argument handling and do the review after 6.0 was released.

PR #923, opened 2017-09-14
General cleanups

### LINKED ISSUE #933: Revamp performance test
There is already a performance test script in: performance/run_performance_test.py
How does this compare to the pull request #893?
Maybe we should document this somewhere.

Also note that the performance test above is not working right now, it was not updates for the new product system for instance.

Xazax-hun:
Clean up unused environment variables, context base class and package configs.
This also resolves #933.


config/config.json
```diff
@@ -2,7 +2,6 @@
   "environment_variables": {
       "env_package_root" : "CC_PACKAGE_ROOT",
       "env_verbose_name" : "CC_VERBOSE_LEVEL",
-      "env_alchemy_verbose_name" : "CC_ALCHEMY_LOG_LEVEL",
```
gyorb: `docs/user_guide.md` should be updated that this option is removed.

config/config.json
```diff
@@ -2,7 +2,6 @@
   "environment_variables": {
       "env_package_root" : "CC_PACKAGE_ROOT",
       "env_verbose_name" : "CC_VERBOSE_LEVEL",
```
whisperity: Is this used? I think commands moved to use the `--verbose` option. @gyorb ?

config/config.json
```diff
@@ -15,7 +14,6 @@
   },
   "package_variables": {
       "default_db_username" : "codechecker",
```
whisperity: I think this is also unused.

config/config.json
gyorb: Yes, `--verbose` should be used right now to set log levels.

config/config.json
gyorb: With the new config/run databases and user handling this can be removed too.

libcodechecker/analyze/analyzer_crash_handler.py
```diff
@@ -12,13 +12,14 @@
 
 from libcodechecker.logger import LoggerFactory
 
+LOG = LoggerFactory.get_new_logger('ANALYZER_CRASH_HANDLER')
```
whisperity: Remove the `_`s from the logger name. Most logger tags have been fixed earlier to not include underscores.

libcodechecker/analyze/analyzer_crash_handler.py
```diff
@@ -12,13 +12,14 @@
 
 from libcodechecker.logger import LoggerFactory
```
gyorb: I think the whole analyzer crash handler can be removed, we do not use it.

libcodechecker/cmd/cmd_line_client.py
```diff
@@ -30,10 +30,7 @@
 
 
 class CmdLineOutputEncoder(json.JSONEncoder):
-    def default(self, obj):
-        d = {}
-        d.update(obj.__dict__)
-        return d
+    pass
```
gyorb: Are you sure we can remove this? [json.JSONEncoder.default](https://docs.python.org/3/library/json.html#json.JSONEncoder.default)

config/config.json
```diff
@@ -1,21 +1,15 @@
...
       "cc_logger_compiles" : "CC_LOGGER_GCC_LIKE",
       "ld_preload" : "LD_PRELOAD",
-      "codechecker_enable_check" : "CODECHECKER_ENABLE_CHECK",
-      "codechecker_disable_check" : "CODECHECKER_DISABLE_CHECK",
       "codechecker_workspace" : "CODECHECKER_WORKSPACE"
   },
   "package_variables": {
-      "default_db_username" : "codechecker",
-      "pgsql_data_dir_name" : "pgsql_data",
       "path_dumps_name" : "dumps"
```
whisperity: Is this used? `CodeChecker debug` (which created "dumps") were removed long ago.

libcodechecker/analyze/analyzers/result_handler_plist_to_stdout.py
```diff
@@ -91,8 +91,8 @@ def __print_bugs(self, reports, files):
                 LOG.debug(report + ' is skipped (in ' + f_path + ")")
                 continue
 
-            bug = {'hash_value':
-                   report.main['issue_hash_content_of_line_in_context'],
+            bug = {'hash_value': report.main[
```
whisperity: Can we reformat this a bit better? That indent is huge and ugly. Perhaps a local variable and then putting only that into the map?

libcodechecker/generic_package_context.py
```diff
@@ -52,16 +40,33 @@ def __init__(self, package_root, pckg_layout, cfg_dict):
...
-        super(Context, self).set_env(env_vars)
+        self._package_root = os.environ.get(env_vars['env_package_root'])
+
+        self._codechecker_workspace = os.environ.get('codechecker_workspace')
+
+        try:
+            with open(self.checkers_severity_map_file) as severity_file:
+                self._severity_map = json.load(severity_file)
+        except (IOError, ValueError):
+            LOG.warning(self.checkers_severity_map_file + " doesn't exist or "
```
whisperity: Reformat this as a `"{0} doesn't exist...".format(self.checkers_...)` so that there aren't so many linebreaks.

libcodechecker/server/permissions.py
```diff
@@ -597,7 +595,6 @@ def require_manager(permission, extra_params, user):
 
     return False
 
-# ---------------------------------------------------------------------------
```
whisperity: I would suggest leaving this line in the code. It is better if the "work logic definition" part of this file (permissions classes and methods) and the "permission declaration" part is separated.

libcodechecker/suppress_handler.py
```diff
@@ -199,7 +199,7 @@ def get_suppressed(self):
         if self.__checker_name in suppress_checkers or \
            suppress_checkers == {'all'}:
 
-            file_path, file_name = os.path.split(self.__source_file)
+            _, file_name = os.path.split(self.__source_file)
```
whisperity: `basename` if the path is not needed

libcodechecker/generic_package_context.py
```diff
@@ -52,16 +38,31 @@ def __init__(self, package_root, pckg_layout, cfg_dict):
...
         Get the environment variables.
         """
-        super(Context, self).set_env(env_vars)
+        self._package_root = os.environ.get(env_vars['env_package_root'])
+
+        self._codechecker_workspace = os.environ.get('codechecker_workspace')
+
+        try:
+            with open(self.checkers_severity_map_file) as severity_file:
+                self._severity_map = json.load(severity_file)
```
gyorb: Shouldn't we set a default empty dict for the `self._severity_map` ?

Xazax-hun: Done.

PR #927, opened 2017-09-14
remove unused Config table from the database

gyorb:


whisperity:
Shouldn't we also remove the `getCheckerConfig()` calls from the API and the implementation too?

whisperity:
Talking about these things:

https://github.com/Ericsson/codechecker/blob/162dde6bcce26d65cfff718f86d3e6624b6f4e75/api/v6/report_server.thrift#L265-L268

https://github.com/Ericsson/codechecker/blob/162dde6bcce26d65cfff718f86d3e6624b6f4e75/api/v6/report_server.thrift#L377-L381

PR #928, opened 2017-09-14
Rename "Baseline" and "Newcheck" labels

### LINKED ISSUE #903: Confusing names in diff
In diff view it is confusing to say new, resolved, unresolved, since these are also the categories of detection status.

We should use instead:
Only in A,
Only in B,
Both in A and B

bruntib:
Fixes #903

### DIFF
--- a/www/scripts/codecheckerviewer/BugFilterView.js
+++ b/www/scripts/codecheckerviewer/BugFilterView.js
@@ -939,10 +939,26 @@ function (declare, Deferred, dom, domClass, all, topic, Standby, Button,
               deferred.resolve(Object.keys(CC_OBJECTS.DiffType)
               .map(function (key, index) {
                 var value = CC_OBJECTS.DiffType[key];
+
+                var label;
+                switch (key) {
+                  case 'NEW':
+                    label = 'Only in Newcheck (New)';
+                    break;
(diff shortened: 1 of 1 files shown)


Xazax-hun:
I think Baseline is fine terminology. NewCheck could be replaced by something.
The main problem was to rename categories such as new, resolved, unresolved, since they are the same as detection status categories.

www/scripts/codecheckerviewer/BugFilterView.js
```diff
@@ -603,10 +603,18 @@ function (declare, Deferred, dom, domClass, all, topic, Standby, Button,
               deferred.resolve(Object.keys(CC_OBJECTS.DiffType)
               .map(function (key, index) {
                 var value = CC_OBJECTS.DiffType[key];
+
+                var label;
+                switch (key) {
+                  case 'NEW':        label = 'Only in run set B';       break;
```
Xazax-hun: I would rather change the categories not just the labels.

bruntib:
Finally we chose not to rename all occurrences of the diff types in the API, GUI and command line, but to leave everything as is. On the GUI we just explain what is considered to be new, resolved and unresolved.

PR #935, opened 2017-09-15
Skip zipping .plist files which contain a missing source file in a path

bruntib:


libcodechecker/libhandlers/store.py
```diff
@@ -221,19 +226,28 @@ def collect_file_hashes_from_plist(plist_file):
...
-                LOG.debug("Copying plist '{0}' to ZIP assembly dir..."
-                          .format(input_path))
-                shutil.copy(input_path, report_dir)
-                collect_file_hashes_from_plist(input_path)
+                if collect_file_hashes_from_plist(input_path):
+                    LOG.debug("Copying plist '{0}' to ZIP assembly dir..."
+                              .format(input_path))
+                    shutil.copy(input_path, report_dir)
+                else:
+                    LOG.debug("Skipping '{0}' because it contains "
```
Xazax-hun: This message here is debug and bellow it is a warning. Also, there is a lot of code duplication here. If the input file is one path you could create a one element list with that path, otherwise do the walk of the directory.

PR #949, opened 2017-09-21
Place files directly to ZIP archive

bruntib:
It is unnecessary to copy the files to a temporary directory for zipping
them later. The files can go directly to the .zip file.

### DIFF
--- a/libcodechecker/libhandlers/store.py
+++ b/libcodechecker/libhandlers/store.py
@@ -182,15 +182,6 @@ def res_handler(results):
 
 
 def assemble_zip(inputs, zip_file, client):
-    temp_dir = tempfile.mkdtemp()
-    LOG.debug("Assembling ZIP contents in '{0}':".format(temp_dir))
-
-    report_dir = os.path.join(temp_dir, 'reports')
-    os.makedirs(report_dir)
-
-    source_dir = os.path.join(temp_dir, 'root')
-    os.makedirs(source_dir)
-
     hash_to_file = {}
     # There can be files with same hash,
     # but different path.
@@ -215,7 +206,7 @@ def assemble_zip(inputs, zip_file, client):
         except Exception as ex:
             LOG.error('Parsing the plist failed: ' + str(ex))
 
-    try:
+    with zipfile.ZipFile(zip_file, 'a') as zipf:
         for input_path in inputs:
             input_path = os.path.abspath(input_path)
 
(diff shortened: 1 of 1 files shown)


libcodechecker/libhandlers/store.py
```diff
@@ -215,71 +206,53 @@ def collect_file_hashes_from_plist(plist_file):
...
-                        shutil.copy(plist_file, report_dir)
+                        zipf.write(plist_file, os.path.join('reports', f))
                     else:
                         LOG.warning("Skipping '{0}' because it contains "
                                     "a missing source file."
                                     .format(plist_file))
 
-        necessary_hashes = client.getMissingContentHashes(hash_to_file.keys())
+    necessary_hashes = client.getMissingContentHashes(hash_to_file.keys())
+    with zipfile.ZipFile(zip_file, 'a') as zipf:
```
Xazax-hun: Why do we reopen this zipfile instead of keeping it open?

libcodechecker/libhandlers/store.py
```diff
@@ -215,71 +206,53 @@ def collect_file_hashes_from_plist(plist_file):
...
-        with open(zip_file, 'wb') as target:
-            target.write(compressed)
-
-        LOG.debug("[ZIP] Mass store zip written at '{0}'".format(zip_file))
-    finally:
-        shutil.rmtree(temp_dir)
+
+                zipf.write(f, os.path.join('root', f.lstrip('/')))
+
+    with zipfile.ZipFile(zip_file, 'a') as zipf:
```
Xazax-hun: Same comment as above.

libcodechecker/libhandlers/store.py
```diff
@@ -215,71 +206,53 @@ def collect_file_hashes_from_plist(plist_file):
...
+        zipf.writestr('content_hashes.json', json.dumps(file_to_hash))
+
+    # Compressing .zip file
+    with open(zip_file, 'rb') as source:
```
Xazax-hun: There is no way to write out compressed files in the first place?

libcodechecker/libhandlers/store.py
bruntib: I think it's rather a question of style because it doesn't really cause performance issues. Anyway I can change it.

libcodechecker/libhandlers/store.py
bruntib: Yes, there is, but as far as I know if there is no way to set the compression rate.

gyorb: I think you can [set the compression](https://docs.python.org/2/library/zipfile.html#zipfile-objects).
` class zipfile.ZipFile(file[, mode[, compression[, allowZip64]]])`

bruntib: Here you can state whether a compression should happen, but the compression rate can't be set. And the default is not the best compression.

PR #970, opened 2017-09-22
Unnecessary plist parsing on server side.

bruntib:
On server side at storage the plist files were parsed twice: once for
gathering the files and once for collecting bug paths and events. The
first phase is not necessary since the .zip file already contains the
file paths and hashes.

### DIFF
--- a/libcodechecker/server/client_db_access_handler.py
+++ b/libcodechecker/server/client_db_access_handler.py
@@ -1731,56 +1731,41 @@ class ThriftRequestHandler(object):
         # Storing file contents from plist.
         file_path_to_id = {}
 
-        _, _, report_files = next(os.walk(report_dir), ([], [], []))
-        for f in report_files:
-            if not f.endswith('.plist'):
+        for file_name, file_hash in filename2hash.items():
+            source_file_name = os.path.join(source_root,
+                                            file_name.strip("/"))
+            source_file_name = os.path.realpath(source_file_name)
+            LOG.debug("Storing source file: " + source_file_name)
+
+            if not os.path.isfile(source_file_name):
+                # The file was not in the ZIP file, because we already have the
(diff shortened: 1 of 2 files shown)


libcodechecker/server/client_db_access_handler.py
```diff
@@ -1731,56 +1731,41 @@ def massStoreRun(self, name, tag, version, b64zip, force):
...
-                continue
-
-            LOG.debug("Parsing input file '" + f + "'")
-
-            try:
-                files, _ = plist_parser.parse_plist(
-                    os.path.join(report_dir, f))
-            except Exception as ex:
-                LOG.error('Parsing the plist failed: ' + str(ex))
+        for file_name, file_hash in filename2hash.iteritems():
```
Xazax-hun: Is this python3 compatible? We try to not to introduce incompatibilities because in the future we want to support python 3 as well.

libcodechecker/server/client_db_access_handler.py
```diff
@@ -1731,56 +1731,41 @@ def massStoreRun(self, name, tag, version, b64zip, force):
...
+            source_file_name = os.path.realpath(source_file_name)
+            LOG.debug("Storing source file: " + source_file_name)
+
+            if not os.path.isfile(source_file_name):
+                # The file was not in the ZIP file, because we already have the
+                # content. Let's check if we already have a file record in the
+                # database or we need to add one.
+
+                LOG.debug(file_name + ' not found or already stored.')
+                fid = store_handler.addFileRecord(self.__Session(),
```
Xazax-hun: Will this call diagnose if the hash is not yet in the database? (To catch mistakes in CodeChecker logic.)

bruntib: Yes

PR #978, opened 2017-09-27
Local compare mode to multiple runs

### LINKED ISSUE #962: run name filter in diff command
Allow to specify run filters in the local compare mode (not only a single run name):

CodeChecker cmd diff -b "**tmux_master***” -n ./reports-john-doe --new --url http://localhost:8555/Default

bruntib:
Fixes #962

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -929,7 +929,7 @@ The `--sqlite` (or `--postgresql` and the various `--db-` arguments) can be
 used to specify where the database, containing the analysis reports is.
 
 `--config-directory` specifies where the server configuration files, such as
-[authentication config](docs/authentication.md) is. For example, one can start
+[authentication config](authentication.md) is. For example, one can start
 two servers with two different product layout, but with the same authorisation
 configuration:
 
@@ -945,7 +945,7 @@ resides the `config.sqlite` file, containing the product configuration.
 If the server is started in `--sqlite` mode and fresh, that is, no product
 configuration file is found, a product named `Default`, using `Default.sqlite`
 in the configuration directory is automatically created. Please see
-[Product management](docs/products.md) for details on how to configure products.
+[Product management](products.md) for details on how to configure products.
 
 ### Master superuser and authentication forcing
 
@@ -1137,7 +1137,10 @@ optional arguments:
   -b BASE_RUN, --basename BASE_RUN
                         The 'base' (left) side of the difference: this
                         analysis run is used as the initial state in the
-                        comparison.
+                        comparison. The basename can be a Python regex which
+                        is meant to cover the whole run name. So if you have
+                        run_1_a_name, run_2_b_name and run_2_c_name then
+                        run_.*_[ab]_name selects the first two.
   -n NEW_RUN, --newname NEW_RUN
                         The 'new' (right) side of the difference: this
                         analysis run is compared to the -b/--basename run.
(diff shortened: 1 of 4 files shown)


libcodechecker/cmd/cmd_line_client.py
```diff
@@ -367,24 +357,29 @@ def printReports(client, reports, output_format):
...
         report_dir_mode = True
     else:
-        run_info = check_run_names(client, [args.basename, args.newname])
-        newid = run_info[args.newname][0]
+        run_info = check_run_names(client, [args.newname])
+        newid = run_info[args.newname].runId
+
+    basename_regex = '^' + args.basename.replace('*', '.*') + '$'
+    baseids = [run.runId for run in client.getRunData(None)
+               if re.match(basename_regex, run.name)]
```
dkrupp: I think the run names should be matched case insensitively.

libcodechecker/cmd/cmd_line_client.py
```diff
@@ -367,24 +357,29 @@ def printReports(client, reports, output_format):
...
+        # If newname is a valid directory we assume that it is a report dir and
+        # we are in local compare mode.
         report_dir_mode = True
     else:
-        run_info = check_run_names(client, [args.basename, args.newname])
-        newid = run_info[args.newname][0]
+        run_info = check_run_names(client, [args.newname])
+        newid = run_info[args.newname].runId
+
+    basename_regex = '^' + args.basename.replace('*', '.*') + '$'
```
Xazax-hun: I think if we only support `*`, in the future it might be more efficient to send this pattern to the SQL server and let it do the pattern matching. Maybe this worth a fixme or a todo comment.

docs/user_guide.md
```diff
@@ -1137,7 +1137,8 @@ optional arguments:
   -b BASE_RUN, --basename BASE_RUN
                         The 'base' (left) side of the difference: this
                         analysis run is used as the initial state in the
-                        comparison.
+                        comparison. The basename can be a regex which is meant
+                        to cover the whole run name.
```
gyorb: A simple example would be nice here.

tests/functional/diff/test_diff.py
```diff
@@ -561,3 +561,38 @@ def test_local_compare_res_count_unresovled(self):
         self.assertEqual(count, 5)
         count = len(re.findall(r'\[unix\.Malloc\]', out))
         self.assertEqual(count, 1)
+
+    def test_local_compare_res_count_unresovled_regex(self):
+        """
+        Count the unresolved results with no filter in local compare mode.
```
gyorb: A test case with a filter would be nice too.

bruntib: There already is a test case for filtering above. This test case checks only the regex in baseline name.

docs/user_guide.md
```diff
@@ -1179,6 +1182,9 @@ be compared to each other:
 CodeChecker cmd diff -p 8001 --basename my_project --newname my_new_checkin --new
 ~~~~~~~~~~~~~~~~~~~~~
 
+There is opportunity to compare a run to multiple baselines. You can simply
+provide a regular expression by `-b` flag which covers the required run names.
```
dkrupp: insert here a link that we specifically support this syntax https://docs.python.org/2/library/re.html

libcodechecker/libhandlers/cmd.py
```diff
@@ -165,7 +165,11 @@ def __register_diff(parser):
                         default=argparse.SUPPRESS,
                         help="The 'base' (left) side of the difference: this "
                              "analysis run is used as the initial state in "
-                             "the comparison.")
+                             "the comparison. The basename can be a regex "
```
dkrupp: instead of regex. say python regex syntax

PR #990, opened 2017-10-03
Analyzer arguments can be provided via check subcommand

### LINKED ISSUE #980: synchronize analyze and check subcommands parameter set
check command lacks some parameters that analyze has:

-ctu argument group
-enable-all 
...
All command line arguments of analyze should be available for the check subcommand tooű

bruntib:
Fixes #980

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -121,7 +121,9 @@ After the results has been printed to the standard output, the temporary files
 used for the analysis are cleaned up.
 
 Please see the individual help for `log`, `analyze` and `parse` (below in this
-_User guide_) for information about the arguments of `quickcheck`.
+_User guide_) for information about the arguments which are not documented
+here. For example the CTU related arguments are documented at `analyze`
+subcommand.
 
 ~~~~~~~~~~~~~~~~~~~~~
 usage: CodeChecker check [-h] [-o OUTPUT_DIR] [-q] [-f]
@@ -144,6 +146,9 @@ optional arguments:
                         Store the analysis output in the given folder. If it
                         is not given then the results go into a temporary
                         directory which will be removed after the analysis.
+  -t {plist}, --type {plist}, --output-format {plist}
+                        Specify the format the analysis results should use.
+                        (default: plist)
   -q, --quiet           If specified, the build tool's and the analyzers'
                         output will not be printed to the standard output.
   -f, --force           Delete analysis results stored in the database for the
@@ -163,11 +168,23 @@ log arguments:
 analyzer arguments:
 
   -j JOBS, --jobs JOBS
+  -c, --clean
   -i SKIPFILE, --ignore SKIPFILE, --skip SKIPFILE
   --analyzers ANALYZER [ANALYZER ...]
   --add-compiler-defaults
+  --capture-analysis-output
   --saargs CLANGSA_ARGS_CFG_FILE
   --tidyargs TIDY_ARGS_CFG_FILE
+
+cross translation unit analysis arguments:
+  These arguments are only available if the Clang Static Analyzer supports
+  Cross-TU analysis. By default, no CTU analysis is run when 'CodeChecker
+  analyze' is called.
+
+  --ctu, --ctu-all
+  --ctu-collect
+  --ctu-analyze
+  --ctu-on-the-fly
 
 checker configuration:
 
@@ -325,7 +342,7 @@ optional arguments:
                         directory. (By default, CodeChecker would keep reports
                         and overwrites only those files that were update by
                         the current build command).
-  -n NAME, --name NAME  Annotate the ran analysis with a custom name in the
+  -n NAME, --name NAME  Annotate the run analysis with a custom name in the
                         created metadata file.
   --verbose {info,debug,debug_analyzer}
                         Set verbosity level. (default: info)
@@ -519,7 +536,7 @@ your own risk!**
 
 Even specifying `--enable-all` will **NOT** enable checkers from some special
 checker groups, such as `alpha.` and `debug.`. `osx.` checkers are only enabled
-if CodeChecker is ran on a macOS machine. `--enable-all` can further be
+if CodeChecker is run on a macOS machine. `--enable-all` can further be
 fine-tuned with subsequent `--enable` and `--disable` arguments, for example
 
 ~~~~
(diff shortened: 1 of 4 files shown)


whisperity:
Please update the documentation so that these arguments are shown for `check`'s argument list.
The help need not be copied over (see the current help MD for example) as they are forwarded arguments with the exact same help as in `analyze`.

docs/user_guide.md
```diff
@@ -317,7 +335,7 @@ optional arguments:
                         directory. (By default, CodeChecker would keep reports
                         and overwrites only those files that were update by
                         the current build command).
-  -n NAME, --name NAME  Annotate the ran analysis with a custom name in the
+  -n NAME, --name NAME  Annotate the run analysis with a custom name in the
```
whisperity: I don't think this change is grammatically correct.

bruntib: I think, it is. This 3rd form makes an adjective from the verb. (I only know its Hungarian name :smile: )

whisperity: My bad. Not having actually *learnt* the language for around a decade can make some bad habits burn in. 🙂

docs/user_guide.md
```diff
@@ -160,12 +167,25 @@ log arguments:
 analyzer arguments:
 
   -j JOBS, --jobs JOBS
+  -c, --clean
+  -n, --name
```
gyorb: Why do we have a `--name` here? `check` does not store anything.

libcodechecker/libhandlers/check.py
```diff
@@ -94,13 +96,32 @@ def add_arguments_to_parser(parser):
     Add the subcommand's arguments to the given argparse.ArgumentParser.
     """
 
+    def is_ctu_capable():
```
gyorb: Shouldn't we move this to a separate module, this is used by `check` and `analyze` commands if I'm correct right?

libcodechecker/libhandlers/check.py
```diff
@@ -317,6 +437,10 @@ def __update_if_key_exists(source, target, key):
                           'add_compiler_defaults',
                           'clangsa_args_cfg_file',
                           'tidy_args_cfg_file',
+                          'name',
```
gyorb: What is the name here?

docs/user_guide.md
bruntib: Maybe we don't need this. I inserted this argument here because `analyze` has a `name` flag too and this is just a wrapper script which runs `analyze` among others. By the way `analyze` uses `name` argument to write this name to `metadata.json` which I think is not used anywhere. We could remove `name` from `analyze` too, but that might break some scripts which use this if any.

libcodechecker/libhandlers/check.py
bruntib: This is the `name` command line argument. I don't really understand why exactly are these listed here.

whisperity: This list contains the argument which **don't** have any default value, neither in `check` nor in the wrapped subcommand. This list contains the variable names (not the optstrings).

The reason behind this is to keep the behaviour of `check` and the wrapped commands together: if the wrapped command does not provide a default value: `default=argparse.SUPPRESS`, meaning that if the optstring is missing, the resulting `Namespace` -- the argparse parsed argument dict -- **won't contain the key** for the argument in particular), then the wrapped command should not have a value set in the `args` dict either.

The wrapped commands are called directly on their `main(args)` method with an `args` specifically constructed from `check`'s `args` based on which argument should be forwarded where.

whisperity: So if you say `check --saargs ./a.txt` but do not specify `--tidyargs`, the wrap to analyze will emulate the behaviour of `analyze --saargs ./a.txt` without `--tidyargs`: just as how `tidy_args_cfg_file` (the `dest` for `--tidyargs`) is missing from `check`'s `args`, it won't be put into `analyze`'s `args` either.

The method which does this translation is basically a `setattr()` call, I think defined in the local scope of this wrapper method at the top.

docs/user_guide.md
whisperity: If a `--name` is given to `analyze`, it will be written into the `metadata.json`. It **is** used. `store` runs over the given report directories, eating up the `metadata.json`s therein, and construct the name for the stored analysis automatically.

https://github.com/Ericsson/codechecker/blob/a096bccae947f8168fceb4ce108efba8354d1ba6/libcodechecker/libhandlers/store.py#L273-L282

Technically, agreeing with @gyorb: `check` does no longer wrap `store`, so for a full-fledged `check` execution, `--name` is not needed.

Version 6 was rolled **without** `--name` in check, so there is no need to introduce it. (The only reason would be "preparing" an output_dir which contains an automatically `store`-able structure.) Old scripts which use `-n/--name` for `check` are using v5 syntax, which means that they **should** break, as that invocation is not compatible with the current behaviour.

gyorb: @bruntib @whisperity I've checked and the analyze  and it has the `--name` argument too, which will write the given name (run name) into the metadata so at the store you do not have to give a value for it.
Based on the help message of the store command if a name is detected in the `metadata.json` it will be used. Does this work still this way?

whisperity: Yes, `analyze` and `store` works that way. (I posted the revelant code snipped from `store` above.) But that doesn't imply `check` needs to have a `--name` argument. Due to it not having a use in a full `check` invocation, I'd rather not have it wrapped.

docs/user_guide.md
```diff
@@ -163,12 +168,24 @@ log arguments:
...
   -j JOBS, --jobs JOBS
+  -c, --clean
   -i SKIPFILE, --ignore SKIPFILE, --skip SKIPFILE
   --analyzers ANALYZER [ANALYZER ...]
   --add-compiler-defaults
+  --capture-analysis-output
   --saargs CLANGSA_ARGS_CFG_FILE
   --tidyargs TIDY_ARGS_CFG_FILE
 
+cross translation unit analysis arguments:
```
gyorb: Please update the user guide with the right output.
```
cross translation unit analysis arguments:
  These arguments are only available if the Clang Static Analyzer supports
  Cross-TU analysis. By default, no CTU analysis is run when 'CodeChecker
  analyze' is called.

  --ctu, --ctu-all      Perform Cross Translation Unit (CTU) analysis, both
                        'collect' and 'analyze' phases. In this mode, the
                        extra files created by 'collect' are cleaned up after
                        the analysis.
  --ctu-collect         Perform the first, 'collect' phase of Cross-TU
                        analysis. This phase generates extra files needed by
                        CTU analysis, and puts them into '<OUTPUT_DIR>/ctu-
                        dir'. NOTE: If this argument is present, CodeChecker
                        will NOT execute the analyzers!
  --ctu-analyze         Perform the second, 'analyze' phase of Cross-TU
                        analysis, using already available extra files in
                        '<OUTPUT_DIR>/ctu-dir'. (These files will not be
                        cleaned up in this mode.)
  --ctu-on-the-fly      If specified, the 'collect' phase will not create the
                        extra AST dumps, but rather analysis will be run with
                        an in-memory recompilation of the source files.
```

PR #1011, opened 2017-10-06
Add printing report hash to parse command steps printing

### LINKED ISSUE #1009: Add the possibility to print the reportHash in "CodeChecker parse"
This could ease the debugging of different issues.

martong:
fixes #1009

### DIFF
--- a/libcodechecker/analyze/analyzers/result_handler_plist_to_stdout.py
+++ b/libcodechecker/analyze/analyzers/result_handler_plist_to_stdout.py
@@ -143,6 +143,7 @@ class PlistToStdout(ResultHandler):
                                                        source_file))
             self.__output.write('\n')
             if self.__print_steps:
+                self.__output.write('  Report hash: ' + report_hash + '\n')
                 self.__output.write('  Steps:\n')
                 for index, event in enumerate(events):
                     self.__output.write(index_format % (index + 1))
--- a/tests/functional/analyze_and_parse/test_files/multi_error.steps.output
+++ b/tests/functional/analyze_and_parse/test_files/multi_error.steps.output
@@ -20,6 +20,7 @@ CHECK#CodeChecker check --build "make multi_error" --output $OUTPUT$ --quiet --a
 [HIGH] multi_error.cpp:3:3: Address of stack memory associated with local variable 'x' returned to caller [core.StackAddressEscape]
   return &x;
   ^
+  Report hash: 9b976691ee1dbc182ca89ecfcae5c8db
   Steps:
     1, multi_error.cpp:10:12: Calling 'foo'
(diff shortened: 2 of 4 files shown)


whisperity:
User guide was not updated with this new option.

Also, I feel like parse is getting a lot of options added to it... What if we just print the hash all the time? Or print it automatically when --print-steps is given, above the bug path list?

Xazax-hun:
Just a minor philosophical question from the user's point of view. 
I was wondering, is the hash something the user is interested in? 
The answer might be yes, since she might want to use this info to suppress a bug.
But if the answer is no for some other reason, we might write out the hash in verbose mode (for debug purposes) and should not introduce a new flag.

whisperity:
@Xazax-hun's idea is even better.

Also: you can no longer really suppress bugs with the bughash format, it is only kept as a backward-compatible feature. I think the consensus was that it's the best when suppressions are in the source code (for which the user only needs the bug's location information, which is printed by `parse` all the time), or via the web browser (with which you also can't suppress via bughash).

dkrupp:
Bug hash may also be used to create a link to a bug to a CodeChecker server. I think too that is it is fine if we print it only when `--print-steps` option is present and we are not introducing a new option for this.

gyorb:
@dkrupp the parse command works only on the report directory yes? Do we want to create a link to a local report directory? There might be no database in this case.

If it is used only for debug purposes I agree with @Xazax-hun to log this information only in debug mode.

libcodechecker/analyze/analyzers/result_handler_plist_to_stdout.py
```diff
@@ -48,6 +48,20 @@ def print_steps(self, value):
         """
         self.__print_steps = value
 
+    @property
+    def print_report_hash(self):
+        """
+        Print the hash of the report.
+        """
+        return self.__print_report_hash
```
gyorb: We should set a default value to this.
Like here:
https://github.com/Ericsson/codechecker/blob/a3d8f812c236bedfcc89a5cb331156e4a4d66ab4/libcodechecker/analyze/analyzers/result_handler_plist_to_stdout.py#L32

martong:
As we discussed with @dkrupp we will print the hash together with the steps.

whisperity:
We definitely have such tests in the "analyze and parse" folder, that has steps output. Or we had a month ago, at least.

martong:
@whisperity what would be the alternatives for `ReportHash`? 
I am thinking about `Report hash`, `report hash`, `report_hash`, `Report Hash`, ... , `RePoRtHaSh`

whisperity:
`Report hash` is fine with me.

PR #1014, opened 2017-10-09
Rename modules in server package to make more sense

whisperity:
> **This is a purely internal change.**

Some server modules from the old era (before all the restructuring) retained names such as `client_db_access_server`. I've renamed and restructured the `libcodechecker.server` package to make more sense. The "server" module in particular not only handles the "client db" stuff, the report and storage server was merged a while back, but we didn't touch this issue so far.

With this change navigating the server folder will be a lot more concise.


gyorb:
@whisperity please rebase your pull request so the osx tests can run too. Thanks!

whisperity:
It has been rebated and tests have run.

PR #1015, opened 2017-10-09
fix unicode value on thrift api

### LINKED ISSUE #1005: CodeChecker store fails (sqlite)
Git tag information    : 6.0.1-8-ga3d8f81   
I am using sqlite. My system: Ubuntu 17.04

I ran the command:
`CodeChecker check -b "make clean;make" -o "/home/peters/codechecker_results"`
which worked fine. At least some .plist files two .json and one .json.lock file have been created in the directory ~/codechecker_results

Then I tried:
`CodeChecker store /home/peters/codechecker_results/ -n test`
which failed with the following error message:
`[13:25] - Storing analysis results for run 'test'`
`[13:25] - Storage failed: 'unicode' does not have the buffer interface`

I executed the store-command with the verbose debug option turned on, which then returns the following log messages:
[store.log](https://github.com/Ericsson/codechecker/files/1362532/store.log)

gyorb:
context.version is read as a unicode value which can not be sent through thrift string type in python2.
The unicode value should be converted to string.

Resolves #1005

### DIFF
--- a/libcodechecker/libhandlers/store.py
+++ b/libcodechecker/libhandlers/store.py
@@ -322,7 +322,7 @@ def main(args):
 
         client.massStoreRun(args.name,
                             args.tag if 'tag' in args else None,
-                            context.version,
+                            str(context.version),
                             b64zip,
                             'force' in args)
 


whisperity:
I think it's because this is an environment specific thing. Maybe on Ubuntu 14 and 16 LTS the value read from the json files were `str` and not `unicode` in type.

I was not able to reproduce the original bug locally either, using 16 LTS and upstream build.

whisperity:
Shouldn't we instead fix that the context module always return a string, just to not run into this issue when context.version is ever used again somewhere?

gyorb:
That is a really good question. It is possible that there is some version mismatch between the thrift compiler and the thrift python packages in the virtualenv.
We are using an older LTS Ubuntu to run the tests.

gyorb:
I thought about that what you mention @whisperity, but I prefer the idea to use unicode everywhere and only encode/decode on the network or file read/write boundaries if needed.

PR #1050, opened 2017-10-17
Add compiler info to result directory

martong:
Add the results of gcc -v like invocations as files to the result dir.
This is needed for later local investigation of any failed analysis.


libcodechecker/analyze/log_parser.py
```diff
@@ -27,13 +28,84 @@
...
     start_mark = "#include <...> search starts here:"
     end_mark = "End of search list."
 
+    include_paths = []
+
+    do_append = False
+    for line in lines.splitlines(True):
+        line = line.strip()
+        if line.startswith(end_mark):
+            do_append = False
```
Xazax-hun: Can there be multiple occurrences of start_mark? If no, could we just break here?

libcodechecker/analyze/log_parser.py
```diff
@@ -47,69 +119,28 @@ def get_compiler_includes(compiler, lang, compile_opts, extra_opts=None):
...
-            line = line.strip().split()
-            if line[0] == target_label:
-                target = line[1]
-
-    except OSError as oerr:
-        LOG.error("Cannot find compiler target: " + oerr.strerror + "\n")
-
-    return target
+    err = get_compiler_err(cmd)
+    dump_compiler_info(output_path, "compiler_target.json",
```
Xazax-hun: Does it make sense to both dump compiler info when getting the target and when getting the includes?

libcodechecker/analyze/log_parser.py
```diff
@@ -27,13 +28,84 @@
...
+
+    for line in lines.splitlines(True):
+        line = line.strip().split()
+        if line[0] == target_label:
+            target = line[1]
+
+    return target
+
+
+def dump_compiler_info(output_path, filename, data):
```
Xazax-hun: If the file already exists, we just append to it.
Is it ensured that multiple concurrent checks (to different result directories) won't interfere?
When will we clean up this file? If we don't do that, this file will just grow with subsequent checks to the same result directory.

libcodechecker/analyze/log_parser.py
```diff
@@ -208,13 +243,13 @@ def parse_compile_commands_json(logfile, add_compiler_defaults=False):
     return actions
 
 
-def parse_log(logfilepath, add_compiler_defaults=False):
+def parse_log(logfilepath, output_path, add_compiler_defaults=False):
```
Xazax-hun: I think `output_path` worth a documentation, what this refers to.

libcodechecker/analyze/log_parser.py
martong: This is the original code for parsing, unfortunately diff shows this as completely new code.
Still, I think we could just break here.

libcodechecker/analyze/log_parser.py
```diff
@@ -47,75 +123,50 @@ def get_compiler_includes(compiler, lang, compile_opts, extra_opts=None):
...
+                           {"compiler": compiler, "target": err})
+    return parse_compiler_target(err)
 
-    except OSError as oerr:
-        LOG.error("Cannot find compiler target: " + oerr.strerror + "\n")
 
-    return target
+def remove_file_if_exists(filename):
+    try:
+        os.remove(filename)  # throws if file does not exsist
```
Xazax-hun: We start comments with capital letters and terminate them with periods.

libcodechecker/analyze/log_parser.py
```diff
@@ -26,14 +28,88 @@
...
+    return target
+
+
+def dump_compiler_info(output_path, filename, data):
+    filename = os.path.join(output_path, filename)
+    append_write = 'w'
+    if os.path.exists(filename):
+        append_write = 'a'
+    with open(filename, append_write) as f:
+        f.write(json.dumps(data))
```
martong: Probably this will not be a proper json in case of multiple compilers.
We have to 

- read the file first with 
- then do a `json.loads` on the content
- then add the new entry
- then write back

martong:
@Xazax-hun, as we discussed two files make the json handling code more compact for our benefit.

libcodechecker/analyze/log_parser.py
```diff
@@ -10,6 +10,8 @@
 import traceback
 import subprocess
 import shlex
+import json
```
whisperity: Import order has been violated here.

libcodechecker/analyze/log_parser.py
```diff
@@ -26,14 +28,90 @@
...
+compiler_includes_dump_file = "compiler_includes.json"
+compiler_target_dump_file = "compiler_target.json"
 
-def get_compiler_includes(compiler, lang, compile_opts, extra_opts=None):
-    """
-    Returns a list of default includes of the given compiler.
-    """
+
+def get_compiler_err(cmd):
+    '''
```
whisperity: `'''` ➡️ `"""`

libcodechecker/analyze/log_parser.py
```diff
@@ -26,14 +28,90 @@
...
 
+compiler_includes_dump_file = "compiler_includes.json"
+compiler_target_dump_file = "compiler_target.json"
 
-def get_compiler_includes(compiler, lang, compile_opts, extra_opts=None):
-    """
-    Returns a list of default includes of the given compiler.
-    """
+
+def get_compiler_err(cmd):
```
whisperity: Also, don't we have a method for arbitrary command invocation and output saving somewhere in `util`? Or am I mixing it up with the `test` library?

libcodechecker/analyze/log_parser.py
```diff
@@ -26,14 +28,90 @@
...
+    for line in lines.splitlines(True):
+        line = line.strip()
+        if line.startswith(end_mark):
+            break
+        if do_append:
+            # On OSX there are framework includes,
+            # where we need to strip the "(framework directory)" string.
+            # For instance:
+            # /System/Library/Frameworks (framework directory)
+            fpos = line.find("(framework directory)")
```
whisperity: `string::find(substr) == -1` can be replaced with `substr not in string`, which is more _Pythonic_.

libcodechecker/analyze/log_parser.py
```diff
@@ -47,75 +125,50 @@ def get_compiler_includes(compiler, lang, compile_opts, extra_opts=None):
...
+                           {"compiler": compiler, "target": err})
+    return parse_compiler_target(err)
 
-    except OSError as oerr:
-        LOG.error("Cannot find compiler target: " + oerr.strerror + "\n")
 
-    return target
+def remove_file_if_exists(filename):
+    try:
+        os.remove(filename)  # Throws if file does not exsist.
```
whisperity: These comments are unneccessary here. Also, I recall we use the following form almost everywhere in the code.

```python
if os.path.isfile(a):  # Or exists()
  os.remove(a)
```

libcodechecker/analyze/log_parser.py
```diff
@@ -47,75 +125,50 @@ def get_compiler_includes(compiler, lang, compile_opts, extra_opts=None):
...
+def parse_compile_commands_json(logfile, output_path=None,
+                                add_compiler_defaults=False):
     import json
     # The add-compiler-defaults is a deprecated argument
     # and we always perform target and include auto-detection.
     add_compiler_defaults = True
     LOG.debug('parse_compile_commands_json: ' + str(add_compiler_defaults))
 
+    if output_path is not None:
+        remove_file_if_exists(os.path.join(output_path,
```
whisperity: Is it always applicable for incremental analysis to remove these files? What if analysis configuration changes?

tests/unit/test_buildcmd_escaping.py
```diff
@@ -38,7 +38,7 @@ def setup_class(cls):
         cls.src_file_name = "main.cpp"
         cls.src_file_path = os.path.join(cls.tmp_dir,
                                          cls.src_file_name)
-        cls.compiler = "g++"
+        cls.compiler = "clang++"
```
whisperity: Why is this change introduced?

tests/functional/analyze/test_analyze.py
```diff
@@ -91,6 +91,71 @@ def __analyze_incremental(self, content_, build_json, reports_dir,
...
 
+    def test_compler_info_files(self):
+        # GIVEN
+        build_json = os.path.join(self.test_workspace, "build_simple.json")
+        reports_dir = os.path.join(self.test_workspace, "reports")
+        source_file = os.path.join(self.test_workspace, "simple.cpp")
+
+        # Create a compilation database.
+        build_log = [{"directory": self.test_workspace,
+                      "command": "gcc -c " + source_file,
```
whisperity: Due to `gcc` not existing on macOS systems, this way of creating a build log on the fly is prone to break macOS tests. See #824.

tests/unit/test_buildcmd_escaping.py
martong: I think this is a bug which is present since a while. `cls.compiler` is called with `--target` and gcc7  exits with an error (return value not equal with 0), however newer gcc like gcc 4.8 just gives a warning and returns with 0.

tests/functional/analyze/test_analyze.py
martong: I thought gcc is a symbolic link to clang on macOS. https://stackoverflow.com/questions/19535422/os-x-10-9-gcc-links-to-clang
Has this changed in systems newer than El Capitan?

libcodechecker/analyze/log_parser.py
martong: It is not a problem. If the analysis configuration changes the compile database still remains the same. Compiler includes and targets do not depend on analysis configuration.

tests/functional/analyze/test_analyze.py
whisperity: I've no macOS machine, so I'm not sure. There was definitely no such thing as `gcc` and `g++` on the Travis macOS instances a few months ago.

libcodechecker/analyze/log_parser.py
martong: there is `util.call_command` but that does not handle the stderr. Also this code is refactored from https://github.com/Ericsson/codechecker/pull/1050/files?diff=split#diff-3edc7a6cc15c42194786d15e8ab5852dL95

libcodechecker/analyze/log_parser.py
martong: Ok, I agree. 
Still, this is again not new code despite it looks like that in the diff; merge conflict resolution brought this in :(

tests/functional/analyze/test_analyze.py
martong: I am not sure if this is really a problem since there are several other uses like this in other test cases:
```
tests/functional/analyze/test_analyze.py|102 col 35| "command": "gcc -c " + source_file,
tests/functional/analyze/test_analyze.py|172 col 35| "command": "gcc -c " + source_file,
tests/functional/analyze/test_analyze.py|205 col 35| "command": "gcc -c " + source_file,
tests/functional/analyze/test_analyze.py|240 col 35| "gcc -c " + source_file)    
tests/functional/analyze/test_analyze.py|303 col 35| "command": "gcc -c " + source_file,
tests/functional/ctu/test_files/buildlog.json|4 col 17| "command": "gcc -c lib.c -o lib.o",  
tests/functional/ctu/test_files/buildlog.json|9 col 17| "command": "gcc -c main.c -o main.o",
```

libcodechecker/analyze/log_parser.py
```diff
@@ -26,14 +27,90 @@
...
+    return target
+
+
+def dump_compiler_info(output_path, filename, data):
+    filename = os.path.join(output_path, filename)
+    all_data = dict()
+    if os.path.exists(filename):
+        with open(filename, 'r') as f:
+            all_data = json.load(f)
+    all_data.update(data)
```
dkrupp: be careful, this will not add a new member in the hash, but update an existing one. Thus you will always have one element in the dict such as `{'target': 'arm', 'compiler': 'gcc-arm'}`

martong: Fixed in a new squashed commit

tests/functional/analyze/test_analyze.py
dkrupp: in this case compilation.json is not created on the fly, but wired in into the test case. So it will not cause problem on Mac.

martong:
We could not squash the merge commit 3df5ecd , so for simplicity we keep the 4 commit.
Next time I will rebase and will not create a separate merge commit.

PR #1060, opened 2017-10-23
Update dependencies

tmsblgh:
I don't know how these changes works on Linux.


whisperity:
Uh... I don't think Thrift 10 will work as we target supporting Ubuntu LTS and they are still not supplying Thrift 0.10 as the upstream version.

We once looked into upgrading Thrift around a month ago but it was deemed infeasible.

tmsblgh:
In this case Thrift version can stay at 0.9.*, BTW why we use three different version?
0.9.1 - in .ci/* files
0.9.2 - in vendor_projects.json file
0.9.3 - in .travis.yaml (If I remember correctly I was who added that line....)

I will modify README file because the downgrade from thrift 0.10.0 to 0.9.3 (or 0.9.1-2) on MacOS need to do two steps:
brew unlink thrift
brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/9d524e4850651cfedd64bc0740f1379b533f607d/Formula/thrift.rb

tmsblgh:
Yes I know these things but I assumed that GitHub does not like this and I did not want to use --force but I can.

whisperity:
@tmsblgh GitHub doesn't like or dislike stuff, it is just a visualiser and access control over a Git supersystem.

You're right that `--force` should never ever be used on repositories others work from. It's dangerous to force-push this repository, as everyone working from it will have to rebase their own stuff, which is usually *Whisperity goes there and rebases stuff for them* :wink:

But on your own repository, on your own branch, as long as these changes are "in development", you can do whatever you like. But in the end, it's better if we don't have 223874 commits merged, in which there are commits which are only there to partially remove earlier changes. It's useless for the project's history.

.ci/basic_python_requirements
```diff
@@ -3,3 +3,4 @@ alembic==0.9.2
 thrift==0.9.1
 portalocker==1.1.0
 psutil==5.2.2
+scan-build==2.0.10
```
gyorb: I would not add `scan-build` to the base requirement. It is required only for OSX now. A separate requirements file would be better.
Which will be added to the OSX install descriptions.

README.md
```diff
@@ -123,16 +123,17 @@ You can turn off SIP on El Capitan this way:
...
 ~~~{.sh}
 # Download and install dependencies.
 brew update
 brew install doxygen gcc git
-brew install homebrew/versions/thrift090
+brew llvm
+
+# Install Thrift 0.9.3 (Note: the general Thrift version is 0.10.0 on macOS High Sierra 10.13)
+brew unlink thrift
+brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/9d524e4850651cfedd64bc0740f1379b533f607d/Formula/thrift.rb
```
gyorb: Couldn't we install thrift [this way](https://github.com/tmsblgh/codechecker/blob/5f035fa55fecefa0a2135fd21b4d9d0f51198c22/.travis.yml#L35) ? In Travis it works.

tmsblgh: You are right we can, just 'brew install homebrew/versions/thrift090' is not working.

.ci/basic_python_requirements
tmsblgh: It is okay for me but I am curious why don't we use 'scan-build' instead of 'LD_Logger'(if I remember correctly to the name) on Linux?

.travis.yml
```diff
@@ -29,18 +29,13 @@ matrix:
 
 before_install:
     # Download macOS specific extra dependencies.
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then git clone https://github.com/llvm-mirror/clang.git ~/llvm --branch master --single-branch --depth 1; fi
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then curl http://releases.llvm.org/3.9.0/clang+llvm-3.9.0-x86_64-apple-darwin.tar.xz -O; fi
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update;               fi
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install thrift@0.9;                fi
+    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
```
gyorb: What is the default clang version in the Travis OSX? Was there an update and we do not have to download it?

tmsblgh: Sorry if it caused any confusion. I used Travis about one and half years ago and don't know how was it updated. I read it somewhere and this is why I am tried it. Travis support did not answered it to me yet.

whisperity: @gyorb https://docs.travis-ci.com/user/reference/osx/#Compilers-and-Build-toolchain There isn't any Clang included in the macOS environment. It was the Ubuntu images that had Clang updated to 3.9 in September.

tmsblgh: @whisperity and in the xcode image? 9.0 & 9.1 is available

whisperity: You can always set up Travis for your own fork and make a custom `.travis.yml` file that tries to use some Clang installed on the system, and ask its version.

Though I fear it might be newer than the Clang we use in the Linux image, which can make testing tricky on bug hashes. (Although between 3.9 and 6.0-trunk, it didn't show any *actual* problems.)

README.md
```diff
@@ -123,23 +122,25 @@ You can turn off SIP on El Capitan this way:
...
-Check out and build LLVM/Clang with extra tools. Follow the [Get Started with
-LLVM/Clang](http://clang.llvm.org/get_started.html) documentation.
+OS X El Capitan 10.11, macOS Sierra 10.12 and macOS High Sierra 10.13.
 
 ~~~{.sh}
 # Download and install dependencies.
 brew update
 brew install doxygen gcc git
-brew install homebrew/versions/thrift090
+brew llvm
```
gyorb: The `clang` compiler is installed by default. Do we need the `llvm` here?

gyorb: Ok, I see, `llvm` will install the `clang-tools-extra` where clang tidy is. In that case the latest points to llvm 5.0 I think we should be in sync with the available clang version. I'm not sure what is the latest available clang version in OSX right now. In Travis I think we should use the same `clang` and `clang-tidy` versions.

Xazax-hun: Does this mean it is ok to merge this as is?

tmsblgh: @gyorb ```Apple LLVM version 9.0.0 (clang-900.0.38)``` from Xcode 9 is based on clang 4.0.

At the moment I am not able to check which version will be installed with ```brew install llvm``` but I will as soon as possible.

Xazax-hun: Hmm, maybe it would be better to force the version of LLVM somehow, similar to thrift. The problem is, we do not want to break the TravisCI when the llvm package is upgraded in the repository.

Issue #799, opened 2017-08-14
Clang-Tidy fixits don't store code neccessary to understand fixit


whisperity:
```
projects/thrift-0.10.0/compiler/cpp/src/thrift/generate/t_rb_generator.cc:91:18: warning: do not use 'compare' to test equality of strings; use the string equality operator instead [misc-string-compare]
      } else if( iter->first.compare("namespaced") == 0) {
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~    ~~
                 iter->first                          "namespaced"
```

Currently, only

    iter->first                          "namespaced"

is stored as the fixit's message (actually, as

    iter->first                          "namespaced" (fixit)

) and showed in the viewer. Clang-tidy outputs the string as above, making use of the surrounding code.

Storing the exact leftside whitespaces, and the full line with the arrow and showing it appropriately in the webGUI would help this issue. Currently, the whitespace between `first` and `"namespaced"` is eliminated. The `plist` contains the whitespace, but the browser gets it as

```
                  
```

instead of equal number of `&nbsp;`s, so the browser eliminates them, resulting in this:

![image](https://user-images.githubusercontent.com/1969470/29270078-4c72aac8-80f5-11e7-9909-736b844a57f3.png)

tmsblgh:
Hi,

I am working on this and I would like to ask your opinion.

Now it's looks like this:
<img width="857" alt="screen shot 2017-10-07 at 14 36 13" src="https://user-images.githubusercontent.com/9061606/31307899-f292b3ac-ab6c-11e7-89bd-96beb1a6b7ca.png">

(1) I would like to move the lines more left till the arrow will under the 'i' character.
(2) The FIXIT label can be on the right side. 

What do you think?

whisperity:
Can you please also provide an example on how it looks in the command-line? (`CodeChecker parse`)

I think it would be better if the labels moved to the left a bit.
The other, even better &ndash; but I think harder to implement &ndash; option is to create a separate tag that somehow fits into the source code better, with less padding. I'm thinking of something that fits as if it were part of the source code, like how the standard output of the analyzer handles it. Because the font sizes and the spacing is so different in these tags and the code viewer...

**But** I'm thinking of something even better, which would make the thing even more kickass. @Xazax-hun, I need your input on how Clang-Tidy works in this regards, but here is what I think:

Considering we already know it is a fix-it tag by some logic...

Considering Clang-Tidy gives us an output like:

```
[...]

PR #1061, opened 2017-10-23
Store fixed code from Clang-Tidy FIXITs

tmsblgh:
For Issue #799 

Please note that these are not the final changes and any comments and suggestions are welcome.

Thanks!

### DIFF
--- a/libcodechecker/analyze/tidy_output_converter.py
+++ b/libcodechecker/analyze/tidy_output_converter.py
@@ -50,24 +50,35 @@ class Message(Note):
     Represents a clang-tidy message with an optional fixit message.
     """
 
-    def __init__(self, path, line, column, message, checker, fixits=None,
+    def __init__(self, path, line, column, message, checker,
+                 code_lines=None, arrow_lines=None, fixits=None,
                  notes=None):
         super(Message, self).__init__(path, line, column, message)
         self.checker = checker
+        self.code_lines = code_lines if code_lines else []
+        self.arrow_lines = arrow_lines if arrow_lines else []
         self.fixits = fixits if fixits else []
         self.notes = notes if notes else []
 
     def __eq__(self, other):
         return super(Message, self).__eq__(other) and \
             self.checker == other.checker and \
+            self.code_lines == other.code_lines and \
+            self.arrow_lines == other.arrow_lines and \
             self.fixits == other.fixits and \
             self.notes == other.notes
 
     def __str__(self):
-        return '%s, checker=%s, fixits=%s, notes=%s' % \
-               (super(Message, self).__str__(), self.checker,
-                [str(fixit) for fixit in self.fixits],
-                [str(note) for note in self.notes])
+        return '%s, checker=%s, code_lines=%s, arrow_lines=%s, ' \
+               'fixits=%s, notes=%s' % (super(Message, self).__str__(),
+                                        self.checker,
+                                        [str(fixed_message)
+                                         for fixed_message
+                                         in self.code_lines],
+                                        [str(arrow_line)
+                                         for arrow_line in self.arrow_lines],
+                                        [str(fixit) for fixit in self.fixits],
+                                        [str(note) for note in self.notes])
 
 
 class OutputParser(object):
@@ -157,8 +168,9 @@ class OutputParser(object):
             match.group('checker').strip())
 
         try:
-            line = next(titer)
-            line = self._parse_code(message, titer, line)
+            line = titer.next()
+            line = self._parse_code_lines(message, titer, line)
+            line = self._parse_arrow_lines(message, titer, line)
             line = self._parse_fixits(message, titer, line)
             line = self._parse_notes(message, titer, line)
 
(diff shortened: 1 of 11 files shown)


whisperity:
Test cases are missing for this. Possibly already existing tests need be modified.

Xazax-hun:
Could you add some screenshots how it looks like? Also please keep the code PEP8 clean.

tmsblgh:
Now it looks like this:

screen shot 2017-10-26 at 17 22 19

Now the code is PEP8 clean but I am still need to create & fix unit tests.

And what you suggested, if we have a 'fixed' code (bubble 1 in the pic) we don't need to show arrow line and the fixit line (bubble 2 & 3 in the pic). I don't know yet where is the best place in the code to solve this, maybe in BugViewer.js.

Xazax-hun:
Also I think fixita should not be numbered since this is not a bug step.

whisperity:
@Xazax-hun Au contraire, I believe those "bubbles" should **not be shown at all**.

Here's why.

This is the current view. I think `(fixed)` should be rephrased, or maybe this "bug step" should be shown as yellow. This "normal" blue colour without any extra annotations isn't the best.
image

**But**

image

These two lines have no meaningful expression power at all in the bug visualisation. They are retrieved from a monospace environment. Remember when it took like 10-15 minutes for the two of us to even figure out what are these things? 😀 

Due to the bubbles' padding, the font different between the code viewer and the annotations, these things can never (or at least only with superfluous effort) could be made so that the arrows and tildes point to the code properly. Along with the _fix_ in the next step.

What I propose: We should only show two lines. First is the `fixed` version in some better colour. Perhaps something green-ish, with an annotation similar to that of `[FIXIT]`, but perhaps worded as `fix`, `fix as:`, `fixed`, ...
And the second line is the actual check message. And that's it. Two lines, just as we handled Tidy fixits initially -- right now we have 4 lines which means for each tidy fixit output we store twice the row count!

whisperity:
My original idea to @tmsblgh in #799 was to actually create a new line **inside** the syntax highlighting... thing... which contains the fixed code, so the same font, paddings, etc, perhaps with a background or some arrows or something, but I understand if that is much, or even unnecessarily, harder to implement.

Xazax-hun:
@whisperity , I think you misunderstood something, we are on the same page removing the redundant data.

whisperity:
Oh. Okay. You just said `should not be numbered` which I thought you still want to show it but without the numbering. 🙁

gyorb:
Will the fixit be parsed and stored into the database as a bug event, so only the UI will handle this differently?
How will the `CodeChecker parse --print-steps` handle the fixits?

tmsblgh:
Just the UI handles it at the moment. @gyorb What do you think about it? Shall we modify the database schema to store these things?

whisperity:
Ah, the age old debate on whether or not the "bug events" should be typed in the DB or not.
@Xazax-hun an input on this? This PR run stale a month ago.

whisperity:
@tmsblgh I think there are two ways to do this:
 1. Let the UI handle the showing of the fixed code.
 2. Modify the *Tidy Plist Parser* logic to store the fixed code in the database. This may or may not require schema modification, but I don't recall it should.

I think the second approach is better: as explained earlier, the `~~~~`s `^token`s are worthless in bug visualisation. They also seem to be dead weight in the database. A Tidy Fix-It contains enough information to "build" the fixed line (am I right about this, @Xazax-hun?). If we give just a bit of extra annotation (perhaps the `(fixed)` is enough for that?) then the database could only contain two bug path elements, the fixed line and the actual checker message. From this, we can
 * on the GUI, show the fixed line and the checker comment (so what is blue and red on your screenshot earlier)
 * in the command-line output, say something like "Don't use compare for string equality. Change line XXX to this: `[line with fixed contents]`.

gyorb:
With our current db schema I would not mix the fixit notes into the `BugPathEvent`. We could handle it more flexible and this information does not belong to a bug path. It belongs to a report and it is optional because only the clang-tidy analyzer can provide these type of information right now.

Beside the fixit information there might be notes from the analyzers in the plist reports what we do not handle right now.
These will be bug path independent information (as the fixits) which can help to understand the report.
I think we should talk about creating two separate tables or one table (for the fixits and notes) and design the new schema based on that.

Xazax-hun:
The problem with having fixit as a BugPathEvent, each client needs to parse each path events to determine if this is a fixit or a proper message to display. I think the API should reflect that if a report hash fixits and what those are. This would make the client side much cleaner. So I am in favor of extending the API. I am also in favor of having a separate table to store the fixits.

tmsblgh:
I did not have as much time as I want so I do not fully understand this and I would like to clarify it.

What is the final decision about this?

At the moment I have changes only in the tidy_output_converter.py and BugViewer.js.

In the python file I added a new "fixed thing" so the plist file contains the "fixed" messages also, before the "arrow_line" and the "fixit" messages. But we do not need the "arrow_line" and the "fixit" anymore, right?
Now it "builds up" the "fixed" from the other two message.

In the js file I need to adapt the things for the final plist file, with/without "arrow_line" and "fixit".

But as I see there is one more thing, the API extending and the new table for the fixits. I need to look in to these :)

whisperity:
I don't think a final decision has been made ever since, due to a lot of other stuff, and holidays going on. But until then, could you snip one or two screenshots about how this feature looks as of now?

Xazax-hun:
API:
I think it would be great to modify the thrift API, so we have a separate FIXIT list in the `ReportDetails` data structure, see the `api` directory. This would make it easier for the clients to consume and display FIXITs.

Web Client:
Could you share a screenshot showing the current state?

Server:
I think it would be better to have a separate table to store FIXITs. Also, if you add a new table you will need to add an alembic migration script.

tmsblgh:
Web Client:
misc-string-compare was removed so I tried to find another example and turned out that the current solution is not good.

This is the current state:

screen shot 2018-01-14 at 14 26 57

The fixed message build from ```arrow``` & ```fix it``` lines but as you can see it is not working in this case. Moreover I need to remove bubbles.

whisperity:
[Buttbuttination and consbreastution](http://www.telegraph.co.uk/news/newstopics/howaboutthat/2667634/The-Clbuttic-Mistake-When-obscenity-filters-go-wrong.html)

However, this might be a logic issue **in the checker**. If the arrow only points to `t` in the type specifier, instead of highlighting the whole thing!

Xazax-hun:
image

The output from clang tidy is correct. So either you are not interpreting the output well or CodeChecker is not converting the tidy output well to plist. Also, misc-string-compare did not disappear, I can still see this among the checks so I do that must be a problem with your setup.

tmsblgh:
Thanks, you were right the problem was with my setup.

The current state:

screen shot 2018-01-25 at 22 59 26

libcodechecker/analyze/tidy_output_converter.py
```diff
@@ -324,6 +354,15 @@ def _create_event_from_note(note, fmap):
             'message': note.message
         }
 
+    @staticmethod
+    def _create_fixit_from_note(note, fmap):
+        return {
+            'kind': 'fixit',
+            'location': PListConverter._create_location(note, fmap),
+            'depth': 0,  # I don't know WTF is this.
```
gyorb: `depth` is used for events to mark how deep they are in the call chain.

libcodechecker/server/api/store_handler.py
```diff
@@ -129,6 +129,37 @@ def collect_paths_events(report, file_ids, files):
...
+                database.
+    files -- A list containing the file paths from the parsed plist file. The
+             order of this list must be the same as in the plist file.
+    """
+    bug_fixits = []
+
+    fixits = filter(lambda i: i.get('kind') == 'fixit', report.bug_path)
+
+    for fixit in fixits:
+        LOG.info('Fixit ... ')
```
gyorb: Debug level logs in the store would be better.

logfile
```diff
@@ -0,0 +1,31 @@
+2018-03-21 23:18:01.145 CET [11568] LOG:  listening on IPv6 address "::1", port 5432
```
gyorb: I do not think we will need this `logfile`

PR #1065, opened 2017-10-25
Workaround for wrong clang-tidy output

### LINKED ISSUE #1064: increasing memory consumption
Parsing the output of the `bugprone-undefined-memory-manipulation` tidy checker causes an infinite loop and consumes more and more memory.

bruntib:
Clang-tidy has an output of the following format:

/path/to/file:<line>:<col> message [checker-name]
code_fragment(42);
^~~~~~~~~~~~~

The output may contain an optional fixit. Several groups of these lines
may follow each other.
In case of bugprone-undefined-memory-manipulation checker it happened
that in the last group only the first line appeared. The tidy output
converter assumed that the following lines are also present. If not then
it was an infinite loop.
Fixes #1064

### DIFF
--- a/libcodechecker/analyze/tidy_output_converter.py
+++ b/libcodechecker/analyze/tidy_output_converter.py
@@ -160,7 +160,7 @@ class OutputParser(object):
 
             return message, line
         except StopIteration:
-            return message, line
+            return message, ''
 
     def _parse_code(self, message, titer, line):
         # Eat code line.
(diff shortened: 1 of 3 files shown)


Xazax-hun:
Could you add a unittest? This would also make it easier to understand this change.

gyorb:
Just a side note, maybe we should report to clang tidy developers or fix the `bugprone-undefined-memory-manipulation` checker. That checker produced this type of report. What do you think @Xazax-hun ?

PR #1069, opened 2017-10-26
Update analyzer mentioned files

martong:
* Since https://github.com/Ericsson/clang/pull/199/files we can track which source files corresponding AST had been loaded before a crash.
Therefore, it is no longer needed to have complex regex matching in `get_analyzer_mentioned_files`.
* We put absolute paths into the failed zip under sources-root, because all analyzer runs are using absolute paths.

### DIFF
--- a/libcodechecker/analyze/analysis_manager.py
+++ b/libcodechecker/analyze/analysis_manager.py
@@ -359,6 +359,9 @@ def check(check_data):
 
                     LOG.debug("Writing dependent files to archive.")
                     for dependent_source in dependencies:
+                        if not os.path.isabs(dependent_source):
+                            dependent_source = \
+                                os.path.abspath(dependent_source)
                         LOG.debug("[ZIP] Writing '" + dependent_source + "' "
                                   "to the archive.")
                         archive_path = dependent_source.lstrip('/')
--- a/libcodechecker/analyze/analyzers/analyzer_clangsa.py
+++ b/libcodechecker/analyze/analyzers/analyzer_clangsa.py
@@ -128,15 +128,13 @@ class ClangSA(analyzer_base.SourceAnalyzer):
                                          '-Xclang', checker_name])
 
             if config.ctu_dir:
-                env = analyzer_env.get_check_env(config.path_env_extra,
-                                                 config.ld_lib_path_extra)
-                triple_arch = ctu_triple_arch.get_triple_arch(self.buildaction,
-                                                              self.source_file,
-                                                              config, env)
                 analyzer_cmd.extend(['-Xclang', '-analyzer-config',
                                      '-Xclang',
-                                     'xtu-dir=' + os.path.join(config.ctu_dir,
-                                                               triple_arch)])
+                                     'xtu-dir=' + self.get_xtu_dir(),
+                                     ])
+                if config.ctu_has_analyzer_display_ctu_progress:
+                    analyzer_cmd.extend(['-Xclang',
+                                         '-analyzer-display-ctu-progress'])
                 if config.ctu_in_memory:
                     analyzer_cmd.extend(['-Xclang', '-analyzer-config',
                                          '-Xclang',
@@ -160,35 +158,38 @@ class ClangSA(analyzer_base.SourceAnalyzer):
             LOG.error(ex)
             return []
 
+    def get_xtu_dir(self):
+        """
+        Returns the path of the xtu directory (containing the triple).
+        """
+        config = self.config_handler
+        env = analyzer_env.get_check_env(config.path_env_extra,
+                                         config.ld_lib_path_extra)
+        triple_arch = ctu_triple_arch.get_triple_arch(self.buildaction,
+                                                      self.source_file,
+                                                      config, env)
+        xtu_dir = os.path.join(config.ctu_dir, triple_arch)
+        return xtu_dir
+
     def get_analyzer_mentioned_files(self, output):
         """
         Parse ClangSA's output to generate a list of files that were mentioned
         in the standard output or standard error.
         """
 
-        # A line mentioning a file in ClangSA's output looks like this:
-        # /home/.../.cpp:L:C: warning: foobar.
-        regex = re.compile(
-            # File path followed by a ':'.
-            '^(?P<path>[\S ]+?):'
-            # Line number followed by a ':'.
-            '(?P<line>\d+?):'
-            # Column number followed by a ':' and a space.
-            # This is optional, as lines beginning with
-            # "In file included from" don't have a column number.
-            '((?P<column>\d+?):\ )?')
+        regex_for_ctu_ast_load = re.compile(
+            "ANALYZE \(CTU loaded AST for source file\): (.*)")
 
         paths = []
 
-        for line in output.splitlines():
-            # Files can also be referenced in a line that
-            # begins as "In file included from /home/...".
-            if line[0] == 'I':
-                line = line.replace("In file included from ", "", 1)
+        xtu_ast_dir = os.path.join(self.get_xtu_dir(), "ast")
 
-            match = re.match(regex, line)
+        for line in output.splitlines():
+            match = re.match(regex_for_ctu_ast_load, line)
             if match:
-                paths.append(match.group('path'))
+                path = match.group(1)
+                if xtu_ast_dir in path:
+                    paths.append(path[len(xtu_ast_dir):])
 
         return set(paths)
 
--- a/libcodechecker/analyze/host_check.py
+++ b/libcodechecker/analyze/host_check.py
@@ -6,6 +6,7 @@
 
 import errno
 import subprocess
+import tempfile
 
 from libcodechecker.logger import LoggerFactory
 
@@ -35,3 +36,26 @@ def check_clang(compiler_bin, env):
             LOG.error(oerr)
             LOG.error('Failed to run: "' + ' '.join(clang_version_cmd) + '"')
             return False
+
+
+def has_analyzer_feature(clang_bin, feature):
+    with tempfile.NamedTemporaryFile() as inputFile:
+        inputFile.write("void foo(){}")
+        inputFile.flush()
+        cmd = [clang_bin, "-x", "c", "--analyze",
+               "-Xclang", feature, inputFile.name, "-o", "-"]
+        LOG.debug('run: "' + ' '.join(cmd) + '"')
+        try:
+            proc = subprocess.Popen(cmd,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE,
+                                    )
+            out, err = proc.communicate()
+
+            LOG.debug("stdout:\n" + out)
+            LOG.debug("stderr:\n" + err)
+
+            return proc.returncode == 0
+        except OSError:
+            LOG.error('Failed to run: "' + ' '.join(cmd) + '"')
+            raise
--- a/tests/functional/ctu/ctu_failure/test_ctu_failure.py
+++ b/tests/functional/ctu/ctu_failure/test_ctu_failure.py
@@ -1,0 +1,176 @@
+#
+# -----------------------------------------------------------------------------
+#                     The CodeChecker Infrastructure
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -----------------------------------------------------------------------------
+""" CTU function test."""
+
+import json
+import os
+import shutil
+import subprocess
+import unittest
+import zipfile
+
+from libtest import env
+from libtest import project
+
+from libcodechecker.analyze import host_check
+
+NO_CTU_MESSAGE = "CTU is not supported"
+NO_DISPLAY_CTU_PROGRESS = "-analyzer-display-ctu-progress is not supported"
+
+
+class TestCtu(unittest.TestCase):
+    """ Test CTU functionality. """
+
+    def setUp(self):
+        """ Set up workspace."""
+
+        # TEST_WORKSPACE is automatically set by test package __init__.py .
+        self.test_workspace = os.environ['TEST_WORKSPACE']
+
+        test_class = self.__class__.__name__
+        print('Running ' + test_class + ' tests in ' + self.test_workspace)
+
+        # Get the CodeChecker cmd if needed for the tests.
+        self._codechecker_cmd = env.codechecker_cmd()
+        self.env = env.codechecker_env()
+        self.report_dir = os.path.join(self.test_workspace, 'reports')
+        os.makedirs(self.report_dir)
+
+        self.test_dir = project.path('ctu_failure')
+
+        # Get if clang is CTU-capable or not.
+        cmd = [self._codechecker_cmd, 'analyze', '-h']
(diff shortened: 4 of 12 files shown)


libcodechecker/analyze/analyzers/analyzer_clangsa.py
```diff
@@ -160,35 +157,35 @@ def construct_analyzer_cmd(self, result_handler):
...
-                line = line.replace("In file included from ", "", 1)
+        xtu_ast_dir = os.path.join(self.get_xtu_dir(), "ast")
 
-            match = re.match(regex, line)
+        for line in output.splitlines():
+            match = re.match(regex_for_ctu_ast_load, line)
             if match:
-                paths.append(match.group('path'))
+                path = match.group(1)
+                assert path.find(xtu_ast_dir) != -1
```
whisperity: Certain python environments can overoptimise code which run with assertions disabled. Also, a Python assert doesn't really carry information. An Exception, and a proper, user-understandable (or at least, via the exception, developer-understandable) error message should be used.

Is this assertion needed in the first place?

Why do we use `find` instead of the `substr in string` syntax?

tests/functional/ctu_failure/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+
+# Test workspace should be initialized in this module.
+TEST_WORKSPACE = None
+
+
+def setup_package():
+    """Setup the environment for testing ctu."""
+
+    global TEST_WORKSPACE
+    TEST_WORKSPACE = env.get_workspace('ctu_failure')
```
whisperity: We have a **lot** of subdirectories under `functional` already. Can't the test module file of this test package instead go as a new python module in the `ctu` folder?

tests/functional/ctu_failure/test_ctu_failure.py
```diff
@@ -0,0 +1,151 @@
...
+        """ Test the failure zip contains the source of imported TU"
+        """
+
+        if not self.ctu_capable:
+            self.skipTest(NO_CTU_MESSAGE)
+
+        output = self.__do_ctu_all(reparse=False,
+                                   extra_args=[
+                                       "--verbose", "debug",
+                                       "-e", "debug.ExprInspection"
```
gyorb: Please add a comment why do we enable/use this checker for this test.

tests/functional/ctu_failure/test_files/lib.c
```diff
@@ -0,0 +1,5 @@
+int foo(int i)
```
gyorb: We have a separate directory for the [test_projects](https://github.com/Ericsson/codechecker/tree/09837984d59daecb1f799290a06eb9a105c6020b/tests/projects) which will be analyzed during the tests.

libcodechecker/analyze/analyzers/analyzer_clangsa.py
```diff
@@ -160,35 +157,35 @@ def construct_analyzer_cmd(self, result_handler):
...
+                                         config.ld_lib_path_extra)
+        triple_arch = ctu_triple_arch.get_triple_arch(self.buildaction,
+                                                      self.source_file,
+                                                      config, env)
+        xtu_dir = os.path.join(config.ctu_dir, triple_arch)
+        return xtu_dir
+
     def get_analyzer_mentioned_files(self, output):
         """
         Parse ClangSA's output to generate a list of files that were mentioned
```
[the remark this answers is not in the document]
whisperity: Technically, yes. It still parses the output, just with a different pattern.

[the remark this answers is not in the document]
Xazax-hun: But it does not collect every mentioned file in the warnings but only files mentioned by a specific log message.

libcodechecker/analyze/analyzers/analyzer_clangsa.py
```diff
@@ -128,15 +128,12 @@ def construct_analyzer_cmd(self, result_handler):
...
-                env = analyzer_env.get_check_env(config.path_env_extra,
-                                                 config.ld_lib_path_extra)
-                triple_arch = ctu_triple_arch.get_triple_arch(self.buildaction,
-                                                              self.source_file,
-                                                              config, env)
                 analyzer_cmd.extend(['-Xclang', '-analyzer-config',
                                      '-Xclang',
-                                     'xtu-dir=' + os.path.join(config.ctu_dir,
-                                                               triple_arch)])
+                                     'xtu-dir=' + self.get_xtu_dir(),
```
gerazo: We've inherited "XTU" name from somewhere else so we made the decision to call everything we do "CTU". Although the legacy parameter here (which was not changed in clang yet) says XTU, we should rather consequently call these CTU in the code. Someone may change it in clang as well in the future. :)

tests/functional/ctu_failure/__init__.py
martong: If I understand the structure of the test dirs correctly then each dir represent a test suite with its own setup and optionally files for the test project. And I think, the ctu_failure directory perfectly fits into this structure.
```
   ▾ delete_runs/            
       __init__.py           
       test_delete_runs.py   
   ▾ detection_status/       
       __init__.py           
       test_detection_status.
   ▾ diff/                   
       __init__.py           
       test_diff.py          
   ▾ func_template/          
       template__init__.py   
       template_test.py      
   ▾ instance_manager/       
       __init__.py           
       test_instances.py     
```

libcodechecker/analyze/analyzers/analyzer_clangsa.py
martong: OK , I removed the assert and replaced with a condition.

tests/functional/ctu_failure/__init__.py
whisperity: I don't think there ever was any structural decision on the functional test folder. The [suppress tests before review status was implemented](https://github.com/Ericsson/codechecker/tree/v5.8/tests/functional/suppress), for example, also tested multiple features at once.

The current layout of the test suite better represents testing individual features or feature sets of CodeChecker. Analyzer handling, analyzer result handling, result querying (report viewer api), diffing, run deletion, product management, etc.

"CTU failure" is not a feature set, it is a new, special test case for CTU. Of course, due to this test being related to analyzer handling, it *could* go into the `analyze` folder, but `CTU` is a bigger, and thus more powerful category.

whisperity: Structurally, nosetest test on the module level, the package level things only handle analysis execution, result storage and server setup.

For example, the current layout is also bad, as every single damn test package runs a full analysis, sometimes on a big project, but at least where it wasn't neccessarry, we cut down on using a simpler project only, thus further quickening up the tests without losing functionality.

martong: Ok, I moved the ctu_failure dir under ctu.

libcodechecker/analyze/analyzers/analyzer_clangsa.py
```diff
@@ -160,35 +157,35 @@ def construct_analyzer_cmd(self, result_handler):
             LOG.error(ex)
             return []
 
+    def get_xtu_dir(self):
+        config = self.config_handler
```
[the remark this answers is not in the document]
martong: Ok, I added a desc.

tests/functional/ctu_failure/test_ctu_failure.py
martong: Ok , I added a comment.

tests/functional/ctu_failure/test_files/lib.c
martong: Ok, I moved it under the projects dir.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -359,6 +359,8 @@ def check(check_data):
 
                     LOG.debug("Writing dependent files to archive.")
                     for dependent_source in dependencies:
+                        if not os.path.isabs(dependent_source):
+                            dependent_source = os.path.abspath(dependent_source)
```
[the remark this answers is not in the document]
martong: The point here is to replicate the host's file system under the sources-root directory.
`dependent_source` contains the list of files reported as dependent by `$CXX -M`, this includes the arg as well.
```
> gcc -E -M -MT __dummy -c main.c
__dummy: main.c /usr/include/stdc-predef.h

> gcc -E -M -MT __dummy -c $(realpath main.c)                                                  
__dummy: /home/egbomrt/WORK/CodeChecker/tests/projects/ctu_failure/main.c \
 /usr/include/stdc-predef.h
```
We assemble the gcc commandline based on the compile database, but the `directory` key from it is not used, therefore the arg will be a relative path (in the example above it is the first case, just `main.c`). 
The cwd did not changed after calling gcc, so we can call `os.path.abspath()` on the arg thus resulting the correct absolute path.

libcodechecker/analyze/analyzers/analyzer_clangsa.py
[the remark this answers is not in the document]
martong: We don't need anymore to catch the warnings. If there would be a warning from a TU, that must had been imported by Clang previously (together with the proper `CTU loaded` log).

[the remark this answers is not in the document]
martong: perhaps a better name could be: `get_analyzer_ctu_loaded_files` ?

libcodechecker/analyze/analyzers/analyzer_clangsa.py
```diff
@@ -128,15 +128,12 @@ def construct_analyzer_cmd(self, result_handler):
...
-                triple_arch = ctu_triple_arch.get_triple_arch(self.buildaction,
-                                                              self.source_file,
-                                                              config, env)
                 analyzer_cmd.extend(['-Xclang', '-analyzer-config',
                                      '-Xclang',
-                                     'xtu-dir=' + os.path.join(config.ctu_dir,
-                                                               triple_arch)])
+                                     'xtu-dir=' + self.get_xtu_dir(),
+                                     '-Xclang',
+                                     '-analyzer-display-ctu-progress',
```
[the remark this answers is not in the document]
martong: Good point. The analysis will fail if the used Clang does not support this switch.
This will cause a problem for CodeChecker in the time window when the CTU patch is accepted but the patch for this switch is still not.
I hope this time window will be short, since the patch for `-analyzer-display-ctu-progress` is very tiny. Still this is a valid danger.
We have the following options:
* Create a framework to discover feature capabilities of the used Clang (this is a huge work, counting in men weeks I estimate)
* If the command fails, then retry without `-analyzer-display-ctu-progress` switch
* Any other ideas?

libcodechecker/analyze/analyzers/analyzer_clangsa.py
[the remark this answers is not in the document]
martong: We can't just rename this function, because this is an inherited abstract function from `SourceAnalyzer`. And it is used in `analyzer_clang_tidy.py` as well.

libcodechecker/analyze/analyzers/analyzer_clangsa.py
gyorb: A `FIXME` or a comment would be nice here, that this parameter needs to be changed, or we should create a separate issue.

martong:
@Xazax-hun  @gyorb @whisperity I think I addressed all your comments, could you please review again?

libcodechecker/analyze/analyzers/analyzer_clangsa.py
martong: https://github.com/Ericsson/clang/issues/208

libcodechecker/analyze/host_check.py
```diff
@@ -35,3 +36,29 @@ def check_clang(compiler_bin, env):
...
+        proc = subprocess.Popen(cmd,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE,
+                                )
+        out, err = proc.communicate()
+
+        LOG.debug("stdout:\n" + out)
+        LOG.debug("stderr:\n" + err)
+
+        rc = proc.returncode
```
Xazax-hun: You could just `return proc.returncode == 0`

libcodechecker/analyze/host_check.py
```diff
@@ -35,3 +36,26 @@ def check_clang(compiler_bin, env):
             LOG.error(oerr)
             LOG.error('Failed to run: "' + ' '.join(clang_version_cmd) + '"')
             return False
+
+
+def has_analyzer_feature(clang_bin, feature):
+    inputFile = tempfile.NamedTemporaryFile()
```
gyorb: the inputFile should be closed or you can use it in a with statement

tests/functional/ctu/ctu_failure/test_ctu_failure.py
```diff
@@ -0,0 +1,178 @@
...
+                                         cwd=self.test_dir, env=self.env)
+        self.ctu_capable = '--ctu-' in output
+        print("'analyze' reported CTU-compatibility? " + str(self.ctu_capable))
+
+        self.ctu_has_analyzer_display_ctu_progress = \
+            host_check.has_analyzer_feature(self.__getClangSaPath(),
+                                            '-analyzer-display-ctu-progress')
+        print("Has -analyzer-display-ctu-progress? " + str(self.ctu_capable))
+
+        # Fix the "template" build JSONs to contain a proper directory
```
gyorb: I would recommend to copy the whole `ctu_failure` test project under the test workspace ad do the modifications there and not in the source tree. Later some other test might use this project which could cause conflicts.
Maybe less `os.chdir` is needed with the test project copy. I'm not a big fan of it, it can be hard to follow in which directory the scripts are running.

tests/functional/host_check/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+import shutil
+
+from libtest import env
+
+# Test workspace should be initialized in this module.
+TEST_WORKSPACE = None
+
+
+def setup_package():
+    """Setup the environment for testing ctu."""
```
gyorb: This is not a ctu test.

tests/functional/host_check/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+
+    global TEST_WORKSPACE
+    TEST_WORKSPACE = env.get_workspace('host_check')
```
gyorb: I do not think `TEST_WORKSPACE` setup is needed for these tests. `setup` and `teardown` could just `pass`.

tests/functional/ctu/ctu_failure/test_ctu_failure.py
martong: Ok, I removed the cwd changing.

libcodechecker/analyze/host_check.py
martong: Ok, I added a `with` stmt.

tests/functional/host_check/__init__.py
martong: Fixed.

tests/functional/host_check/__init__.py
martong: Fixed.

Issue #1095, opened 2017-11-06
Sanitize relative path handling in codechecker


Xazax-hun:
Right now CodeChecker does not use consistently the directory argument of the build action. The current working directory should be the directory defined in the build action for every tool invocation.

martong:
For example in:
* libcodechecker/analyze/analysis_manager.py `create_dependencies`

PR #1097, opened 2017-11-07
Use absolute path in logger.

bruntib:
In case of some compiler flags (-I, -L, -isystem) a path can be given.
If this is a relative path then it is possible that in analyse phase we
can't find an include path if the analysis is started from another
directory, not where the logging happened.

### DIFF
--- a/vendor/build-logger/src/ldlogger-tool-gcc.c
+++ b/vendor/build-logger/src/ldlogger-tool-gcc.c
@@ -28,7 +28,11 @@ typedef enum _GccArgsState
   /**
    * After a -o paramater.
    */
-  InOutputArg
+  InOutputArg,
+  /**
+   * After a flag which requires a path after it, e.g. -I or -L.
+   */
+  PathRequired
 } GccArgsState;
 
 /**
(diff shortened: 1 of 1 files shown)


bruntib:
You're probably right. But this fix is created for a report from one of our users. We are already changed the paths to absolute (when using `-I` or `-L` flags and their parameters were added without a separating space character), this pull request just generalize this behavior for all cases.

Xazax-hun:
I agree, but I wonder if changing the working directory of the analyzer invocation would also fix this problem. In case it does. it might also solve other problems and we might also remove additional logic to convert every kind of paths to absolute in multiple places within the project and we might end up simplifying the project. See https://github.com/Ericsson/codechecker/issues/1095

dkrupp:
Let's change the cwd of the clang invocation to the "directory" parameter of the compilation db. That should enable the handling of relative paths. That should be a simpler change anyway.

If we want to change logger we should not do it without adding test cases.

gyorb:
Relative path can be used in the compilation database because the analyzer will be run from the same directory as the original compiler.

PR #1099, opened 2017-11-08
Compilation target is registered per build action.

dkrupp:
This fixes a bug where if a compilation db contained multiple
different compilation targets, only one of them was used for
all files.


libcodechecker/log/option_parser.py
```diff
@@ -290,6 +291,14 @@ def arch(self, value):
         self._arch = value
 
     @property
+    def target(self):
+        return self._arch
```
martong: should be `self._target`

martong:
LGTM

PR #1100, opened 2017-11-08
Compiler info from file

martong:
Make it possible to load the compiler system includes and target info from a file.
This is good for two reasons:
* We can ctu collect the content of a failed report, since the system includes are stored in `sources-root` dir
* We can analyze independently anytime after a build log. If for some reason the compiler is not available, we can just simply feed the previously saved compiler includes and target files (this is the case with the investigation of a fail zip as well)

### DIFF
--- a/libcodechecker/analyze/log_parser.py
+++ b/libcodechecker/analyze/log_parser.py
@@ -106,7 +106,17 @@ def dump_compiler_info(output_path, filename, data):
         f.write(json.dumps(all_data))
 
 
-def get_compiler_includes(compiler, lang, compile_opts, output_path=None,
+def load_compiler_info(filename, compiler):
+    with open(filename, 'r') as f:
+        data = json.load(f)
+    value = data.get(compiler)
+    if value is None:
+        LOG.error("Could not find compiler %s in file %s" %
+                  (compiler, filename))
+    return value
+
+
+def get_compiler_includes(parseLogOptions, compiler, lang, compile_opts,
                           extra_opts=None):
     """
     Returns a list of default includes of the given compiler.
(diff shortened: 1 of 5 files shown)


libcodechecker/analyze/log_parser.py
```diff
@@ -106,7 +106,13 @@ def dump_compiler_info(output_path, filename, data):
...
 
 
-def get_compiler_includes(compiler, lang, compile_opts, output_path=None,
+def load_compiler_info(filename, compiler):
+    with open(filename, 'r') as f:
+        data = json.load(f)
+    return data[compiler]
+
+
+def get_compiler_includes(parseLogOptions, compiler, lang, compile_opts,
```
gyorb: If we introduce a new argument at the first place in a function argument it can cause weird and hard to debug errors in python if not all of the function calls are updated.
Please recheck if there are no wrong function calls for the modified functions: `get_compiler_includes`, `get_compiler_target`, `parse_compile_commands_json`

libcodechecker/analyze/log_parser.py
```diff
@@ -106,7 +106,13 @@ def dump_compiler_info(output_path, filename, data):
         f.write(json.dumps(all_data))
 
 
-def get_compiler_includes(compiler, lang, compile_opts, output_path=None,
+def load_compiler_info(filename, compiler):
+    with open(filename, 'r') as f:
+        data = json.load(f)
+    return data[compiler]
```
gyorb: This will throw a KeyError exception if the key is missing. I prefer to use `.get(key)` which will return `None` by default if the key is missing or a default value can be returned too if needed.

libcodechecker/analyze/log_parser.py
martong: Checked and everthing looks OK.
```
libcodechecker/analyze/log_parser.py|24 col 19| # be forwarded to get_compiler_includes and get_compiler_defines so the 
libcodechecker/analyze/log_parser.py|119 col 5| def get_compiler_includes(parseLogOptions, compiler, lang, compile_opts,
libcodechecker/analyze/log_parser.py|253 col 21| get_compiler_includes(parseLogOptions, results.compiler,

libcodechecker/analyze/log_parser.py|152 col 5| def get_compiler_target(parseLogOptions, compiler):    
libcodechecker/analyze/log_parser.py|259 col 21| get_compiler_target(parseLogOptions, results.compiler)

libcodechecker/analyze/log_parser.py|177 col 5| def parse_compile_commands_json(logfile, parseLogOptions):
libcodechecker/analyze/log_parser.py|181 col 16| LOG.debug('parse_compile_commands_json: ' + str(add_compiler_defaults))
libcodechecker/analyze/log_parser.py|294 col 23| actions = parse_compile_commands_json(logfile, parseLogOptions)
tests/unit/test_buildcmd_escaping.py|86 col 31| return log_parser.parse_compile_commands_json(text, ParseLogOptions())                    

```

libcodechecker/analyze/log_parser.py
martong: Ok, I changed it to `.get`. If it is `None` we log an error.

PR #1108, opened 2017-11-10
Fix a typo in the doc. It is -c and not -C to execute psql commands

sylvestre:


gyorb:
Thanks for fixing this!

PR #1109, opened 2017-11-10
Make 'make package' unconditionally copy the files

sylvestre:


whisperity:
The `; \` at the end is not needed if there is only a single command.

**Are we sure `build_package.py` works well if the output folders already exist?** I usually do `rm -rf build/*; find . -name "*.pyc" -delete; make` (aliased to `r` or `reb`) to deploy CodeChecker locally.

sylvestre:
This was super confusing for me. I spend quite some time to understand why bin/codechecker wasn't the same than from the git repo and why make package didn't fix my problem.

Xazax-hun:
I agree that it is confusing and I am in favor of this change.

gyorb:
I think this was done because of our internal packaging, I need to recheck if it is still needed.

gyorb:
I think the reason for that check in the main Makefile was that running the tests should not rebuild the package if it is already there, because we loose some configuration changes done after packaging.
We should remove the `package` dependencies from the test targets in [tests/Makefile](https://github.com/Ericsson/codechecker/blob/7fbb80c1326feec7878f7203d3ae0e6adcaabd2f/tests/Makefile) so the package will not be overwritten.
This will change the current behavior a bit, because right now you could run the tests even if there was no package available but with this change you will need to run `make package` manually before the tests can run.

sylvestre:
@gyorb if I remove the package dependencies, will you accept this PR? I have been bitten with this issue one more time...

gyorb:
@sylvestre I've created another pull request #1414 with some additional changes.
I still need to verify that it will work with our internal test infrastructure, but it will do a full package clean for the `package` target as in this change.

whisperity:
Will be superseded by #1414.

PR #1113, opened 2017-11-10
Add debug tools for ctu

martong:
These scripts may be used to aid debugging of clang crashes during analysis.
The goal is to reproduce the crash locally, i.e. using the source files in the
failure zip. During the process we take the original compile command database,
compiler includes and compiler target files and we create the modified version
of them (with `_DEBUG` postfix).  These modified files contain the paths
prefixed with the root of the source directory which holds the dependent source
files (`sources-root`).

`prepare_analyzer_cmd.py` creates a new analyzer command which may be executed
immediately, if cross translation unit (CTU) was disabled.  However, to debug
CTU analysis related crashed in clang we have to synthetise the AST dump files
too, so other phases are needed which are done in `prepare_all_cmd_for_ctu.py`.
It executes `CodeChecker` with ctu-collect, therefore `CodeChecker` must be in
the `PATH` and the proper venv must be set.

Example session for debugging a clang crash:
```
$ cd reports/failed
$ unzip main.c_4c7feffae4c2b887abcdc37a3c88b2e5.plist.zip
$ ~w/CodeChecker/debug_tools/prepare_analyzer_cmd.py --clang ~w/llvm/build/debug/bin/clang --clang_plugin_name libericsson --clang_plugin_path ~w/codechecker_core_ws/build/debug/libericsson-checkers.so
$ bash analyzer-command_DEBUG
```

Example session for debuggin a CTU clang crash:
```
$ export PATH=~w/CodeChecker/build/CodeChecker/bin:$PATH
$ source ~w/CodeChecker/venv_dev/bin/activate
$ cd reports/failed
$ unzip main.c_4c7feffae4c2b887abcdc37a3c88b2e5.plist.zip
$ ~w/CodeChecker/debug_tools/prepare_all_cmd_for_ctu.py --clang ~w/llvm/build/debug/bin/clang --clang_plugin_name libericsson --clang_plugin_path ~w/codechecker_core_ws/build/debug/libericsson-checkers.so
$ bash analyzer-command_DEBUG
```

### DIFF
--- a/scripts/debug_tools/README.md
+++ b/scripts/debug_tools/README.md
@@ -1,0 +1,37 @@
+Debug Tools
+===========
+
+These scripts may be used to aid debugging of clang crashes during analysis.
+The goal is to reproduce the crash locally, i.e. using the source files in the
+failure zip. During the process we take the original compile command database,
+compiler includes and compiler target files and we create the modified version
+of them (with `_DEBUG` postfix).  These modified files contain the paths
+prefixed with the root of the source directory which holds the dependent source
+files (`sources-root`).
+
+`prepare_analyzer_cmd.py` creates a new analyzer command which may be executed
+immediately, if cross translation unit (CTU) was disabled.  However, to debug
+CTU analysis related crashed in clang we have to synthetise the AST dump files
+too, so other phases are needed which are done in `prepare_all_cmd_for_ctu.py`.
+It executes `CodeChecker` with ctu-collect, therefore `CodeChecker` must be in
+the `PATH` and the proper venv must be set.
+
+Example session for debugging a clang crash:
+```sh
+$ export WS=/your_own_path
+$ cd reports/failed
+$ unzip main.c_4c7feffae4c2b887abcdc37a3c88b2e5.plist.zip
+$ $WS/CodeChecker/debug_tools/prepare_analyzer_cmd.py --clang $WS/llvm/build/debug/bin/clang --clang_plugin_name libericsson --clang_plugin_path $WS/codechecker_core_ws/build/debug/libericsson-checkers.so
+$ bash analyzer-command_DEBUG
+```
+
+Example session for debuggin a CTU clang crash:
+```sh
+$ export WS=/your_own_path
+$ export PATH=$WS/CodeChecker/build/CodeChecker/bin:$PATH
+$ source $WS/CodeChecker/venv_dev/bin/activate
+$ cd reports/failed
+$ unzip main.c_4c7feffae4c2b887abcdc37a3c88b2e5.plist.zip
+$ $WS/CodeChecker/debug_tools/prepare_all_cmd_for_ctu.py --clang $WS/llvm/build/debug/bin/clang --clang_plugin_name libericsson --clang_plugin_path $WS/codechecker_core_ws/build/debug/libericsson-checkers.so
+$ bash analyzer-command_DEBUG
+```
--- a/scripts/debug_tools/README.md
+++ b/scripts/debug_tools/README.md
@@ -9,12 +9,13 @@ of them (with `_DEBUG` postfix).  These modified files contain the paths
 prefixed with the root of the source directory which holds the dependent source
 files (`sources-root`).
 
-`prepare_analyzer_cmd.py` creates a new analyzer command which may be executed
-immediately, if cross translation unit (CTU) was disabled.  However, to debug
-CTU analysis related crashed in clang we have to synthetise the AST dump files
-too, so other phases are needed which are done in `prepare_all_cmd_for_ctu.py`.
-It executes `CodeChecker` with ctu-collect, therefore `CodeChecker` must be in
-the `PATH` and the proper venv must be set.
+`prepare_analyzer_cmd.py` creates a new clang static analyzer command
+(`analyzer-command_DEBUG`) which may be executed immediately if cross
+translation unit (CTU) was disabled.  However, to debug CTU analysis related
+crashed in clang we have to synthetise the AST dump files too, so other phases
+are needed which are done in `prepare_all_cmd_for_ctu.py`.  It executes
+`CodeChecker` with ctu-collect, therefore `CodeChecker` must be in the `PATH`
+and the proper venv must be set.
 
 Example session for debugging a clang crash:
 ```sh
--- a/scripts/debug_tools/failure_lib.py
+++ b/scripts/debug_tools/failure_lib.py
@@ -1,0 +1,59 @@
+# -------------------------------------------------------------------------
+#                     The CodeChecker Infrastructure
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+import json
+import os
+
+
+def find_path_end(string, path_begin):
+    """
+    Find one path in string.
+    Returns the pair of the begin and end pos.
+    """
+    path_end = string.find(' ', path_begin)
+    if path_end == -1:
+        path_end = len(string)
+    path = string[path_begin:path_end]
+    return path, path_end
+
+
+def change_paths(string, pathModifierFun):
+    """
+    Scan through the string and possibly replace all found paths.
+    Returns the modified string.
+    """
+    result = []
+    i = 0
+    while i < len(string):
+        # Everything is a path wich starts with a '/' and there is a whitespace
+        # after that.
+        # Note, this supports only POSIX paths.
+        if string[i] == '/':
+            path, path_end = find_path_end(string, i)
+            path = pathModifierFun(path)
+            result += path
+            i = path_end - 1
+        else:
+            result.extend(string[i])
+        i = i + 1
+    return ''.join(result)
+
+
+class IncludePathModifier:
+    def __init__(self, sources_root):
+        self.sources_root = sources_root
+
+    def __call__(self, path):
+        return os.path.join(
+            self.sources_root,
+            os.path.normpath(
+                path.lstrip(
+                    os.path.sep)))
+
+
+def load_json_file(filename):
+    with open(filename, 'r') as f:
+        data = json.load(f)
+    return data
--- a/scripts/debug_tools/prepare_all_cmd_for_ctu.py
+++ b/scripts/debug_tools/prepare_all_cmd_for_ctu.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python
+# -------------------------------------------------------------------------
+#                     The CodeChecker Infrastructure
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+import argparse
+import json
+import os
+import subprocess
+
+import prepare_compile_cmd
+import prepare_compiler_includes
+import prepare_compiler_target
+import prepare_analyzer_cmd
+
+
+def execute(cmd):
+    print("Executing command: " + ' '.join(cmd))
+    try:
+        proc = subprocess.Popen(cmd,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+        out, err = proc.communicate()
+
+        print("stdout:\n\n" + out.decode("utf-8"))
+        print("stderr:\n\n" + err.decode("utf-8"))
+
+        if proc.returncode != 0:
+            print('Unsuccessful run: "' + ' '.join(cmd) + '"')
+            raise Exception("Unsuccessful run of command.")
+        return out
+    except OSError:
+        print('Failed to run: "' + ' '.join(cmd) + '"')
+        raise
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Prepare all commands '
+        'to execute in local environmennt for debugging.')
+    parser.add_argument(
+        '--sources_root',
+        default='./sources-root',
+        help="Path of the source root.")
+    parser.add_argument(
+        '--report_dir',
+        default='..',
+        help="Path of the report dir.")
+    parser.add_argument(
+        '--clang',
+        required=True,
+        help="Path to the clang binary.")
+    parser.add_argument(
+        '--clang_plugin_name', default=None,
+        help="Name of the used clang plugin.")
+    parser.add_argument(
+        '--clang_plugin_path', default=None,
+        help="Path to the used clang plugin.")
+    args = parser.parse_args()
+
+    compile_cmd_debug = "compile_cmd_DEBUG.json"
+    with open(compile_cmd_debug, 'w') as f:
+        f.write(
+            json.dumps(
+                prepare_compile_cmd.prepare(
+                    os.path.join(args.report_dir, "compile_cmd.json"),
+                    args.sources_root),
+                indent=4))
+
+    compiler_includes_debug = "compiler_includes_DEBUG.json"
+    with open(compiler_includes_debug, 'w') as f:
+        f.write(
+            json.dumps(
+                prepare_compiler_includes.prepare(
+                    os.path.join(args.report_dir, "compiler_includes.json"),
+                    args.sources_root),
+                indent=4))
+
+    compiler_target_debug = "compiler_target_DEBUG.json"
+    with open(compiler_target_debug, 'wb') as f:
+        f.write(
+            json.dumps(
+                prepare_compiler_target.prepare(
+                    os.path.join(args.report_dir, "compiler_target.json"),
+                    args.sources_root),
+                indent=4))
+
+    # ctu-collect
+    out = execute(["CodeChecker", "analyze", "--ctu-collect",
+                   compile_cmd_debug,
+                   "--compiler-includes-file", compiler_includes_debug,
+                   "--compiler-target-file", compiler_target_debug,
+                   "-o", "report_debug",
+                   "--verbose", "debug"])
+
+    analyzer_command_debug = "analyzer-command_DEBUG"
+    with open(analyzer_command_debug, 'w') as f:
+        f.write(
+            prepare_analyzer_cmd.prepare(
+                "./analyzer-command",
+                prepare_analyzer_cmd.PathOptions(
+                    args.sources_root,
+                    args.clang,
+                    args.clang_plugin_name,
+                    args.clang_plugin_path,
+                    "./report_debug/ctu-dir/x86_64")))
+
+    print(
+        "Preparation of files for debugging is done. "
+        "Now you can execute the generated analyzer command. "
+        "E.g. $ bash % s" %
+        analyzer_command_debug)
--- a/scripts/debug_tools/prepare_analyzer_cmd.py
+++ b/scripts/debug_tools/prepare_analyzer_cmd.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python
+# -------------------------------------------------------------------------
+#                     The CodeChecker Infrastructure
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+import argparse
+import re
+import failure_lib as lib
+import os
+import sys
+
+
+def getFirstLineOfFile(fname):
+    with open(fname) as f:
+        return f.readline()
+
+
+class AnalyzerCommandPathModifier:
+    def __init__(self, opts):
+        self.opts = opts
+
+    def __call__(self, path):
+
+        if re.search('clang$', path):
+            return self.opts.clang
+
+        if self.opts.clang_plugin_name is not None and\
+                re.search(self.opts.clang_plugin_name, path):
+            if self.opts.clang_plugin_path is None:
+                print("clang_plugin_name is in a path, "
+                      "but clang_plugin_path is not given in the options")
+                sys.exit(-1)
+            return self.opts.clang_plugin_path
+
+        if re.search('ctu-dir', path):
+            if self.opts.ctu_dir is None:
+                print('ctu-dir is in a path, but not in the options!')
+                sys.exit(-1)
+            return self.opts.ctu_dir
+
+        return os.path.join(
+            self.opts.sources_root,
+            os.path.normpath(
+                path.lstrip(
+                    os.path.sep)))
+
+
+class PathOptions:
+    def __init__(
+            self,
+            sources_root,
+            clang,
+            clang_plugin_name,
+            clang_plugin_path,
+            ctu_dir):
+        self.sources_root = sources_root
+        self.clang = clang
+        self.clang_plugin_name = clang_plugin_name
+        self.clang_plugin_path = clang_plugin_path
+        self.ctu_dir = ctu_dir
+
+
+def prepare(analyzer_command_file, pathOptions):
+    res = lib.change_paths(getFirstLineOfFile(analyzer_command_file),
+                           AnalyzerCommandPathModifier(pathOptions))
+    return res
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Prepare analyzer-command '
+                                     'to execute in local environmennt.')
+    parser.add_argument(
+        'analyzer_command_file',
+        help="The stored analyzer command.")
+    parser.add_argument(
+        '--sources_root',
+        default='./sources-root',
+        help="Path of the source root.")
+    parser.add_argument(
+        '--ctu_dir',
+        default=None,
+        help="Path of the used ctu-dir.")
+    parser.add_argument(
+        '--clang',
+        required=True,
+        help="Path to the clang binary.")
+    parser.add_argument(
+        '--clang_plugin_name', default=None,
+        help="Name of the used clang plugin.")
+    parser.add_argument(
+        '--clang_plugin_path', default=None,
+        help="Path to the used clang plugin.")
+    args = parser.parse_args()
+
+    print(
+        prepare(
+            args.analyzer_command_file,
+            PathOptions(
+                args.sources_root,
+                args.clang,
+                args.clang_plugin_name,
+                args.clang_plugin_path,
+                args.ctu_dir)))
--- a/scripts/debug_tools/prepare_analyzer_cmd.py
+++ b/scripts/debug_tools/prepare_analyzer_cmd.py
@@ -11,7 +11,7 @@ import os
 import sys
 
 
-def getFirstLineOfFile(fname):
+def get_first_line_of_file(fname):
     with open(fname) as f:
         return f.readline()
 
@@ -62,7 +62,7 @@ class PathOptions:
 
 
 def prepare(analyzer_command_file, pathOptions):
-    res = lib.change_paths(getFirstLineOfFile(analyzer_command_file),
+    res = lib.change_paths(get_first_line_of_file(analyzer_command_file),
                            AnalyzerCommandPathModifier(pathOptions))
     return res
 
--- a/scripts/debug_tools/prepare_compile_cmd.py
+++ b/scripts/debug_tools/prepare_compile_cmd.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python
+# -------------------------------------------------------------------------
+#                     The CodeChecker Infrastructure
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+import argparse
+import json
+import os
+
+import failure_lib as lib
+
+
+def existsInSourcesRoot(entry, sources_root):
+    """
+    Returns true if the given file in the compile commands really available
+    in the sources-root dir
+    """
+    if os.path.isabs(entry['file']):
+        real_path = os.path.join(
+            sources_root,
+            entry['file'].lstrip(os.path.sep))
+        return os.path.exists(real_path)
+
+    real_path = os.path.join(
+        sources_root,
+        entry['directory'].lstrip(
+            os.path.sep),
+        entry['file'])
+    return os.path.exists(real_path)
+
+
+def prepare(compile_command_json, sources_root):
+    """
+    Read a compile cmd json file and change all paths with a prefix
+    (sources_root).  Returns the modified json data.
+    """
+    json_data = lib.load_json_file(compile_command_json)
+    result_json = []
+    sources_root_abs = os.path.abspath(sources_root)
+    for entry in json_data:
+        if not existsInSourcesRoot(entry, sources_root):
+            continue
+
+        entry['directory'] =\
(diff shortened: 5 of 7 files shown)


Xazax-hun:
Is this really CTU related? I think this script should be applicable for non-CTU debugging as well. Also, I think it should be clarified, this is for debugging clang/plugin issues and not debugging CodeChecker issues. I would only mention CTU at the beginning stating that CTU analysis is also supported. I also would not use the CTU abbreviation in the documentation.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,48 @@
+import json
+
+
+def findPath(string, path_begin):
+    """ Find one path in string.
+    @return: the pair of the begin and end pos
```
Xazax-hun: Comments are not full sentences.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    result = ''
+    i = 0
+    while (i < len(string)):
+        # Everything is a path wich starts with a '/' and there is a whitespace
+        # after that
+        if string[i] == '/':
+            (path, path_end) = findPath(string, i)
+            path = pathModifierFun(path)
+            # print(path, end='')
```
Xazax-hun: Do not check in commented out code.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,48 @@
...
+    return (path, path_end)
+
+
+def changePaths(string, pathModifierFun):
+    """ Scan through the string and possibly replace all found paths.
+    @return: The modified string
+    """
+    result = ''
+    i = 0
+    while (i < len(string)):
```
Xazax-hun: Redundant parens in while.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,48 @@
...
+
+def findPath(string, path_begin):
+    """ Find one path in string.
+    @return: the pair of the begin and end pos
+    """
+    path_end = string.find(' ', path_begin)
+    if path_end == -1:
+        path_end = len(string)
+    path = string[path_begin:path_end]
+    return (path, path_end)
```
Xazax-hun: Do you need the parens here?

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,48 @@
...
+    """ Scan through the string and possibly replace all found paths.
+    @return: The modified string
+    """
+    result = ''
+    i = 0
+    while (i < len(string)):
+        # Everything is a path wich starts with a '/' and there is a whitespace
+        # after that
+        if string[i] == '/':
+            (path, path_end) = findPath(string, i)
```
Xazax-hun: Redundant parens.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,48 @@
...
+        i = i + 1
+    return result
+
+
+class IncludePathModifier:
+    def __init__(self, sources_root):
+        self.sources_root = sources_root
+
+    def __call__(self, path):
+        return ''.join((self.sources_root, path))
```
Xazax-hun: Isn't it better to use os.path.join here?

scripts/debug_tools/prepare_all_cmd.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python3
```
Xazax-hun: Is it a good idea to hardcore python3? Most of the code should work both with python3 and 2. We should not require both of them to be installed. Once all our dependencies are python3 compatible we can stop supporting python2 and move on to 3 only.

scripts/debug_tools/prepare_all_cmd.py
```diff
@@ -0,0 +1,83 @@
...
+import os
+import subprocess
+
+
+def execute(cmd):
+    print("Executing command: " + ' '.join(cmd))
+    try:
+        proc = subprocess.Popen(cmd,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE,
```
Xazax-hun: Redundant comma and new line?

scripts/debug_tools/prepare_all_cmd.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import subprocess
+
+
+def execute(cmd):
```
Xazax-hun: Don't we already have a similar utility somewhere?

scripts/debug_tools/prepare_all_cmd.py
```diff
@@ -0,0 +1,83 @@
...
+        print('Failed to run: "' + ' '.join(cmd) + '"')
+        raise
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Prepare all commands '
+        'to execute in local environmennt for debugging.')
+    parser.add_argument('--clang', required=True)
+    parser.add_argument('--libericsson', required=True)
```
Xazax-hun: Should we expose ericsson in APIs? Maybe we should call this clangplugin? Because that library is actually a clang plugin. Also, I would make this optional, so open source users who do not use any clang plugin can also benefit from this.

scripts/debug_tools/prepare_all_cmd.py
```diff
@@ -0,0 +1,83 @@
...
+    args = parser.parse_args()
+
+    if (("reports/failed" not in str(os.getcwd())) or
```
Xazax-hun: Lots of parens in this condition.

scripts/debug_tools/prepare_all_cmd.py
```diff
@@ -0,0 +1,83 @@
...
+            (str(os.path.basename(os.path.normpath(os.getcwd()))) !=
```
Xazax-hun: Do you need the str conversion?

scripts/debug_tools/prepare_all_cmd.py
```diff
@@ -0,0 +1,83 @@
...
+        f.write(out)
+
+    out = execute([os.path.join(tools_dir, "prepare_compiler_target.py"),
+                   "../compiler_target.json"])
+    compiler_target_debug = "compiler_target_DEBUG.json"
+    with open(compiler_target_debug, 'wb') as f:
+        f.write(out)
+
+    # ctu-collect
+    out = execute(["CodeChecker", "analyze", "--ctu-collect",
```
Xazax-hun: I would make ctu related stuff optional and depend on a command line argument.

scripts/debug_tools/prepare_analyzer_cmd.py
```diff
@@ -0,0 +1,77 @@
...
+    content = ''
+    with open(fname) as f:
+        content = f.readlines()
+    # this file has one line only
+    return content[0]
+
+
+# This is duplicated from the libcodechecer module
+# TODO could we have a common module for both libcodechecker and these debug
+# tools?
```
Xazax-hun: I think it should be safe to reuse codechecker modules in this script.

martong: We had a conversation with @gyorb that these files should rather not be part of the package as this is not useful for the regular users.
We cannot just reuse modules from the package if we are not part of the package.

I am open to discuss again this decision, because being part of the package would solve this TODO.

Xazax-hun: One reason I prefer to have this in the package it is possible to debug an old version of a codechecker just having the package without having to check out the corresponding version from the source control. It does not have to reside in the bin directory where other tools are.

scripts/debug_tools/prepare_all_cmd.py
martong: Changed it to be optional.

scripts/debug_tools/failure_lib.py
martong: Yes, we could use os.path.join here.
But, currently these debug tools can handle only POSIX paths, they won't work with windows paths.
This is because currently we handle a path as a list of consecutive characters from `/` to a space character.

Xazax-hun: I prefer not to build technical debt. We cannot know whether windows will be a supported platform in the future and while we do not test things on windows, we should not introduce something on purpose that might cause problems there.

scripts/debug_tools/prepare_all_cmd.py
martong: Yes we have, but that is too specific. The only generic enough utility is `Popen` itself.

scripts/debug_tools/prepare_all_cmd.py
martong: Removed `--libericsson`, added `--clang_plugin_name` and `--clang_plugin_path`

scripts/debug_tools/prepare_compile_cmd.py
```diff
@@ -0,0 +1,36 @@
...
+    parser.add_argument('--sources_root', default='./sources-root')
+    args = parser.parse_args()
+
+    json_data = None
+    with open(args.compile_command_json) as data_file:
+        json_data = json.load(data_file)
+
+    result_json = []
+    for entry in json_data:
+        if existsInSourcesRoot(entry, args.sources_root):
```
Xazax-hun: Using early continue would decrease nesting.

scripts/debug_tools/prepare_analyzer_cmd.py
```diff
@@ -0,0 +1,55 @@
...
+                exit(-1)
+            return self.args.clang_plugin_path
+
+        if re.search('ctu-dir', path):
+            if (args.ctu_dir is None):
+                print('ctu-dir is in a path, but not in the options!')
+                exit(-1)
+            return self.args.ctu_dir
+
+        return ''.join((self.args.sources_root, path))
```
Xazax-hun: This should also be an os.path.join.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,88 @@
...
+                   "--compiler-target-file", compiler_target_debug,
+                   "-o", "report_debug",
+                   "--verbose", "debug"])
+
+    prepare_analyzer_cmd = [os.path.join(tools_dir, "prepare_analyzer_cmd.py"),
+                            "--clang", args.clang,
+                            "--ctu_dir", "./report_debug/ctu-dir/x86_64",
+                            "./analyzer-command"]
+    if args.clang_plugin_name is not None:
+        prepare_analyzer_cmd = prepare_analyzer_cmd + \
```
Xazax-hun: Maybe using `extend` is more pythonic in this case.

scripts/debug_tools/failure_lib.py
martong: Don't get me wrong, i am not against using os.path.join, and I already changed that. But I am against supporting windows paths now. Atm that won't help us at all in debugging, but would make these scripts more harder to maintain and write.

Xazax-hun: I do not see why is it harder to maintain. I think it does not really have any downsides but it might help us in the future. Also, I find it more readable because that makes it clear it is an operation on paths and not just random strings without looking at the arguments.

scripts/debug_tools/prepare_all_cmd.py
martong: Ok, changed it back to default `usr/bin/env python`

scripts/debug_tools/prepare_analyzer_cmd.py
martong: Done.

scripts/debug_tools/prepare_compile_cmd.py
martong: Ok, changed.

scripts/debug_tools/failure_lib.py
martong: I think there is a misunderstanding here.
I agree to use os.path.join, and the code is already changed in that way.

martong: Note, that in the while loop here https://github.com/Ericsson/codechecker/pull/1113/files#diff-3e4a613d13e84f2ba4cc26a770517e5fR22 we are still bound to POSIX paths and I don't really know how could we easily handle windows paths as well. I am open to suggestions, but I'd like to avoid ugly regexes and complex implementation.

martong: The crux of the problem is how to detect for a substring if that is a path either in windows or in unix?

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,47 @@
+import json
```
gyorb: Please add license.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def changePaths(string, pathModifierFun):
+    """ Scan through the string and possibly replace all found paths.
+    @return: the modified string
+    """
+    result = ''
+    i = 0
+    while i < len(string):
+        # Everything is a path wich starts with a '/' and there is a whitespace
```
gyorb: We should note here that this is not an OS independent path detection.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,47 @@
...
+        # after that
+        if string[i] == '/':
+            path, path_end = findPath(string, i)
+            path = pathModifierFun(path)
+            result += path
+            i = path_end - 1
+        else:
+            result += string[i]
```
gyorb: If there will be a lot of items in the `result` using a `list` and `join` before the return can be more performant.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,47 @@
...
+    return path, path_end
+
+
+def changePaths(string, pathModifierFun):
+    """ Scan through the string and possibly replace all found paths.
+    @return: the modified string
+    """
+    result = ''
+    i = 0
+    while i < len(string):
```
gyorb: Using a for loop `for c in string` would not be easier? It is harder to make it wrong and create an infinite loop.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python
```
gyorb: Please add license.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,88 @@
...
+    parser.add_argument('--clang', required=True)
+    parser.add_argument('--clang_plugin_name', default=None)
+    parser.add_argument('--clang_plugin_path', default=None)
+    args = parser.parse_args()
+
+    if ("reports/failed" not in str(os.getcwd()) or
+        os.path.basename(os.path.normpath(os.getcwd())) !=
+            "failed"):
+        print("Please cd to .../reports/failed")
+        exit(-1)
```
gyorb: Shouldn't this be `sys.exit(1)` ?

scripts/debug_tools/prepare_compile_cmd.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
```
gyorb: Please add license.

gyorb: Is this worth to create a separate script, there is only one method here? The same question applies to `prepare_compiler_includes.py` and `prepare_compiler_target.py`

scripts/debug_tools/prepare_compiler_includes.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python
```
gyorb: Please add license.

scripts/debug_tools/failure_lib.py
Xazax-hun: There is `os.path.commonpath`, `os.path.commonprefix`, `os.path.split`. Does any of that help to make this cleaner?

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,47 @@
+import json
+import os
+
+
+def findPath(string, path_begin):
```
Xazax-hun: Maybe this could be renamed to `findPathEnd`?

scripts/debug_tools/failure_lib.py
Xazax-hun: This is not resolved yet.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,88 @@
...
+
+    tools_dir = os.path.dirname(os.path.realpath(__file__))
+
+    out = execute([os.path.join(tools_dir, "prepare_compile_cmd.py"),
```
Xazax-hun: Maybe instead of executing other python scripts using this method you could just import them and make a function call. I think this would be a cleaner solution.

scripts/debug_tools/prepare_analyzer_cmd.py
```diff
@@ -0,0 +1,56 @@
...
+import argparse
+import re
+import failure_lib as lib
+import os
+
+
+def getFirstLineOfFile(fname):
+    content = ''
+    with open(fname) as f:
+        content = f.readlines()
```
Xazax-hun: You could use `readline` to read only the first line: https://docs.python.org/2/library/stdtypes.html#file.readline
This way you do not need to read the whole file.

scripts/debug_tools/prepare_analyzer_cmd.py
```diff
@@ -0,0 +1,56 @@
...
+class AnalyzerCommandPathModifier:
+    def __init__(self, args):
+        self.args = args
+
+    def __call__(self, path):
+
+        if re.search('clang$', path):
+            return self.args.clang
+
+        if (args.clang_plugin_name is not None and
```
Xazax-hun: Redundant parens.

scripts/debug_tools/prepare_analyzer_cmd.py
```diff
@@ -0,0 +1,56 @@
...
+                re.search(args.clang_plugin_name, path)):
+            if (args.clang_plugin_path is None):
```
Xazax-hun: Redundant parens.

scripts/debug_tools/prepare_analyzer_cmd.py
```diff
@@ -0,0 +1,56 @@
...
+                print("clang_plugin_name is in a path, "
+                      "but clang_plugin_path is not given in the options")
+                exit(-1)
+            return self.args.clang_plugin_path
+
+        if re.search('ctu-dir', path):
+            if (args.ctu_dir is None):
```
Xazax-hun: Same as above.

scripts/debug_tools/prepare_compile_cmd.py
```diff
@@ -0,0 +1,37 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Prepare compile cmd json '
+                                     'to execute in local environmennt.')
+    parser.add_argument('compile_command_json')
+    parser.add_argument('--sources_root', default='./sources-root')
+    args = parser.parse_args()
+
+    json_data = None
+    with open(args.compile_command_json) as data_file:
```
Xazax-hun: You already have a utility function for this.

scripts/debug_tools/failure_lib.py
martong: Here we work with the positions and we jump forward in the range to skip a path, so the range based for cannot be used.

scripts/debug_tools/prepare_compile_cmd.py
martong: During debugging each of the compile_cmd, compiler_includes, compiler_target files might be used separately, so I think it is worth to have them as separate files.

scripts/debug_tools/failure_lib.py
martong: Well, it finds a path, and returns that too. (On the other hand it returns the path end too.) So I would keep the name.

Xazax-hun: Yeah, but the beginning of the path is the same as the argument and is found by a separate piece of code.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,53 @@
...
+
+def changePaths(string, pathModifierFun):
+    """ Scan through the string and possibly replace all found paths.
+    Returns the modified string.
+    """
+    result = ''
+    i = 0
+    while i < len(string):
+        # Everything is a path wich starts with a '/' and there is a whitespace
+        # after that
```
Xazax-hun: No period at the end of the sentence.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,109 @@
...
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Prepare all commands '
+        'to execute in local environmennt for debugging.')
+    parser.add_argument('--clang', required=True)
+    parser.add_argument('--clang_plugin_name', default=None)
+    parser.add_argument('--clang_plugin_path', default=None)
+    args = parser.parse_args()
+
+    if ("reports/failed" not in str(os.getcwd()) or
```
Xazax-hun: I think the parens might be redundant here.

scripts/debug_tools/README.md
```diff
@@ -0,0 +1,32 @@
+These scripts may be used to aid debugging of clang crashes during analysis.
```
whisperity: Start the README with some sort of a header so that it looks like a proper markdown file:

```md
CTU Debugging
=============
```

scripts/debug_tools/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+`prepare_analyzer_cmd.py` creates a new analyzer command which may be executed
+immediately, if cross translation unit (CTU) was disabled.  However, to debug
+CTU analysis related crashed in clang we have to synthetise the AST dump files
+too, so other phases are needed which are done in `prepare_all_cmd_for_ctu.py`.
+It executes `CodeChecker` with ctu-collect, therefore `CodeChecker` must be in
+the `PATH` and the proper venv must be set.
+
+Example session for debugging a clang crash:
+```
```
whisperity: If this is a shell script-like thing, we might want to turn on some syntax highlighting. Type `sh` after the three backticks.

scripts/debug_tools/README.md
```diff
@@ -0,0 +1,32 @@
...
+$ cd reports/failed
+$ unzip main.c_4c7feffae4c2b887abcdc37a3c88b2e5.plist.zip
+$ ~w/CodeChecker/debug_tools/prepare_analyzer_cmd.py --clang ~w/llvm/build/debug/bin/clang --clang_plugin_name libericsson --clang_plugin_path ~w/codechecker_core_ws/build/debug/libericsson-checkers.so
+$ bash analyzer-command_DEBUG
+```
+
+Example session for debuggin a CTU clang crash:
+```
+$ export PATH=~w/CodeChecker/build/CodeChecker/bin:$PATH
```
whisperity: What is `~w`?

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,53 @@
+# -------------------------------------------------------------------------
+#                     The CodeChecker Infrastructure
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+import json
+import os
+
+
+def findPathEnd(string, path_begin):
```
whisperity: Python functions use the naming convention of `lower_letters_and_underscores`.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,53 @@
...
+    """ Find one path in string.
```
whisperity: Please format the docstrings either

```py
"""This way."""
```

or 

```
"""
This way.
"""
```

I vote in favor of the second version.

scripts/debug_tools/failure_lib.py
whisperity: Agreed. String building seems really expensive here.

scripts/debug_tools/failure_lib.py
```diff
@@ -0,0 +1,53 @@
...
+
+class IncludePathModifier:
+    def __init__(self, sources_root):
+        self.sources_root = sources_root
+
+    def __call__(self, path):
+        return os.path.join(self.sources_root, path.lstrip(os.path.sep))
+
+
+def load_json_file(filename):
```
whisperity: Do we really want a utility function for this code snippet?
If so, do we want it here?

What if the file opening fails? Where should it be handled?

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,109 @@
...
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+import argparse
+import json
+import os
+import subprocess
+import sys
+
+import prepare_compile_cmd
```
whisperity: What are these imports, where do they come from?
If they come from the current package, I would rather see `from . import xyz` so it is not mixed up with a "system" (or virtualenv) import on first glance.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,109 @@
...
+    except OSError:
+        print('Failed to run: "' + ' '.join(cmd) + '"')
+        raise
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Prepare all commands '
+        'to execute in local environmennt for debugging.')
+    parser.add_argument('--clang', required=True)
```
whisperity: Please add some minor help texts for these optstrings.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,109 @@
...
+    parser.add_argument('--clang_plugin_name', default=None)
```
whisperity: Usually, command line arguments don't contain `_`, but rather `-` is used.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,109 @@
...
+    sources_root = "./sources-root"
+
+    compile_cmd_debug = "compile_cmd_DEBUG.json"
+    with open(compile_cmd_debug, 'w') as f:
+        f.write(
+            json.dumps(
+                prepare_compile_cmd.prepare(
+                    "../compile_cmd.json",
+                    sources_root),
+                indent=4))
```
whisperity: `indent=2` is still readable and produces a smaller file.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,109 @@
...
+
+    compiler_includes_debug = "compiler_includes_DEBUG.json"
+    with open(compiler_includes_debug, 'w') as f:
```
whisperity: We have a JSON read helper function, but many parts of this code is duplicated here. You could try passing the `prepare_XYZ` as an argument to a wrapper? What are these types?

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -0,0 +1,109 @@
...
+                    "../compiler_target.json",
+                    sources_root),
+                indent=4))
+
+    # ctu-collect
+    out = execute(["CodeChecker", "analyze", "--ctu-collect",
+                   compile_cmd_debug,
+                   "--compiler-includes-file", compiler_includes_debug,
+                   "--compiler-target-file", compiler_target_debug,
+                   "-o", "report_debug",
```
whisperity: When scripting, for the sake of debugging, it's nicer if argument names are written out as a whole `"--output", "report_debug"`. Is this folder cleared at start or is incrementally analyzing into it the expected behaviour?

scripts/debug_tools/prepare_analyzer_cmd.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python
+# -------------------------------------------------------------------------
+#                     The CodeChecker Infrastructure
+#   This file is distributed under the University of Illinois Open Source
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+import argparse
+import re
+import failure_lib as lib
```
whisperity: This is not a system-level import, so let's format it accordingly, and below all system imports.

scripts/debug_tools/prepare_analyzer_cmd.py
```diff
@@ -0,0 +1,85 @@
...
+import os
+import sys
+
+
+def getFirstLineOfFile(fname):
```
whisperity: Function naming conventions violated here.

scripts/debug_tools/failure_lib.py
martong: My goal is to handle the file opening, if the file is not in place, the default OSError exception is perfectly fine here.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
martong: As we discussed it will not work.

scripts/debug_tools/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+These scripts may be used to aid debugging of clang crashes during analysis.
+The goal is to reproduce the crash locally, i.e. using the source files in the
+failure zip. During the process we take the original compile command database,
+compiler includes and compiler target files and we create the modified version
+of them (with `_DEBUG` postfix).  These modified files contain the paths
+prefixed with the root of the source directory which holds the dependent source
+files (`sources-root`).
+
+`prepare_analyzer_cmd.py` creates a new analyzer command which may be executed
```
dkrupp: exend so: creates a `Clang Static Analyzer` command (`analyzer-command_DEBUG`)

scripts/debug_tools/prepare_all_cmd_for_ctu.py
martong: We use only the ctu-dir from this folder, does not matter if its incremental or not.

PR #1115, opened 2017-11-10
Handle run database connection failure while server is running

whisperity:
Previously, the server only checked the availability of run databases at server start time. If the database was unavailable, an error was given, and if users tried to connect to the product which database was offline even after a retry, access was prohibited.

But if the server thought a database connection was valid and it, for any reason, terminated, the only indication of this was from the users' perspective, when, at issuing a query, they got an exception back on Thrift.

```
$ CodeChecker cmd runs
Database error on server
(psycopg2.OperationalError) could not connect to server: Connection refused
        Is the server running on host "localhost" (::1) and accepting
        TCP/IP connections on port 6250?
could not connect to server: Connection refused
        Is the server running on host "localhost" (127.0.0.1) and accepting
        TCP/IP connections on port 6250?

[16:17] - Server error.
```

This patch changes the behaviour so these errors are properly caught and put the product in particular back to "connection failed" state.

**NOTE!** This status check is not preemptive! An actual database operation must fail due to connection issues for this status change to be registered.

But this means that if the database dies, the first user trying to do anything to it will get an exception, but  any subsequent queries will, instead of the same exception, get a message that the product's database is missing.

> Image taken after starting a CodeChecker server and terminating the database server behind `pp`.

error message

Of course, an "unconnected" database is, after this, subject to the same rules as if the database wasn't connected at server start. Once every 5 minutes (I think this is the grace period configured) a connection attempt can be made, when users query the product list.

### DIFF
--- a/libcodechecker/server/api/report_server.py
+++ b/libcodechecker/server/api/report_server.py
@@ -414,6 +414,55 @@ class StorageSession:
         return getattr(self.instance, name)
 
 
+class DatabaseException(Exception):
+    """
+    Represents an Exception that was created by a handler method that
+    accessed the database.
+    """
+    def __init__(self, alchemy_ex):
+        """
+        Create a new DatabaseException encapsulating the SQLAlchemy-given
+        exception as its "inner exception".
+        """
+        super(DatabaseException, self).__init__(str(alchemy_ex))
+        self.alchemy_ex = alchemy_ex
+
+        # OperationalError encapsulates lower level DB-API instances of
+        # operational errors, such as database server disconnects, memory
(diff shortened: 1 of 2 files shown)


libcodechecker/server/api/report_server.py
```diff
@@ -414,6 +414,55 @@ def __getattr__(self, name):
         return getattr(self.instance, name)
 
 
+class DatabaseException(Exception):
```
Xazax-hun: What is the value of wrapping SQLAlchemy exceptions?

whisperity:
Superseded by #1166.

PR #1116, opened 2017-11-13
Change directory before analysis

### LINKED ISSUE #1095: Sanitize relative path handling in codechecker

bruntib:
Fixes #1095

### DIFF
--- a/libcodechecker/analyze/analysis_manager.py
+++ b/libcodechecker/analyze/analysis_manager.py
@@ -145,7 +145,9 @@ def create_dependencies(action):
 
         LOG.debug("Crafting build dependencies from GCC or Clang!")
 
-        output, rc = util.call_command(command, env=os.environ)
+        output, rc = util.call_command(command,
+                                       env=os.environ,
+                                       cwd=action.directory)
         if rc == 0:
             # Parse 'Makefile' syntax dependency output.
             dependencies = output.replace('__dummy: ', '') \
@@ -366,6 +368,8 @@ def check(check_data):
 
                     LOG.debug("Writing dependent files to archive.")
                     for dependent_source in dependencies:
+                        dependent_source = os.path.join(action.directory,
+                                                        dependent_source)
                         if not os.path.isabs(dependent_source):
                             dependent_source = \
                                 os.path.abspath(dependent_source)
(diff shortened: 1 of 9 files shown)


Xazax-hun:
I am not sure actually changing the working directory of python script (and also not restoring it) is a good idea. I would expect to pass the working directory parameter to the function that executes the analyzer command. See the cwd argument of Popen constructor (https://docs.python.org/2/library/subprocess.html). What do you think?

libcodechecker/analyze/analysis_manager.py
```diff
@@ -190,6 +190,8 @@ def check(check_data):
     skipped = False
     reanalyzed = False
     try:
+        os.chdir(action.directory)
```
dkrupp: instead of this, maybe we should specify the cwd parameter of the SourceAnalyzer.run_proc(analyzer_cmd,
                                                               env, cwd)
in analyzer_base.py.
I guess that would be cleaner.

bruntib:
I hope, I managed to find all places where changing the current working directory counts :smile:

libcodechecker/analyze/ctu_manager.py
```diff
@@ -182,6 +187,7 @@ def collect_build_action(params):
         ctu_temp_fnmap_folder = params
 
     try:
+        os.chdir(action.directory)
```
gyorb: Why do we need the directory change here?

libcodechecker/util.py
```diff
@@ -128,10 +128,14 @@ def get_binary_in_path(basename_list, versioning_pattern, env):
...
+def call_command(command, env=None, cwd=None):
     """ Call an external command and return with (output, return_code)."""
 
     try:
+        old_cwd = os.getcwd()
+        if cwd:
+            os.chdir(cwd)
+
         LOG.debug('Run ' + ' '.join(command))
         out = subprocess.check_output(command,
```
gyorb: check_output has a `cwd` parameter, saving and restoring the old directory should not be required.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -366,6 +368,8 @@ def check(check_data):
 
                     LOG.debug("Writing dependent files to archive.")
                     for dependent_source in dependencies:
+                        dependent_source = os.path.join(action.directory,
```
Xazax-hun: Can we do this unconditionally? The `dependent_source` is never absolute?

PR #1119, opened 2017-11-14
When check name is empty, shows in which file it is empty

sylvestre:
For some reason, on Firefox, I do experience that in some files. Don't now why but having the plist file is necessary to do anything wrt that.


Xazax-hun:
It is a clang bug. We submitted a patch to fix that, but unfortunately the review is still going on: https://reviews.llvm.org/D37437

PR #1151, opened 2017-11-18
Don't start server if session_config.json is syntactically incorrect

### LINKED ISSUE #1149: Server should not start in case of incorrectly formated json file
I have a typo in my configuration file:
```
[WARNING] [10:13:23] {SESSION MANAGER} - session_manager.py:149 load_session_cfg() - No JSON object could be decoded
[WARNING] [10:13:23] {SESSION MANAGER} - session_manager.py:151 load_session_cfg() - Not valid user authentication file: /var/lib/jenkins/.codechecker/session_config.json

```

In that case, the server should not start:
* it is risky 
* this is clearly not what I want when I am making a mistake in the configuration

whisperity:
> Fixes #1149.

If the **clientside** file is badly formatted, execution can still continue, because worst thing that can happen is that there is no automatic password login for the user, which will result in a valid prompt of "Please authenticate".

However, serverside, as @sylvestre pointed out, it's indeed dangerous to default back to unauthenticated and whatever mode.


libcodechecker/server/server.py
```diff
@@ -748,6 +749,12 @@ def start_server(config_directory, package_data, port, db_conn_string,
                      .format(root_file))
             root_sha = __make_root_file(root_file)
 
+    try:
+        manager = session_manager.SessionManager(root_sha, force_auth)
+    except (IOError, ValueError):
```
Xazax-hun: Do we need the parens?

whisperity: Actually, we don't.

PR #1154, opened 2017-11-20
fix type mismatch

gyorb:
after json load the value type in the dictionary will be unicode
this can not be used in a os.path.join or other string expressions
where a string is expected

In our current case this would fail:
https://github.com/Ericsson/codechecker/blob/8a84dbeca2f0c61acce685b7e8f28dd9c67f738a/libcodechecker/analyze/analysis_manager.py#L371-L372

### DIFF
--- a/libcodechecker/analyze/log_parser.py
+++ b/libcodechecker/analyze/log_parser.py
@@ -191,6 +191,7 @@ def parse_compile_commands_json(logfile, parseLogOptions):
     filtered_build_actions = {}
 
     logfile.seek(0)
+    # WARN: Every value will be unicode.
     data = json.load(logfile)
 
     compiler_includes = {}
@@ -198,7 +199,7 @@ def parse_compile_commands_json(logfile, parseLogOptions):
 
     counter = 0
     for entry in data:
-        sourcefile = entry['file']
+        sourcefile = str(entry['file'])
 
         if not os.path.isabs(sourcefile):
             # Newest versions of intercept-build can create the 'file' in the
@@ -213,7 +214,7 @@ def parse_compile_commands_json(logfile, parseLogOptions):
 
         action = build_action.BuildAction(counter)
         if 'command' in entry:
-            command = entry['command']
+            command = str(entry['command'])
 
             # Old versions of intercept-build (confirmed to those shipping
             # with upstream clang-5.0) do escapes in another way:
@@ -229,9 +230,11 @@ def parse_compile_commands_json(logfile, parseLogOptions):
             command = ' '.join(entry['arguments'])
         else:
             raise KeyError("No valid 'command' or 'arguments' entry found!")
+
+        command = str(command)
         results = option_parser.parse_options(command)
 
-        action.original_command = command
+        action.original_command = str(command)
         action.analyzer_options = results.compile_opts
 
         action.lang = results.lang
@@ -266,8 +269,8 @@ def parse_compile_commands_json(logfile, parseLogOptions):
             action.skip = False
 
         # TODO: Check arch.
-        action.directory = entry['directory']
-        action.sources = sourcefile
+        action.directory = str(entry['directory'])
+        action.sources = str(sourcefile)
         # Filter out duplicate compilation commands.
         unique_key = action.cmp_key
         if filtered_build_actions.get(unique_key) is None:


libcodechecker/analyze/log_parser.py
```diff
@@ -191,14 +191,15 @@ def parse_compile_commands_json(logfile, parseLogOptions):
     filtered_build_actions = {}
 
     logfile.seek(0)
+    # WARN: Every value will be unicode.
```
martong: How did it work before? Do we have this problem only when the compile command file is utf encoded?

gyorb: `json.load` will always produce unicode type values, I'm not sure why this was not a problem so far.
Do we have functional test for the dependency collection? The test should have failed.

martong: Yes we have tests for dependency collection:
tests/functional/test_analyze.py `test_failure()`

martong: Would be good to know why these tests do not fail currently.

whisperity: I recall a previous case where suddenly this unicode-ness became a problem, and it wasn't earlier. Perhaps it's environment-specific.

@gyorb What are you basing this statement on?
> json.load will always produce unicode type values

I can't find the issue in particular, but I recall the error only happened to some particular user and we couldn't even reproduce it initially. I remember myself trying to find something related to this in the Python changelog...

whisperity: Found the issue meanwhile. It turns out you can use `commenter:name` on GitHub to search for issues you commented on! :raised_hands:

#1005 was the issue, #1015 was a patch on it. Sounds rather environmental to me. Perhaps Python can internally automatically convert `unicode` and `str` in some functions back and forth, if we have a proper ASCII string encoded in a `unicode` variable, and the issue only happens if we have a byte pair which can't be auto-encoded properly? Or perhaps only **some** modules are written in a way that if a `unicode` is given and it can be silently transformed to `str`, it just does it without reporting an error.

@gyorb What was the original, actual Exception trace?

whisperity: Nothing says in `os.path.join`, nor its code (`/usr/lib/python2.7/posixpath.py`) that it needs to be unicode. It gets variables, it uses `.endswith()` and `+=`. The implementation for string concatenation lies in the Python interpreter's world, in `C` files. Something tells me there is a conversion going on in there.

gyorb: I use python 2.7.12 and the dev_venv.

If I load a json from a string:
```
>>> json.loads('{ "name":"John", "age":31, "city":"New York" }')
{u'city': u'New York', u'age': 31, u'name': u'John'}
```
or from a file
```
with open("test.json") as json_file:
...     json.load(json_file)
... 
{u'city': u'New York', u'age': 31, u'name': u'John'}
```
the values will be unicode.
A json string will be unicode in python you can checkout this [table](https://docs.python.org/2/library/json.html#encoders-and-decoders).

I run the analyze functional tests when I get this error. 
The problem is with the type mismatch in the join one of the value is str and the other is unicode.
```
  File "...libcodechecker/analyze/analysis_manager.py", line 372, in check
    dependent_source)
  File ".../lib/python2.7/posixpath.py", line 73, in join
    path += '/' + b
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1: ordinal not in range(128)
```

whisperity: I tried to run

```py
os.path.join(os.getcwd(), u'whisperity', "gyorb")
```

In the same environment and it worked.

So I think my explanation is the closest to reality (if not **the** explanation) on why we didn't see this error earlier: Python can essentially convert from unicorn to string and the error is only raised if we have a special character is in there and thus it can't convert.

Though notice that if at least one of the operands for the join is a unicorn, the result is also a unicorn, even if no non-ASCII chars are there which would make using non-utf strings reasonable.

whisperity: `0xe2` is the `â` letter according to a Google search.

gyorb: Ok, but in this case we would still fail if there is a special character in the path.

PR #1166, opened 2017-11-22
database status and migration feature

### LINKED ISSUE #351: Schema migration support of product databases
When a new CodeChecker server version is started, there are 2 different types of schema migration is needed

* Migration of the configuration database (where CodeChecker configuration is stored)
* Migration of each product databases (where the reports for the products are stored)

Migration should not happen automatically, it should always be confirmed by the user, who starts the server, as a faulty migration may corrupt the database.

gyorb:
Introduce new database statuses.
Status updates are done at the database connect attempts.
Last connection date is removed because a simple refresh
will try to reconnect to the databases.

* Database statuses are shown at server start
* GUI support for database status
* Server command line support for migration and db status check

Resolves #351

### DIFF
--- a/api/shared.thrift
+++ b/api/shared.thrift
@@ -35,3 +35,17 @@ enum Permission {
   PRODUCT_ACCESS   = 17,        // scope: PRODUCT
   PRODUCT_STORE    = 18         // scope: PRODUCT
 }
+
+/**
+* Status information about the database backend.
+*/
+enum DBStatus {
+  OK,                           // Everything is ok with the database.
+  MISSING,                      // The database is missing.
+  FAILED_TO_CONNECT,            // Failed to connect to the database.
+  SCHEMA_MISMATCH_OK,           // Schema mismatch between the server and the database.
+  SCHEMA_MISMATCH_NO,           // No automatic migration is available.
+  SCHEMA_MISSING,               // Schema is missing from the database.
+  SCHEMA_INIT_ERROR,            // Failed to create initial database schema.
+  SCHEMA_UPGRADE_FAILED         // Failed to upgrade schema.
+}
--- a/api/v6/products.thrift
+++ b/api/v6/products.thrift
@@ -31,13 +31,15 @@ typedef list<ProductConfiguration> ProductConfigurations
 
 /* Product carries data to the end user's product list and tasks. */
 struct Product {
-  1: i64    id,
-  2: string endpoint,
-  3: string displayedName_b64,
-  4: string description_b64,
-  5: bool   connected,         // Indicates that the server could set up the database connection properly.
-  6: bool   accessible,        // Indicates whether the current user can access this product.
-  7: bool   administrating     // Indicates that the current user can administrate the product.
+  1: i64               id,
+  2: string            endpoint,
+  3: string            displayedName_b64,
+  4: string            description_b64,
+  5: bool              connected,         // !DEPRECATED FLAG it will be always False.
+                                          // databaseStatus is used to get the status of the database.
+  6: bool              accessible,        // Indicates whether the current user can access this product.
+  7: bool              administrating     // Indicates that the current user can administrate the product.
+  8: shared.DBStatus   databaseStatus     // Indicates the database backend status.
 }
 typedef list<Product> Products
 
--- a/db_migrate/env.py
+++ b/db_migrate/env.py
@@ -17,7 +17,7 @@ except ImportError:
 
     sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__),
                                                  "..")))
-    from libcodechecker.server.run_db_model import Base
+    from libcodechecker.server.database.run_db_model import Base
 
 target_metadata = Base.metadata
 
--- a/libcodechecker/cmd/product_client.py
+++ b/libcodechecker/cmd/product_client.py
@@ -15,9 +15,11 @@ from ProductManagement_v6.ttypes import *
 from libcodechecker.libclient.client import setup_product_client
 from libcodechecker.logger import LoggerFactory
 from libcodechecker.output_formatters import twodim_to_str
+from libcodechecker.server.database import database_status
 from libcodechecker.util import split_server_url
 
 from cmd_line_client import CmdLineOutputEncoder
+
 
 LOG = LoggerFactory.get_new_logger('CMD')
 
@@ -33,16 +35,22 @@ def handle_list_products(args):
             results.append({product.endpoint: product})
         print(CmdLineOutputEncoder().encode(results))
     else:  # plaintext, csv
-        header = ['Status', 'Endpoint', 'Name', 'Description']
+        header = ['Database status', 'Endpoint', 'Name', 'Description']
         rows = []
         for product in products:
             name = base64.b64decode(product.displayedName_b64) \
                 if product.displayedName_b64 else ''
             description = base64.b64decode(product.description_b64) \
                 if product.description_b64 else ''
-            rows.append(('Database error' if not product.connected
-                         else 'No access' if not product.accessible
-                         else '',
+
+            if not product.accessible:
+                db_status_msg = 'No access.'
+            else:
+                db_status = product.databaseStatus
+                db_status_msg = database_status.db_status_msg.get(
+                    db_status, 'Unknown database status')
+
+            rows.append((db_status_msg,
                          product.endpoint, name, description))
 
         print(twodim_to_str(args.output_format, header, rows))
--- a/libcodechecker/libclient/client.py
+++ b/libcodechecker/libclient/client.py
@@ -228,9 +228,10 @@ def setup_product_client(protocol, host, port, product_name=None):
     return product_client
 
 
-def setup_client(product_url):
+def setup_client(product_url, product_client=False):
     """
-    Setup the Thrift client and check API version and authentication needs.
+    Setup the Thrift Product or Service client and
+    check API version and authentication needs.
     """
 
     try:
@@ -243,9 +244,8 @@ def setup_client(product_url):
     _, session_token = setup_auth_client(protocol, host, port)
 
     # Check if the product exists.
-    product_client = setup_product_client(protocol, host, port,
-                                          product_name=None)
-    product = product_client.getProducts(product_name, None)
+    client = setup_product_client(protocol, host, port, product_name=None)
+    product = client.getProducts(product_name, None)
     product_error_str = None
     if not (product and len(product) == 1):
         product_error_str = "It does not exist."
@@ -254,16 +254,34 @@ def setup_client(product_url):
             # Only a "substring" match was found. We explicitly reject it
             # on the command-line!
             product_error_str = "It does not exist."
-        elif not product[0].connected:
-            product_error_str = "The database has issues, or the connection " \
-                                "is badly configured."
+
         elif not product[0].accessible:
             product_error_str = "You do not have access."
+
+        elif product[0].databaseStatus != shared.ttypes.DBStatus.OK:
+            product_error_str = "The database has issues, or the connection " \
+                                "is badly configured."
 
     if product_error_str:
         LOG.error("The given product '{0}' can not be used! {1}"
                   .format(product_name, product_error_str))
         sys.exit(1)
+
+    if product_client:
+        LOG.debug("returning product client")
+        return client
+    else:
+        LOG.debug("returning service client")
+        # Service client was requested, setup
+        # and return it.
+        return setup_service_client(protocol,
+                                    host,
+                                    port,
+                                    product_name,
+                                    session_token)
+
+
+def setup_service_client(protocol, host, port, product_name, session_token):
 
     client = thrift_helper.ThriftClientHelper(
         protocol, host, port,
--- a/libcodechecker/libclient/product_helper.py
+++ b/libcodechecker/libclient/product_helper.py
@@ -98,6 +98,10 @@ class ThriftProductHelper(object):
     def getCurrentProduct(self):
         pass
 
+    @ThriftClientCall
+    def getProductConfiguration(self, product_id):
+        pass
+
     # -----------------------------------------------------------------------
     @ThriftClientCall
     def addProduct(self, product):
--- a/libcodechecker/libhandlers/server.py
+++ b/libcodechecker/libhandlers/server.py
@@ -15,6 +15,12 @@ import os
 import socket
 import sys
 
+from alembic import config
+from alembic import script
+from sqlalchemy.orm import sessionmaker
+
+from shared.ttypes import DBStatus
+
 from libcodechecker import generic_package_context
 from libcodechecker import generic_package_suppress_handler
 from libcodechecker import host_check
@@ -27,8 +33,11 @@ from libcodechecker.logger import add_verbose_arguments
 from libcodechecker.server import server
 from libcodechecker.server import instance_manager
 from libcodechecker.server.database import database
+from libcodechecker.server.database import database_status
 from libcodechecker.server.database.config_db_model \
     import IDENTIFIER as CONFIG_META
+from libcodechecker.server.database.config_db_model \
+    import Product as ORMProduct
 from libcodechecker.server.database.run_db_model \
     import IDENTIFIER as RUN_META
 
@@ -247,6 +256,47 @@ def add_arguments_to_parser(parser):
                                 help="Stops all of your running CodeChecker "
                                      "server instances.")
 
+    database_mgmnt = parser.add_argument_group(
+            "Database management arguments.",
+            """WARNING these commands needs to be called with the same
+            workspace and configuration arguments as the server so the
+            configuration database will be found which is required for the
+            schema migration. Migration can be done without a running server
+            but pay attention to use the same arguments which will be used to
+            start the server.
+            NOTE:
+            Before migration it is advised to create a full a backup of
+            the product databases.
+            """)
+
+    database_mgmnt = database_mgmnt. \
+        add_mutually_exclusive_group(required=False)
+
+    database_mgmnt.add_argument('--db-status',
+                                type=str,
+                                dest="status",
+                                action='store',
+                                default=argparse.SUPPRESS,
+                                required=False,
+                                help="Name of the product to get "
+                                     "the database status for. "
+                                     "Use 'all' to list the database "
+                                     "statuses for all of the products.")
+
+    database_mgmnt.add_argument('--db-upgrade-schema',
+                                type=str,
+                                dest='product_to_upgrade',
+                                action='store',
+                                default=argparse.SUPPRESS,
+                                required=False,
+                                help="Name of the product to upgrade to the "
+                                     "latest database schema available in "
+                                     "the package. Use 'all' to upgrade all "
+                                     "of the products."
+                                     "NOTE: Before migration it is advised"
+                                     " to create a full backup of "
+                                     "the product databases.")
+
     add_verbose_arguments(parser)
 
     def __handle(args):
@@ -342,6 +392,193 @@ def add_arguments_to_parser(parser):
     parser.set_defaults(func=__handle)
 
 
+def print_prod_status(prod_status):
+    """
+    Print the database statuses for each of the products.
+    """
+
+    header = ['Product endpoint', 'Database status',
+              'Database location',
+              'Schema version in the database',
+              'Schema version in the package']
+    rows = []
+
+    for k, v in prod_status.items():
+        db_status, schema_ver, package_ver, db_location = v
+        db_status_msg = database_status.db_status_msg.get(db_status)
+        if schema_ver == package_ver:
+            schema_ver += " (up to date)"
+        rows.append([k, db_status_msg, db_location, schema_ver, package_ver])
+
+    prod_status = output_formatters.twodim_to_str('table',
+                                                  header,
+                                                  rows,
+                                                  sortby_column_number=0)
+    print(prod_status)
+
+
+def get_schema_version_from_package(migration_root):
+    """
+    Returns the latest schema version in the package.
+    """
+
+    cfg = config.Config()
+    cfg.set_main_option("script_location", migration_root)
+    pckg_schema_ver = script.ScriptDirectory.from_config(cfg)
+    return pckg_schema_ver.get_current_head()
+
+
+def check_product_db_status(cfg_sql_server, context):
+    """
+    Check the products for database statuses.
+
+    :returns: dictionary of product endpoints with database statuses
+    """
+
+    migration_root = context.run_migration_root
+
+    engine = cfg_sql_server.create_engine()
+    config_session = sessionmaker(bind=engine)
+    sess = config_session()
+
+    try:
+        products = sess.query(ORMProduct).all()
+    except Exception as ex:
+        LOG.debug(ex)
+        LOG.error("Failed to get product configurations from the database.")
+        LOG.error("Please check your command arguments.")
+        sys.exit(1)
+
+    package_schema = get_schema_version_from_package(migration_root)
+
+    db_errors = [DBStatus.FAILED_TO_CONNECT,
+                 DBStatus.MISSING,
+                 DBStatus.SCHEMA_INIT_ERROR,
+                 DBStatus.SCHEMA_MISSING]
+
+    prod_status = {}
+    for pd in products:
+        db = database.SQLServer.from_connection_string(pd.connection,
+                                                       RUN_META,
+                                                       migration_root,
+                                                       interactive=False)
+        db_location = db.get_db_location()
+        ret = db.connect()
+        s_ver = db.get_schema_version()
+        if s_ver in db_errors:
+            s_ver = None
+        prod_status[pd.endpoint] = (ret, s_ver, package_schema, db_location)
+
+    sess.commit()
+    sess.close()
+    engine.dispose()
+
+    return prod_status
+
+
+def __db_status_check(cfg_sql_server, context, product_name=None):
+    """
+    Check and print database statuses for the given product.
+    """
+    if not product_name:
+        return 0
+
+    LOG.debug("Checking database status for " + product_name +
+              " product.")
+
+    prod_statuses = check_product_db_status(cfg_sql_server, context)
+
+    if product_name != 'all':
+        avail = prod_statuses.get(product_name)
+        if not avail:
+            LOG.error("No product was found with this endpoint: " +
+                      str(product_name))
+            return 1
+
+        prod_statuses = {k: v for k, v in prod_statuses.items()
+                         if k == product_name}
+
+    print_prod_status(prod_statuses)
+    return 0
+
+
+def __db_migration(cfg_sql_server, context, product_to_upgrade='all'):
+    """
+    Handle database management.
+    Schema checking and migration.
+    """
+    LOG.info("Preparing schema upgrade for " + str(product_to_upgrade))
+    product_name = product_to_upgrade
+
+    prod_statuses = check_product_db_status(cfg_sql_server, context)
+    prod_to_upgrade = []
+
+    if product_name != 'all':
+        avail = prod_statuses.get(product_name)
+        if not avail:
+            LOG.error("No product was found with this endpoint: " +
+                      product_name)
+            return 1
+        prod_to_upgrade.append(product_name)
+    else:
+        prod_to_upgrade = list(prod_statuses.keys())
+
+    migration_root = context.run_migration_root
+
+    LOG.warning("Please note after migration only "
+                "newer CodeChecker versions can be used "
+                "to start the server")
+    LOG.warning("It is advised to make a full backup of your "
+                "run databases.")
+    for prod in prod_to_upgrade:
+        LOG.info("========================")
+        LOG.info("Checking: " + prod)
+        engine = cfg_sql_server.create_engine()
+        config_session = sessionmaker(bind=engine)
+        sess = config_session()
+
+        product = sess.query(ORMProduct).filter(
+                ORMProduct.endpoint == prod).first()
+        db = database.SQLServer.from_connection_string(product.connection,
+                                                       RUN_META,
+                                                       migration_root,
+                                                       interactive=False)
+
+        db_status = db.connect()
+
+        msg = database_status.db_status_msg.get(db_status,
+                                                'Unknown database status')
+
+        LOG.info(msg)
+        if db_status == DBStatus.SCHEMA_MISSING:
+            question = 'Do you want to initialize a new schema for ' \
+                        + product.endpoint + '? Y(es)/n(o) '
+            if util.get_user_input(question):
+                ret = db.connect(init=True)
+                msg = database_status.db_status_msg.get(
+                    ret, 'Unknown database status')
+            else:
+                LOG.info("No schema initialization was done.")
+
+        elif db_status == DBStatus.SCHEMA_MISMATCH_OK:
+            question = 'Do you want to upgrade to new schema for ' \
+                        + product.endpoint + '? Y(es)/n(o) '
+            if util.get_user_input(question):
+                LOG.info("Upgrading schema ...")
+                ret = db.upgrade()
+                LOG.info("Done.")
+                msg = database_status.db_status_msg.get(
+                    ret, 'Unknown database status')
+            else:
+                LOG.info("No schema migration was done.")
+
+        sess.commit()
+        sess.close()
+        engine.dispose()
+        LOG.info("========================")
+    return 0
+
+
 def __instance_management(args):
     """Handles the instance-manager commands --list/--stop/--stop-all."""
 
@@ -421,9 +658,6 @@ def main(args):
             not os.path.isdir(os.path.dirname(args.sqlite)):
         os.makedirs(os.path.dirname(args.sqlite))
 
-    suppress_handler = generic_package_suppress_handler. \
-        GenericSuppressHandler(None, False)
-
     if 'reset_root' in args:
         try:
             os.remove(os.path.join(args.config_directory, 'root.user'))
@@ -446,20 +680,82 @@ def main(args):
     check_env = analyzer_env.get_check_env(context.path_env_extra,
                                            context.ld_lib_path_extra)
 
+    cfg_sql_server = database.SQLServer.from_cmdline_args(
+        vars(args), CONFIG_META, context.config_migration_root,
+        interactive=True, env=check_env)
+
+    LOG.info("Checking configuration database ...")
+    db_status = cfg_sql_server.connect()
+    db_status_msg = database_status.db_status_msg.get(db_status)
+    LOG.info(db_status_msg)
+
+    if db_status == DBStatus.SCHEMA_MISSING:
+        LOG.debug("Config database schema is missing, initializing new.")
+        db_status = cfg_sql_server.connect(init=True)
+        if db_status != DBStatus.OK:
+            LOG.error("Config database initialization failed!")
+            LOG.error("Please check debug logs.")
+            sys.exit(1)
+
+    if db_status == DBStatus.SCHEMA_MISMATCH_NO:
+        LOG.debug("Configuration database schema mismatch.")
+        LOG.debug("No schema upgrade is possible.")
+        sys.exit(1)
+
+    if db_status == DBStatus.SCHEMA_MISMATCH_OK:
+        LOG.debug("Configuration database schema mismatch.")
+        LOG.debug("Schema upgrade is possible.")
+        LOG.warning("Please note after migration only "
+                    "newer CodeChecker versions can be used"
+                    "to start the server")
+        LOG.warning("It is advised to make a full backup of your "
+                    "configuration database")
+
+        LOG.warning(cfg_sql_server.get_db_location())
+
+        question = 'Do you want to upgrade to the new schema?' \
+                   ' Y(es)/n(o) '
+        if util.get_user_input(question):
+            print("Upgrading schema ...")
+            ret = cfg_sql_server.upgrade()
+            msg = database_status.db_status_msg.get(
+                ret, 'Unknown database status')
+            print(msg)
+            if ret != DBStatus.OK:
+                LOG.error("Schema migration failed")
+                syst.exit(ret)
+        else:
+            LOG.info("No schema migration was done.")
+            sys.exit(0)
+
+    if db_status == DBStatus.MISSING:
+        LOG.error("Missing configuration database.")
+        LOG.error("Server can not be started.")
+        sys.exit(1)
+
+    # Configuration database setup and check is needed before database
+    # statuses can be checked.
+    try:
+        if args.status:
+            ret = __db_status_check(cfg_sql_server, context, args.status)
+            sys.exit(ret)
+    except AttributeError:
+        LOG.debug('Status was not in the arguments.')
+
+    try:
+        if args.product_to_upgrade:
+            ret = __db_migration(cfg_sql_server, context,
+                                 args.product_to_upgrade)
+            sys.exit(ret)
+    except AttributeError:
+        LOG.debug('Product upgrade was not in the arguments.')
+
     # Create the main database link from the arguments passed over the
     # command line.
     default_product_path = os.path.join(args.config_directory,
                                         'Default.sqlite')
     create_default_product = 'sqlite' in args and \
-                             not os.path.exists(args.sqlite) and \
                              not os.path.exists(default_product_path)
-
-    sql_server = database.SQLServer.from_cmdline_args(
-        vars(args), CONFIG_META, context.config_migration_root,
-        interactive=True, env=check_env)
-
-    LOG.debug("Connecting to product configuration database.")
-    sql_server.connect(context.product_db_version_info, init=True)
 
     if create_default_product:
         # Create a default product and add it to the configuration database.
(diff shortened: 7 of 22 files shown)


api/v6/products.thrift
```diff
@@ -31,13 +31,14 @@ typedef list<ProductConfiguration> ProductConfigurations
...
-  3: string displayedName_b64,
-  4: string description_b64,
-  5: bool   connected,         // Indicates that the server could set up the database connection properly.
-  6: bool   accessible,        // Indicates whether the current user can access this product.
-  7: bool   administrating     // Indicates that the current user can administrate the product.
+  1: i64               id,
+  2: string            endpoint,
+  3: string            displayedName_b64,
+  4: string            description_b64,
+  5: bool              connected,         // Indicates that the server could set up the database connection properly.
```
whisperity: Is this attribute still relevant if `databaseStatus` conveys such values?

libcodechecker/cmd/product_client.py
```diff
@@ -33,16 +35,29 @@ def handle_list_products(args):
...
+            if db_status == DBStatus.MISSING:
+                db_status_msg = 'Missing database'
+            elif db_status == DBStatus.FAILED_TO_CONNECT:
+                db_status_msg = 'Connection failed'
+            elif db_status == DBStatus.SCHEMA_MISMATCH:
+                db_status_msg = 'Database schema mismatch'
+            elif db_status == DBStatus.SCHEMA_INITIALIZATION_ERROR:
+                db_status_msg = 'Schema initialization error'
+
+            if db_status == DBStatus.OK and not product.accessible:
```
whisperity: The fact that the user has no access to the database is irrelevant to the database's status itself, as this value is stored in the config db. This condition hides the "even if the database would reconnect, you'd still have no access..." state.

libcodechecker/libhandlers/server.py
```diff
@@ -247,6 +252,28 @@ def add_arguments_to_parser(parser):
                                 help="Stops all of your running CodeChecker "
                                      "server instances.")
 
+    database_mgmnt = parser.add_argument_group("""\nDatabase management arguments.
```
whisperity: `add_argument_group` takes a `name` and a `description` option.

api/v6/products.thrift
gyorb: I do not wanted to break the api in a backward compatible way, but yes this value is not needed.

api/shared.thrift
```diff
@@ -35,3 +35,17 @@ enum Permission {
   PRODUCT_ACCESS   = 17,        // scope: PRODUCT
   PRODUCT_STORE    = 18         // scope: PRODUCT
 }
+
+/**
+* Status information about the database backend.
+*/
+enum DBStatus {
+  OK,                           // everything is ok
```
Xazax-hun: Translate these comments to full sentences.

libcodechecker/cmd/product_client.py
```diff
@@ -33,16 +35,35 @@ def handle_list_products(args):
...
                 if product.displayedName_b64 else ''
             description = base64.b64decode(product.description_b64) \
                 if product.description_b64 else ''
-            rows.append(('Database error' if not product.connected
-                         else 'No access' if not product.accessible
-                         else '',
+
+            db_status_msg = 'OK'
+            db_status = product.databaseStatus
+            if db_status == DBStatus.MISSING:
```
Xazax-hun: I'd prefer to have a map and index into that.

libcodechecker/libclient/client.py
```diff
@@ -243,9 +244,9 @@ def setup_client(product_url):
     _, session_token = setup_auth_client(protocol, host, port)
 
     # Check if the product exists.
-    product_client = setup_product_client(protocol, host, port,
-                                          product_name=None)
-    product = product_client.getProducts(product_name, None)
+    prod_client = setup_product_client(protocol, host, port,
```
Xazax-hun: Is this change intended? Why abbreviate the variable?

libcodechecker/libclient/client.py
```diff
@@ -265,6 +264,20 @@ def setup_client(product_url):
                   .format(product_name, product_error_str))
         sys.exit(1)
 
+    if product_client:
```
Xazax-hun: `prod_client` vs `product_client` issue?

libcodechecker/libhandlers/server.py
```diff
@@ -247,6 +252,28 @@ def add_arguments_to_parser(parser):
...
+
+    database_mgmnt.add_argument('--status',
+                                type=str,
+                                dest="status",
+                                action='store',
+                                required=False,
+                                help="Name of the product to get "
+                                     "the database status for.")
+
+    database_mgmnt.add_argument('--upgrade-schema',
```
Xazax-hun: Why do we want to have a commend for this? Shouldn't this be automatic?

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +369,193 @@ def arg_match(options):
     parser.set_defaults(func=__handle)
 
 
+def handle_connect_ret(ret):
+    """
+    Print out a message based on the database statuses.
+    """
+    if ret == DBStatus.OK:
+        return "Database is OK."
+    elif ret == DBStatus.FAILED_TO_CONNECT:
```
Xazax-hun: I do not like that this is partially duplicated with the other big `elif` above

api/v6/products.thrift
whisperity: Understandable. Though, I'd rather have a comment here at least about the status of this variable.

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +369,193 @@ def arg_match(options):
...
+            LOG.info("Schema mismatch")
+            question = 'Do you want to upgrade to new schema? y[es]/Y[es] '
+            if get_user_input(question):
+                print("Upgrading schema ...")
+                return db.upgrade()
+        elif db_status == DBStatus.SCHEMA_MISMATCH_NO_MIGRATION:
+            LOG.info("Automatic migration is not available")
+            continue
+        elif db_status == DBStatus.MISSING:
+            LOG.info("Database is missing")
```
Xazax-hun: Duplication here as well.

www/scripts/productlist/ListOfProducts.js
```diff
@@ -213,15 +214,46 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
...
+        case DBStatus.SCHEMA_MISMATCH_OK_MIGRATION:
+            db_status_text = "Schema mismatch: migration is possible."
+            break;
+        case DBStatus.SCHEMA_MISMATCH_NO_MIGRATION:
+            db_status_text = "Schema mismatch: migration not available."
+            break;
+        case DBStatus.SCHEMA_MISSING:
+            db_status_text = "Schema is missing."
+            break;
+        case DBStatus.SCHEMA_INITIALIZATION_ERROR:
```
Xazax-hun: This is also duplication. I think it is possible to define constants in thrift. Using a map constant from enum to string might reduce the duplication also across languages.

api/shared.thrift
```diff
@@ -35,3 +35,17 @@ enum Permission {
   PRODUCT_ACCESS   = 17,        // scope: PRODUCT
   PRODUCT_STORE    = 18         // scope: PRODUCT
 }
+
+/**
+* Status information about the database backend.
+*/
+enum DBStatus {
```
csordasmarton: - Use full sentences in comments.
- Place OK/NO at the end of enum values: `SCHEMA_MISMATCH_MIGRATION_OK`, `SCHEMA_MISMATCH_MIGRATION_NO`
- Use short enum values: `SCHEMA_INIT_FAILED`
```thrift
enum DBStatus {
  OK,                           // Everything is okay.
  MISSING,                      // Database is missing.
  CONNECT_FAILED,               // Failed to connect to the database.
  SCHEMA_MISMATCH_OK,           // Schema mismatch between the server and the database.
  SCHEMA_MISMATCH_NO,           // No automatic migration is available.
  SCHEMA_MISSING,               // Schema is missing from the database.
  SCHEMA_INIT_FAILED,           // Failed to create initial database schema.
  SCHEMA_UPGRADE_FAILED         // Failed to upgrade schema.
}
```

libcodechecker/libclient/product_helper.py
```diff
@@ -93,6 +93,10 @@ def getProducts(self, product_endpoint_filter, product_name_filter):
     def getCurrentProduct(self):
         pass
 
+    @ThriftClientCall
+    def getProductConfiguration(self, product_id):
```
csordasmarton: Why did you add this to the product helper. Do you want to use it in the client?

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +369,193 @@ def arg_match(options):
     parser.set_defaults(func=__handle)
 
 
+def handle_connect_ret(ret):
+    """
+    Print out a message based on the database statuses.
+    """
+    if ret == DBStatus.OK:
```
csordasmarton: You use this block of code in multiple places (https://github.com/gyorb/codechecker/blob/16b2583583258cf5a5935a01f3385b1c7ac9a388/libcodechecker/cmd/product_client.py#L46-L64). Create a helper function and use it in both places.

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +369,193 @@ def arg_match(options):
...
+        return "Database is missing"
+
+
+def get_user_input(msg):
+    """
+    Get the user input.
+
+    :returns: True/False based on the asnwer from the user
+    """
+    yes = ['y', 'Y', 'yes', 'Yes']
```
csordasmarton: It would be better to convert the user input to lower case and check that input. In this case we can handle the YES case too.
```py
yes = ['y', 'yes']
answer = raw_input(msg)
if answer.lower() in yes:
    return True
return False
```

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +369,193 @@ def arg_match(options):
...
+    rows = []
+
+    for k, v in prod_status.items():
+        rows.append([k, handle_connect_ret(v)])
+
+    prod_status = output_formatters.twodim_to_str('table',
+                                                  header,
+                                                  rows,
+                                                  sortby_column_number=0)
+    print(prod_status)
```
csordasmarton: Use logger debug instead of print.

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +369,193 @@ def arg_match(options):
...
+
+
+def __db_status_check(cfg_sql_server, context, product_name=None):
+    """
+    Check and print database statuses for the given product.
+    """
+    if not product_name:
+        return 0
+
+    print(product_name)
```
csordasmarton: Same as above. Use logger instead of print.

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +369,193 @@ def arg_match(options):
...
+    Schema checking and migration.
+    """
+    LOG.info("Migrating schema for " + str(product_to_upgrade))
+    product_name = product_to_upgrade
+
+    prod_statuses = check_product_db_status(cfg_sql_server, context)
+    prod_to_upgrade = []
+
+    if product_name != 'all':
+        avail = prod_statuses.get(product_name)
```
csordasmarton: You use this piece of code in multiple places: https://github.com/gyorb/codechecker/blob/16b2583583258cf5a5935a01f3385b1c7ac9a388/libcodechecker/libhandlers/server.py#L469-L473
Create a helper function for this.

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +369,193 @@ def arg_match(options):
...
+        product = sess.query(ORMProduct).filter(
+                ORMProduct.endpoint == prod).first()
+        db = database.SQLServer.from_connection_string(product.connection,
+                                                       RUN_META,
+                                                       migration_root,
+                                                       interactive=False)
+
+        db_status = db.connect()
+
+        if db_status == DBStatus.OK:
```
csordasmarton: To print the database status use the helper function what I have mentioned above at `handle_connect_ret`. Handle special database status cases after the log. 
```py
Log.info(util.db_status_to_str(db_status))

if db_status == DBStatus.SCHEMA_MISSING:
    question = ...
elif db_status == DBStatus.SCHEMA_MISMATCH_OK_MIGRATION:
    question = ...
```

libcodechecker/server/database/database.py
```diff
@@ -53,55 +60,190 @@ def __init__(self, model_meta, migration_root):
...
+            cfg.attributes["connection"] = connection
+
+            scriptt = script.ScriptDirectory.from_config(cfg)
+            mcontext = migration.MigrationContext.configure(connection)
+            database_schema_revision = mcontext.get_current_revision()
+            LOG.debug('Schema revision in the database: ' +
+                      str(database_schema_revision))
+
+            if database_schema_revision:
+                LOG.debug('Database schema was found'
```
csordasmarton:
```py
LOG.debug('Database schema was found. '
          'No need to initialize new.')
```

www/scripts/productlist/ListOfProducts.js
```diff
@@ -167,10 +167,11 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
 
     onRowClick : function (evt) {
       var item = this.getItem(evt.rowIndex);
+      var db_status = item.db_status == 0
```
csordasmarton: Missing `;` from the end of the statement. Also use `===` to equality check.

www/scripts/productlist/ListOfProducts.js
```diff
@@ -213,15 +214,46 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
         + name[0].toUpperCase()
         + '</span></div>';
 
-      if (!item.connected || !item.accessible) {
+
+      var dbStatus = item.databaseStatus
+
+      var db_status_text = ""
```
csordasmarton: This is a python style variable name :smile:. Use upper cased variable names in JavaScript.

www/scripts/productlist/ListOfProducts.js
```diff
@@ -230,6 +262,7 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
         }
       } else {
         name = '<span class="link">' + name + '</span>';
+        description = '<span>' + '</span><br />' + description;
```
csordasmarton: What is the main purpose of an empty `span`?

www/scripts/productlist/ListOfProducts.js
```diff
@@ -213,15 +214,46 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
         + name[0].toUpperCase()
         + '</span></div>';
 
-      if (!item.connected || !item.accessible) {
+
+      var dbStatus = item.databaseStatus
+
+      var db_status_text = ""
+
+      switch(dbStatus) {
```
csordasmarton: Maybe it would be better to create a converter function (`dbStatusFromCodeToString`) in the `util.js` for this..

csordasmarton:
Also fix Codacy warnings.

libcodechecker/cmd/product_client.py
```diff
@@ -33,16 +35,37 @@ def handle_list_products(args):
...
+                db_status_msg = \
+                    'Database schema mismatch migration is not available.'
+            elif db_status == DBStatus.SCHEMA_MISSING:
+                db_status_msg = 'Database schema is missing.'
+            elif db_status == DBStatus.SCHEMA_INITIALIZATION_ERROR:
+                db_status_msg = 'Schema initialization error.'
+            elif db_status == DBStat us.SCHEMA_UPGRADE_FAILED:
+                db_status_msg = 'Schema upgrade failed.'
+
+            if db_status == DBStatus.OK and not product.accessible:
```
whisperity: The header says `Database status` yet this isn't a *database* status. Also... is the database error or the users' no access more "powerful" or the no access?

I think: if the user has no access to a product (which information is always available in the configuration database if the server is alive), should they see database statuses?

libcodechecker/libhandlers/server.py
```diff
@@ -247,6 +256,30 @@ def add_arguments_to_parser(parser):
                                 help="Stops all of your running CodeChecker "
                                      "server instances.")
 
+    database_mgmnt = parser.add_argument_group(
+            "Database management arguments.",
+            """WARNING these commands needs to be called with the same
+            workspace and configuration arguments as the server
```
whisperity: workspace has been in the process of being removed, I'd rather not mention it in a new help thing.

libcodechecker/libhandlers/server.py
```diff
@@ -247,6 +256,30 @@ def add_arguments_to_parser(parser):
                                 help="Stops all of your running CodeChecker "
                                      "server instances.")
 
+    database_mgmnt = parser.add_argument_group(
+            "Database management arguments.",
+            """WARNING these commands needs to be called with the same
```
whisperity: The help should be better here. Do I need to first start a server, then from another shell invoke CodeChecker server --config-directory ... --postgresql ... --status?

libcodechecker/libhandlers/server.py
gyorb: The schema upgrade will be automatic only for the configuration database, for the run database it will be always a manual.

libcodechecker/libhandlers/server.py
gyorb: Using debug log would mess up the generated table output here.

libcodechecker/libhandlers/server.py
gyorb: Using debug log would mess up the generated table output here.

csordasmarton:
It would be great to create a user guide for this feature.

libcodechecker/libhandlers/server.py
```diff
@@ -247,6 +256,47 @@ def add_arguments_to_parser(parser):
...
+            start the server.
+            NOTE:
+            Before migration it is advised to create a full a backup of
+            the product databases.
+            """)
+
+    database_mgmnt = database_mgmnt. \
+        add_mutually_exclusive_group(required=False)
+
+    database_mgmnt.add_argument('--status',
```
csordasmarton: Update the user guide with these new arguments: https://github.com/Ericsson/codechecker/blob/master/docs/user_guide.md#server

libcodechecker/libhandlers/server.py
```diff
@@ -247,6 +256,47 @@ def add_arguments_to_parser(parser):
...
+                                type=str,
+                                dest="status",
+                                action='store',
+                                default=argparse.SUPPRESS,
+                                required=False,
+                                help="Name of the product to get "
```
csordasmarton: It would be better to set the default value to `all`.

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +392,186 @@ def arg_match(options):
...
+    Print the database statuses for each of the products.
+    """
+
+    header = ['Product endpoint', 'Database status',
+              'Schema version in the database',
+              'Schema version in the package']
+    rows = []
+
+    for k, v in prod_status.items():
+        db_status, schema_ver, package_ver = v
```
csordasmarton: For me it would be better to see what product has the same schema version as the package:
```bash
----------------------------------------------------------------------------------------------------------------------------------------
Product endpoint | Database status                                      | Schema version in the database | Schema version in the package
----------------------------------------------------------------------------------------------------------------------------------------
Default          | Database is OK.                                      | 82ca43f05c10 (up to date)      | 82ca43f05c10                 
Test             | Database is OK.                                      | 82ca43f05c10 (up to date)      | 82ca43f05c10                 
Xerces           | Database schema mismatch migration is not available. | 8268fc7ca7f4                   | 82ca43f05c10                 
----------------------------------------------------------------------------------------------------------------------------------------
```

libcodechecker/libhandlers/server.py
```diff
@@ -247,6 +256,47 @@ def add_arguments_to_parser(parser):
...
+                                help="Name of the product to get "
+                                     "the database status for. "
+                                     "Use 'all' to list the database "
+                                     "statuses for all of the products.")
+
+    database_mgmnt.add_argument('--upgrade-schema',
+                                type=str,
+                                dest='product_to_upgrade',
+                                action='store',
+                                default=argparse.SUPPRESS,
```
csordasmarton: Set the default value to `all`.

libcodechecker/util.py
```diff
@@ -373,3 +373,17 @@ def get_line(file_name, line_no):
             return ''
     except IOError:
         return ''
+
+
+def get_user_input(msg):
```
csordasmarton: Can we start the server by hitting `Enter`? So not only for `y`/`yes`.

libcodechecker/server/database/database_status.py
```diff
@@ -0,0 +1,24 @@
...
+#   License. See LICENSE.TXT for details.
+# -------------------------------------------------------------------------
+
+"""
+Map between database statuses and the corresponding messages.
+"""
+from shared.ttypes import DBStatus
+
+db_status_msg = {
+    DBStatus.OK: "Database is OK.",
```
csordasmarton: Instead of `Database is OK.` can we use `Database is up to date.`?

csordasmarton:
Also fix the pep8 error.

libcodechecker/libclient/client.py
```diff
@@ -243,9 +244,8 @@ def setup_client(product_url):
     _, session_token = setup_auth_client(protocol, host, port)
 
     # Check if the product exists.
-    product_client = setup_product_client(protocol, host, port,
-                                          product_name=None)
-    product = product_client.getProducts(product_name, None)
+    client = setup_product_client(protocol, host, port, product_name=None)
```
Xazax-hun: I am a bit confused. We set up product client also if `product_client` is `False`? Maybe the naming is not the best?

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +392,186 @@ def arg_match(options):
...
+
+
+def get_schema_version_from_package(migration_root):
+    """
+    Returns the schema version in the package.
+    """
+
+    cfg = config.Config()
+    cfg.set_main_option("script_location", migration_root)
+    scriptt = script.ScriptDirectory.from_config(cfg)
```
Xazax-hun: I don't like the name of `scriptt` variable.

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +392,186 @@ def arg_match(options):
...
+                                                       RUN_META,
+                                                       migration_root,
+                                                       interactive=False)
+
+        db_status = db.connect()
+
+        msg = database_status.db_status_msg.get(db_status,
+                                                'Unknown database status')
+
+        nothing_to_do = [DBStatus.OK,
```
Xazax-hun: Do you need this? Instead of listing everything when there is nothing to do this could be just an else branch.

libcodechecker/server/database/database.py
```diff
@@ -53,62 +58,224 @@ def __init__(self, model_meta, migration_root):
...
+        ret = self.check_schema()
+
+        migration_ok = [DBStatus.SCHEMA_MISMATCH_OK,
+                        DBStatus.SCHEMA_MISSING]
+        if ret not in migration_ok:
+            # schema migration is not possible
+            return ret
+
+        try:
+            engine = self.create_engine()
```
Xazax-hun: This pattern of getting the engine, session, connection is very frequent. Maybe it is worth to have a utility function for this?

www/scripts/productlist/ListOfProducts.js
```diff
@@ -213,20 +214,51 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
         + name[0].toUpperCase()
         + '</span></div>';
 
-      if (!item.connected || !item.accessible) {
+
+      var dbStatus = item.databaseStatus;
+
+      var dbStatusMsg = '';
+
+      switch(dbStatus) {
```
Xazax-hun: This could also be transformed into a map/object kind of thing.

libcodechecker/libclient/client.py
gyorb: Yes we set it up because the availability of the product needs to be checked even if a service client is requested.

libcodechecker/util.py
gyorb: I would keep this as it is. This feature changes/modifies data and the user should not accidentally do this.

libcodechecker/server/database/database.py
```diff
@@ -376,7 +542,6 @@ def connect(self, db_version_info, init=False):
                     '-h', self.host,
                     '-p', str(self.port),
                     '-U', self.user,
-                    '-d', self.database,
```
whisperity: If I start the server like this:

`CodeChecker server --verbose debug --dbpost 6251 --dbuser whisperity --dbname codechecker_config_2`

(Where `codechecker_config_2` is an existing database which has just been created!), this happens:

```
[12:25:51] {DATABASE HANDLER} [4514] <139729816016640> - database.py:539 connect() - Checking if database is running at [localhost:6251]
[12:25:51] {UTIL} [4514] <139729816016640> - util.py:135 call_command() - Run psql -h localhost -p 6251 -U whisperity -c SELECT version();
[12:25:51] {UTIL} [4514] <139729816016640> - util.py:144 call_command() - Running command "psql -h localhost -p 6251 -U whisperity -c SELECT version();" Failed.
[12:25:51] {UTIL} [4514] <139729816016640> - util.py:145 call_command() - 2
[12:25:51] {UTIL} [4514] <139729816016640> - util.py:146 call_command() - psql: FATAL:  database "whisperity" does not exist
[12:25:51] {SERVER} - server.py:683 main() - Connection failed.
[12:25:51] {SERVER} [4514] <139729816016640> - server.py:730 main() - Status was not in the arguments.
[12:25:51] {SERVER} [4514] <139729816016640> - server.py:738 main() - Product upgrade was not in the arguments.
[12:25:51] {SERVER} [4514] <139729816016640> - server.py:437 check_product_db_status() - Checking config database
[... stack trace ...]
ProgrammingError: (psycopg2.ProgrammingError) relation "products" does not exist
LINE 2: FROM products
             ^
 [SQL: 'SELECT products.id AS products_id, products.endpoint AS products_endpoint, products.connection AS products_connection, products.display_name AS products_display_name, products.description AS products_description \nFROM products']
```

whisperity:
Also, if I crate the database named `whisperity` I can now add products, but the initial status will be `Schema is missing.` even though the run schema exists &mdash; and it already contains some check-ins.

whisperity:
Okay, if I do everything from scratch:

 1. Init two new databases (`cc_config` and `cc`, **along with `whisperity`**), then start server with `cc_config`.
 2. Add new product, run the `--upgrade-schema` command.
 3. Kill server.
 4. Add a new table, create a mig.script, rebuild package, start server, `Schema upgrade possible` is shown
 5. Run `--upgrade-schema` again

Then it works as intended.

libcodechecker/server/database/database_status.py
```diff
@@ -0,0 +1,24 @@
...
+
+"""
+Map between database statuses and the corresponding messages.
+"""
+from shared.ttypes import DBStatus
+
+db_status_msg = {
+    DBStatus.OK: "Database is up to date.",
+    DBStatus.MISSING: "Missing database.",
+    DBStatus.FAILED_TO_CONNECT: "Connection failed.",
```
whisperity: I think to emphasize the difference between this and the previous, we should reword some keys here: `Connection failed to server.`.

libcodechecker/server/database/database_status.py
```diff
@@ -0,0 +1,24 @@
...
+# -------------------------------------------------------------------------
+
+"""
+Map between database statuses and the corresponding messages.
+"""
+from shared.ttypes import DBStatus
+
+db_status_msg = {
+    DBStatus.OK: "Database is up to date.",
+    DBStatus.MISSING: "Missing database.",
```
whisperity: `Connected to server, but database is missing`

libcodechecker/server/database/database_status.py
```diff
@@ -0,0 +1,24 @@
...
+    DBStatus.FAILED_TO_CONNECT: "Connection failed.",
+    DBStatus.SCHEMA_MISMATCH_OK:
+        "Database schema mismatch: migration is available.",
```
whisperity: We use `migration` here, and in the code, but the command is called **`upgrade`** schema. Perhaps it would be more user-friendly if we said: `Database schema mismatch! Possible to upgrade.`

libcodechecker/server/database/database_status.py
```diff
@@ -0,0 +1,24 @@
...
+    DBStatus.SCHEMA_MISMATCH_NO:
+        "Database schema mismatch: migration is not available.",
```
whisperity: `Database schema mismatch! NOT possible to upgrade!`

www/scripts/productlist/ListOfProducts.js
```diff
@@ -213,20 +214,37 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
...
-          description = '<span class="product-description-error database">'
-            + 'The database connection for this product could not be made!'
-            + '</span><br />' + description;
-        } else if (!item.accessible) {
+        if (!item.accessible) {
           statusIcon = '<span class="customIcon product-noaccess"></span>';
           description = '<span class="product-description-error access">'
             + 'You do not have access to this product!'
             + '</span><br />' + description;
+        } else if (dbStatus !== 0) {
```
whisperity: Maybe if the viewing user goes into administrator mode, we should say that the `server` command can be used to upgrade these things?

whisperity:
Another question, @gyorb: How does schema upgrading the *config* database work after this patch?

whisperity:
Another issue: I just rebuilt the package. Seemingly everything is fine, but it still says "issues". (I am using SQLite now.)

```
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:398 from_cmdline_args() - Using SQLite:
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:400 from_cmdline_args() - Database at /home/<username>/.codechecker/config.sqlite
[14:39:39] {SERVER} - server.py:680 main() - Checking configuration database ...
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:168 check_schema() - Schema revision in the database: 8268fc7ca7f4
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:174 check_schema() - Checking schema versions in the package.
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:193 check_schema() - Schema in the package and in the database is the same.
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:195 check_schema() - No schema modification is needed.
[14:39:39] {SERVER} - server.py:683 main() - Database is up to date.
[14:39:39] {SERVER} [937] <139620825134848> - server.py:730 main() - Status was not in the arguments.
[14:39:39] {SERVER} [937] <139620825134848> - server.py:738 main() - Product upgrade was not in the arguments.
[14:39:39] {SERVER} [937] <139620825134848> - server.py:437 check_product_db_status() - Checking config database
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:398 from_cmdline_args() - Using SQLite:
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:400 from_cmdline_args() - Database at /home/<username>/.codechecker/Default.sqlite
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:168 check_schema() - Schema revision in the database: 82ca43f05c10
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:174 check_schema() - Checking schema versions in the package.
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:193 check_schema() - Schema in the package and in the database is the same.
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:195 check_schema() - No schema modification is needed.
[14:39:39] {DATABASE HANDLER} [937] <139620825134848> - database.py:127 get_schema_version() - Schema revision in the database: 82ca43f05c10
-----------------------------------------------------------------------------------------------------------
Product endpoint | Database status         | Schema version in the database | Schema version in the package
-----------------------------------------------------------------------------------------------------------
Default          | Database is up to date. | 82ca43f05c10 (up to date)      | 82ca43f05c10                 
-----------------------------------------------------------------------------------------------------------
There are some database issues. Do you want to start the server? y[es]/Y[es] 
```

gyorb:
@whisperity  schema upgrade for the config database should be done at server start before checking the product databases.
The user will be prompted at the server start that there is a schema mismatch and upgrade is needed.

libcodechecker/libhandlers/server.py
whisperity: Why would you? Also it will prevent starting a new server because the `all` will al(l)ways be there, even when I just want to start a server without any schema meddling.

csordasmarton:
@whisperity I couldn't reproduce this state: "Also, if I crate the database named whisperity I can now add products, but the initial status will be Schema is missing. even though the run schema exists — and it already contains some check-ins.". 
I can reproduce this with psql by doing the following:
  - Creating a `whisperity` database
  - Creating a new product with this database
  - Removing `whisperity` database
  - Creating `whisperity` database again
  - Restart the server
The server gives me the `Database schema is missing.` for the product with this database.

libcodechecker/server/database/database_status.py
csordasmarton: I think `Database is missing.` text is enough following the DBStatus.OK/DBStatus.SCHEMA_MISSING status messages.

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +392,192 @@ def arg_match(options):
     parser.set_defaults(func=__handle)
 
 
+def print_prod_status(prod_status):
```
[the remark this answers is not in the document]
gyorb: Checking the database status is before initialization, this is why the table is empty.

libcodechecker/libhandlers/server.py
```diff
@@ -342,6 +392,192 @@ def arg_match(options):
...
+
+
+def check_product_db_status(cfg_sql_server, context):
+    """
+    Check the products for database statuses.
+
+    :returns: dictionary of product endpoints with database statuses
+    """
+
+    LOG.info("Checking product database")
```
whisperity: `Checking product databases...` would be better here.

libcodechecker/libhandlers/server.py
```diff
@@ -446,6 +679,70 @@ def main(args):
...
+
+    if db_status == DBStatus.SCHEMA_MISMATCH_OK:
+        LOG.debug("Configuration database schema mismatch.")
+        LOG.debug("Schema upgrade is possible.")
+        LOG.warning("Please note after migration only "
+                    "newer CodeChecker versions can be used"
+                    "to start the server")
+        LOG.warning("It is advised to make a full backup of your "
+                    "configuration database")
+        if util.get_user_input(question):
```
whisperity: `question` is undefined here.

 1. Create default config database in SQLite.
 2. Stop server.
 3. Create a random migration script for config db
 4. Start server.

www/scripts/productlist/ListOfProducts.js
```diff
@@ -213,20 +214,37 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
...
+          statusIcon = '<span class="customIcon product-error"></span>';
+          description = '<span class="product-description-error database">'
+            + dbStatusMsg + ' (use server command for schema upgrades)' +
```
csordasmarton: Remove `+` from the end of line. This will remove `NaN` string from the description.
cc_products_error

libcodechecker/libclient/client.py
```diff
@@ -265,6 +263,22 @@ def setup_client(product_url):
                   .format(product_name, product_error_str))
         sys.exit(1)
 
+    if product_client:
+        LOG.debug("returning product client")
+        return client
+    else:
+        LOG.debug("returning service client")
+        # Service client was requestes, setup
```
whisperity: Typo: requeste**d**

libcodechecker/libclient/client.py
```diff
@@ -254,9 +254,7 @@ def setup_client(product_url):
             # Only a "substring" match was found. We explicitly reject it
             # on the command-line!
             product_error_str = "It does not exist."
-        elif not product[0].connected:
-            product_error_str = "The database has issues, or the connection " \
-                                "is badly configured."
+
```
whisperity: This condition went missing. How will the client handle if a product has database issues, when, e.g. `CodeChecker cmd runs --url MyOfflineDatabase` is used?

www/scripts/productlist/ListOfProducts.js
```diff
@@ -213,20 +214,37 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
         + name[0].toUpperCase()
         + '</span></div>';
 
-      if (!item.connected || !item.accessible) {
+      var dbStatus = item.databaseStatus;
+
+      var dbStatusMessages = {};
```
csordasmarton: I don't like this. Create a helper function in the util.js:
```js
    /**
     * Converts a Thrift API database status id to human readable string.
     *
     * @param {String|Number} dbStatus Thrift API DBStatus id.
     * @return Human readable review status string.
     */
    dbStatusFromCodeToString : function (dbStatus) {
      switch (parseInt(dbStatus)) {
        case DBStatus.OK:
          return 'Database is up to date.';
        case DBStatus.MISSING:
          return 'Database is missing.';
        case DBStatus.FAILED_TO_CONNECT:
          return 'Failed to connect to the database.';
        case DBStatus.SCHEMA_MISMATCH_OK:
          return 'Schema mismatch: migration is possible.';
        case DBStatus.SCHEMA_MISMATCH_NO:
          return 'Schema mismatch: migration not available.';
        case DBStatus.SCHEMA_MISSING:
          return 'Schema is missing.';
        case DBStatus.SCHEMA_INIT_ERROR:
          return 'Schema initialization error.';
        default:
          console.warn(
            'Non existing database status code: ',
            dbStatus);
          return 'N/A';
      }
    },
```
Use this function after it was added:
```js
var dbStatusMsg = util.dbStatusFromCodeToString(dbStatus);
```

whisperity:
The "use server command" text should only be shown if the server command could actually handle the error. A connection failure to the database is not such.

image

whisperity:
I don't think we should use words like "schema version" and "database migration" in a server startup warning. It's just too scary, and perhaps for the average user a not explained, and I think not need to be explained concept. The words "outdated" and "upgrade" is fine and feel more friendly.

The `y(es)/Y(es)` prompt should be changed to `Y(es)/n(o)` It's standard in Linux command lines to capitalise the default (enter without input) action. If both say yes it feels like the user has no choice but to answer yes. The fact that both lower and upper case is accepted comes yet again as a "usual" feature it need not be emphasised. If we only emphasise the yes, the extra step taken to type in a capital letter will hopefully give the user an extra moment of thinking _"Do I really want this?"_.

@dkrupp Take a backup -> make a backup

What happens if I have a product endpoint `all`? I presume `all` takes priority as specifying everything? How does `CodeChecker cmd sum` handle a run named `all`? Maybe the `all` token should be changed to `--all` if argparse is able to capture that as the input for `--db-update-schema`?

db_migrate/env.py
```diff
@@ -17,7 +17,7 @@
 
     sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__),
                                                  "..")))
-    from libcodechecker.server.run_db_model import Base
+    from libcodechecker.server.database.run_db_model import Base
```
csordasmarton: Do this modification also in the config_db_migrate:
https://github.com/Ericsson/codechecker/blob/50ba32b02b06b90c86eabf50554fe9bb45752f3a/config_db_migrate/env.py#L22

gyorb: done.

www/scripts/productlist/ListOfProducts.js
```diff
@@ -167,10 +167,11 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
 
     onRowClick : function (evt) {
       var item = this.getItem(evt.rowIndex);
+      var databaseStatus = item.databaseStatus === 0;
```
whisperity: What does this line, and the one in L174 do? This initialises the variable to a boolean checking if the API-reported status is `0`?

www/scripts/productlist/ListOfProducts.js
```diff
@@ -167,10 +167,11 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
...
 
       switch (evt.cell.field) {
         case 'name':
-          if (item.connected[0] && item.accessible[0]) {
+          if (databaseStatus === DBStatus.OK && item.accessible[0]) {
```
whisperity: Then we check the boolean against the enum value?

www/scripts/productlist/ListOfProducts.js
```diff
@@ -213,20 +214,33 @@ function (declare, domClass, domConstruct, ItemFileWriteStore, topic,
...
             + 'You do not have access to this product!'
             + '</span><br />' + description;
+        } else if (dbStatus !== DBStatus.OK) {
+
+          var upgradeMsg = '';
+          statusIcon = '<span class="customIcon product-error"></span>';
+
+          if(dbStatus === DBStatus.SCHEMA_MISMATCH_OK ||
+             dbStatus === DBStatus.SCHEMA_MISSING){
+            upgradeMsg = ' (use server command for schema '
```
whisperity: I think we should put server between a `<tt>` to indicate it's a command.

tests/functional/products/test_products.py
```diff
@@ -139,8 +139,9 @@ def test_get_product_data(self):
                          "The product's endpoint is improper.")
         self.assertTrue(pr_data.id > 0, "Product didn't have valid ID")
 
-        self.assertTrue(pr_data.connected, "Products should be connectible "
-                        "in a test environment.")
+        self.assertFalse(pr_data.connected,
+                         "Deprecated value will take always False."
```
whisperity: Will these tests be updated later on? **We are losing a dimension here** to see if calling `editProduct` actually makes the server use the **new** (edited) connection information under the same product endpoint.

At least introduce a TODO in this test to explain this!

libcodechecker/util.py
```diff
@@ -373,3 +373,17 @@ def get_line(file_name, line_no):
...
+    """
+    Get the user input.
+
+    :returns: True/False based on the asnwer from the user
+    """
+    yes = ['y', 'yes']
+
+    answer = raw_input(msg)
+    if answer.lower() in yes:
+        return True
```
whisperity: Pure boolean return with condition can be simplified. Actually, this whole method can be a lot more simplified:

```py
def m(msg):
  return raw_input(msg).lower() in ['y', 'yes']
```

libcodechecker/libhandlers/store.py
```diff
@@ -240,6 +240,8 @@ def collect_file_hashes_from_plist(plist_file):
             LOG.warning("There is no report to store. After uploading these "
                         "results the previous reports become resolved.")
 
+        LOG.error(client)
```
whisperity: Possible debug intention left as `error` log here.

libcodechecker/libhandlers/server.py
```diff
@@ -470,18 +766,55 @@ def main(args):
...
+    upgrade_available = {}
+    for k, v in prod_statuses.items():
+        db_status, _, _, _ = v
+        if db_status == DBStatus.SCHEMA_MISMATCH_OK or \
+                db_status == DBStatus.SCHEMA_MISSING:
+            upgrade_available[k] = v
+
+    if upgrade_available:
+        print_prod_status(prod_statuses)
+        LOG.warning("Products can be upgraded take a backup!")
```
whisperity: And please, change this line. `Databases can be upgraded, please make a backup!` (Products??) (And `take` a backup doesn't sound right!)

www/scripts/productlist/ListOfProducts.js
gyorb: HTML <tt> [is not supported](https://www.w3schools.com/tags/tag_tt.asp) in HTML5.

tests/functional/products/test_products.py
gyorb: I've added comment to the other test change too before. That the schema will not be automatically created in the new database after editing a product configuration.

gyorb:
@whisperity since git 2.7.0 you can change the date like this too: `git commit --amend --date=now`

www/scripts/productlist/ListOfProducts.js
gyorb: I think this was left there this way by accident.

www/scripts/productlist/ListOfProducts.js
whisperity: One of the things which are *officially* not supported but *unofficially* forever will be. Then let's use `<samp>` or `<kbd>` instead, which give teletype format.

Issue #1159, opened 2017-11-22
Failure zip does not contain all dependent headers (CTU)


martong:
In case of CTU analysis failure we should call `compiler -M` for each imported TU, not just the actual TU.

PR #1170, opened 2017-11-23
Fix ctu collect

martong:
Collect headers of dependent TUs.

### DIFF
--- a/libcodechecker/analyze/analysis_manager.py
+++ b/libcodechecker/analyze/analysis_manager.py
@@ -179,6 +179,19 @@ def init_worker(checked_num, action_num):
     progress_actions = action_num
 
 
+def get_dependent_headers(buildaction, archive):
+    LOG.debug("Generating dependent headers via compiler...")
+    try:
+        dependencies = set(create_dependencies(buildaction))
+    except Exception as ex:
+        LOG.debug("Couldn't create dependencies:")
+        LOG.debug(str(ex))
+        # TODO append with buildaction
+        archive.writestr("no-sources", str(ex))
+        dependencies = set()
+    return dependencies
+
+
 def check(check_data):
     """
     Invoke clang with an action which called by processes.
@@ -187,7 +200,7 @@ def check(check_data):
     skiplist handler is None if no skip file was configured.
     """
 
-    action, context, analyzer_config_map, \
+    actions_map, action, context, analyzer_config_map, \
         output_dir, skip_handler, quiet_output_on_stdout, \
         capture_analysis_output = check_data
 
@@ -338,14 +351,8 @@ def check(check_data):
                         if not quiet_output_on_stdout:
                             LOG.debug_analyzer('\n' + rh.analyzer_stderr)
 
-                    LOG.debug("Generating dependent headers via compiler...")
-                    try:
-                        dependencies = set(create_dependencies(rh.buildaction))
-                    except Exception as ex:
-                        LOG.debug("Couldn't create dependencies:")
-                        LOG.debug(str(ex))
-                        archive.writestr("no-sources", str(ex))
-                        dependencies = set()
+                    dependencies = get_dependent_headers(rh.buildaction,
+                                                         archive)
 
                     LOG.debug("Fetching other dependent files from analyzer "
                               "output...")
@@ -366,15 +373,39 @@ def check(check_data):
                         LOG.debug(str(ex))
                         other_files = set()
 
-                    dependencies.update(other_files)
+                    mentioned_files_dependent_files = set()
+                    for of in other_files:
+                        mentioned_file = os.path.abspath(
+                            os.path.join(action.directory, of))
+                        # Use the target of the original build action
+                        key = mentioned_file, action.target
+                        mentioned_file_action = actions_map.get(key)
+                        if mentioned_file_action is not None:
+                            mentioned_file_deps = get_dependent_headers(
+                                mentioned_file_action, archive)
+                            mentioned_files_dependent_files.\
+                                update(mentioned_file_deps)
+                        else:
+                            LOG.debug("Could not find %s in build actions"
+                                      % key)
 
-                    LOG.debug("Writing dependent files to archive.")
+                    dependencies.update(other_files)
+                    dependencies.update(mentioned_files_dependent_files)
+
+                    # `dependencies` might contain absolute and relative paths
+                    # for the same file, so make everything absolute by
+                    # joining with the the absolute action.directory.
+                    # If `dependent_source` is absolute it remains absolute
+                    # after the join, as documented on docs.python.org .
+                    dependencies_copy = set()
                     for dependent_source in dependencies:
                         dependent_source = os.path.join(action.directory,
                                                         dependent_source)
-                        if not os.path.isabs(dependent_source):
-                            dependent_source = \
-                                os.path.abspath(dependent_source)
+                        dependencies_copy.add(dependent_source)
+                    dependencies = dependencies_copy
+
+                    LOG.debug("Writing dependent files to archive.")
+                    for dependent_source in dependencies:
                         LOG.debug("[ZIP] Writing '" + dependent_source + "' "
                                   "to the archive.")
                         archive_path = dependent_source.lstrip('/')
@@ -433,6 +464,25 @@ def check(check_data):
         return 1, skipped, reanalyzed, action.analyzer_type, None
 
 
+def create_actions_map(actions):
+    """
+    Create a dict for the build actions.
+    Key of the dict: (source_file, target).
+    """
+    result = dict()
+    for act in actions:
+        if act.source_count > 1:
+            LOG.error("Multiple sources for one build action: " +
+                      str(act.sources))
+        source = os.path.join(act.directory, act.sources.next())
+        key = source, act.target
+        if key in result:
+            LOG.error("Multiple entires in compile database "
+                      "with the same (source, target) pair: (%s, %s)" % key)
+        result[key] = act
+    return result
+
+
 def start_workers(actions, context, analyzer_config_map,
                   jobs, output_path, skip_handler, metadata,
                   quiet_analyze, capture_analysis_output):
@@ -440,6 +490,8 @@ def start_workers(actions, context, analyzer_config_map,
     Start the workers in the process pool.
     For every build action there is worker which makes the analysis.
     """
+
+    actions_map = create_actions_map(actions)
 
     # Handle SIGINT to stop this script running.
     def signal_handler(*arg, **kwarg):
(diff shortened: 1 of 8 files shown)


libcodechecker/analyze/analysis_manager.py
```diff
@@ -364,15 +371,41 @@ def check(check_data):
                         LOG.debug(str(ex))
                         other_files = set()
 
+                    mentioned_files_dependent_files = set()
+                    for of in other_files:
+                        mentioned_file = os.path.abspath(of)
```
Xazax-hun: This should be made absolute with respect to the working directory of the build action and not the working directory of the CodeChecker script.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -364,15 +371,41 @@ def check(check_data):
                         LOG.debug(str(ex))
                         other_files = set()
 
+                    mentioned_files_dependent_files = set()
```
Xazax-hun: I would rather pick a name like `ctu_dependencies`. I think it might be more clear than mentioned files.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -364,15 +371,41 @@ def check(check_data):
...
+                    # `dependencies` might contain absolute and relative paths
+                    # for the same file, so make everything absolute.
+                    dependencies_copy = set()
                     for dependent_source in dependencies:
                         dependent_source = os.path.join(action.directory,
                                                         dependent_source)
                         if not os.path.isabs(dependent_source):
                             dependent_source = \
                                 os.path.abspath(dependent_source)
+                        print(dependent_source)
```
Xazax-hun: This is a leftover print?

libcodechecker/analyze/analysis_manager.py
```diff
@@ -431,6 +464,26 @@ def check(check_data):
...
+    """
+    Create a dict for the build actions.
+    Key of the dict: (source_file, target).
+    """
+    result = dict()
+    for act in actions:
+        if act.source_count > 1:
+            LOG.error("Multiple sources for one build action: " +
+                      str(act.sources))
+        source = os.path.abspath(os.path.join(act.directory,
```
Xazax-hun: Is it possible for `act.directory` to be relative? Usually, `os.path.abspath` is not the best solution most of the time since this is the path where CodeChecker was invoked and not the path where the compiler was invoked. (Which might be different).

libcodechecker/analyze/analysis_manager.py
```diff
@@ -431,6 +464,26 @@ def check(check_data):
         return 1, skipped, reanalyzed, action.analyzer_type, None
 
 
+def create_actions_map(actions):
+    """
+    Create a dict for the build actions.
+    Key of the dict: (source_file, target).
```
Xazax-hun: Is this a feasible solution in the future?
What if we compiler the same file twice but with different defines or once for shared object and once for static library but for the same target.

martong: We had a discussion with @dkrupp and @whisperity about this, and there is no better general solution we could provide atm.

In the future we might support different defines, we might have a key e.g. based on the output flag (-o). But that still would not enough to support special builds where an object is erased then recreated (Linux kernel).

Fortunately `-fpic` and its friends affect only the CodeGen but not the AST, so we can safely choose the entry either for the .so or for the .a .

libcodechecker/analyze/analysis_manager.py
martong: From here https://clang.llvm.org/docs/JSONCompilationDatabase.html:
`directory`: The working directory of the compilation. All paths specified in the `command` or `file` fields must be either absolute or relative to this directory.

If it was not an absolute path then all the other paths in `command` could not be relative paths, which is a contradiction with the definition of `directory`.
Consequently, it must be an absolute path.

libcodechecker/analyze/analysis_manager.py
martong: In case of clangsa we get abs paths from Clang, but in case of clang-tidy it may be relative. So yep I am changing it.

libcodechecker/analyze/analysis_manager.py
martong: There is an other path, where we get the files from a clang-tidy run.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -179,6 +179,19 @@ def init_worker(checked_num, action_num):
     progress_actions = action_num
 
 
+def get_dependent_headers(buildaction, archive):
+    LOG.debug("Generating dependent headers via compiler...")
+    try:
+        dependencies = set(create_dependencies(buildaction))
```
Xazax-hun: Does it make sense to have this very thin wrapper that only calls `create_dependencies`? And maybe having `create_dependencies` return a set would be better than convert here?

libcodechecker/analyze/analysis_manager.py
```diff
@@ -366,15 +373,39 @@ def check(check_data):
...
+
                     dependencies.update(other_files)
+                    dependencies.update(mentioned_files_dependent_files)
 
-                    LOG.debug("Writing dependent files to archive.")
+                    # `dependencies` might contain absolute and relative paths
+                    # for the same file, so make everything absolute.
+                    dependencies_copy = set()
                     for dependent_source in dependencies:
                         dependent_source = os.path.join(action.directory,
```
Xazax-hun: I know this is unrelated to this patch but this does not seem to be correct for me at all.
If `dependent_source` is absolute we should not join it to `action.directory`. If it is relative, the result of join must be absolute, since `action.directory` must be absolute. So the next check is redundant.

libcodechecker/analyze/analysis_manager.py
Xazax-hun: Ok, this means the `os.path.abspath` call there is redundant and should be removed.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -433,6 +464,26 @@ def check(check_data):
...
+    result = dict()
+    for act in actions:
+        if act.source_count > 1:
+            LOG.error("Multiple sources for one build action: " +
+                      str(act.sources))
+        source = os.path.abspath(os.path.join(act.directory,
+                                              act.sources.next()))
+        key = source, act.target
+        if key in result:
+            LOG.error(("Multiple entires in compile database "
```
Xazax-hun: Do you need the extra parens around the text?

tests/functional/ctu/ctu_failure/test_ctu_failure.py
```diff
@@ -87,7 +89,61 @@ def test_ctu_logs_ast_import(self):
...
+        if not self.ctu_has_analyzer_display_ctu_progress:
+            self.skipTest(NO_DISPLAY_CTU_PROGRESS)
+
+        output = self.__do_ctu_all(reparse=False,
+                                   extra_args=[
+                                       "--verbose", "debug",
+                                       # The below special checker crashes when
+                                       # a function with a specified name is
+                                       # analyzed.
+                                       "-e", "debug.ExprInspection"
```
Xazax-hun: Maybe hoisting the comment would make it more readable.

tests/functional/ctu/ctu_failure/test_ctu_failure.py
```diff
@@ -149,21 +206,27 @@ def __do_ctu_all(self, reparse, extra_args=None):
...
-
-        cmd = [self._codechecker_cmd, 'analyze', '-o', self.report_dir,
-               '--analyzers', 'clangsa', '--ctu-analyze']
-        if reparse:
-            cmd.append('--ctu-on-the-fly')
-        cmd.append(self.buildlog)
-        output = subprocess.check_output(cmd, stderr=subprocess.STDOUT,
-                                         cwd=self.test_dir, env=self.env)
-        return output
+        def show(out, err):
```
gyorb: You could put this under `libtest` it might be useful for other tests.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -433,6 +464,26 @@ def check(check_data):
...
 
 
+def create_actions_map(actions):
+    """
+    Create a dict for the build actions.
+    Key of the dict: (source_file, target).
+    """
+    result = dict()
+    for act in actions:
+        if act.source_count > 1:
```
gyorb: I think this filtering and splitting up is already done in the logger. If there are multiple source files in one build action it will be added as multiple build actions into the compile command json file. Do we have to check this again here?

libcodechecker/analyze/analysis_manager.py
```diff
@@ -187,7 +200,7 @@ def check(check_data):
     skiplist handler is None if no skip file was configured.
     """
 
-    action, context, analyzer_config_map, \
+    actions_map, action, context, analyzer_config_map, \
```
gyorb: Could we find out a better name for this. `actions_map` does not contain any compilation actions, only extra information for them.

gyorb: Another comment do we have to forward the whole dictionary for each of the processes in the process pool? Can we forward only the data with belongs to the build action handled by the current process?

libcodechecker/analyze/analysis_manager.py
martong: :+1:

libcodechecker/analyze/analysis_manager.py
martong: Actually, the purpose of this function is to handle any error which may happen during `create_dependencies`. Currently we just write the last exception string into the zip file, but on a long term it would be better to collect all the errors and pour them in once, that should be implemented in this function.

libcodechecker/analyze/analysis_manager.py
martong: Yes, right it will be always absolute, so the check is redundant, i am removing it.

Other than that, if the `dependent_source` is absolute than it remains absolute after the join, that is how `os.path.join` behaves and documented:
> If a component is an absolute path, all previous components are thrown away and joining continues from the absolute path component.

libcodechecker/analyze/analysis_manager.py
martong: Actually, `actions_map` does contain all the build actions indexed by a (source, target) pair.
Unfortunately we need this map for all processes in case of CTU because we get the build action for a referenced TU from this map in case of the failure.

libcodechecker/analyze/analysis_manager.py
martong: It is just some sanity check for asserting invariants and alerting on possible programming errors.
First I wanted `assert` or `sys.exit` but later I decided we just log and survive by choosing the first source.

tests/functional/ctu/ctu_failure/test_ctu_failure.py
martong: :+1:

martong:
Closes #1159

martong:
Fix #1159

whisperity:
@martong The `Fix #XXX` needs to be put into the opening comment's body to make GitHub automatically resolve the merge of the RP to the close of the issue.

PR #1173, opened 2017-11-23
Fix resource dir issue

### LINKED ISSUE #1171: Correctly set resource dir when not specified explicitly
Right now we detect default include directories of gcc. This might interfere with clang resource dirs. We need to include the clang resource dir with higher priority than the detected gcc includes (but potentially lower than user include dirs?). To address the priority issue we specify the clang resource dir with the `-isystem` flag. This, however, is not specified when the resource dir is not set in the package layout. From clang 5.0 it is easy to query the resource dir from clang. We should query the resource dir from clang in case it is not specified in the package layout config.

martong:
Closes #1171.

### DIFF
--- a/libcodechecker/analyze/analyzer_env.py
+++ b/libcodechecker/analyze/analyzer_env.py
@@ -70,7 +70,8 @@ def get_check_env(path_env_extra, ld_lib_path_extra):
     return new_env
 
 
-def extend_analyzer_cmd_with_resource_dir(analyzer_cmd, compiler_resource_dir):
+def extend_analyzer_cmd_with_resource_dir(analyzer_cmd,
+                                          compiler_resource_dir):
     """
     GCC and Clang handles the certain compiler intrinsics for specific
     instruction sets (like SSE, AVX) differently.  They have their own set of
@@ -95,6 +96,14 @@ def extend_analyzer_cmd_with_resource_dir(analyzer_cmd, compiler_resource_dir):
     and describes our intentions more cleanly if we just simply disable the
     inludes from the resource dir.
     """
-    if len(compiler_resource_dir) > 0:
-        analyzer_cmd.extend(['-nobuiltininc',
-                             '-isystem', compiler_resource_dir])
+    if len(compiler_resource_dir) == 0:
+        return
+
+    resource_inc = compiler_resource_dir
+    # Resource include path must end with "/include".
+    basename = os.path.basename(os.path.normpath(resource_inc))
+    if basename != 'include':
+        resource_inc = os.path.join(resource_inc, "include")
+
+    analyzer_cmd.extend(['-nobuiltininc',
+                         '-isystem', resource_inc])
--- a/libcodechecker/analyze/analyzers/analyzer_types.py
+++ b/libcodechecker/analyze/analyzers/analyzer_types.py
@@ -266,6 +266,19 @@ def __replace_env_var(cfg_file):
     return replacer
 
 
+def __get_compiler_resource_dir(context, analyzer_binary):
+    resource_dir = ''
+    if len(context.compiler_resource_dir) > 0:
+        resource_dir = context.compiler_resource_dir
+    # If not set then ask the binary for the resource dir.
+    else:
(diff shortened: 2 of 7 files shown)


libcodechecker/analyze/analyzer_env.py
```diff
@@ -95,6 +96,14 @@ def extend_analyzer_cmd_with_resource_dir(analyzer_cmd, compiler_resource_dir):
...
     """
-    if len(compiler_resource_dir) > 0:
-        analyzer_cmd.extend(['-nobuiltininc',
-                             '-isystem', compiler_resource_dir])
+    if len(compiler_resource_dir) == 0:
+        return
+
+    resource_inc = compiler_resource_dir
+    # Resource include path must end with "/include".
+    basename = os.path.basename(os.path.normpath(resource_inc))
```
Xazax-hun: Maybe better to move this logic to `__get_compiler_resource_dir`?

libcodechecker/analyze/analyzers/analyzer_types.py
```diff
@@ -337,7 +351,24 @@ def __build_clang_tidy_config_handler(args, context):
...
-    config_handler.compiler_resource_dir = context.compiler_resource_dir
+
+    # FIXME We cannot get the resource dir from the clang-tidy binary,
+    # therefore now we get a clang binary which is a sibling of the clang-tidy.
+    # TODO Support "clang-tidy -print-resource-dir" .
+    check_env = analyzer_env.get_check_env(context.path_env_extra,
+                                           context.ld_lib_path_extra)
+    # Overwrite PATH to contain only the parent of the clang binary.
+    if os.path.isabs(config_handler.analyzer_binary):
+        check_env['PATH'] = os.path.dirname(config_handler.analyzer_binary)
```
Xazax-hun: Is it a good idea to overwrite the PATH? What is the motivation behind this?

libcodechecker/analyze/analyzer_env.py
Xazax-hun: This is not resolved.

libcodechecker/analyze/analyzers/analyzer_types.py
Xazax-hun: Why not just extend it?

Xazax-hun: Or in case the point is to always resolve to the absolute clang binary, maybe we could just set the binary instead of invoking the function that looks up the binary in the path?

martong: The idea is to lookup `clang` from $PATH when it is a relative path. This is handled with the local object on the stacked with the name `check_env`.
However, when `clang-tidy` is absolute we overwrite the PATH in `check_env` so we lookup only in the parent dir of `clang-tidy`. 

It might be confusing to append to the path, because we may end up with /a/b/c/clang for /x/y/z/clang-tidy. This is true nevertheless in the relative case too, but that we cannot work around.

libcodechecker/analyze/analyzer_env.py
martong: On a long run we plan to use a function which takes both the system includes and the resource dir as a parameter, so we could implement just in one place the madness around these headers.

libcodechecker/analyze/analyzers/analyzer_types.py
Xazax-hun: I think if we want tidy and clang from the same directory we should modify the lookup function. 
Overwriting the path might have negative consequences when we use this environment. (E.g.: we can not even issue a `cd` command if we pass this environment to a Popen or similar functions.)

martong: Simply setting the binary will miss the binary e.g. `clang-4.0`, this the reason to use `resolve_missing_binary`.

martong: We don't write over the effective PATH in the global env. We use a local variable and we overwrite the path in that local variable.

PR #1193, opened 2017-11-29
Fetch include paths from environment variables

### LINKED ISSUE #1184: Include paths from environment variables
C/C++ include paths may also come from `CPATH`, `C_INCLUDE_PATH`, `CPLUS_INCLUDE_PATH` and `OBJC_INCLUDE_PATH` environment variables. In ld-logger we should capture these too. This documentation describes how to use these variables and where the paths in these variables are placed in the compile command: https://gcc.gnu.org/onlinedocs/cpp/Environment-Variables.html.

bruntib:
Include paths can also be given via environment variables. The logger
should collect these too.
Fixes #1184

### DIFF
--- a/vendor/build-logger/src/ldlogger-tool-gcc.c
+++ b/vendor/build-logger/src/ldlogger-tool-gcc.c
@@ -39,6 +39,20 @@ static const char* const srcExts[] = {
 };
 
 /**
+ * List of compiler name infixes belonging to C compilers.
+ */
+static const char* const cCompiler[] = {
+  "gcc", "cc", "clang", NULL
+};
+
+/**
+ * List of compiler name infixes belonging to C++ compilers.
+ */
+static const char* const cppCompiler[] = {
+  "g++", "c++", "clang++", NULL
+};
+
+/**
  * Check if the given path is a gcc libpath or not.
  *
  * @param path_ an absolute directory path.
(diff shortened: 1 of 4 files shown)


vendor/build-logger/src/ldlogger-tool-gcc.c
```diff
@@ -38,6 +38,20 @@ static const char* const srcExts[] = {
   "c", "cc", "cpp", "cxx", "h", "hh", "hxx", "hpp", "o", "so", "a", NULL
 };
 
+/**
+ * List of compiler name infixes belonging to C compilers.
+ */
+static const char* const cCompiler[] = {
```
Xazax-hun: Why is this hardcoded? Shouldn't we use the GCC_LIKE and similar environment variables?

gyorb: Does clang support these environment variables?

Xazax-hun: It does, see the environment section of https://clang.llvm.org/docs/CommandGuide/clang.html

gyorb: That means if we do not remove them the analyzers environment clang will use them during the analysis. Does clang-tidy support this too?

Xazax-hun: This does not mean that. The environment of logging and the analysis might differ.

gyorb: Yes I know that, but if these environment variables are set before codechecker is run than the same same environment is used for analysis with some extensions. In that case the analysis environment would contain these variables.

bruntib: @gyorb: This pull request is about inserting the content of these environment variables to the build command in build.json. The analyzers will get this extended command, so it is not required that these variables are set in analysis phase.

@Xazax-hun: These are hard-coded because we have to distinguish C and C++ compilers. `CC_LOGGER_GCC_LIKE` contains these together. However I extend this list by the other elements of `CC_LOGGER_GCC_LIKE` to handle those too.

Xazax-hun: In case the build system sets these variables, it might not be the same.
Build system might start a subshell that is used for compilation command, but the same subshell is not used for the analysis actions.

gyorb: @Xazax-hun I think we are speaking about completely different environment setups. I understand the cases you described and why this is needed on those cases.

bruntib:
Are you sure about this? I can't compile even a simple C++ "Hello World!" program with `clang` binary.

Xazax-hun:
I am able to compile it, I will get a link error for sure, but in case the link command is separate from the compilation command it should work.

vendor/build-logger/src/ldlogger-tool-gcc.c
```diff
@@ -278,20 +341,46 @@ int loggerGccParserCollectActions(
   else  /* Compiler was not found in path, log the binary name only. */
   	  loggerVectorAdd(&action->arguments, loggerStrDup(toolName_));
 
+  /* Determine programming language based on compiler name. */
```
dkrupp: In CodeChecker def get_language(extension).
in option_parser.py we determine the programming language based on the file extentsion.
It well may be that someone compiles a file with a g++ compiler in c mode.

get_language(extension):
    mapping = {'.c': 'c',
               '.cp': 'c++',
               '.cpp': 'c++',
               '.cxx': 'c++',
               '.txx': 'c++',
               '.cc': 'c++',
               '.C': 'c++',
               '.ii': 'c++',
               '.m': 'objective-c',
               '.mm': 'objective-c++'}
    lang = mapping.get(extension) if extension in mapping else None
    return lang

Of course the -x switch should override this choice, but wouldn't it be better to detect the language in the same way?

PR #1232, opened 2017-12-07
Change log error to debug

### LINKED ISSUE #1230: don't log as error if multiple source and triple is present in the log file
This is not an error, but a limitation in our processing script (that linking is not handled).
Make this a debug message.

[ERROR] [18:09] - Multiple entires in compile database with the same (source, target) pair: (/home/ednikru/work/codechecker/TinyXML/xmltest.cpp, x86_64-linux-gnu)

martong:
Fixes #1230


martong:
Fixes #1230

PR #1234, opened 2017-12-07
Don't attempt to add the same file multiple times to the ZIP

whisperity:
> Related to #1219. *Not sure yet if this closes it!*

When a file is added multiple times to a ZIP, at the `.close()` of the `ZipFile` context manager, warnings are raised about this. (ZIP files seemingly can contain the same path twice, but clients can get nondeterministic about which actual content to use if the contents differ between the paths!)

I am absolutely unsure **why** the analysis manager for @sylvestre attempts to put the same file twice into the ZIP. But instead of letting the warnings flow at the end of failure-zip creation, we preemptively check if the file already contains a path, and show a message about this fact.

### DIFF
--- a/libcodechecker/analyze/analysis_manager.py
+++ b/libcodechecker/analyze/analysis_manager.py
@@ -377,7 +377,7 @@ def check(check_data):
                     for of in other_files:
                         mentioned_file = os.path.abspath(
                             os.path.join(action.directory, of))
-                        # Use the target of the original build action
+                        # Use the target of the original build action.
                         key = mentioned_file, action.target
                         mentioned_file_action = actions_map.get(key)
                         if mentioned_file_action is not None:
@@ -386,8 +386,8 @@ def check(check_data):
                             mentioned_files_dependent_files.\
                                 update(mentioned_file_deps)
                         else:
-                            LOG.debug("Could not find %s in build actions"
-                                      % key)
+                            LOG.debug("Could not find {0} in build actions."
+                                      .format(key))
 
                     dependencies.update(other_files)
                     dependencies.update(mentioned_files_dependent_files)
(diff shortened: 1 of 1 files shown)


libcodechecker/server/api/report_server.py
```diff
@@ -1921,19 +1921,18 @@ def massStoreRun(self, name, tag, version, b64zip, force):
                 LOG.debug("Storing source file: " + source_file_name)
 
                 if not os.path.isfile(source_file_name):
-                    # The file was not in the ZIP file, because we already have the
-                    # content. Let's check if we already have a file record in the
-                    # database or we need to add one.
+                    # The file was not in the ZIP file, because we already
```
whisperity: Changes in this file irrelevant to the PR, but the commit failed to generate locally for me on pep8 errors:

```
libcodechecker/server/api/report_server.py:1924:80: E501 line too long (83 > 79 characters)
libcodechecker/server/api/report_server.py:1925:80: E501 line too long (82 > 79 characters)
libcodechecker/server/api/report_server.py:1934:80: E501 line too long (80 > 79 characters)
```

gyorb: I fixed them in #1166 too.

whisperity:
@gyorb I think we are safe to assume that once the user has `/home/foo/bar/baz.cpp` and we save it as `sources-root/...`, it won't change. Hopefully. If it does, we are done for...

libcodechecker/server/api/report_server.py
whisperity: Okay, undid the changes.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -408,14 +408,26 @@ def check(check_data):
...
 
+                        archive_path = os.path.join('sources-root',
+                                                    archive_subpath)
                         try:
-                            archive.write(
-                                dependent_source,
-                                os.path.join("sources-root",
-                                             archive_path),
-                                zipfile.ZIP_DEFLATED)
+                            _ = archive.getinfo(archive_path)
```
Xazax-hun: I wonder if this can be a bottleneck? An alternative approach would be to query the file list upfront and do not query the info for each path. In case this query is cheap, it should not be a problem.

whisperity: The `ZipFile` class (see `/usr/lib/python2.7/zipfile.py`) contains a `NameToInfo` dict which maps paths to `ZipInfo` objects. Dicts have `O(n)` worst-case lookup time.

However, this is currently a strange patch. If you look at the code, you can see we use `set`s everywhere for file lists, which also have a linear worst-case complexity. As explained earlier, I have no idea **why** some files are marked to be added multiple times to the ZIP. Perhaps something is wrong with parsing the analyzer output, perhaps this simply happens from @sylvestre using `--enable debug` checkers which produce *strange* output.

whisperity: And we are using string keys, so this is hopefully fast enough. We also aren't expecting **a lot** of analyses to fail. The ZIP file already goes over its internal structure at `.close()` (to finish up the ZIP structure, calculte CRCs and such), where it already figures out path duplications (hence the warning in #1219).

Xazax-hun:
Maybe it would be great to open a bug to investigate the underlying issue.

whisperity:
There already is. And it's mentioned in this PR in the first line.

PR #1236, opened 2017-12-07
Fix a bunch of Codacy-reported code issues in Python

whisperity:
> **INTERNAL CHANGES**

Fixed all the Codacy issues in the Python code which are not related to `subprocess`. Obviously those can't be fixed as we have to start external processes...

There were some issues left behind which were either mediocre or deliberate (such as using a dummy `/tmp` path in a test.)

Funny thing: Codacy picks up the bad C++ codes we have for testing Clang. :stuck_out_tongue_winking_eye:

### DIFF
--- a/libcodechecker/libclient/thrift_helper.py
+++ b/libcodechecker/libclient/thrift_helper.py
@@ -125,10 +125,6 @@ class ThriftClientHelper(object):
         pass
 
     @ThriftClientCall
-    def getRunResultCount(self, runIds, reportFilters, cmp_data):
-        pass
-
-    @ThriftClientCall
     def getRunResultTypes(self, runId, reportFilters):
         pass
 
--- a/libcodechecker/server/server.py
+++ b/libcodechecker/server/server.py
@@ -269,49 +269,46 @@ class RequestHandler(SimpleHTTPRequestHandler):
 
             SimpleHTTPRequestHandler.do_GET(self)
 
-    def __check_prod_db(self, product):
+    @staticmethod
+    def __check_prod_db(product):
         """
         Check the product database status.
         Try to reconnect in some cases.
 
-        Returns if everything is ok with the database or
-        throw an exception with the error message if something
-        is wrong with the database.
-
+        Returns if everything is ok with the database or throw an exception
+        with the error message if something is wrong with the database.
         """
-        if product:
-            # Try to reconnect in these cases.
-            # Do not try to reconnect if there is a schema mismatch.
-            # If the product is not connected, try reconnecting...
-            reconnect_cases = [DBStatus.FAILED_TO_CONNECT,
-                               DBStatus.MISSING,
-                               DBStatus.SCHEMA_INIT_ERROR]
 
-            if product.db_status == DBStatus.OK:
-                # no reconnection needed
-                return
-            elif product.db_status in reconnect_cases:
-                LOG.error("Request's product '{0}' is not connected! "
-                          "Attempting reconnect..."
-                          .format(product.endpoint))
-                product.connect()
-                if product.db_status != DBStatus.OK:
-                    # If the reconnection fails,
-                    # send an error to the user.
-                    LOG.debug("Product reconnection failed.")
-                    error_msg = "'{0}' database connection " \
-                        "failed!".format(product.endpoint)
-                    LOG.error(error_msg)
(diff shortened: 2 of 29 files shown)


whisperity:
For some reason, GitHub seems to not fire the event for Codacy to start analysing the PR. Or perhaps it's in some queue. The results:

image

libcodechecker/server/server.py
```diff
@@ -269,49 +269,48 @@ def do_GET(self):
...
+                # If the reconnection fails send an error to the user.
+                LOG.debug("Product reconnection failed.")
                 error_msg = "'{0}' database connection " \
                     "failed!".format(product.endpoint)
                 LOG.error(error_msg)
                 raise ValueError(error_msg)
-
         else:
-            error_msg = "The product " + str(product) + " does not exists"
+            # send an error to the user.
```
Xazax-hun: The comment starts with lower case letter.

libcodechecker/server/server.py
```diff
@@ -269,49 +269,48 @@ def do_GET(self):
...
+
+        if not product:
+            error_msg = "The product " + str(product) + " does not exists"
+            LOG.error(error_msg)
+            raise ValueError(error_msg)
+
+        # Try to reconnect in these cases.
+        # Do not try to reconnect if there is a schema mismatch.
+        # If the product is not connected, try reconnecting...
+        reconnect_cases = [DBStatus.FAILED_TO_CONNECT,
```
Xazax-hun: Maybe this could be moved closer to the usage.

libcodechecker/libclient/thrift_helper.py
```diff
@@ -124,10 +124,6 @@ def getLinesInSourceFileContents(self, lines_in_files_requested, encoding):
     def getDiffResultsHash(self, run_ids, report_hashes, diff_type):
         pass
 
-    @ThriftClientCall
```
gyorb: Why is this removed?

libcodechecker/util.py
```diff
@@ -45,15 +45,14 @@ def is_localhost(address):
...
-        pass
+    except socket.herror:
+        LOG.debug("Failed to get IP address for localhost.")
 
     try:
         valid_values.append(socket.gethostbyname(socket.gethostname()))
-    except Exception:
-        # Failed to get ip address for host_name.
-        pass
+    except socket.herror:
```
gyorb: `socket.error`?

libcodechecker/libclient/thrift_helper.py
whisperity: It was defined multiple times.

libcodechecker/util.py
whisperity: There is no indication that [`socket.herror`](https://docs.python.org/2/library/socket.html#socket.herror) would be a subtype of `socket.error`.

whisperity:
I have a clever idea about squelching the subprocess-related errors. As we **have** to use subprocesses but it looks like Codacy isn't using call flow analysis. So if we'd put all subprocess creation behind a wrapper script, we could localise the subprocess usage to a single file, on which Codacy would keep reporting, but not on a good 10-15% of our files.

Xazax-hun:
There is no way to turn a specific check off? I would prefer that to modify the code for the sake of reducing findings. If there are no other solutions I am fine with isolating the calls to subprocess but only as a last resort.

whisperity:
The Codacy project is registered under @gyorb's name, so I think it's him who could alter the configuration. There is a *Code patterns* overview which allows you to specify which rules to check against. (By default, *most* of Python ones are enabled, but there are some disableds, just like how we set up the checker list in CodeChecker.)
Allegedly, there is a way to put a configuration file into the repository which sets the rules Codacy should use.

whisperity:
DNM until all the other big fixes of the next release are committed, as per gyorb's request.

gyorb:
I've turned off the subprocess related warnings.

Xazax-hun:
The release is out. Is this ready to merge or wait for the next point release?

gyorb:
After the conflicts are resolved it can be merged.

whisperity:
@gyorb Resolved the small conflicts that were there. Let's wait for a green light from Travis and Codacy to see if I haven't messed anything up in the meantime.

PR #1237, opened 2017-12-07
--include flags shouldn't be skipped

bruntib:


whisperity:
Is this flag valid for G++, Clang? Documentations use `-I` everywhere, with a single hyphen.

bruntib:
`-include` and `--include` flags are given a header file instead of a directory. This file will be included in the translation unit implicitly during the preprocessing phase.

PR #1255, opened 2017-12-13
Remove alpha.cpluscplus.IteratorRange from all profiles

baloghadamsoftware:
This is the first step of a complex checker, it does not do any useful in its current state.


gyorb:
the next time please a create branch on your fork and send a pull request from that, thanks!

PR #1266, opened 2017-12-19
Fixing the argument parsing gcc arguments

dkrupp:
Some of the gcc compilation options that
accept files or directories
as arguments (-I,-idirafter, -include ...)
are accepted with and without space between the option
and the option argument. The handling of these
options corrected now.


dkrupp:
This patch fixes #1267

PR #1272, opened 2017-12-27
 Add guide about an important class of false negatives and a guide about  false positives with loops.

Xazax-hun:


docs/false_positives.md
```diff
@@ -235,6 +236,84 @@ the analyzer better. Note that the compiler might generate the same code
...
+}
+
+int main() {
+  return strlen(0);
+}
+```
+
+The analyzer will be able to find the null dereference in the code above.
+In some cases, however, the author of the functions add defensive checks to
+avoid crashes when some clients to dno respect the precondition.
```
rnkovacs: typo: clients **do not** respect

docs/false_positives.md
```diff
@@ -235,6 +236,84 @@ the analyzer better. Note that the compiler might generate the same code
...
+  return res;
+}
+
+int main() {
+  return strlen(0);
+}
+```
+
+The analyzer will be able to find the null dereference in the code above.
+In some cases, however, the author of the functions add defensive checks to
```
rnkovacs: nit: the author ... add**s** defensive checks

docs/false_positives.md
```diff
@@ -235,6 +236,84 @@ the analyzer better. Note that the compiler might generate the same code
...
+```
+
+The analyzer will be able to find the null dereference in the code above.
+In some cases, however, the author of the functions adds defensive checks to
+avoid crashes when some clients do not respect the precondition.
+
+```cpp
+int strlen(const char *c) {
+  if (c == 0)
+    return 0; // Or might throw exception in case of C++.
```
whisperity: Perhaps mention setting an error flag, which is a more C-ic behaviour?

docs/false_positives.md
```diff
@@ -235,6 +236,84 @@ the analyzer better. Note that the compiler might generate the same code
 for the two implementations, so it might make sense to use only the more
 obvious one.
 
+## <a name="defensive-checks"></a> Defensive checks
+
+This section is an odd one because we describe a technique to reduce false
```
whisperity: Should the user really be interested in the deep reasoning behind why a particular subsection is part of this docfile, and not the other one?

docs/false_positives.md
```diff
@@ -235,6 +236,84 @@ the analyzer better. Note that the compiler might generate the same code
...
+int main() {
+  return strlen(0);
+}
+```
+
+In the code above the analyzer will not be able to find any errors since it has
+no way to tell whether the defensive checks are due to contract violations or
+they are part of the defined behavior.
+
+In order to reduce false negatives due to those safety checks we have several
```
whisperity: I would elaborate more on how this particular case results in a false negative. (It might be me being out of the loop as I haven't looked at code for weeks now, but can't seem to figure out at first or second read.)

docs/false_positives.md
```diff
@@ -235,6 +236,84 @@ the analyzer better. Note that the compiler might generate the same code
...
+```cpp
+int strlen(const char * _Nonnull c);
+```
+* We can just omit the safety check. This might not be feasible in every 
+scenario. But in case the only purpose of the check is debugging using 
+sanitizers and other dynamic analysis tools is always a viable alternative.
+* Guard the safety checks with a macro.
+```cpp
+int strlen(const char *c) {
+#ifndef __clang_analyzer__
```
whisperity: Is this particular macro explained anywhere in this file?
(Is this given automatically and in a well defined manner when Clang SA is being run? What about Clang-Tidy?)

docs/false_positives.md
Xazax-hun: I think the main information here is that this is about how to reduce false negatives, not false positives. But I will remove the last sentence from this paragraph.

docs/false_positives.md
Xazax-hun: The paragraph above this one supposed to explain this.

docs/false_positives.md
Xazax-hun: Added an explanation to the first mention of this macro.

docs/false_positives.md
```diff
@@ -128,6 +130,52 @@ Other macros or builtins expressing unreachable code may be used.
 Note that the rewritten code is also safer, since debug builds now check for
 more precondition violations. 
 
+### <a name="loops"></a> Loops
+
+Some loops are guaranteed to execute at least once and this is a dynamic
+invariant of the program.
+In some cases the analyzer cannot prove this property of a loop and it will
+simulate the path when the loop is not executed at all. It can be due to the
+lack of context or the code being to complex to the analyzer.
```
whisperity: s/to/too

docs/false_positives.md
```diff
@@ -128,6 +130,52 @@ Other macros or builtins expressing unreachable code may be used.
...
+  int count = 0;
+  for(; l != 0; l = l->next) {
+    sum += l->data;
+    ++count;
+  }
+  return sum/count; // Warning, division by zero.
+}
+```
+
+Without a calling context the analyzer cannot now that the `List` guaranteed to
```
whisperity: **k**now that ~~the~~ `List` **is**

docs/false_positives.md
```diff
@@ -128,6 +130,52 @@ Other macros or builtins expressing unreachable code may be used.
...
+    ++count;
+  }
+  return sum/count; // No warning.
+}
+```
+
+Adding the `assert` will make the code cleaner for the reader because it makes
+an important invariant of the program explicit. Moreover, it will make the
+false positive finding disappear. This will also provide the users of the code
+with an explicit check in the debug build which can help us find bugs.
```
whisperity: *us* is not needed here

docs/false_positives.md
```diff
@@ -177,6 +225,8 @@ turn off checks from the `core` package.
 How to suppress a specific dead store warning from the Clang Static Analyzer
 and more useful tips can be found [here](https://clang-analyzer.llvm.org/faq.html).
 Alternatively, the `__clang_analyzer__` macro can be used to introduce usages.
+This macro is automatically defined when the code is analyzed using the clang
```
whisperity: It's a name: Clang Static Analyzer

PR #1276, opened 2018-01-03
use manager to share data between processes

gyorb:
Using a plain dictionary shared between processes
is not safe and can cause performance degradation too.
Proxy objects should be used.

Move BuildAction preprocessing to analyzer module where
the other build action processing is done.

### DIFF
--- a/libcodechecker/analyze/analyzer.py
+++ b/libcodechecker/analyze/analyzer.py
@@ -6,7 +6,9 @@
 """
 Prepare and start different analysis types
 """
+
 import copy
+from multiprocessing import Manager
 import os
 import shlex
 import shutil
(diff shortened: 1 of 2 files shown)


libcodechecker/analyze/analyzer.py
```diff
@@ -122,13 +147,26 @@ def perform_analysis(args, context, actions, metadata):
...
 
+    # Use Manager to create data objects which can be
+    # safely shared between processes.
+    manager = Manager()
+
+    cfg_mgr = manager.dict()
+    for k, v in config_map.items():
+        cfg_mgr[k] = v
+
+    config_map = cfg_mgr
```
bruntib: These few lines could be replaced by this:
```.py
config_map = manager.dict(config_map)
```
And why isn't `config_map` a `manager.dict` originally?

gyorb: Until it is not shared between multiple processes it is not necessary to create a proxy object.
I just wanted to create the proper types before they are shared between processes. This way it is more explicit, if it was created in another module it would be less visible.

libcodechecker/analyze/analyzer.py
```diff
@@ -13,6 +13,8 @@
 import subprocess
 import time
 
+from multiprocessing import Manager
```
whisperity: Import at bad location.

whisperity:
Is this a general improvement, or a fix? Did the previous method of sharing have any visible symptoms?

gyorb:
@whisperity it is a bugfix because there was a performance degradation on older Linux kernels (2.6.32)

PR #1279, opened 2018-01-08
Ignore -Werror

bruntib:
In some cases Clang is stricter than GCC regarding the warning messages.
-Werror flag converts these to error messages which thus fail the
analysis. That's why -Werror will be ignored from now on.

### DIFF
--- a/libcodechecker/log/option_parser.py
+++ b/libcodechecker/log/option_parser.py
@@ -120,7 +120,6 @@ IGNORED_OPTION_MAP = {
     '-MD': 0,
     '-MV': 0,
     '-MMD': 0,
-    '-fsyntax-only': 0,
     '-save-temps': 0,
     '-install_name': 1,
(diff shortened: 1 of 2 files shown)


libcodechecker/log/option_parser.py
```diff
@@ -134,7 +134,11 @@
     '-sectorder': 3,
     '--param': 1,
     '-u': 1,
-    '--serialize-diagnostics': 1
+    '--serialize-diagnostics': 1,
+    # If Clang gives warning on which GCC doesn't then during checking it is
+    # considered an error and fails the analysis if the original build contains
+    # -Werror flag.
```
dkrupp: This is not a well written English sentence.
I suggest rewriting the comment to:
"Clang gives different warnings than GCC. Thus if this flag is kept, the analysis with Clang can fail even if the compilation passes with GCC. "

PR #1286, opened 2018-01-11
refactor serverside of storing runs

gyorb:
* StorageSession was removed an upcoming patch will it make obsolate
  (used to check if there is an already running store for a run,
  this feature is removed with the cleanup)

* massStoreRun was refactored to multiple methods

### DIFF
--- a/libcodechecker/server/api/report_server.py
+++ b/libcodechecker/server/api/report_server.py
@@ -423,78 +423,6 @@ def get_report_path_hash(report, files):
     return hashlib.md5(report_path_hash.encode()).hexdigest()
 
 
-class StorageSession:
-    """
-    This class is a singleton which helps to handle a transaction which
-    belong to the checking of an entire run. This class holds the SQLAlchemy
-    session for the run being checked and the set of touched reports. This
-    latter one is needed so at the end the detection status of the rest reports
-    can be set to "resolved".
-    """
-
-    class __StorageSession:
-        def __init__(self):
-            self.__sessions = dict()
-            self._timeout_sessions()
-
-        def start_run_session(self, run_id, transaction):
-            self.__sessions[run_id] = {
-                'touched_reports': set(),
-                'transaction': transaction,
-                'timer': time.time()}
-
-        def end_run_session(self, run_id):
-            this_session = self.__sessions[run_id]
-            transaction = this_session['transaction']
-
-            transaction.commit()
-            transaction.close()
-
-            del self.__sessions[run_id]
-
-        def abort_session(self, run_id):
-            transaction = self.__sessions[run_id]['transaction']
-            transaction.rollback()
-            transaction.close()
-            del self.__sessions[run_id]
-
-        def touch_report(self, run_id, report_id):
-            self.__sessions[run_id]['touched_reports'].add(report_id)
-
-        def is_touched(self, run_id, report_id):
-            return report_id in self.__sessions[run_id]['touched_reports']
-
-        def has_ongoing_run(self, run_id):
-            return run_id in self.__sessions
-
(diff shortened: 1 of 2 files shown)


gyorb:
The run locking feature will be available with #1147 again.

libcodechecker/server/api/store_handler.py
```diff
@@ -234,52 +231,42 @@ def addCheckerRun(session, storage_session, command, name, tag, username,
...
 
         run_history = RunHistory(run_id, tag, username, run_history_time)
         session.add(run_history)
-        session.flush()
 
-        storage_session.start_run_session(run_id, session)
+        session.flush()
         return run_id
     except Exception as ex:
-        session.close()
```
Xazax-hun: What is the reason this line being removed?

gyorb: The session will be closed automatically when the execution leaves the `DBSession` context. There is no need to close it there.

whisperity:
@Xazax-hun Considering that there most likely won't be a release between the two patches, no functionality will be lost.

libcodechecker/server/api/report_server.py
```diff
@@ -1827,19 +1754,168 @@ def getMissingContentHashes(self, file_hashes):
             return list(set(file_hashes) -
                         set(map(lambda fc: fc.content_hash, q)))
 
+    def __store_source_files(self, source_root, filename2hash):
```
whisperity: Shouldn't we use `_to_` instead of `2`?

libcodechecker/server/api/report_server.py
```diff
@@ -1827,19 +1754,168 @@ def getMissingContentHashes(self, file_hashes):
...
+        """
+        Storing file contents from plist.
+        """
+
+        file_path_to_id = {}
+
+        for file_name, file_hash in filename2hash.items():
```
whisperity: Geeky wording here too.

libcodechecker/server/api/report_server.py
whisperity: If there is a file named `store_handler`, shouldn't these codes be moved there instead?

libcodechecker/server/api/report_server.py
```diff
@@ -1865,180 +1949,33 @@ def massStoreRun(self, name, tag, version, b64zip, force):
...
-                                             # Round the duration to seconds.
-                                             int(sum(check_durations)))
+                                             durations)
 
-            # The database session initialized in addCheckerRun should be
-            # closed in finishCheckerRun after the report storage is done.
-            store_handler.finishCheckerRun(self.__storage_session, run_id)
+                store_handler.finishCheckerRun(session, run_id)
+
+                session.commit()
```
whisperity: This is definitely the only part we `commit`, yes?

libcodechecker/server/api/store_handler.py
whisperity: What are the commit rules for these? When `DBSession.__exit__`, is it just a `close()` without a commit, right?

libcodechecker/server/api/report_server.py
gyorb: I do not wanted to move it there so all of the database session creations are visible in the report server.

whisperity: Understandable.

libcodechecker/server/api/report_server.py
gyorb: The massStore is split up to two main store sections, at first the source files are stored (each file in a new session). Everything after the source files are stored is done in one session (storing the reports), it should be successful and committed once or it should fail and rollback.

PR #1292, opened 2018-01-15
Improvements to false positive guide. Some of them is based on feedba…

Xazax-hun:
…ck from reddit.

### DIFF
--- a/docs/false_positives.md
+++ b/docs/false_positives.md
@@ -22,6 +22,7 @@ Table of Contents
    * [Partial functions](#partial-functions)
    * [Loops](#loops)
    * [Prefer standard functions](#prefer-standard-functions)
+   * [Use `const` whenever possible](#use-const)
    * [Do not turn off `core` checks](#do-not-turn-off-core-checks)
    * [Suppress specific dead store warnings](#suppress-specific-dead-store-warnings)
    * [Alternative implementations](#alternative-implementations)
@@ -130,6 +131,22 @@ Other macros or builtins expressing unreachable code may be used.
 Note that the rewritten code is also safer, since debug builds now check for
 more precondition violations. 
 
+In case of C++11 or later, another option is to use Immediately-Invoked
+Function Expression (IIFE) to avoid assigning a meaningless value.
+
+```cpp
+int f(MyEnum Val) {
+  const int x = [&] { // Note the lambda.
+    switch (Val) {
+      case MyEnumA: return 1;
+      case MyEnumB: return 5;
+      default: assert(false); return 0;
+    }
+  } ();
+  return 5/x; // No warning.
+}
+```
+
 ### <a name="loops"></a> Loops
 
 Some loops are guaranteed to execute at least once and this is a dynamic
@@ -146,7 +163,7 @@ Let us look at the the following example:
 int avg(List *l) {
   int sum = 0;
   int count = 0;
-  for(; l != 0; l = l->next) {
+  for(; l != NULL; l = l->next) {
     sum += l->data;
     ++count;
   }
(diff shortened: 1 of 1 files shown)


docs/false_positives.md
```diff
@@ -247,6 +266,41 @@ The result is easier to read for other developers who might not be familiar
 with the custom version of the function. The standard library functions also
 tend to be faster and more correct than custom solutions.
 
+### <a name="use-const"></a> Use const whenever possible
+
+It is useful to tell the analyzer when the state is not going to change.
+This will make the analysis more precise and the code more readable.
+Mark methods that are not going to change the state of the object as const.
+
+Also, avoid using global variables they are harder to reason about.
```
rnkovacs: perhaps: avoid using global variables **as/because** they are harder to reason about

docs/false_positives.md
```diff
@@ -130,6 +131,20 @@ Other macros or builtins expressing unreachable code may be used.
 Note that the rewritten code is also safer, since debug builds now check for
 more precondition violations. 
 
+In case of C++11 or later, another option is to use immediately evaluated
+function expression (IEFE) to avoid assigning a meaningless value.
```
rnkovacs: Googling this expression does not seem to throw up anything. Is it possible that you meant [Immeadiately-Invoked Function Expression (IIFE)](https://cppnow2015.sched.com/event/2np7/iife-in-c-for-performance-and-safety), which seems to be a widely used term?

docs/false_positives.md
```diff
@@ -130,6 +131,22 @@ Other macros or builtins expressing unreachable code may be used.
 Note that the rewritten code is also safer, since debug builds now check for
 more precondition violations. 
 
+In case of C++11 or later, another option is to use Immediately-Invoked
+Function Expression (IIFE) to avoid assigning a meaningless value.
+
+```cpp
+int f(MyEnum Val) {
+  const int x = [&] {
```
whisperity: This code snipper looks too much alike the previous example. I think you should put a comment in this line (and at the call site of the lambda) to mark that this is the difference.

docs/false_positives.md
```diff
@@ -247,6 +269,41 @@ The result is easier to read for other developers who might not be familiar
 with the custom version of the function. The standard library functions also
 tend to be faster and more correct than custom solutions.
 
+### <a name="use-const"></a> Use const whenever possible
+
+It is useful to tell the analyzer when the state is not going to change.
+This will make the analysis more precise and the code more readable.
+Mark methods that are not going to change the state of the object as const.
```
whisperity: `const` is a keyword, use backticks.

whisperity:
Where is this reddit thread? Do we have to explicity mention the site in the commit history?

docs/false_positives.md
```diff
@@ -247,6 +269,41 @@ The result is easier to read for other developers who might not be familiar
 with the custom version of the function. The standard library functions also
 tend to be faster and more correct than custom solutions.
 
+### <a name="use-const"></a> Use const whenever possible
```
whisperity: Backtick const

docs/false_positives.md
```diff
@@ -247,6 +269,41 @@ The result is easier to read for other developers who might not be familiar
...
+
+It is useful to tell the analyzer when the state is not going to change.
+This will make the analysis more precise and the code more readable.
+Mark methods that are not going to change the state of the object as `const`.
+
+Also, avoid using global variables as they are harder to reason about.
+Mark global constants as consts.
```
whisperity: Backtick const

docs/false_positives.md
```diff
@@ -247,6 +269,41 @@ The result is easier to read for other developers who might not be familiar
...
+Tha analyzer might not be able to tell that the value of `allc` is always
+the same as the length of the array `allv`. Rewriting the code and marking
+`allc` const will solve this issue.
+
+```cpp
+static char *allv[] = { "prog", "arg1", "arg2" };
+static const int allc = sizeof(allv) / sizeof(allv[0]);
+static void f(void) {
+  for (int i = 1; i < allc; i++) {
+    const char *p = allv[i];  // Warning, out of bounds.
```
whisperity: Is the issue solved? Then why do we get the warning here too?

PR #1296, opened 2018-01-16
Fix python-ldap not throwing exception on anonymous binds when it should

whisperity:
See for further details: https://mail.python.org/pipermail/python-ldap/2012q4/003183.html

### DIFF
--- a/libcodechecker/libauth/cc_ldap.py
+++ b/libcodechecker/libauth/cc_ldap.py
@@ -99,7 +99,7 @@ def log_ldap_error(ldap_error):
     """
     Log the LDAP error details in debug mode.
     """
-    toprint = ''
+    toprint = 'LDAP Error: '
     if ldap_error.message:
         if 'info' in ldap_error.message:
             toprint = toprint + ldap_error.message['info']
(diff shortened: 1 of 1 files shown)


libcodechecker/libauth/cc_ldap.py
```diff
@@ -218,14 +218,27 @@ def __init__(self, ldap_config, who=None, cred=None):
...
                     res = self.connection.simple_bind_s()
                     LOG.debug(res)
                 else:
-                    # Bind with the given credentials.
-                    LOG.debug(who)
+                    LOG.debug("Binding with credential: " + who)
                     res = self.connection.simple_bind_s(who, cred)
+                    whoami = self.connection.whoami_s()
                     LOG.debug(res)
+                    LOG.debug(whoami)
```
gyorb: Could this leak some sensitive information to the logs?

whisperity: No. It is either `None` or the username, which is already logged above.

whisperity:
@gyorb Added, and also explained it a bit in more detail. Having mailing list links in the code is not too good, what if the mailing list goes out of business?

gyorb:
@whisperity yes I know it is possible that the mailing list will not be accessible but I still prefer the idea that most of the information should be in the source.
I'm not sure if there will be any other problem with ldap authentication everybody will search for the pull request in github to read the comments here.

PR #1298, opened 2018-01-17
Add better diagnostics for ctu tests in case of failure

martong:
Trying to fix the mysterious Jenkins error by having better output in case of failing tests.


tests/functional/ctu/ctu_failure/test_ctu_failure.py
```diff
@@ -9,7 +9,6 @@
 import json
 import os
 import shutil
-import subprocess
 import unittest
```
gyorb: Shouldn't call_command be imported here?

PR #1300, opened 2018-01-17
Explicitly show version information in the build script output

whisperity:
Add the following output to the build script to better trace release creation.

```
[INFO] - This is CodeChecker v6.5 (df359f07632960f12ccb32609ea011ae476fbee4)
[INFO] - Built from Git tags: v6.4 (v6.4-11-gdf359f0)
[INFO] - Baked with Git tag information merged with configured version: 6.5 (6.5-11-gdf359f0)
```

### DIFF
--- a/scripts/build_package.py
+++ b/scripts/build_package.py
@@ -668,13 +668,15 @@ def build_package(repository_root, build_package_config, env=None):
                                                      cwd=repository_root)
         git_describe_dirty = str(git_describe_dirty.rstrip())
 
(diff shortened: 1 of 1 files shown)


gyorb:
Won't be showing the previous git tag a bit confusing?

whisperity:
I don't think so. It's better for the releaser to trace the content from which the thing was created. From the Git history's perspective, manually overriding the version in the git describe output (to have the config file version, and not the git tag, as it *SHOULD* be) can be more confusing.

PR #1305, opened 2018-01-18
ldap hardening and tests

gyorb:
* better error handling in case of
  - wrong ldap config
  - username or credentials are missing
* add ldap authentication test

### DIFF
--- a/libcodechecker/libauth/cc_ldap.py
+++ b/libcodechecker/libauth/cc_ldap.py
@@ -190,9 +190,10 @@ class LDAPConnection(object):
         None if initialization failed.
         """
 
-        ldap_server = ldap_config['connection_url']
+        ldap_server = ldap_config.get('connection_url')
         if ldap_server is None:
             LOG.error('Server address is missing from the configuration')
+            self.connection = None
             return
 
         referals = ldap_config.get('referals', False)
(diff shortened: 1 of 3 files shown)


libcodechecker/libauth/cc_ldap.py
```diff
@@ -273,6 +274,11 @@ def auth_user(ldap_config, username=None, credentials=None):
     """
     Authenticate a user.
     """
+    if not username:
```
whisperity: Should an error be printed here? Either way, I think for this we could express it shorter:

```python
if not username or not credentials:
```

tests/requirements.txt
```diff
@@ -1,2 +1,3 @@
 nose
 pycodestyle
+mockldap
```
whisperity: Is there a reason behind the order here? If not, could we make it lexicographical, so `mockldap` is the first?

tests/unit/test_ccldap.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def setUp(self):
+        # Patch ldap.initialize
+        self.mockldap.start()
+        self.ldapobj = self.mockldap['ldap://localhost/']
+
+    def tearDown(self):
+        # Stop patching ldap.initialize and reset state.
+        self.mockldap.stop()
+        del self.ldapobj
```
whisperity: Is this for explicit garbage collection triggering?

gyorb: Yes it's there so each test runs with a clean state.
For each test it will be initialized:
```
def setUp(self):
    # Patch ldap.initialize
    self.mockldap.start()
    self.ldapobj = self.mockldap['ldap://localhost/']
```
It could be set on class level, so all the tests use the same mock, but I think it is cleaner this way.

PR #1308, opened 2018-01-22
Use arch of the analizer machine instead of the original one.

bruntib:
When creating the content of the ctu-dir the target architecture was
used to which compilation happened. If the analysis runs on a different
architecture (e.g. for debugging reasons) then the analyzer can't find
the necessary files of which the paths are named after the architecture.

### DIFF
--- a/scripts/debug_tools/prepare_all_cmd_for_ctu.py
+++ b/scripts/debug_tools/prepare_all_cmd_for_ctu.py
@@ -7,6 +7,7 @@
 import argparse
 import json
 import os
+import platform
 import subprocess
 
 import prepare_compile_cmd
@@ -33,6 +34,18 @@ def execute(cmd):
     except OSError:
         print('Failed to run: "' + ' '.join(cmd) + '"')
         raise
+
+
+def get_triple_arch(analyze_command_file):
+    with open(analyze_command_file) as f:
+        cmd = f.readline()
+
+    cmd = cmd.split()
+    for flag in cmd:
+        if flag.startswith('--target='):
+            return flag[9:].split('-')[0]  # 9 == len('--target=')
+
+    return platform.machine()
 
 
 if __name__ == '__main__':
@@ -95,6 +108,7 @@ if __name__ == '__main__':
                    "--verbose", "debug"])
 
     analyzer_command_debug = "analyzer-command_DEBUG"
+    target = get_triple_arch('./analyzer-command')
     with open(analyzer_command_debug, 'w') as f:
         f.write(
             prepare_analyzer_cmd.prepare(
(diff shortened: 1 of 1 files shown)


gerazo:
It is by design that we use the target architecture since this is the only way to have multiple directories for multiple analysis on the same machine if the code is cross-compiled to different architectures at once. This is why it is important to get this arch setting from clang itself instead of relying on the current machine we are running on. The ctu-dir is supposed to have temporary stuff which is only good for the same clang arch and version. If you use these internals in a different version of clang (as I understand, you want to use an other clang on a different platform), there is no guarantee that things don't break.

gerazo:
(...and I'm afraid we don't have tests for cross-compiled, multiple-arch-in-one-build projects, so be aware!)

bruntib:
I thought it had a reason. But the other option fixing this would be a fix in the Clang code base where this externalFnMap is read.

whisperity:
So what is the issue itself? Clang tries to read the externalFnMap from its own triple path instead of using the triple generated from the command that is getting analysed?

bruntib:
OK, other solution uploaded.

scripts/debug_tools/prepare_all_cmd_for_ctu.py
```diff
@@ -35,6 +36,18 @@ def execute(cmd):
...
+def get_triple_arch(analyze_command_file):
+    with open(analyze_command_file) as f:
+        cmd = f.readline()
+
+    cmd = cmd.split()
+    for flag in cmd:
+        if flag.startswith('--target='):
+            return flag[9:].split('-')[0]  # 9 == len('--target=')
+
+    return platform.machine()
```
martong: `platform.machine()` would return the same string what clang uses when `--target` is not given?

martong:
@gerazo Recently we changed CTU code in Clang to not allow merging ASTs from different targets.
https://github.com/Ericsson/clang/pull/244/files

scripts/debug_tools/prepare_all_cmd_for_ctu.py
whisperity: Debatable. What happens if the architecture of Python and Clang differs, exactly one of them is 32-bit?

bruntib: This is just a fallback in case there is no `--target` given in the `clang` command invocation. This `platform.machine()` returns the architecture of the current machine where this Python script is run. If it differs from the one on which the build action was run, then it still doesn't work, like before this patch.

martong: Ok, let's merge it now because we need this badly and most often `--target` is given.

PR #1326, opened 2018-01-25
Fix non-existing current working directory

bruntib:
For some reason sometimes the folders in "directory" attributes of the
compile_cmd.json don't exist, however these are required to be used as a
current working directory during the analysis.

### DIFF
--- a/libcodechecker/analyze/analyzer.py
+++ b/libcodechecker/analyze/analyzer.py
@@ -143,6 +143,15 @@ def perform_analysis(args, context, actions, metadata):
         LOG.error("CTU directory:'" + ctu_dir + "' does not exist.")
         return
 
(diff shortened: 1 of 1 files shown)


gyorb:
Is it possible that the build system (make ...) deletes these directories, so they exist when the log is run but they are missing at the analysis phase?

bruntib:
OK, I created another solution which is at analysis level.
By the way the first solution was created in CTU related scripts because this error happened only when CTU related subprocesses were run, like creating the `externalFnMap` file, etc.

bruntib:
This directory creation is only necessary for the debug scripts, thus it is not neccessary to solve this issue in the analyzer.
For the debug scripts there was another solution in fc912e0, so this is done.

PR #1331, opened 2018-01-25
Add Clang include path in debug scripts

bruntib:
Some standard header files should be present during the analysis from
the standard library. The location of these is now added to the prepared
analysis command.

### DIFF
--- a/scripts/debug_tools/prepare_analyzer_cmd.py
+++ b/scripts/debug_tools/prepare_analyzer_cmd.py
@@ -64,7 +64,26 @@ class PathOptions:
 def prepare(analyzer_command_file, pathOptions):
     res = lib.change_paths(get_first_line_of_file(analyzer_command_file),
                            AnalyzerCommandPathModifier(pathOptions))
-    return res
+
+    if '-nobuiltininc' not in res:
+        return res
+
+    # Find Clang include path
+    clang_include_path = lib.get_resource_dir(pathOptions.clang) + '/include'
+
+    if clang_include_path is None:
+        clang_lib_path = os.path.dirname(pathOptions.clang) + '/../lib'
+        clang_include_path = ''
+        for path, _, files in os.walk(clang_lib_path):
+            if 'stddef.h' in files:
+                clang_include_path = path
+                break
+
+    if clang_include_path is None:
+        return res
+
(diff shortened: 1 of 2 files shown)


scripts/debug_tools/prepare_analyzer_cmd.py
```diff
@@ -64,7 +64,24 @@ def __init__(
...
     res = lib.change_paths(get_first_line_of_file(analyzer_command_file),
                            AnalyzerCommandPathModifier(pathOptions))
-    return res
+
+
+    if '-nobuiltininc' not in res:
+        return res
+
+    # Find Clang include path
+    clang_lib_path = os.path.dirname(pathOptions.clang) + '/../lib'
```
gyorb: Isn't there an option for clang like in gcc to get the builtin includes?

martong: Yes there is an option: `-print-resource-dir` , but that is only available in Clang >= 5.0 .

scripts/debug_tools/prepare_analyzer_cmd.py
```diff
@@ -64,7 +64,24 @@ def __init__(
...
+    clang_include_path = ''
+    for path, _, files in os.walk(clang_lib_path):
```
martong: Would `os.walk` go down in the directory tree into lower levels? E.g. in my case the file is here: `../lib/clang/5.0.0/include/stddef.h` ?

scripts/debug_tools/prepare_analyzer_cmd.py
martong: I think we could call clang with `-print-resource-dir` and if that is not supported (which is very unlikely, since we use at least 5.0 for CTU) then fall back to `os.walk`.
See: https://github.com/Ericsson/codechecker/blob/master/libcodechecker/analyze/host_check.py#L64

scripts/debug_tools/prepare_analyzer_cmd.py
bruntib: Yes, that's the goal of `os.walk()`. It visits all subdirectories recursively.

martong:
Travis still fails because of some style stuff :(

bruntib:
Done

PR #1334, opened 2018-01-29
refactor plist to plaintext formatting (parse)

gyorb:
Plist parsing and formatting used by the old architecture is not
the way it should be handled now. It was not flexible and it was hard
to improve it.
It should be easier to test and setup the output formatting after the
changes.

Skip lists were not supported by the parse command, this
commit will add the skip files support for parse too.


libcodechecker/analyze/plist_parser.py
```diff
@@ -268,3 +271,188 @@ def skip_report_from_plist(plist_file, skip_handler):
...
+            sp_handler = suppress_handler.SourceSuppressHandler(source_file,
+                                                                report_line,
+                                                                report_hash,
+                                                                checker_name)
+
+            # Check for suppress comment.
+            suppress_data = sp_handler.get_suppressed()
+            if suppress_data:
+                if self.suppress_handler:
+                    LOG.info("Writing source-code suppress at '{0}:{1}' to "
```
whisperity: Is this output true? This should only be printed if `--export-suppress` is given to `parse`. What happens if a suppress file is given but we don't want to export into it?

gyorb: If a suppress file is given than the report will be printed by parse only if it is not suppressed in the suppress file.

whisperity: I understand that. What I meant is the log output here if it holds in execution. I feel this log shouldn't be here.

gyorb: I've changed it to debug log level. It might be useful if there is some problem.

PR #1377, opened 2018-02-15
bug event and point to report id index

gyorb:
With the new indexes deleting the reports should be much faster.
Storing the reports in update mode (using the same run name) and
deleting a run should be much faster.

### DIFF
--- a/db_migrate/versions/101a9cb747de_add_bug_event_and_point_report_index.py
+++ b/db_migrate/versions/101a9cb747de_add_bug_event_and_point_report_index.py
@@ -1,0 +1,24 @@
+"""add bug event and point report index
+
+Revision ID: 101a9cb747de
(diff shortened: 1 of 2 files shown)


gyorb:
Some measurement reports using PostgreSQL:

Total number of reports: 37874
Unique reports: 15741

Without the indexes:
First store:  9,42s user 0,18s system 7% cpu **2:13,49 total**
Second store on the same run: 9,46s user 0,18s system 1% cpu **11:47,78 total**
Deleting the run: **601677.908 ms**

With indexes:
First store:  9,47s user 0,15s system 6% cpu **2:18,24 total**
Second store on the same run: 9,20s user 0,22s system 6% cpu **2:20,77 total**
Deleting the run: **383.538 ms**

PR #1379, opened 2018-02-16
move plist parsing out of the database session

gyorb:
Plist parsing can take some time and keeping a database session
until it is done might not be the best solution however with the
change we need to parse up all the reports and store them in memory
until they will be written to the database.

### DIFF
--- a/libcodechecker/server/api/report_server.py
+++ b/libcodechecker/server/api/report_server.py
@@ -1817,8 +1817,9 @@ class ThriftRequestHandler(object):
                                                      None)
         return file_path_to_id
 
-    def __store_reports(self, session, report_dir, source_root, run_id,
-                        file_path_to_id, run_history_time, severity_map):
+    def __store_reports(self, session, source_root, run_id,
+                        file_path_to_id, run_history_time, severity_map,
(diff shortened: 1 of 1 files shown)


Xazax-hun:
Did you check how much more memory do we need? What are the side effects to keep the session open for longer? This is a trade-off, so it would be good to know whether it is really worth it.

gyorb:
I've used [guppy](https://pypi.python.org/pypi/guppy) to measure the heap size and
 it was done at the end of the `massStoreRun` api call in the server.

Storing ~15000 results in 200 runs.
The memory usage where plist parsing is done one by one inside a db session was around 30MB
Parsing all the plist files before the db session the memory usage was around 180MB

Xazax-hun:
I see, in this case I do prefer to abandon this change. While we keep the database connection open a bit longer this enables us to keep the memory usage down. Moreover, I do not see any problems with keeping the connection open unless we plan to handle lots of stores for the same product from different users at the same time. I think this is not the use case we are trying to support at the moment.

gyorb:
changing to this solution would increase the memory usage to much for a store, we should investigate other options later if needed

PR #1442, opened 2018-03-09
cleanup some pylint warnings

gyorb:


libcodechecker/server/api/report_server.py
```diff
@@ -13,12 +13,15 @@
...
 import hashlib
 import json
+import os
 import sys
 import tempfile
 import zipfile
 import zlib
 
 import sqlalchemy
+from sqlalchemy.sql.expression import or_, and_, func, \
```
Xazax-hun: Do we need this change and the `import os` above? I do not see any usages in the changes but I might miss something.

libcodechecker/server/api/store_handler.py
```diff
@@ -5,6 +5,7 @@
 # -------------------------------------------------------------------------
 
 import base64
+from datetime import datetime
```
Xazax-hun: I also do not see any usages of this import. Is this a fix for a crash?

libcodechecker/server/api/report_server.py
gyorb: The previous import with `*` imported the `os` too that is why we could use it.
Importing `*` is a bad practice because you do not know precisely what is imported.

It is the same case with importing `datetime`.

libcodechecker/server/api/store_handler.py
gyorb: Please see my answer at the `os` import question.

PR #1446, opened 2018-03-10
add initial pylintrc

gyorb:


.pylintrc
```diff
@@ -0,0 +1,378 @@
...
+
+# List of strings which can identify a callback function by name. A callback
+# name must start or end with one of those strings.
+callbacks=cb_,_cb
+
+
+[BASIC]
+
+# List of builtins function names that should not be used, separated by a comma
+bad-functions=map,filter,input
```
Xazax-hun: Are we banning `map` and `filter` on purpose?

gyorb: This is just a generated initial config file which should be modified, most of the values kept as they are. I've done only some small changes.
I've put it there so if any of us runs pylint we will get the same result.

whisperity:
@gyorb Will this tool be added to the test makefiles?

PR #1457, opened 2018-03-12
update gitignore

gyorb:


gyorb:
we already have some editor specific entries, I'm not against them

PR #1490, opened 2018-03-29
Add ctu-clang6 compatibility

gerazo:
Clang 6.0 has new parameters to trigger CTU functionality. Also the API was renamed from XTU to CTU. This puts both 5.0 and 6.0 compatible parameters into the clangsa call, so all clang versions will work.

### DIFF
--- a/libcodechecker/analyze/analyzers/analyzer_clangsa.py
+++ b/libcodechecker/analyze/analyzers/analyzer_clangsa.py
@@ -172,9 +172,18 @@ class ClangSA(analyzer_base.SourceAnalyzer):
                                          '-Xclang', checker_name])
 
             if config.ctu_dir and not self.__disable_ctu:
+                # ctu-clang5 compatibility
                 analyzer_cmd.extend(['-Xclang', '-analyzer-config',
                                      '-Xclang',
-                                     'xtu-dir=' + self.get_xtu_dir()])
+                                     'xtu-dir=' + self.get_ctu_dir()])
+                # ctu-clang6 compatibility (5.0 and 6.0 options work together)
+                analyzer_cmd.extend(['-Xclang', '-analyzer-config',
+                                     '-Xclang',
+                                     'experimental-enable-naive-ctu-analysis'
+                                     '=true',
+                                     '-Xclang', '-analyzer-config',
+                                     '-Xclang',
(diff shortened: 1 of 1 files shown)


libcodechecker/analyze/analyzers/analyzer_clangsa.py
```diff
@@ -172,9 +172,18 @@ def construct_analyzer_cmd(self, result_handler):
...
+                # ctu-clang5 compatibility
                 analyzer_cmd.extend(['-Xclang', '-analyzer-config',
                                      '-Xclang',
-                                     'xtu-dir=' + self.get_xtu_dir()])
+                                     'xtu-dir=' + self.get_ctu_dir()])
+                # ctu-clang6 compatibility (5.0 and 6.0 options work together)
+                analyzer_cmd.extend(['-Xclang', '-analyzer-config',
+                                     '-Xclang',
+                                     'experimental-enable-naive-ctu-analysis'
+                                     '=true',
```
Xazax-hun: I do not like this line break before `=true`. Maybe starting the list in the next line would help to reduce the indent? This way we could keep this string together.

gerazo:
Yes. With this, no remaining XTU references are there in the code.

libcodechecker/analyze/analyzers/analyzer_clangsa.py
gerazo: pep8, flake8 and pylint liked this.

gyorb:
@gerazo this will work with the current clang5 ctu branch too right?

gerazo:
Yes, it will.

2018. március 29. 16:35:07 CEST, Gyorgy Orban <notifications@github.com> írta:

libcodechecker/analyze/analyzers/analyzer_clangsa.py
whisperity: Agreed, this is very bad for the eye. You could try the following:

```python
analyzer_cmd.extend(
    ['-Xclang', '-analyzer-config',
     '-Xclang', 'experimental-enable-naive-ctu-analysis=true'])
```

Or if it doesn't work, you could also create a string above with this literal. If everything else fails, as a module "constant".

gerazo:
New formatting is in place.

PR #1491, opened 2018-03-29
skip dependency generator compiler calls from the analysis

gyorb:


tests/unit/test_log_parser.py
```diff
@@ -141,3 +142,64 @@ def test_new_intercept_build(self):
...
+            "file": "/tmp/a.cpp" }]
+        ''')
+        build_actions = \
+            log_parser.parse_compile_commands_json(preprocessor_actions,
+                                                   ParseLogOptions())
+        for build_action in build_actions:
+            self.assertTrue(build_action.skip)
+
+    def test_keep_compile_and_dep(self):
+        """ Keep the compile command if -MD is set.
```
martong: Why do we keep `-MD` ? If `-MD` is present there must be a a same compile command without the `-MD`. Typically the build systems first execute the dep generation and then they execute the real command which generates the .o.

gyorb: The `-MD` command does compilation (generates the .o file) together with the dependency generation if `-E` is not set, in that case you do not need a separate command to compile the source file.

gyorb: @martong do you have any other questions? I would keep the `-MD` because it does compilation if `-E` is not set.

martong: Okay.

martong: LGTM!

PR #1524, opened 2018-04-17
Apply ignore first, and ignore -flto flag.

bruntib:
Some standard headers are not found by clang-tidy if -flto flag is
given.


bruntib:
I ran it on an open source project and I didn't see any regression.

PR #1553, opened 2018-05-08
Issues with relative path in the compile json

lorincbalog:
If the logged build command contains an include like "-I relative_path", it is logged into the compile.json as relative to the build directory. However, if other CodeChecker commands try to use this compile.json outside the build directory, includes with relative path will not be available at the defined location. This fix tries to solve this by prepending the build directory if the include path does not represent a directory on the disk.


Xazax-hun:
Wouldn't this work if the working directory of the compiler/analyzer is set properly out of the box?
I.e. making sure to invoke the analyzer in the correct working directory.

PR #1581, opened 2018-05-14
Setting up checker severities

### LINKED ISSUE #1568: some of the clang6 checkers severity are not classified
```
 - alpha.core.DynamicTypeChecker                                  clangsa    UNSPECIFIED Check for cases where the dynamic and the static type of an object are unrelated.
 - alpha.core.StackAddressAsyncEscape                             clangsa    UNSPECIFIED Check that addresses to stack memory do not escape the function
 - alpha.cplusplus.DeleteWithNonVirtualDtor                       clangsa    UNSPECIFIED Reports destructions of polymorphic objects with a non-virtual destructor in their base class
 - alpha.security.taint.TaintPropagation                          clangsa    UNSPECIFIED Generate taint information used by other checkers
 + apiModeling.google.GTest                                       clangsa    UNSPECIFIED Model gtest assertion APIs
 - core.NonnilStringConstants                                     clangsa    UNSPECIFIED Assume that const string-like globals are non-null
 + cplusplus.SelfAssignment                                       clangsa    UNSPECIFIED Checks C++ copy and move assignment operators for self assignment
 - optin.portability.UnixAPI                                      clangsa    UNSPECIFIED Finds implementation-defined behavior in UNIX/Posix functions
 + unix.StdCLibraryFunctions                                      clangsa    UNSPECIFIED Improve modeling of the C standard library functions
 - bugprone-copy-constructor-init                                 clang-tidy UNSPECIFIED 
 + bugprone-integer-division                                      clang-tidy UNSPECIFIED 
 - bugprone-misplaced-operator-in-strlen-in-alloc                 clang-tidy UNSPECIFIED 
 - bugprone-virtual-near-miss                                     clang-tidy UNSPECIFIED 
 - cppcoreguidelines-c-copy-assignment-signature                  clang-tidy UNSPECIFIED 
 - cppcoreguidelines-owning-memory                                clang-tidy UNSPECIFIED 
 - fuchsia-default-arguments                                      clang-tidy UNSPECIFIED 
 - fuchsia-overloaded-operator                                    clang-tidy UNSPECIFIED 
 - fuchsia-virtual-inheritance                                    clang-tidy UNSPECIFIED 
 - google-runtime-references                                      clang-tidy UNSPECIFIED 
 - hicpp-deprecated-headers                                       clang-tidy UNSPECIFIED 
 - hicpp-move-const-arg                                           clang-tidy UNSPECIFIED 
 - hicpp-no-array-decay                                           clang-tidy UNSPECIFIED 
 - hicpp-no-malloc                                                clang-tidy UNSPECIFIED 
 - hicpp-static-assert                                            clang-tidy UNSPECIFIED 
 - hicpp-use-auto                                                 clang-tidy UNSPECIFIED 
 - hicpp-use-emplace                                              clang-tidy UNSPECIFIED 
 - hicpp-use-noexcept                                             clang-tidy UNSPECIFIED 
 - hicpp-use-nullptr                                              clang-tidy UNSPECIFIED 
 - hicpp-vararg                                                   clang-tidy UNSPECIFIED 
 - modernize-unary-static-assert                                  clang-tidy UNSPECIFIED 
 - readability-static-accessed-through-instance                   clang-tidy UNSPECIFIED 

```

dkrupp:
> Closes #1568

-Specifying severities for yet uncategorized checkers.
Many checkers are alias of each other. For them, the same
severity is specified.

-alpha.cplusplus.DeleteWithNonVirtualDtor is put into default
checker package as it seems to be working with a low false positive
rate.

-optin.portability.UnixAPI is put into the default package
as it warns for errorneous basic posix function calls (malloc with 0 size).

-new core.NonnilStringConstants is put into the default package as
it models basic const char strings, which otherwise would bring in false positives.


whisperity:
In the long run, checker aliases should be implemented either better in clang or in a wrapper layer in CodeChecker. What happens if the user turns off the same checker from name `A` but turns on name `B`?

Xazax-hun:
This question applies to clang tidy as well and CodeChecker should do whatever clang tidy does. We should not add extra semantics here in my opinion but only forward the arguments.

PR #1605, opened 2018-05-24
handle gcc-toolchain flag

gyorb:
--gcc-toolchain flag was not forwarded to the analyzers which
caused that not the right header files were detected

### DIFF
--- a/libcodechecker/analyze/log_parser.py
+++ b/libcodechecker/analyze/log_parser.py
@@ -13,6 +13,7 @@ import sys
 import traceback
 
 # TODO: This is a cross-subpackage import!
+from libcodechecker.analyze import gcc_toolchain
 from libcodechecker.log import build_action
 from libcodechecker.log import option_parser
 from libcodechecker.logger import get_logger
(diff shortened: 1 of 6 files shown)


libcodechecker/analyze/log_parser.py
```diff
@@ -277,12 +276,31 @@ def parse_compile_commands_json(logfile, parseLogOptions):
...
+        # gcc-toolchain parameter.
+        # Detecting extra include paths from the host compiler could
+        # conflict with this.
+
+        # For example if the compiler in the compile command is clang
+        # and gcc-toolchain is set we will get the include paths
+        # for clang and not for the compiler set in gcc-toolchain.
+        # This can cause missing headers during the analysis.
+
+        for compiler_option in action.analyzer_options:
```
martong: We must add the resource includes of Clang with -isystem before adding any of the system headers of the original compiler.
And we do that in `extend_analyzer_cmd_with_resource_dir`. Could you please check if `--gcc-toolchain` is set it works well with those settings? I mean ,the files from Clangs' resource dir must be the first one on the include search path.

libcodechecker/analyze/log_parser.py
```diff
@@ -277,12 +276,31 @@ def parse_compile_commands_json(logfile, parseLogOptions):
...
+            if '--gcc-toolchain' in compiler_option:
+                add_compiler_defaults = False
```
martong: If `add_compiler_defaults` is `False` then `get_compiler_includes` and `get_compiler_target` is never invoked. And that is a problem. Those functions create two additional json files which are essential in the failure zip; we use them to reproduce any failure locally.

Couldn't we just swallow the `--gcc-toolchain` param?

PR #1608, opened 2018-05-24
fix skip file arguments

### LINKED ISSUE #1607: AttributeError: 'Namespace' object has no attribute 'skip_file'
Using version 6.7.0
I am trying to run CodeChecker on the Clangd code base with a "skipfile" argument.

/foo/llvm/build$ CodeChecker parse -i SKIPFILE ./reports
  File "/bar/CodeChecker/codechecker-64bit-6.7.0/cc_bin/CodeChecker.py", line 93, in main
    args.func(args)
  File "/bar/CodeChecker/codechecker-64bit-6.7.0/lib/python2.7/libcodechecker/libhandlers/parse.py", line 160, in __handle
    main(args)
  File "/bar/CodeChecker/codechecker-64bit-6.7.0/lib/python2.7/libcodechecker/libhandlers/parse.py", line 294, in main
    with open(args.skip_file, 'r') as skip_file:
AttributeError: 'Namespace' object has no attribute 'skip_file'

The SKIPFILE has:
```
-*/llvm/include/*
-*/llvm/tools/clang/include/*
-/usr/include/*
-/usr/local/include/*
```

gyorb:
* wrong skip file argument name was used in parse command
* fix the test which should have trigger the error
* skip argument was not forwarded by the check command

resolves #1607

### DIFF
--- a/libcodechecker/libhandlers/check.py
+++ b/libcodechecker/libhandlers/check.py
@@ -466,6 +466,7 @@ def main(args):
         )
         __update_if_key_exists(args, parse_args, 'print_steps')
         __update_if_key_exists(args, parse_args, 'verbose')
+        __update_if_key_exists(args, parse_args, 'skipfile')
 
         parse_module = __load_module('parse')
         LOG.debug("Calling PARSE with args:")
--- a/libcodechecker/libhandlers/parse.py
+++ b/libcodechecker/libhandlers/parse.py
@@ -291,7 +291,7 @@ def main(args):
 
     skip_handler = None
     if 'skipfile' in args:
-        with open(args.skip_file, 'r') as skip_file:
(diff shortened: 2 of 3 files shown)


MarkZ3:
Is it possible to add a test for the "skip file" argument? It's the second bug that it had in not too long and a simple smoke test would have found those.

gyorb:
@MarkZ3 we have a test for the `check` command which includes testing `analyze` and `parse` (with skip files too).
We found another error where the check did not forward the skip argument properly that is why our tests did not fail and did not found the problem. The `parse` command gets tested through the `check` command right now.
I can add separate tests for `parse` just to make sure.

gyorb:
I fixed the problem in the test cases. That is why we did not detect the error.

MarkZ3:
Thank you!

PR #1612, opened 2018-05-29
send sigabort to clang in case of timeout

gyorb:
With sigabrt signal a core dump will be created which could help the
debugging in case of an error.

### DIFF
--- a/libcodechecker/analyze/analysis_manager.py
+++ b/libcodechecker/analyze/analysis_manager.py
@@ -536,7 +536,7 @@ def check(check_data):
                     called. Set up a timeout for the analysis.
                     """
                     timeout_cleanup[0] = util.setup_process_timeout(
-                        analyzer_process, analysis_timeout, signal.SIGKILL)
+                        analyzer_process, analysis_timeout, signal.SIGABRT)
             else:
                 def __create_timeout(analyzer_process):
                     # If no timeout is given by the client, this callback


whisperity:
Is `ulimit -c 0` before calling CodeChecker inherited into the subprocesses started by Python? In this case, the fact that core dumps are getting generated should be well-documented, and if users don't want massive core dumps, turn generating core dumps off.

But this is very invasive... can we reasonably manage "free space left on the device"? You need to take into account quotas and mount points and such, which, I know like 3 different ways to configure it on a Linux machine...

If we roll with core dumps, perhaps the best option would be to fail the entire `analyze` invocation once a subprocess times out. Or generate a single core dump only...

There doesn't seem to be a good answer, if we don't want to introduce more command-line options... :scream:.

Are massive hangs so prevalent? This is an easy line and an easy place to fix. Perhaps users who have their environment crapping up should simply replace this line in their own environment, run the analysis, the contact us or other Clang gurus to figure out why things went sideways. Then change the line back, if they want. We don't, can't, and shouldn't know about everyone's environment... Soon I think maybe a "CodeChecker OS" fork of Linux should just be developed.

gyorb:
Keeping it as it is for now, until a better solution is available.

irishrover:
> Keeping it as it is for now, until a better solution is available.

Anyway, there's  no `signal` param  in `def setup_process_timeout(proc, timeout,  failure_callback=None):` since my fix for timeout.

PR #1618, opened 2018-05-31
Yet another unknown GCC flag (-mfloat-gprs=double)

bruntib:


libcodechecker/log/option_parser.py
```diff
@@ -148,7 +148,9 @@
     # Link Time Optimization:
     '^-flto': 0,
     # MicroBlaze Options:
-    '^-mxl': 0
+    '^-mxl': 0,
+    # PowerPC SPE Option:
+    '^-mfloat-gprs': 0
```
gyorb: Please add a test for it `-mfloat-gprs=double`:
https://github.com/Ericsson/codechecker/blob/29df12c79fe4bc8d980442e85580bc6006e766ff/tests/unit/test_option_parser.py#L191

PR #1629, opened 2018-06-06
Taking out some checks from the sensitive profile

dkrupp:
* cppcoreguidelines-no-malloc is removed as when legacy C code is transformed into C++ there could be many malloc usage that is correct, but was not converted to RAII* cppcoreguidelines-pro-type-reinterpret-cast should be taken out from the sensitive profile as there may be cases when such potentially dangerous casts are necessary
* cppcoreguidelines-owning-memory This checker propagates the usage of the cppcoreguideline library to indicate ownership of standard (non-smart pointers). In general gsl::owner<> is not required to be used in Ericsson codebase. So this check is not generic enough.
* google-build-using-namespace is only style suggestion is not generic enough


Xazax-hun:
I very rarely find cases where reinterpret_cast is necessary. Otherwise looks good.

dkrupp:
we found many reinterpret_cast warnings in the code generate d by the google protobuf in the GOOGLE_PROTOBUF_GENERATED_MESSAGE_FIELD_OFFSET macro.

And also some in other places in production code which seemed to be valid.

Xazax-hun:
3rd party libs can be added to skiplists. But if that is what the users want I am fine with this.

PR #1668, opened 2018-07-11
sysroot parameter can be given multiple ways

bruntib:
The --sysroot flag can be given a directory these ways, but only the
first one was supported by CodeChecker when fetching implicit include
paths:

--sysroot=/path/to/sysroot
--sysroot /path/to/sysroot

### DIFF
--- a/libcodechecker/analyze/log_parser.py
+++ b/libcodechecker/analyze/log_parser.py
@@ -170,8 +170,14 @@ def get_compiler_includes(parseLogOptions, compiler, lang, compile_opts,
     # The first sysroot flag found among the compilation options is added
     # to the command below to give a more precise default include list.
     # Absence of any sysroot flags results in an empty string.
-    sysroot = next(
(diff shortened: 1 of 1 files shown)


libcodechecker/analyze/log_parser.py
```diff
@@ -170,8 +170,14 @@ def get_compiler_includes(parseLogOptions, compiler, lang, compile_opts,
     # The first sysroot flag found among the compilation options is added
     # to the command below to give a more precise default include list.
     # Absence of any sysroot flags results in an empty string.
-    sysroot = next(
-        (item for item in compile_opts if item.startswith("--sysroot=")), "")
+    sysroot = ""
```
dkrupp: Shouldn't we fix it in the https://github.com/Ericsson/codechecker/blob/master/libcodechecker/log/option_parser.py
instead? There is a code block which already handles the sysroot parameters and merging of them.

PR #1672, opened 2018-07-12
Ignore target dependent -mabi compiler option.

### LINKED ISSUE #1673: -mabi=spe results in unknown target ABI error
Depending on the target triple passed to clang, -mabi could result in unknown target ABI error. Since -mabi does not affect the analysis, CodeChecker might just ignore it.
Example:
--target=powerpc-montavista-linux-gnuspe -m32 -mcpu=8548 -mabi=spe
error: unknown target ABI 'spe'

lorincbalog:
Resolves #1673

### DIFF
--- a/libcodechecker/log/option_parser.py
+++ b/libcodechecker/log/option_parser.py
@@ -153,8 +153,9 @@ IGNORED_OPTION_MAP_REGEX = {
     '^-flto': 0,
     # MicroBlaze Options:
     '^-mxl': 0,
-    # PowerPC SPE Option:
-    '^-mfloat-gprs': 0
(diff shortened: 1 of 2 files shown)


lorincbalog:
#1673

whisperity:
At this point I think it would be beneficial to introduce a command-line option to Clang which turns off the *unknown argument* error. And CodeChecker could call Clang first with the normal argument list (and log out the unknown argument error), and if there is an error, call it with this extra flag, and it will just pass through.

gyorb:
yes, it would be nice to have some support for an improved argument handling from the Clang side until that we just increase our list

PR #1689, opened 2018-07-27
restructure python requirements files

gyorb:
GitHub can automatically detect vulnerabilities in
3pp dependecies if the requirements files follow
the naming conventions


whisperity:
Generally looks good. Can you test it on a dummy repository where you push these to master (or on your own fork, even... if dependency checks work for forks.)?

My only concern is, shouldn't these files go under the `vendor` directory, in any case? The naming of the top-level folder feels weird.

gyorb:
I've tested it on my fork and GitHub detects the files correctly. `requirements.txt` is kept usually top level but because we have multiple files for different purposes I though we should keep at least the directory top level.

PR #1716, opened 2018-08-22
update dojotoolkit link for download

gyorb:


whisperity:
Do they no longer work over HTTP? The link loads for me. HTTPS is problematic if the machine that does the builds has outdated root certificates and such.

gyorb:
I've tested it and it works, I've done some extra curl configuration changes in #1718 to handle redirects.

PR #1719, opened 2018-08-24
Query reports only when shown.

bruntib:
When opening a product, all reports are queried even if "All reports"
tab is not active. In case of >200000 reports this runs a huge query in
the background which occupies a whole thread for a relatively long time.
With this modification the reports are queried only when the grid is
shown.


bruntib:
This modification seems to be necessary in case of too many reports under a product (see the commit message). However, I think the greedy behavior in case of "All reports" was intentional, because the idea might have been that by the time the user clicks on "All reports" tab, the list will have been loaded.

PR #1722, opened 2018-08-30
Zombie processes remain on analysis interruption

bruntib:
When the analysis stops with Ctrl-C then the subprocesses in the process
pool are not terminated. The reason is that the shared data structures
managed by multiprocess.managers.SyncManager() type die before the
termination of processes. This causes an exception in pool.terminate()
which makes the subprocesses of the pool zombies.

### DIFF
--- a/libcodechecker/analyze/analyzer.py
+++ b/libcodechecker/analyze/analyzer.py
@@ -11,10 +11,11 @@ from __future__ import division
 from __future__ import absolute_import
 
 import copy
-from multiprocessing import Manager
+from multiprocessing.managers import SyncManager
 import os
 import shlex
 import shutil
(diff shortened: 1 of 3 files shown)


libcodechecker/analyze/analyzer.py
```diff
@@ -105,6 +106,10 @@ def __get_skip_handler(args):
         LOG.debug_analyzer('Skip file was not set in the command line')
 
 
+def __mgr_init():
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
```
gyorb: Please add a small documentation about ignoring the SIGINT signal here, so the manager will not stop before the pool.

PR #1723, opened 2018-08-30
calculate bug path length at store (schema change)

gyorb:
for performace improvement calculate bug path lengths
during the storage not in the report queries

### DIFF
--- a/db_migrate/versions/e89887e7d3f0_add_bug_path_length.py
+++ b/db_migrate/versions/e89887e7d3f0_add_bug_path_length.py
@@ -1,0 +1,33 @@
+"""Add bug path length
+
+Revision ID: e89887e7d3f0
+Revises: 3793e361a752
+Create Date: 2018-08-28 11:11:07.533906
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'e89887e7d3f0'
+down_revision = '3793e361a752'
+branch_labels = None
+depends_on = None
+
+from alembic import op
+import sqlalchemy as sa
+
+def upgrade():
+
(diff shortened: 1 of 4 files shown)


db_migrate/versions/e89887e7d3f0_add_bug_path_length.py
```diff
@@ -0,0 +1,33 @@
...
+branch_labels = None
+depends_on = None
+
+from alembic import op
+import sqlalchemy as sa
+
+def upgrade():
+
+    op.add_column('reports',
+                  sa.Column('path_length', sa.Integer(), nullable=True, default=0))
```
whisperity: I think we always come back to this at such changes, but SQLAlchemy still don't support views, right?

Nontheless, I am worried about allowing `nullable=True` here. The rest looks good.

libcodechecker/server/api/report_server.py
```diff
@@ -806,20 +806,12 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
...
-
             is_unique = report_filter is not None and report_filter.isUnique
             if is_unique:
                 sort_types, sort_type_map, order_type_map = \
                     get_sort_map(sort_types, True)
 
-                selects = [func.max(Report.id).label('id'),
-                           func.min(path_len_q.c.path_length)
-                               .label('bug_path_length')]
+                selects = [func.max(Report.id).label('id')]
```
whisperity: Some sorting code or something disappeared here. Will sorting by BPL still work, or was that always a clientside thing on the webapp?

PR #1744, opened 2018-09-16
Skip '-Xclang <opt>'

irishrover:
Essential for checking the Chromium project.
Without this patch `-Xclang -mllvm -Xclang -instcombine-lower-dbg-declare=0 -fcomplete-member-pointers`
becomes (for clangsa options parser) `-mllvm -fcomplete-member-pointers` which is incorrect.

### DIFF
--- a/libcodechecker/log/option_parser.py
+++ b/libcodechecker/log/option_parser.py
@@ -144,7 +144,9 @@ IGNORED_OPTION_MAP = {
     # '-Werror', '-pedantic-errors' the analysis with Clang can fail even
     # if the compilation passes with GCC.
     '-Werror': 0,
-    '-pedantic-errors': 0
+    '-pedantic-errors': 0,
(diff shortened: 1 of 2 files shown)


gyorb:
Hi, so you would like to filter out all the `-Xclang` flags from the compilation commands?
(There is a typo in the commit message: Chroimum ;)

irishrover:
> Hi, so you would like to filter out all the `-Xclang` flags from the compilation commands?
Correct.

> (There is a typo in the commit message: Chroimum ;)
Thanks 👍

irishrover:
Thanks, Gyorgy!

gyorb:
@irishrover could you attach some logs, which part and how it failed exactly?

gyorb:
We were thinking on a better solution to dynamically get the available compilation flags from clang and filter based on that instead of a hard coded regex list of arguments in the option parser #1747

PR #1756, opened 2018-10-04
Add missing requirement.

Xazax-hun:
In order to run make test we also need pycodestyle, so I think it would make sense to have it in the dev requirements.


Xazax-hun:
Added the version requirement.

PR #1770, opened 2018-10-12
handle character decoding problems (locale mismatch)

gyorb:
In case the decoding to unicode fails for some reason replace the
problematic characters.
The problem could be caused if the host locale settings do not match
the files encoding opened for read (text mode).
Python tries to decode the strings based on the host locale settings.

NOTE: this change could modify the hash generated for clang tidy in case there is some
character which can not be decoded (replaced with `?`) in the source line used for hash generation.


gyorb:
This should be fixed with the porting to python3.

PR #1782, opened 2018-10-24
Convert relative include paths to absolute

bruntib:
In pull request #1553 the relative include paths have been changed to absolute based on the working directory from the build json file. The intention was to accomplish as less modifications on the provided build command as possible. However, leaving relative paths in a plist interferes with the skipfile functionality:

project files:
```
/absolute/path/to/the/project/a/b/f.h
/absolute/path/to/the/project/main.cpp
```

good build command: `g++ -I/absolute/path/to/the/project/a main.cpp`
bad build command: `g++ -I./a main.cpp`

main.cpp:
```cpp
#include "b/f.h"
int main() { ... }
```

skipfile:
```
-/absolute/path/to/the/project/a*
```

This configuration doesn't skip the reports from subdirectory "a" with the bad build command.

### DIFF
--- a/libcodechecker/analyze/log_parser.py
+++ b/libcodechecker/analyze/log_parser.py
@@ -370,16 +370,16 @@ def parse_compile_commands_json(log_data, parseLogOptions):
 
         action.original_command = command
 
-        # If the original include directory could not be found
-        # in the filesystem, it is possible that it was provided
-        # relative to the working directory in the compile json.
+        # If the include directory is given as relative path then the working
+        # directory is prepended.
         compile_opts = results.compile_opts
         for i, opt in enumerate(compile_opts):
             if opt.startswith('-I'):
                 inc_dir = opt[2:].strip()
-                if not os.path.isdir(inc_dir):
+                if not os.path.isabs(inc_dir):
                     compile_opts[i] = '-I' + \
-                        os.path.join(entry['directory'], inc_dir)
+                        os.path.normpath(os.path.join(
+                            entry['directory'], inc_dir))
 
         action.analyzer_options = compile_opts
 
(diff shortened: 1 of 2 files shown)


Xazax-hun:
I usually prefer avoiding expanding relative oaths to absolute as much as possible to have fidelity with the original process and remain as portable as possible. I would expect the skipfiles to contain relative paths to be portable. Will we work in that scenario?

bruntib:
In a skipfile what should the paths be relative to? Because in a compilation database the build commands may vary the current working directory even for every single command, as the build system jumps between directories. Moreover the user shouldn't be aware whether a path is provided with a relative or absolute path after a `-I` flag.
The skipfile contains regexes and maybe that could handle this relative path case.

gyorb:
Will clang print the absolute file paths into the plist files? In this case if there is an error in `/absolute/path/to/the/project/a/b/f.h` and the compilation command used for analysis is `g++ -I./a main.cpp`?

bruntib:
Yes, I tried it, and this is the point of this pull request.

gyorb:
Ok, just to clear up the current clang behavior. If clang gets a relative include path the file path in the plist report file will be relative. If clang gets an absolute include path the file path in the plist will be absolute.

gyorb:
Could we use this clang argument `-fdiagnostics-absolute-paths`? Does `clang-tidy` print out absolute paths in the diagnostics?

bruntib:
Unfortunately this flag doesn't solve the problem. This flag refers only to the diagnostic messages (i.e. compilation errors, warnings, notes. etc.):

`./hello/f.h:3:13: warning: division by zero is undefined [-Wdivision-by-zero]`
vs.
`/absolute/path/hello/f.h:3:13: warning: division by zero is undefined [-Wdivision-by-zero]`

PR #1788, opened 2018-10-26
Constructing config handler is the analyzer classes' responsibility

Xazax-hun:
Another step towards https://github.com/Ericsson/codechecker/issues/1760

Also fixed a bug along the way. If clang-6.0 was available in the path but not clang, and clang-6.0 was not configured in the package config, we failed on the statistics feature check (not finding the binary).

It is hard to add a test for this scenario as there are lots of dependencies on the environment.

### DIFF
--- a/libcodechecker/analyze/analyzers/analyzer_clang_tidy.py
+++ b/libcodechecker/analyze/analyzers/analyzer_clang_tidy.py
@@ -14,12 +14,16 @@ import re
 import shlex
 import subprocess
 
+from libcodechecker.analyze import analyzer_env
+from libcodechecker.analyze import host_check
 from libcodechecker.analyze.analyzers import analyzer_base
-from libcodechecker.logger import get_logger
-from libcodechecker.util import get_binary_in_path
-from libcodechecker.analyze.analyzer_env import\
-    extend_analyzer_cmd_with_resource_dir
+from libcodechecker.analyze.analyzers import config_handler_clang_tidy
 from libcodechecker.analyze.analyzers import result_handler_clang_tidy
+from libcodechecker.analyze.analyzers.analyzer_clangsa import ClangSA
+from libcodechecker.analyze.analyzer_env import \
+    extend_analyzer_cmd_with_resource_dir
+from libcodechecker.logger import get_logger
+from libcodechecker.util import get_binary_in_path, replace_env_var
 
 LOG = get_logger('analyzer')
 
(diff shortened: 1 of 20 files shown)


libcodechecker/analyze/analyzers/analyzer_clang_tidy.py
```diff
@@ -14,12 +14,16 @@
...
+from libcodechecker.analyze import analyzer_env
+from libcodechecker.analyze import host_check
 from libcodechecker.analyze.analyzers import analyzer_base
-from libcodechecker.logger import get_logger
-from libcodechecker.util import get_binary_in_path
-from libcodechecker.analyze.analyzer_env import\
-    extend_analyzer_cmd_with_resource_dir
+from libcodechecker.analyze.analyzers import config_handler_clang_tidy
 from libcodechecker.analyze.analyzers import result_handler_clang_tidy
+from libcodechecker.analyze.analyzers.analyzer_clangsa import ClangSA
```
gyorb: Shouldn't we create a separate binary detection part for each analyzers so the tidy analyzer will not depend on the `analyzer_clangsa` modul?

libcodechecker/analyze/pre_analysis_manager.py
```diff
@@ -49,7 +48,7 @@ def collect_statistics(action, source, config, environ, statistics_data):
     LOG.debug_analyzer(cmd)
 
     ret_code, analyzer_out, analyzer_err = \
-        analyzer_base.SourceAnalyzer.run_proc(cmdstr)
+        analyzer_base.SourceAnalyzer.run_proc(cmdstr, env=environ)
```
gyorb: Maybe I just missed it but where is `environ` set to some value? It was an argument before.

libcodechecker/analyze/analyzers/analyzer_clang_tidy.py
Xazax-hun: We definitely should. It is a workaround below in this file, that we cannot get the resource dir from clang-tidy at the moment. As soon as we come up with a way to get the resource dir without looking at the clang binary, we can get rid of this dependency. There is a fixme for that in this file.

libcodechecker/analyze/pre_analysis_manager.py
Xazax-hun: It is still an argument. It was just unused before. `build_stat_coll_cmd` did not use it.

Xazax-hun:
ping :) I just resolved some merge conflicts, it would be great to have this in to avoid further rebasing

Xazax-hun:
ping

Xazax-hun:
I improved the commit messages and moved the fix into a separate commit. I would keep all off the commits as they are reflecting different individual steps of the refactoring.

Xazax-hun:
ping

PR #1789, opened 2018-10-27
Improve option_parser.py::arg_check to reduce number of regexps compiled

irishrover:
The change merges mulptiple regexps into a single one composed by '|'-ing them.
For Chromium (40k+ targets) 'arg_check' time has dropped from 657 to 227 seconds.


whisperity:
It should be verified if this join properly escapes inner regexes and doesn't break when new components are introduced later on. If any time the original map gets added a regex which contains a `|` and isn't parenthesised, it won't match properly.

irishrover:
> If any time the original map gets added a regex which contains a `|` and isn't parenthesised, it won't match properly.

Namely these regexps are used in OR semantics only, so any `|` inside are safe: "a | b | c" is the same as "a | (b | c)".

whisperity:
@irishrover Agreed, I gave it some more thought and it seems sound enough. Any disjunctive conjunction would need to expressed with parens within the "original" regex already.

Side question: what did you use to generate the waterfall diagram with? :grinning:

irishrover:
> Side question: what did you use to generate the waterfall diagram with?
cProfile + https://jiffyclub.github.io/snakeviz/
Really nice and simple tool by the way.

PR #1790, opened 2018-10-27
Don't parse command line options for skipped files.

irishrover:
It eliminates a lot of useless work for large projects like Chromium with 40K+ targets.
Also it seems to reduce an amount of RAM consumed.

### DIFF
--- a/libcodechecker/libhandlers/analyze.py
+++ b/libcodechecker/libhandlers/analyze.py
@@ -24,6 +24,7 @@ from libcodechecker.analyze import log_parser
 from libcodechecker.analyze.analyzers import analyzer_types
 from libcodechecker.util import RawDescriptionDefaultHelpFormatter, \
     load_json_or_empty
+from libcodechecker.analyze import skiplist_handler
 
 LOG = logger.get_logger('system')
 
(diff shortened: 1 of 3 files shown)


irishrover:
Thanks, Gábor!

It was always surprising me why CodeChecker first calls clang-tidy for Chromium project only on ~15th minute after start (my PC: core-i7 8700K, 16GB RAM, Samsung 970 SSD M.2). And now I know the answer - a lot of useless work while parsing options :-) Now the same takes ~1 minute and for "-j8" about 5Gb instead of 9Gb for "-j4" before.

irishrover:
>  Both of the users: log_parser.py and analyzer.py are in the analyze subfolder so I would prefer to keep the implementation somewhere there.

log_parser.py is not a user :-) the users are libcodechecker/analyze/analyzer.py  & libcodechecker/libhandlers/analyze.py with no common dirs.

Xazax-hun:
Do not worry about Codecy too much. We do not treat that as a hard requirement to merge a pull request. We usually check if a new functional error, redundant import etc is introduced but do not treat the complexity changes too seriously.

whisperity:
(Trust me, we'd be most happy if Codacy supported shutting their bot up on individual comments. :grimacing: )

libcodechecker/libhandlers/analyze.py
```diff
@@ -470,6 +471,37 @@ def __init__(self, args=None):
...
+            LOG.debug_analyzer("Creating skiplist handler.")
+            with open(args.skipfile) as skip_file:
+                return skiplist_handler.SkipListHandler(skip_file.read())
+    except AttributeError:
+        LOG.debug_analyzer('Skip file was not set in the command line')
+
+    return None
+
+
+def __remove_skipfile(args):
```
gyorb: Maybe it would be a nice addition to the refactoring to rename this function to something like `replace_skip_file` or `update_skip_file` at the output directory.
What we do here is we copy the user's skip file to the output directory so it will be sent to the storage server when the results are stored.

libcodechecker/libhandlers/analyze.py
```diff
@@ -510,6 +542,9 @@ def main(args):
     LOG.debug("args: " + str(args))
     LOG.debug("Output will be stored to: '" + args.output_path + "'")
 
+    # Generate skip list
```
gyorb: Maybe `'Process skip list'` would be a bit better comment here.

libcodechecker/libhandlers/analyze.py
irishrover: Done.

libcodechecker/libhandlers/analyze.py
irishrover: Done.

irishrover:
Hm, "continuous-integration/travis-ci/pr" says "Pending", but https://travis-ci.org/Ericsson/codechecker/builds/447884477?utm_source=github_status&utm_medium=notification shows everything is done.

irishrover:
Hi, @gyorb, is there anything here you want me to change?

irishrover:
Gentle ping :-)

gyorb:
@irishrover nope, thanks for the changes!

PR #1814, opened 2018-11-12
Option parser refactoring

bruntib:


libcodechecker/log/build_action.py
```diff
@@ -12,6 +12,11 @@
 
 
 class BuildAction(object):
+    LINK = 0
+    COMPILE = 1
+    PREPROCESS = 2
```
Xazax-hun: Can't we have build actions that do multiple things? I.e. both compile and link? Is it sufficient to have these constants? Shouldn't they rather be powers of two so we can have a set of purpose for each build action?

libcodechecker/libhandlers/analyze.py
```diff
@@ -569,9 +520,9 @@ def main(args):
                       "exist!")
             continue
 
-        parseLogOptions = ParseLogOptions(args)
-        actions += log_parser.parse_log(log_file, parseLogOptions,
-                                        skip_handler)
+        actions += log_parser.parse_log(
```
Xazax-hun: You no longer pass the skip_handler to the parser. What is the reason? We do not want to process the build actions for skipped files. Skippink this work was a huge performance gain for chromium, see https://github.com/Ericsson/codechecker/pull/1790 for details.

libcodechecker/log/option_parser.py
```diff
@@ -3,302 +3,194 @@
...
-
-Keys are the option name, value is the number of options to skip.
-
-Possible improvements:
-    - modular option handling system configuring possibility from config file.
-
-"""
-from __future__ import print_function
-from __future__ import division
-from __future__ import absolute_import
```
Xazax-hun: Why did we remove these imports?

libcodechecker/log/option_parser.py
```diff
@@ -308,218 +200,228 @@ def item(self):
...
 
-    result_map = OptionParserResult()
+    if m:
+        target.append(flag_iterator.item)
+        buildaction.analyzer_options.append(flag_iterator.item)
 
-    # The first element in the list is the compiler skip it from parsing.
-    for it in OptionIterator(shlex.split(args)[1:]):
-        arg_check(it, result_map)
+        if len(m.group(0)) == len(flag_iterator.item):
```
Xazax-hun: Could you add a comment what is this check actually doing?

libcodechecker/log/option_parser.py
```diff
@@ -308,218 +200,228 @@ def item(self):
...
+
+    buildaction.action_type = buildaction.COMPILE
+    buildaction.compiler = gcc_command[0]
+    if '++' in buildaction.compiler or 'cpp' in buildaction.compiler:
+        buildaction.lang = 'c++'
+
+    clang_command = [gcc_command[0]]
+
+    flag_transformers = [
+        __get_arch,
```
Xazax-hun: I think we should put transformers that are likely to be more common to the beginning and transformers that are likely to seldom fire to the end.

libcodechecker/log/option_parser.py
```diff
@@ -308,218 +200,228 @@ def item(self):
...
+        for flag_transformer in flag_transformers:
+            if flag_transformer(it, clang_command, buildaction):
+                break
+        else:
+            clang_command.append(it.item)
+            #print('Unhandled argument: ' + it.item)
+
+    # TODO: It is difficult to fetch source files from a compilation command if
+    # they contain a space character.
+    #for source_file in buildaction.sources:
```
Xazax-hun: Is this a regression? Or this was buggy before? I think if source files contains spaces they should be either escaped or in quotes. Shouldn't this already be handled by shsplit or some similar facility?

libcodechecker/log/option_parser.py
bruntib: This whole file has been removed.

libcodechecker/libhandlers/analyze.py
bruntib: Sorry, this is my fault, I'll fix it. However, it is strange, how could the tests pass without skip handler.

bruntib:
I don't think it is necessary to create two pull requests for the two commits, because you can comment on the modifications of a specific commit by selecting it.
It would have been reasonable to solve this refactoring in multiple commits for easing the review process. I was thinking of it, but it was hard to segregate small modifications while moving code fragments from one place to another. The main line of refactoring is that the `log_parser` and `option_parser` modules have been merged to a common file: `log/log_parser.py`. So the vast majority of the code is the same. The rest should preserve the earlier semantics.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,791 @@
...
+    """
+    # TODO: These dicts are mapping compiler to the corresponding information.
+    # It may not be enough to use the compiler as a key, because the implicit
+    # information depends on other data like language or target architecture.
+    compiler_includes = {}
+    compiler_target = {}
+    compiler_standard = {}
+
+    def __init__(self, compiler_info_file=None):
+        self.compiler_info_file = compiler_info_file
```
Xazax-hun: Do we still need the `compiler_info_file`? I see the corresponding option deleted from the command line.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,791 @@
...
+
+    def __dump_compiler_info(self, compiler, attr, data):
```
Xazax-hun: This function is not only dumping compiler info but update its value. The name suggest more like a logging function rather than something that does business logic. I think we should rename this.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,791 @@
...
+            if do_append:
+                # On OSX there are framework includes,
+                # where we need to strip the "(framework directory)" string.
+                # For instance:
+                # /System/Library/Frameworks (framework directory)
+                fpos = line.find("(framework directory)")
+                if fpos == -1:
+                    include_paths.append(line)
+                else:
+                    include_paths.append(line[0:fpos-1])
```
Xazax-hun: We can omit the `0` before the colon.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,791 @@
...
+                break
+        else:
+            pass
+            # print('Unhandled argument: ' + it.item)
+
+    def abs_path_maker(p):
+        return os.path.normpath(
+            os.path.join(compilation_db_entry['directory'], p))
+
+    merge_space_files(buildaction.sources, compilation_db_entry['file'])
```
Xazax-hun: Why do we need this? Shouldn't we see an escaped filename in the compilation command (that should be kept as one distinct file by shlex)?

libcodechecker/libhandlers/analyze.py
bruntib: I got it. Skip handler is passed to the analyzer manager a few lines below so it is not necessary to skip the actions at this point.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,791 @@
...
+    buildaction.sources = map(abs_path_maker, buildaction.sources)
+
+    compiled_file = abs_path_maker(compilation_db_entry['file'])
+    if compiled_file != '.' and \
+            compiled_file not in map(abs_path_maker, buildaction.sources):
```
Xazax-hun: Why do we convert the paths to absolute twice?
Also, how is it possible to not have `compiled_file` in the `buildaction.source`? Shouldn't this be a sanity check rather than part of the logic?

libcodechecker/libhandlers/analyze.py
Xazax-hun: The reason why we want to do the skipping earlier is because we do not even want to create the build action if we will not use it anyways. It does take significant amount of time to create the build action, so this is an optimization rather than a functional feature. See the pull request I linked for details. 

Quote from the patch author: "It was always surprising me why CodeChecker first calls clang-tidy for Chromium project only on ~15th minute after start (my PC: core-i7 8700K, 16GB RAM, Samsung 970 SSD M.2). And now I know the answer - a lot of useless work while parsing options :-) Now the same takes ~1 minute and for "-j8" about 5Gb instead of 9Gb for "-j4" before (I skip thousands of autogenerated files, tests etc)."

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,791 @@
...
+        buildaction.sources.append(compiled_file)
+
+    for source_file in buildaction.sources:
+        lang = get_language(os.path.splitext(source_file)[1])
+        if lang:
+            if buildaction.lang is None:
+                buildaction.lang = lang
```
Xazax-hun: If there are multiple files the language can be different for each file!
Do we have a separate entry in the compilation database for each file in this case? If so, we should rather use the compilation_db_entry['file'] to determine the language rather than the list of source. Moreover, if this is the case, we should not have a list of sources per build action but rather only have one single source.

libcodechecker/log/option_parser.py
bruntib: True. I keep `print_function` and `absolute_import`. The `division` is not necessary.

libcodechecker/libhandlers/analyze.py
Xazax-hun: Just to summarize, I think we do need the skip handler here to skip source files early and avoid expensive command line parsing, but we do need skip handler later on as well, to be able to skip storing reports from headers that are included from non-skipped cpp files.

bruntib: These are not good news. The goal would be to provide this log parser module as a 3rd party vendor project so it can be used as a standalone tool too. So in the longer terms all libcodechecker dependencies should be eliminated. I'll figure out something...

libcodechecker/log/option_parser.py
bruntib: This commented code fragment is outdated by now. In the new version this is more intuitive.

libcodechecker/log/log_parser.py
bruntib: The decision here is that we collect the implicit compiler information to this file for debugging reasons for ourselves in the report directory. However, this can't be an input of the analysis like before this commit lacking any reasonable use-case.

libcodechecker/log/log_parser.py
bruntib: Based on our discussion this should be solved in another issue/pull request.

libcodechecker/log/build_action.py
Xazax-hun: We could make it an optional dependency. If we use the tool from CodeChecker using the API we could have skip_handler as an optional parameter, but it would not be exposed when used from command line.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,791 @@
...
+        """
+        Returns a list of default includes of the given compiler.
+
+        action -- A BuildAction object containing the details of compilation.
+        """
+        # The first sysroot flag found among the compilation options is added
+        # to the command below to give a more precise default include list.
+        # Absence of any sysroot flags results in an empty string.
+        sysroot = ""
+        for i, item in enumerate(action.analyzer_options):
```
Xazax-hun: Should we check how the compiler works if multiple sysroot is passed? Which one takes precedence.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,791 @@
...
+def __skip(flag_iterator, buildaction):
+    """
+    This function skips the flag pointed by the given flag_iterator with its
+    parameters if any.
+    """
+    if IGNORED_OPTIONS.match(flag_iterator.item):
+        return True
+
+    for pattern, arg_num in IGNORED_PARAM_OPTIONS.items():
+        if re.match(pattern, flag_iterator.item):
```
Xazax-hun: We would compiler the pattern for the regular expression here in the hot loop. Maybe we should first compile each key in the map once when this module is loaded?

libcodechecker/log/log_parser.py
bruntib: It was really a mistake to convert these paths to absolute twice.
The use-case of this check was that in case of source files containing space in their names, it was impossible to determine whether those are two separate arguments or a filename with space in it. We also have a test case on this. Nonetheless, I'll check if it is still relevant.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,791 @@
...
+    This function determines whether this is a preprocessing, compilation or
+    linking action and sets it in the buildaction object.
+    """
+    if flag_iterator.item == '-c':
+        buildaction.action_type = BuildAction.COMPILE
+        return True
+    elif flag_iterator.item.startswith('-print-prog-name'):
+        buildaction.action_type = BuildAction.INFO
+        return True
+    elif re.match('-(E|M[T|Q|F|J|P|V|M]*)$', flag_iterator.item):
```
Xazax-hun: This is also hot code. I think we should compile this regular expression once when the module is loaded and use the compiled object to do the matching here.

libcodechecker/libhandlers/analyze.py
Xazax-hun: We could make it an optional dependency. If we use the tool from CodeChecker using the API we could have skip_handler as an optional parameter, but it would not be exposed when used from command line.

libcodechecker/log/option_parser.py
bruntib: This is an outdated modification too. But the issue is relevant, this will go to a new ticket about handling spaces in file names by the logger.
(Side note: what about reverse compatibility with older log files? This extra logic seems to be necessary for the first view, but we'll see later.)

Xazax-hun: I think we should not care about reverse compatibility in the case. In CI loops the users should always regenerate the build log since files might have been removed or added in commits. I cannot really see a usecase where the users are not generating new compilation database unless someone is running a buildbot for codechecker itself.

bruntib: I don't know... In `tests/unit/logparser_test_files` you can see what kind of old log file formats are supported. I think the idea of CodeChecker is that it is an infrastructure of multiple tools. It supports several analyzers and it seems that is supports several logger tools as well, not just the delivered one.

libcodechecker/log/log_parser.py
bruntib: The last one will take precedence, so I'll fix it.

bruntib:
I've fixed the code based on the comments. I created a new commit for the elimination of the multiple source files from the `BuildAction` object.

Xazax-hun:
Great, thanks! Just a sanity check, for a build command like `g++ a.cpp b.c`, intercept build will create two separate entries in the compilation database right?

bruntib:
Yes, intercept build (at least its latest version) creates two build actions in the compilation database.

libcodechecker/analyze/analysis_manager.py
```diff
@@ -565,19 +565,17 @@ def check(check_data):
...
-                LOG.debug_analyzer(source_file_name + ' is skipped')
-                skipped = True
-                continue
-
-            source = util.escape_source_path(source)
+        if skip_handler and skip_handler.should_skip(action.source):
+            LOG.debug_analyzer(source_file_name + ' is skipped')
+            skipped = True
+        else:
+            source = util.escape_source_path(action.source)
```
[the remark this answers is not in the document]
bruntib: What fix are you thinking of? By the way I don't think it matters how our logger works. If there are some old versioned intercept-build logger or any other logger tool that we'd like to support then handling these space characters is necessary.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,752 @@
...
+        __get_language,
+        __get_output]
+
+    for it in OptionIterator(gcc_command[1:]):
+        for flag_transformer in flag_transformers:
+            if flag_transformer(it, buildaction):
+                break
+        else:
+            pass
+            # print('Unhandled argument: ' + it.item)
```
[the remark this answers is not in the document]
bruntib: There may be GCC flags which are not handled in this option parser. Their existence is not an error and may not even be necessary for the analysis (e.g. some architecture dependent flag). Actually these unhandled arguments were printed to the debug logs earlier, but some users interpreted them as an erroneous compilation command messages. So this logging has been removed intentionally. This commented print has just remained in the source code which of course would be nice to remove.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,752 @@
...
+        implicit_compiler_info = ImplicitCompilerInfo()
+
+        for entry in compilation_database:
+            action = parse_options(entry)
+
+            if not action.lang:
+                continue
+            if action.action_type != build_action.BuildAction.COMPILE:
+                continue
+            if skip_handler and skip_handler.should_skip(action.source):
```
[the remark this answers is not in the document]
bruntib: I don't understand this comment. This part in log parser happens before the analysis. This is the part which filters the compilation database by removing actions to skip.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,752 @@
...
+
+            toolchain = \
+                gcc_toolchain.toolchain_in_args(action.analyzer_options)
+
+            # Store the compiler built in include paths and defines.
+            if not toolchain:
+                implicit_compiler_info.set_implicit_compiler_info(action)
+
+            # Filter out duplicate compilation commands.
+            if action.cmp_key not in filtered_build_actions:
```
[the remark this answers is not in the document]
bruntib: We do use the keys to check if it is contained by the map. And we use the values later to produce the filtered list of build actions. By the way map is also a hash data structure in Python.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,752 @@
...
+            LOG.error('The compile database is empty.')
+        else:
+            LOG.error('The compile database is not valid.')
+        LOG.debug(traceback.format_exc())
+        LOG.debug(ex)
+        sys.exit(1)
+
+    LOG.debug('Parsing log file done.')
+
+    return actions
```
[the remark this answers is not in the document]
bruntib: No. Not to mention that this variable doesn't even exist :) Thanks, I'll fix it.

libcodechecker/log/log_parser.py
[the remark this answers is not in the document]
Xazax-hun: We do not only log errors. It is a debugging aid.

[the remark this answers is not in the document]
Xazax-hun: We do not only log errors. It is a debugging aid.

libcodechecker/log/log_parser.py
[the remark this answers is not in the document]
Xazax-hun: Calling `parse_options` can be a lot ot work. Why would we call it above just to throw away the result if we could skip the whole operation instead?

libcodechecker/log/log_parser.py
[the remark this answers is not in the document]
Xazax-hun: If we use the keys only to filter duplicates than we could actually use a set which does exactly the same and is a cleaner way to express the intent.

tests/functional/analyze/test_analyze.py
```diff
@@ -155,100 +154,6 @@ def test_compiler_info_files(self):
                 self.fail("json.load should successfully parse the file %s"
                           % info_File)
 
-    def test_compiler_includes_file_is_loaded(self):
```
gyorb: why was this test removed?

libcodechecker/log/build_action.py
```diff
@@ -12,64 +12,46 @@
...
 class BuildAction(object):
-    def __init__(self, build_action_id=0):
-        self._id = build_action_id
+    """
+    The objects of this class hold information which is the input of the
+    analyzer engines.
+
+    !!!WARNING!!!
+    This class defines a mutable type, however, it has a __hash__() method. It
+    is against the "hashable" concept, since it leads to undefined behavior.
```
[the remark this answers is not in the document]
Xazax-hun: I think the original construct was just as dangerous. We stored the hashes in a map, but we could change the corresponding value (the build action object) without updating the hash.

tests/functional/analyze/test_analyze.py
bruntib: Because we eliminated the `--compiler-includes-file` flag.

Xazax-hun:
Could you rebase this on top of master? There are some merge conflicts.

libcodechecker/log/build_action.py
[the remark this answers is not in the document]
gyorb: @bruntib please make this class immutable.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,748 @@
...
+            # Skip GCC's fixinclude dir
+            if os.path.basename(
+                    os.path.normpath(include_dir)) == "include-fixed":
+                continue
+            if contains_intrinsic_headers(include_dir):
+                continue
+            result.append(include_dir)
+        return result
+
+    def get_compiler_includes(self, action):
```
gyorb: compiler, lang and analyzer options should be enough as an argument we do not use any other information from the buildaction.

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,748 @@
...
+        extra_opts = filter(pattern.match, action.analyzer_options)
+        cmd = action.compiler + " " + ' '.join(extra_opts) \
+            + " -E -x " + action.lang + " " + sysroot + " - -v "
+
+        include_dirs = self.__filter_compiler_includes(
+            self.__parse_compiler_includes(self.__get_compiler_err(cmd)))
+
+        return ["-isystem " + os.path.normpath(idir) for idir in include_dirs]
+
+    def get_compiler_target(self, action):
```
gyorb: giving the compiler as an argument should be enough here

libcodechecker/log/log_parser.py
```diff
@@ -0,0 +1,748 @@
...
+        target = ""
+
+        for line in lines.splitlines(True):
+            line = line.strip().split()
+            if len(line) > 1 and line[0] == target_label:
+                target = line[1]
+
+        return target
+
+    def get_compiler_standard(self, action):
```
gyorb: can we change this not to use the whole buildaction as an argument, just what is needed from it?

tests/unit/test_log_parser.py
```diff
@@ -52,11 +51,9 @@ def test_old_ldlogger(self):
         # option_parser, so here we aim for a non-failing stalemate of the
         # define being considered a file and ignored, for now.
 
-        build_action = log_parser.parse_log(logfile, ParseLogOptions())[0]
-        results = option_parser.parse_options(build_action.original_command)
+        build_action = log_parser.parse_log(load_json_or_empty(logfile))[0]
 
-        self.assertEqual(' '.join(results.files),
-                         r'"-DVARIABLE="some value"" /tmp/a.cpp')
+        self.assertEqual(build_action.source, r'/tmp/a.cpp')
```
gyorb: Do we still have a test for compiler flags like this `"-DVARIABLE="some value"" /tmp/a.cpp`

libcodechecker/analyze/analysis_manager.py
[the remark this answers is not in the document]
Xazax-hun: My suspicion is that other loggers does the escaping for us, but of course we need to check that. This should be addressed separately though.

libcodechecker/log/build_action.py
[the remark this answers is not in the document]
Xazax-hun: https://stackoverflow.com/questions/4828080/how-to-make-an-immutable-object-in-python

This has some interesting ideas, like the `slots` version.

libcodechecker/log/log_parser.py
Xazax-hun: I think this is for getting the default include paths from the compiler as it might be different from clang's default include paths. These include paths might not appear in the compilation command, we might still want to add them.

tests/unit/test_log_parser.py
Xazax-hun: Note that, if we want to support windows in the future we might want to use a Python standard library function to get a temporary file rather than hardcoding a name. This would also reduce the chance of colliding file names (even when multiple test instances are running in parallel on the same machine).

libcodechecker/log/build_action.py
```diff
@@ -12,64 +12,57 @@
...
 
 class BuildAction(object):
-    def __init__(self, build_action_id=0):
-        self._id = build_action_id
+    """
+    The objects of this class hold information which is the input of the
+    analyzer engines.
+
+    !!!WARNING!!!
+    This class defines a mutable type, however, it has a __hash__() method. It
```
Xazax-hun: Is this comment still relevant?

Xazax-hun:
Ping.

PR #1821, opened 2018-11-21
ugrade boost before thrift install

gyorb:
If the boost package is installed by the thrift package (dependency)
it will be build from source which takes to much time on travis without
any output so the travis jobs fail.


Xazax-hun:
While I agree that this is a good workaround, but I wonder if there is a way to avoid building packages to improve the speed of the CI.

gyorb:
As far as I know boost and thrift are not part of the default [OSX environment](https://docs.travis-ci.com/user/reference/osx). I do not know any other options what we could use install OSX packages except [Homebrew](https://brew.sh/). The [boost formulae](https://github.com/Homebrew/homebrew-core/blob/master/Formula/boost.rb) builds it from source. I'm open to a faster or better solution if there is any.

PR #1822, opened 2018-11-23
fix slow boost build in travis

gyorb:
boost build is really slow and times out with xcode8
upgrading to a newer xcode fixes the problem

the previous try to split the boost build/upgrade
into a separate step was not enough #1821

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,17 +22,17 @@ matrix:
       python: "2.7"
       env: DATABASE=psql_pg8000
     - os: osx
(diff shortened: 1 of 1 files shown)


Xazax-hun:
In case this is still slow we could experiment with some options seen here: https://formulae.brew.sh/formula/boost

Maybe `--without-single` or `--without-static` would decrease the compile time?

Xazax-hun:
`--c++11` flag could also influence the build time.

PR #1843, opened 2019-01-07
Add bugprone-parent-virtual-call check for clang-tidy v7+

irishrover:
See https://clang.llvm.org/extra/clang-tidy/checks/bugprone-parent-virtual-call.html


Xazax-hun:
I am not sure if it should have high priority as such code might be intentional. Of course, this is only my personal opinion, the patch looks good!

irishrover:
> Should we add this to the default checkers profile? Is it stable enough and does not produce many false positives?

Yes, we can start from `default` and see. Is it OK for you?

irishrover:
Switched to `sensitive`+`extreme` and `HIGH`->`MEDIUM`

PR #1844, opened 2019-01-08
Fix --timeout as clang-sa spawned child processes are not killed now

irishrover:
Clang-sa spawns a child process which in turn spawns another child process.
Using os.kill kills only the parent letting child clang-sa to continue.
It's not what we expect from timeout.
Using killpg kills parent, its children and their children too.

### DIFF
--- a/libcodechecker/analyze/analysis_manager.py
+++ b/libcodechecker/analyze/analysis_manager.py
@@ -604,7 +604,7 @@ def check(check_data):
                     called. Set up a timeout for the analysis.
                     """
                     timeout_cleanup[0] = util.setup_process_timeout(
-                        analyzer_process, analysis_timeout, signal.SIGKILL)
+                        analyzer_process, analysis_timeout)
             else:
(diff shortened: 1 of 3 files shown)


irishrover:
For my case htop looks like:

default

Xazax-hun:
Thanks for the fix! This is a really nice catch!

One problem I see is that os.killpg is only available on unix. While we do not really support Windows at the moment, I do not like the idea of adding Unix only features to make porting harder in the future.

I think the some could be achieved using psutil which is already a dependency we have. Could you rewrite the patch using psutil?

It could be something like:
```
parent = psutil.Process(parent_pid)
for child in parent.children(recursive=True):  # or parent.children() for recursive=False
    child.kill()
parent.kill()
```

I do agree it is more verbose, but it should be portable.

Thanks in advance!

irishrover:
Well, 

> Simpsons Already Did It

I mean we have `kill_process_tree` few lines later in the same `util.py`.

PR #1864, opened 2019-01-22
Handling ccache

bruntib:


bruntib:
WIP, because #1814 should be accepted first.

libcodechecker/log/log_parser.py
```diff
@@ -206,6 +206,21 @@ class ImplicitCompilerInfo(object):
     compiler_target = {}
     compiler_standard = {}
 
+    compiler_isexecutable = {}
+
+    def is_executable_compiler(self, compiler):
```
gyorb: A more general version of this `is_executable` could be made, which could be used at other places too. I think where we detect `scan-build` on osx something similar was done.

libcodechecker/log/log_parser.py
```diff
@@ -206,6 +206,21 @@ class ImplicitCompilerInfo(object):
...
+        if compiler not in self.compiler_isexecutable:
+            proc = subprocess.Popen(['which', compiler],
```
gyorb: This is quite linux specific there are other methods to [find an executable](https://stackoverflow.com/questions/377017/test-if-executable-exists-in-python) please make a note for a python3 solution. Later we can upgrade to it. Using `distutils.spawn.find_executable` might be better in this case.

tests/unit/test_option_parser.py
```diff
@@ -258,3 +258,21 @@ def test_compiler_toolchain(self):
...
+
+    def test_ccache_compiler(self):
+        action = {
+            'file': 'main.cpp',
+            'command': 'ccache main.cpp',
+            'directory': ''}
+
+        action['command'] = 'ccache g++ main.cpp'
+        res = log_parser.parse_options(action)
+        self.assertEqual(res.original_command, 'g++ main.cpp')
```
gyorb: Do we want his behavior where the original command is modified? There is some information loss.

PR #1865, opened 2019-01-22
Add __clang_analayzer__ macro to CTU imported tus

dkrupp:
When running clang analysis, the __clang_analyzer__ macro
needs to be defined. This should also be true for CTU imported TUs.

### DIFF
--- a/libcodechecker/analyze/ctu_manager.py
+++ b/libcodechecker/analyze/ctu_manager.py
@@ -106,7 +106,8 @@ def generate_ast(triple_arch, action, source, config, env):
             pass
 
     cmd = ctu_triple_arch.get_compile_command(action, config, source)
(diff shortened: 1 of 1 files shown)


libcodechecker/analyze/ctu_manager.py
```diff
@@ -106,7 +106,7 @@ def generate_ast(triple_arch, action, source, config, env):
             pass
 
     cmd = ctu_triple_arch.get_compile_command(action, config, source)
-    cmd.extend(['-emit-ast', '-w', '-o', ast_path])
+    cmd.extend(['-emit-ast', '-D__clang_analyzer__', '-w', '-o', ast_path])
```
gyorb: Please add a short comment in the source code too why this extra flag is required.

gyorb:
please wait for the osx tests to finish before merge

PR #1867, opened 2019-01-22
Default standard in first CTU phase.

### LINKED ISSUE #1838: [CTU] Compiler default C++ dialect is not added when generating the PCH files
**Describe the bug**
Compiler default C++ dialect is added to the analyzer command, but it is not added to the compiler invocation which generates the PCH ast files.
This causes C++ language dialect mismatches and consequently false positive ODR errors during CTU analysis.

**CodeChecker version**
All

**To Reproduce**
Steps to reproduce the behavior
0. Apply this patch:
```
diff --git a/lib/CrossTU/CrossTranslationUnit.cpp b/lib/CrossTU/CrossTranslationUnit.cpp
index 4423fed..d15cc85 100644
--- a/lib/CrossTU/CrossTranslationUnit.cpp
+++ b/lib/CrossTU/CrossTranslationUnit.cpp
@@ -342,6 +342,7 @@ CrossTranslationUnitContext::importDefinition(const FunctionDecl *FD) {
                     [&](const ImportError &IE) {
                       switch (IE.Error) {
                       case ImportError::NameConflict:
+                        llvm_unreachable("NameConflict");
                         ++NumNameConflicts;
                         break;
                       case ImportError::UnsupportedConstruct:
```
1. Ctu analyse bitcoin with gcc5 on Ubuntu 16.04 with boost 1.65
    This will create a PCH file for univalue.cpp with the default C++ dialect, which is C++14 in case of Clang7.
2. Analyse the univalue_get.cpp file (which has gnu++98 set). This loads `univalue.cpp.ast` and there we have the mismatch (and unreachable fires).

bruntib:
CTU analyzes in two phases. The first phase where the AST dumps are created
should happen with the same C/C++ standard version which may be implicit.
Fixes #1838

### DIFF
--- a/libcodechecker/analyze/analyzers/ctu_triple_arch.py
+++ b/libcodechecker/analyze/analyzers/ctu_triple_arch.py
@@ -35,6 +35,10 @@ def get_compile_command(action, config, source='', output=''):
         cmd.extend(['-o', output])
     if source:
         cmd.append(source)
+
+    if all(not opt.startswith('-std=') for opt in action.analyzer_options):
+        cmd.append(action.compiler_standard)
+
     return cmd
 
 


bruntib:
Yes, it will, it is the point of the patch. If there is no `--std=` flag in the build command then the `action.compiler_standard` contains the default standard of the compiler.

gyorb:
Do we always set `action.compiler_standard`?

bruntib:
Yes, it is always set unless there is a GCC toolchain, in which case no implicit information is added (implicit include paths, target architecture, default standard dialect).

PR #1886, opened 2019-01-25
log optimization

gyorb:
Most of the logging part is updated but not everywhere.
See https://docs.python.org/3/howto/logging.html#optimization


libcodechecker/libclient/client.py
```diff
@@ -147,8 +147,7 @@ def handle_auth(protocol, host, port, username, login=False):
             pwd = getpass.getpass("Please provide password for user '{0}': "
                                   .format(username))
 
-        LOG.debug("Trying to login as {0} to {1}:{2}"
-                  .format(username, host, port))
+        LOG.debug("Trying to login as %s to %s:%s", username, host, port)
```
bruntib: I think port is an integer if you follow it back.

libcodechecker/libhandlers/server.py
```diff
@@ -905,8 +904,8 @@ def server_init_start(args):
     except socket.error as err:
         if err.errno == errno.EADDRINUSE:
             LOG.error("Server can't be started, maybe the given port number "
-                      "({}) is already used. Check the connection "
-                      "parameters.".format(args.view_port))
+                      "(%s) is already used. Check the connection "
+                      "parameters.", args.view_port)
```
bruntib: I'm not sure, but this might also be integer.

libcodechecker/source_code_comment_handler.py
```diff
@@ -185,9 +185,8 @@ def get_source_line_comments(self, bug_line):
         for a bug line.
         """
         source_file = self.__source_file
-        LOG.debug("Checking for source code comments in the source file '{0}'"
-                  "at line {1}".format(self.__source_file,
-                                       bug_line))
+        LOG.debug("Checking for source code comments in the source file '%s'"
+                  "at line %s", self.__source_file, bug_line)
```
bruntib: As far as I've seen, this `bug_line` really is a string, but I wouldn't have counted on that.

libcodechecker/util.py
```diff
@@ -444,8 +444,8 @@ def split_product_url(url):
     except Exception:
         raise ValueError("The specified product URL is invalid.")
 
-    LOG.debug("Result: With '{0}' on server '{1}:{2}', product '{3}'"
-              .format(protocol, host, port, product_name))
+    LOG.debug("Result: With '%s' on server '%s:%s', product '%s'",
+              protocol, host, port, product_name)
```
bruntib: Port number is an integer.

PR #1888, opened 2019-01-28
Fix escaping in command line arguments

bruntib:
In this commit we follow the rule of BuildAction objects: these should contain
the final form of every used value. In this case the analyzer options should be
escaped so it will be passed correctly to the analyzer.

### DIFF
--- a/libcodechecker/log/log_parser.py
+++ b/libcodechecker/log/log_parser.py
@@ -651,6 +651,11 @@ def parse_options(compilation_db_entry):
         details['original_command'] = ' '.join(gcc_command)
     elif 'command' in compilation_db_entry:
         details['original_command'] = compilation_db_entry['command']
+        # This is needed so shlex.split() leaves the quotation mark in the
+        # output list:
+        # gcc -DHELLO="hello world" main.cpp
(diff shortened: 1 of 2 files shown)


Xazax-hun:
What does the other loggers (scan-build-py, bear, cmake compilation database export) do? Because if our logger is the only one that is not doing any escaping, maybe we should fix our logger instead. 

Also, will it work with escaping loggers? Won't we add extra slashes to already escaped strings?

gyorb:
scan-build can produce a newer compile command json format which is better in handling the spaces in the arguments. On the long term it would be better if our logger would produce the same new format. Until that we should keep the current behavior which was broken in #1814 this a fix for that.

bruntib:
@gyorb Do you think of this format? https://github.com/Ericsson/codechecker/blob/5bab5177fba7b9e19a1e5a780187a5f980490089/tests/unit/logparser_test_files/intercept-new.json
Personally I don't like it. It's much more verbose, results bigger log files, and completely unnecessary as `shlex.split()` produces this format. Moreover LLVM can consume our format directly.

PR #1892, opened 2019-01-28
Make ImplicitCompilerInfo's methods static.

rnkovacs:
All of `ImplicitCompilerInfo`'s members are class attributes, and the whole class 
is used in a static way. Therefore it makes sense to make its methods static. Also,
shorter getter and setter method names avoid some redundancy.

### DIFF
--- a/libcodechecker/log/log_parser.py
+++ b/libcodechecker/log/log_parser.py
@@ -206,17 +206,18 @@ class ImplicitCompilerInfo(object):
     compiler_includes = {}
     compiler_target = {}
     compiler_standard = {}
-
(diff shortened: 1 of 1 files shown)


whisperity:
Instead of staticmethod, you could use classmethod, and then you'll get `cls` as first argument (instead of `self`) and it behaves the same way for this case. Would omit the ugly `ICI = ...` lines, and also offers better syntax highlight in IDEs.

Xazax-hun:
While having a class method would make it one line shorter, I think that is unnecessary complexity. Such code suggests that the method could be called with other classes and that is not the case.

PR #1900, opened 2019-02-04
Upload compiler_info.json

bruntib:
Earlier the implicit compiler information went to compiler_target.json
and compiler_includes.json. Now there is only one file:
compiler_info.json. This has to be uploaded with fail zips too.


libcodechecker/libhandlers/store.py
```diff
@@ -336,6 +336,7 @@ def get_analysis_statistics(inputs, limits):
                     statistics_files.append(compilation_db)
             elif inp_f in ['compiler_includes.json',
```
Xazax-hun: Do we still have these files? Shouldn't we remove these?

PR #1906, opened 2019-02-05
Resurrect --compiler-info-file analyze flag

rnkovacs:
This option was removed by 5519409 under the assumption that it is not
used by the users. The truth is that this flag is very useful for debugging
purposes and some debug tools in the `/scripts` directory do use it.
This commit re-introduces the `--compiler-info-file` option on top of the
refactored log parser.

### DIFF
--- a/libcodechecker/libhandlers/store.py
+++ b/libcodechecker/libhandlers/store.py
@@ -301,14 +301,27 @@ def assemble_zip(inputs, zip_file, client):
             map(lambda f_: " - " + f_, missing_source_files)))
 
 
+def should_be_zipped(input_file, input_files):
+    """
+    Determine whether a given input file should be included in the zip.
+    Compiler includes and target files should only be included if there is
+    no compiler info file present.
+    """
+    return (input_file in ['metadata.json', 'compiler_info.json']
+            or (input_file in ['compiler_includes.json',
+                               'compiler_target.json']
+                and 'compiler_info.json' not in input_files))
+
+
 def get_analysis_statistics(inputs, limits):
     """
     Collects analysis statistics information and returns them.
 
     This function will return the path of failed zips and the following files:
       - compile_cmd.json
-      - compiler_includes.json
-      - compiler_target.json
+      - either
+        - compiler_info.json, or
+        - compiler_includes.json and compiler_target.json
       - metadata.json
 
     If the input directory doesn't contain any failed zip this function will
@@ -342,10 +355,7 @@ def get_analysis_statistics(inputs, limits):
                     LOG.debug("Copying file '%s' to analyzer statistics "
                               "ZIP...", compilation_db)
                     statistics_files.append(compilation_db)
-            elif inp_f in ['compiler_includes.json',
-                           'compiler_target.json',
-                           'compiler_info.json',
-                           'metadata.json']:
+            elif should_be_zipped(inp_f, files):
                 analyzer_file = os.path.join(input_path, inp_f)
                 statistics_files.append(analyzer_file)
         for inp_dir in dirs:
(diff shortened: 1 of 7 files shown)


gyorb:
This feature should be merged after the release.

rnkovacs:
Rebased and fixed unit tests.

libcodechecker/libhandlers/store.py
```diff
@@ -342,9 +341,7 @@ def get_analysis_statistics(inputs, limits):
                     LOG.debug("Copying file '%s' to analyzer statistics "
                               "ZIP...", compilation_db)
                     statistics_files.append(compilation_db)
-            elif inp_f in ['compiler_includes.json',
-                           'compiler_target.json',
-                           'compiler_info.json',
+            elif inp_f in ['compiler_info.json',
```
gyorb: Maybe we should keep the files. If an older version was used to analyze the project, but a newer to store those files will be left out from the zip. @martong @bruntib can we remove those files from the zip?

libcodechecker/log/log_parser.py
```diff
@@ -435,21 +436,39 @@ def get_compiler_standard(compiler, language):
         return standard
 
     @staticmethod
-    def set(details):
+    def load_compiler_info(filename, compiler):
+        contents = load_json_or_empty(filename, {})
+        compiler_info = contents.get(compiler)
+        if compiler_info is None:
+            LOG.error("Could not find compiler %s in file %s" %
```
gyorb: pylint did not warn about this log format? Did you rebase to the latest master?
python can do some logging optimization if the right format is used like this:
`LOG.error("Could not find compiler %s in file %s", compiler, filename)`

libcodechecker/log/log_parser.py
```diff
@@ -435,21 +436,39 @@ def get_compiler_standard(compiler, language):
         return standard
 
     @staticmethod
-    def set(details):
+    def load_compiler_info(filename, compiler):
```
gyorb: Please review #1908 there is an extension which will get the compiler info for both languages which should change the `compiler_info.json` loading introduced here if I'm right.

libcodechecker/libhandlers/store.py
martong: We should have either the new `compiler_info.json` or the old pair of includes and target json in the failed zip. In both cases we can reproduce the failure. If the info file is missing and there are only an include and target file then put those into the zip.

rnkovacs: Created a small helper function for this.

libcodechecker/log/log_parser.py
rnkovacs: Fixed, thanks.

libcodechecker/log/log_parser.py
rnkovacs: OK, I'll rebase on top of it and add the logic when it lands.

rnkovacs: I'll also need to update the recently added `renew_info_files.py` debug script.

rnkovacs: And the other debug scripts in #1941.

martong:
Ping

gyorb:
Please rebase the changes, there was a lot of file movement. The repository is split up to two main parts analyzer and web. I think most of the changes are in the analyzer part but the store is affected too which is in the web part. There is codechecker_common for shared code but i would recommend to minimize it and only share it if it really necessary.
Thanks

dkrupp:
I rebased this patch to comply with the restructured codechecker
https://github.com/Ericsson/codechecker/pull/2039

So we can abandon this PR.

PR #1908, opened 2019-02-05
Collect compiler info for both languages C/C++

gyorb:
Collect compiler includes and standards for both
language (C/C++) independent from the actual compilation
command and select the right values at analyzer command
build time.

It is possible that in a compile command json the same
compiler binary is used to build C and C++ source.
In this case the include paths and the standard will be
detected for the first command but it will be used for the
command too which is wrong.

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -197,7 +197,7 @@ class ImplicitCompilerInfo(object):
     """
     # TODO: These dicts are mapping compiler to the corresponding information.
     # It may not be enough to use the compiler as a key, because the implicit
-    # information depends on other data like language or target architecture.
+    # information depends on other data like target architecture.
     compiler_includes = {}
     compiler_target = {}
     compiler_standard = {}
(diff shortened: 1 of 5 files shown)


libcodechecker/log/log_parser.py
```diff
@@ -692,8 +719,8 @@ def parse_options(compilation_db_entry):
 
     details = {
         'analyzer_options': [],
-        'compiler_includes': [],
-        'compiler_standard': '',
+        'compiler_includes': {},  # For each language C/C++.
+        'compiler_standard': {},  # For each anguage C/C++.
```
bruntib: Small typo.

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -694,8 +721,8 @@ def parse_options(compilation_db_entry):
 
     details = {
         'analyzer_options': [],
-        'compiler_includes': [],
-        'compiler_standard': '',
+        'compiler_includes': {},  # For each language C/C++.
+        'compiler_standard': {},  # For each anguage C/C++.
```
bruntib:
```suggestion
        'compiler_standard': {},  # For each language C/C++.
```

gyorb:
New PR #2193 was created, we can close this.

PR #1930, opened 2019-02-12
make StringIO usage py3 compatible at stats

gyorb:
Use StringIO from io package to make stats collector
source code parts py3 compatible.


gyorb:
Should we put this into 6.9.1 or to 6.10? Statistical collectors might need further testing. I've tested it on tmux and xerces and it works ok.

PR #1941, opened 2019-02-19
Use compiler_info.json file in debug scripts.

rnkovacs:
The old pair of `compiler_[includes/target].json` files is not supported
anymore by new CodeChecker versions. Update debug tools.

Needs to be updated when #1908 lands.


scripts/debug_tools/failure_lib.py
```diff
@@ -40,6 +40,11 @@ def change_paths(string, pathModifierFun):
         # Note, this supports only POSIX paths.
         if string[i] == '/':
             path, path_end = find_path_end(string, i)
+            # Make sure that the prospective output folder exists.
+            if string[(i-3):(i-1)] == "-o":
```
martong: Is `-o/path/to/somewhere/a.o` valid? Or `-o` requires a space always after?
Also, `-o` could be followed by several whitespaces before the path.

rnkovacs: Sorry, that was a hotfix, I should have revised it. Replaced with regex matching.

martong:
Ping

gyorb:
Please check the style errors (pycodestyle, pylint) Travis failed because of that. Thansk

rnkovacs:
Fixed style errors.

martong:
@gyorb Thanks for the merge, #2039 is related and needed for the debug scripts to work, could you please take a look on that too?

PR #2001, opened 2019-03-21
update some python deps

gyorb:
psycopg2 update is required to run on ubuntu 18.04

### DIFF
--- a/web/requirements_py/osx/requirements.txt
+++ b/web/requirements_py/osx/requirements.txt
@@ -1,5 +1,5 @@
 alembic==0.9.2
 portalocker==1.1.0
 psutil==5.2.2
-sqlalchemy==1.1.11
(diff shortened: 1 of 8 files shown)


web/requirements_py/osx/requirements.txt
```diff
@@ -1,5 +1,5 @@
 alembic==0.9.2
 portalocker==1.1.0
 psutil==5.2.2
```
Xazax-hun: PSUtil not updated here.

web/requirements_py/osx/requirements.txt
```diff
@@ -1,5 +1,5 @@
 alembic==0.9.2
 portalocker==1.1.0
```
Xazax-hun: What about alembic, portalocker and pycodestyle?

PR #2003, opened 2019-03-21
Keep clang flags

bruntib:
#1970 should go first!

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
@@ -201,10 +201,17 @@ class ClangSA(analyzer_base.SourceAnalyzer):
                     analyzer_cmd.extend(['-Xclang',
                                          '-analyzer-display-ctu-progress'])
 
-            # Set language.
-            analyzer_cmd.extend(['-x', self.buildaction.lang])
-            if self.buildaction.target != "":
+            def has_flag(flag):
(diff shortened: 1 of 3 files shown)


analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -330,6 +332,8 @@ def construct_config_handler(cls, args, context):
                     re.sub(r'\$\((.*?)\)',
                            replace_env_var(args.clangsa_args_cfg_file),
                            sa_cfg.read().strip())
+                handler.analyzer_extra_arguments = \
+                    shlex.split(handler.analyzer_extra_arguments)
```
gyorb: Couldn't we do the `shlex.split`  inside the config handler?

analyzer/tests/unit/test_option_parser.py
```diff
@@ -229,8 +229,10 @@ def test_preprocess_and_compile(self):
     def test_ignore_flags(self):
         """
         Test if special compiler options are ignored properly.
+        In case the compiler is clang then the original command should be
```
gyorb: I think we should create a separate test case for not modifying clang compilation actions.

With this change you modify the tests for the gcc/g++ case and remove the information that the ` "-MT hello", "-M",` flags are filtered out in that case.

bruntib: These precompilation related flags are filtered in case of clang too.

PR #2039, opened 2019-04-08
Resurrect --compiler-info-file analyze flag.

dkrupp:
This option was removed by 5519409 under the assumption that it is not
used anywhere. The truth is that this flag is very useful for debugging
purposes and some debug tools in the /scripts directory do use it.
This commit re-introduces the --compiler-info-file option on top of the
refactored log parser.

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -19,6 +19,7 @@ import tempfile
 import traceback
 
 from codechecker_common.logger import get_logger
+from codechecker_common.util import load_json_or_empty
 
 from .. import gcc_toolchain
 from .build_action import BuildAction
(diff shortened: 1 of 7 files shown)


analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -549,6 +557,13 @@ def main(args):
     if 'none' in args.compile_uniqueing and 'ctu_phases' in args:
         args.compile_uniqueing = "alpha"
 
+    compiler_info_file = None
+    if 'compiler_info_file' in args:
+        LOG.debug("Compiler info is read from: %s", args.compiler_info_file)
+        compiler_info_file = args.compiler_info_file
```
gyorb: Should we check if the value of `args.compiler_info_file` is not `None`? If it is set on the command line without a value what will be the value here?

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -699,7 +718,7 @@ def f(flag_iterator, details):
     return f
 
 
-def parse_options(compilation_db_entry):
+def parse_options(compilation_db_entry, compiler_info_file=None):
```
gyorb: Shouldn't we set the default value `compiler_info_file=None` on the class staticmethod?
```
    @staticmethod
    def set(details, compiler_info_file):
```

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -818,6 +838,7 @@ class CompileActionUniqueingType(object):
 
 
 def parse_unique_log(compilation_database,
+                     report_dir,
```
gyorb: please texted the documentation that a `compiler_info.json` is witten out to the reports directory too.
Why do we need to do that here? It has not much connection to log parsing.

gyorb: Not depending on that the log parsing tests do not need any changes,

gyorb:
Just a remark so we should not forget:
In #1908 there is an extension which will get the compiler info for both languages which should change the `compiler_info.json` loading introduced here.
The recently added `renew_info_files.py` debug script needs to be updated too.
And the other debug scripts in #1941.

analyzer/codechecker_analyzer/buildlog/log_parser.py
dkrupp: it is documented now. Later on, we may want to factor out compiler auto-detection from parse_unique_log.

analyzer/codechecker_analyzer/cmd/analyze.py
dkrupp: if compiler_info_file is in args then it cannot be None.
If no parameter is given to --compiler_info_file the argparser complains that --compiler-info-file must have 1 argument.

I extended this with the verification wether args.compiler_info_file exists.

PR #2062, opened 2019-04-10
The skipped flags are skipped in case of Clang too

bruntib:
The idea is that in skip-list we are enumerating the flags which
are consumed by GCC but not by Clang. Theoretically there shouldn't
be any skipped flag if the original command used Clang since these
are known only by GCC.
However, in case of Clang we'd like to skip flags starting with
-Xclang which is also listed in our skip-list. -Xclang -emit-llvm is
such a flag that makes Clang emit an LLVM IR instead of the PList file.

### DIFF
--- a/analyzer/tests/unit/test_option_parser.py
+++ b/analyzer/tests/unit/test_option_parser.py
@@ -232,9 +232,7 @@ class OptionParserTest(unittest.TestCase):
         """
         ignore = ["-Werror", "-fsyntax-only",
                   "-mfloat-gprs=double", "-mfloat-gprs=yes",
-                  "-mabi=spe", "-mabi=eabi",
-                  '-Xclang', '-mllvm',
-                  '-Xclang', '-instcombine-lower-dbg-declare=0']
+                  "-mabi=spe", "-mabi=eabi"]
         action = {
             'file': 'main.cpp',
             'command': "g++ {} main.cpp".format(' '.join(ignore)),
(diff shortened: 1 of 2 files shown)


bruntib:
It is not `-e$` which ignores this flag but `-Xclang` with one argument. **Clang doesn't have a skip list**, because originally we shouldn't skip anything from a Clang command. The reason why I chose this solution is that `-Xclang` was already listed for skipping, because we must have had a similar issue earlier. The issue came back after refactoring the log parser module. `-Xclang` is the only Clang related flag, and it was a workaround even at its introduction to the skip list. We should think about the final solution.

I was thinking of keeping the test functions separate for Clang and GCC, but after this change the skipping action is common for both compilers: we are expecting to omit the same flags for both compilers. Although I agree with consolidating the log parser in a second refactoring phase, but we should discuss about how clever CodeChecker wants to be about flag handling. This should be done together with the correct handling of preprocessing build actions.

analyzer/tests/unit/test_option_parser.py
```diff
@@ -257,9 +254,21 @@ def test_ignore_flags_clang(self):
             'command': "clang++ {} main.cpp".format(' '.join(ignore)),
             'directory': ''}
         res = log_parser.parse_options(action)
-        ignore.remove("-E")
-        ignore.remove("-M")
-        self.assertEqual(res.analyzer_options, ignore)
+        self.assertEqual(res.analyzer_options, ["-fsyntax-only"])
+
+    @unittest.skip("This will be enabled when we distinguish -Xclang params.")
+    def test_ignore_groups(self):
```
gyorb:
```suggestion
    def test_ignore_xclang_groups(self):
```

Please add a short documentation to the test.

PR #2072, opened 2019-04-14
[clang] Skip sources argument when parsing the precompilation options

irishrover:
Precompilation options are needed for clang too.
Example from the Chromium project:
```
...
 "command": "bin/clang++ -MMD -MF new_tab_page_navigation_throttle.o.d -DUSE_UDEV -DUSE_AURA=1
...
```


bruntib:
Good catch, thanks for the fix!

PR #2076, opened 2019-04-16
Add --trim_path_prefix option for parser command

hpwxf:
This --trim_path_prefix option allows to trim path of reported files as
it acts in store command.

Example:
CodeChecker parse --trim-path-prefix=/path -e html -o ./html_report ./results

### DIFF
--- a/analyzer/codechecker_analyzer/cmd/parse.py
+++ b/analyzer/codechecker_analyzer/cmd/parse.py
@@ -142,6 +142,19 @@ def add_arguments_to_parser(parser):
                              "consult the User guide on how a Skipfile "
                              "should be laid out.")
 
+    parser.add_argument('--trim-path-prefix',
+                        type=str,
+                        nargs='*',
+                        dest="trim_path_prefix",
+                        required=False,
+                        default=argparse.SUPPRESS,
+                        help="Removes leading path from files which will be "
+                             "printed. So if you have /a/b/c/x.cpp and "
+                             "/a/b/c/y.cpp then by removing \"/a/b/\" prefix "
+                             "will print files like c/x.cpp and c/y.cpp. "
+                             "If multiple prefix is given, the longest match "
+                             "will be removed.")
+
     logger.add_verbose_arguments(parser)
 
     def __handle(args):
@@ -290,6 +303,16 @@ def main(args):
         with open(args.skipfile, 'r') as skip_file:
             skip_handler = SkipListHandler(skip_file.read())
 
+    trim_path_prefixes = args.trim_path_prefix if \
+        'trim_path_prefix' in args else None
+
+    def trim_path_prefixes_handler(source_file):
+        """
+        Callback to util.trim_path_prefixes to prevent module dependency
+        of plist_to_html
+        """
+        return util.trim_path_prefixes(source_file, trim_path_prefixes)
+
     html_builder = None
 
     for input_path in args.input:
@@ -312,7 +335,8 @@ def main(args):
                               output_path,
                               context.path_plist_to_html_dist,
                               skip_html_report_data_handler,
-                              html_builder)
+                              html_builder,
+                              trim_path_prefixes_handler)
             continue
 
         files = []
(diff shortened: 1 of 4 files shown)


analyzer/codechecker_analyzer/cmd/parse.py
```diff
@@ -142,6 +142,19 @@ def add_arguments_to_parser(parser):
...
+    parser.add_argument('--trim-path-prefix',
+                        type=str,
+                        nargs='*',
+                        dest="trim_path_prefix",
+                        required=False,
+                        default=argparse.SUPPRESS,
+                        help="Removes leading path from files which will be "
+                             "reported. So if you have /a/b/c/x.cpp and "
+                             "/a/b/c/y.cpp then by removing \"/a/b/\" prefix "
+                             "will store files like c/x.cpp and c/y.cpp. "
```
gyorb: Please update the help message. This is done for parse not for store.

tools/plist_to_html/plist_to_html/PlistToHtml.py
```diff
@@ -490,7 +501,7 @@ def main():
     html_builder = HtmlBuilder(args.layout_dir)
     for input_path in args.input:
         changed_files = parse(input_path, args.output_dir, args.layout_dir,
-                              None, html_builder)
+                              None, html_builder, None)
```
gyorb: You set the default value to `None` in `parse` you do not need to explicitly set it here too.

tools/plist_to_html/plist_to_html/PlistToHtml.py
```diff
@@ -340,7 +348,7 @@ def update_source_file(file_id):
 
 
 def plist_to_html(file_path, output_path, html_builder,
-                  skip_report_handler=None):
+                  skip_report_handler=None, trim_path_prefixes=None):
```
gyorb: You do not need to set `trim_path_prefixes` to `None` here. It is already set as a default value in `get_report_data_from_plist`.

tools/plist_to_html/plist_to_html/PlistToHtml.py
```diff
@@ -19,6 +19,8 @@
 from string import Template
 from xml.parsers.expat import ExpatError
 
+from codechecker_common import util
```
gyorb: @csordasmarton what do you think about this module dependency? PlistToHtml.py should be an independent tool. Should we copy the `trim_path_prefixes` function to this tool?

tools/plist_to_html/plist_to_html/PlistToHtml.py
hpwxf: I added the new argument `trim_path_prefixes` after `skip_report_handler` to minimize the impact of such signature change. To define `trim_path_prefixes` as non-optional, I have to move it before `skip_report_handler` (which was already optional). Do you prefer that ?

hpwxf:
Ok; I'm going to add the path trim feature for `stdout` output.

tools/plist_to_html/plist_to_html/PlistToHtml.py
hpwxf: This may explain the following error in Travis CI:
`Failure: ImportError (No module named codechecker_common) ... ERROR`

hpwxf: I fixed that illegal dependency by passing `trim_path_prefixes` using a `trim_path_prefixes_handler` callback (which can call `codechecker_common.util.trim_path_prefixes`)

tools/plist_to_html/plist_to_html/PlistToHtml.py
gyorb: I think we can keep it as it is in your last commit.

gyorb:
What is might not be visible right now is that we are trying to split up our repository for various tools (plist to html, tu_collector, analyzer and web parts.) which is not finished yet.

The analyzer part includes the plist to html converter tool but generally the plist to html tool should be independent of the analyzer or common modules.
Splitting is not finished yet, that is why the skip list handler is provided from the analyzer module, but it is missing if you build or use the plist to html tool separately.

For now we can keep the current minimal dependency but on the long term we should split them up (we might need to replicate some functionality in the plist to html tool). I will create a separate issue for that.

hpwxf:
I have rewritten the history of these commits as one squash commit on top of the updated master.

gyorb:
Great, Thanks!

PR #2090, opened 2019-04-22
Various Python 3 compatibility fixes.

Xazax-hun:
With these changes I was able to run CodeChecker check command on a file with Python 3 interpreter successfully.`


analyzer/codechecker_analyzer/analysis_manager.py
```diff
@@ -625,7 +625,7 @@ def signal_handler(*arg, **kwarg):
                            callback=lambda results: worker_result_handler(
                                results, metadata, output_path,
                                context.analyzer_binaries)
-                           ).get(float('inf'))
```
Xazax-hun: Python 3 will throw an exception for inf, but not providing a timeout will result in no timeout exception.

analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -79,7 +79,8 @@ def __get_analyzer_version(context, analyzer_config_map):
         version = [analyzer_bin, u' --version']
         try:
             output = subprocess.check_output(shlex.split(' '.join(version)),
-                                             env=check_env)
+                                             env=check_env,
+                                             universal_newlines=True)
```
Xazax-hun: The name of this argument is not too intuitive. The point is, these kind of subprocess calls will return `str`s in Python2 and 'byte`s in Python3. It would be a lot of work to make CodeChecker handle both `str` and `byte` everywhere. Using the `universal_newlines` argument, we can force Python3 to return text rather then bytes.

analyzer/codechecker_analyzer/analysis_manager.py
gyorb: Was this just a "temporary" solution because of some problem in the map_async in python2? If it works without is we can remove it.

Xazax-hun: I do not recall any reason why was this there. The tests pass in Python 2, so I would vote for removing this argument.

analyzer/codechecker_analyzer/analyzer.py
Xazax-hun: Currently, in Python 2, we also get text with the default encoding (and not bytes) from subprocess calls. This added feature will just preserve the current semantics under Python 3. If we want to change the semantics, I think it should be done in a separate pull requests. But nowadays, I think almost all systems with reasonable settings should have unicode as the default encoding.

Xazax-hun:
> Please set travis ci and pylint for python3 to test everything (python 3.6).
> Did you check if the python2 client can store results to a python3 sever and a python3 client can store the results on a python2 server. Can thrift handle this right?

With this patch only the basic features are tested and working, so we are not ready yet to run the full test-suite in a python 3 environment. This is only a small incremental step towards Python 3 support. I think setting travis ci and pylint should be among the last steps when we are close to the goal.

whisperity:
> I think setting travis ci and pylint should be among the last steps when we are close to the goal.

Seems like a good direction, especially considering you can't mark things *optional* in Travis AFAIK.

whisperity:
Python3 interpreter, with Python2 virtualenv, or...? What happened to the dependencies and such?

Xazax-hun:
I created a Python 3 virtualenv with updated thrift.

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -418,16 +419,15 @@ def get_compiler_standard(compiler, language):
...
             with source.file as f:
                 f.write(VERSION_C if language == 'c' else VERSION_CPP)
 
             try:
-                proc = subprocess.Popen([compiler, source.name],
-                                        stdout=subprocess.PIPE,
-                                        stderr=subprocess.PIPE)
-                _, err = proc.communicate()  # Wait for execution.
+                err = ImplicitCompilerInfo.\
+                    __get_compiler_err(" ".join([compiler, source.name]))
```
gyorb: `__get_compiler_err` can return None if there is some OSError, will that cause problems with the `re.search` later?

Previously it did not run re.search if there was an exception.

analyzer/codechecker_analyzer/analyzers/result_handler_base.py
```diff
@@ -94,7 +94,7 @@ def analyzer_result_file(self):
                 self.buildaction.original_command
 
             out_file_name = analyzed_file_name + '_' + \
-                hashlib.md5(build_info).hexdigest() + '.plist'
+                hashlib.md5(build_info.encode()).hexdigest() + '.plist'
```
gyorb: Is it possible that a different hash will be generated with this for the same build_info?
If yes it can mess up the existing report directories a bit, because the old plist files will not be overwritten.

analyzer/codechecker_analyzer/analysis_manager.py
gyorb: Did you try to kill the pool during the analysis? Every process did stop? I would keep until we fully switch to a python3 interpreter see the [stackoverflow discussion]( https://stackoverflow.com/questions/11312525/catch-ctrlc-sigint-and-exit-multiprocesses-gracefully-in-python) and the [python bug](https://bugs.python.org/issue8296)

gyorb:
As a next step to py3 this is ok, this should still run with with python2.
I'm working on the full switch to python3.

analyzer/codechecker_analyzer/analysis_manager.py
Xazax-hun: I see. In this case we could either store the resulting object and pass an argument conditionally or we could always pass a very big timeout like one year which will not throw an exception in python 3 (the exception says inf is too big to be converted to a timestamp).

analyzer/codechecker_analyzer/analyzers/result_handler_base.py
Xazax-hun: I think if the string was a unicode string this should not change it at all (in Python 2). For Python 3, we need this to avoid getting an exception.

analyzer/codechecker_analyzer/analysis_manager.py
gyorb: Setting a large timeout should be fine, we can remove it later when only py3 is supported.

analyzer/codechecker_analyzer/analyzers/result_handler_base.py
gyorb: I do not think `build_info` is a unicode string it is constructed a few lines above. There are some places with unicode strings but not everywhere.
please set the encoding to `utf8` (`.encode(encoding='UTF-8',errors='ignore')`) if for some reason the interpreter can not detect the locale settings for some reason it will fallback to ansi-c which could cause problems (analyzed file name with a non ascii character in it).

analyzer/codechecker_analyzer/buildlog/log_parser.py
gyorb: Any thoughts on this?

analyzer/codechecker_analyzer/analyzers/result_handler_base.py
Xazax-hun: The `utf-8` is the default if we do not set it.

analyzer/codechecker_analyzer/buildlog/log_parser.py
Xazax-hun: Agreed.

whisperity:
Let me get this clear: merging these changes will still keep us running on Python 2.y.z with the usual installation guide, right?

As to ease the migration to Python 3, I think you should put up somewhere (maybe in a follow-up **draft PR** ([GitHub now supports such!](https://github.blog/2019-02-14-introducing-draft-pull-requests/))?) so we have a solid ground to move forward with, in terms of the 3rd party dependencies we use.

Xazax-hun:
Sure. Everything should still support Python 2. Even Travis still runs in a Python 2 env.

Xazax-hun:
Commit message extended.

PR #2126, opened 2019-05-15
Fix ccache handling

bruntib:
CCache basically has two ways to use:

ccache <compiler> <flags>
ccache <flags>

Currently we intend to support the first case. For this we have to
determine whether it is a compiler at the second position. Earlier
we checked if it matches on CC_LOGGER_GCC_LIKE. However, if this
environment variable is set to "ccache" (since the user wanted to
log ccache invocations directly) then the compiler at the second
position won't match. In this commit we consider the second parameter
a compiler if it is executable in the current environment.


bruntib:
Our logger converts the compiler to canonical path. Should we convert the compiler after "ccache" canonical too? What was the reason of this conversion in the logger?

gyorb:
@bruntib probably the reason for the canonical path is that when we collect the compiler information, the key is right now the path to the compiler.
If the compiler is changed for various parts of the project during one build to collect the right information we will need to separate the compilers right.

PR #2131, opened 2019-05-17
Don't follow symlink in the logger

bruntib:
The build environment can be set so "g++" is a symlink to
/usr/bin/ccache. CCache can detect whether it was run through this
symlink or it was run directly. In the former case CCache forwards
the command line arguments to the original g++ compiler. This way
we can query the implicit include paths from the compiler later in
CodeChecker. If we resolve the symlink, then the implicit include
path getter command line arguments go to CCache binary. The solution
is not to resolve the symlinks in the logger.

### DIFF
--- a/analyzer/tools/build-logger/src/ldlogger-tool-gcc.c
+++ b/analyzer/tools/build-logger/src/ldlogger-tool-gcc.c
@@ -354,7 +354,7 @@ int loggerGccParserCollectActions(
   size_t i;
   /* Position of the last include path + 1 */
   char full_prog_path[PATH_MAX+1];
-  char *path_ptr;
+  char *path_ptr = NULL;
 
   size_t lastIncPos = 1;
   size_t lastSysIncPos = 1;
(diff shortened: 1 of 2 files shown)


analyzer/tools/build-logger/src/ldlogger-tool-gcc.c
```diff
@@ -364,16 +364,11 @@ int loggerGccParserCollectActions(
   char* keepLinkVar = getenv("CC_LOGGER_KEEP_LINK");
   int keepLink = keepLinkVar && strcmp(keepLinkVar, "true") == 0;
 
-  /* If prog_ is a relative path we try to
-   * convert it to absolute path.
-   */
-  path_ptr = realpath(prog_, full_prog_path);
-
-  /* If we cannot convert it, we try to find the
+  /* If toolName_ is not an absolute path, we try to find it as an
```
dkrupp: Please add a comment here that we do not resolve symlinks on purpose, to support ccache use-case.

csordasmarton:
@bruntib :exclamation: :exclamation: :exclamation: If you modify the code, force push something to a pull request which will resolve all the review comments please **request a new review** from the user who requested changes:
image

PR #2133, opened 2019-05-20
Update of the doc to show more recent version of clang

sylvestre:


docs/README.md
```diff
@@ -176,8 +176,8 @@ binaries you intend to use.
 
 ```json
 "analyzers" : {
-  "clangsa" : "/path/to/clang/bin/clang-4.0",
-  "clang-tidy" : "/path/to/clang/bin/clang-tidy-3.8"
+  "clangsa" : "/path/to/clang/bin/clang-8",
+  "clang-tidy" : "/path/to/clang/bin/clang-tidy-7"
```
gyorb: I think we can move to clang-tidy-8 here too. Otherwise LGTM.

PR #2191, opened 2019-06-26
Skipping -specs flag in TU collector

bruntib:
This flag can be given a .specs file which contains the config options of
cc1, cc1plus, as, ld, etc. Sometimes this file is just a temporary during
the compilation. However, if the file doesn't exist, this flag fails the
compilation. Since this flag is not necessary for dependency generation,
we can skip it.

### DIFF
--- a/tools/tu_collector/tests/project/compile_command.json
+++ b/tools/tu_collector/tests/project/compile_command.json
@@ -1,7 +1,7 @@
 [
 	{
 		"directory": "/tmp",
-		"command": "g++ -o /dev/null main.cpp",
+		"command": "g++ -specs=NON_EXISTENT_FILE -o /dev/null main.cpp",
 		"file": "main.cpp"
 	},
   {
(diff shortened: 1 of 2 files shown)


Xazax-hun:
I do not get it. If the file can contain compilation options, it should be necessary for the TU collector phase. If we compile the TUs with a set of flags and import it with another set of flags, we might end up with inconsistent AST (especially if these flags included language options, macro definitions).

bruntib:
I'm not so familiar with the internal details of a compiler, but as far as I have seen these configuration options are not traditional compiler flags. You can see the default config of your compiler with the following command:
`g++ -dumpspecs`
I don't know if these are necessary for the analysis but there are a few things here:
1. Our user's specs file was under `/tmp`, so I believe they generated it. By the time of the analysis this file was not there, so if we want to save them, then this has to be summoned in the logger phase.
2. Clang doesn't have a `-specs` flag. This config must be given differently (https://stackoverflow.com/questions/24491386/is-there-an-equivalent-to-spec-files-for-clang-llvm-and-where-can-i-find-a-refe).
3. In this pull request I didn't add this flag to the black list of compiler flags. This modification is in the TU collector where we'd like to gather the list of files in a translation unit. I was thinking of adding it, but Clang is just giving a warning on this flag. The real issue is during "fail zip" generation which uses GCC.

PR #2193, opened 2019-06-28
Collect compiler information for multiple languages

gyorb:
If the same compiler was used multiple times in a
compile command json but compiled multiple source
files for differerent languages (C and C++) the implicit
compiler information was collected only when first detected
(only for C).

With this change when a compiler is first found the include
paths, default build targets and compiler standards
will be detected for both languages. So the information can
be used when the same compiler is compiling for a
different language.

Additional changes were done to set the include paths,
build targets and compiler standard at multiple analysis
stages like, statistis collection, ctu pre analysis and
for setting the clang tidy arguments.

New PR replacing #1908

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/ctu_triple_arch.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/ctu_triple_arch.py
@@ -15,6 +15,7 @@ import shlex
 from codechecker_analyzer.env import extend_analyzer_cmd_with_resource_dir
 
 from .. import analyzer_base
+from ..flag import has_flag
 
 
 def get_compile_command(action, config, source='', output=''):
@@ -22,13 +23,20 @@ def get_compile_command(action, config, source='', output=''):
     for other operations. """
 
     cmd = [config.analyzer_binary]
-    if action.target != "":
-        cmd.append("--target=" + action.target)
+
+    compile_lang = action.lang
+
+    if not has_flag('--target', cmd) and \
+            action.target[compile_lang] != "":
+        cmd.append("--target=" + action.target[compile_lang])
+
     extend_analyzer_cmd_with_resource_dir(cmd,
                                           config.compiler_resource_dir)
-    cmd.extend(action.compiler_includes)
+
+    cmd.extend(action.compiler_includes[compile_lang])
     cmd.append('-c')
-    cmd.extend(['-x', action.lang])
+    if not has_flag('-x', cmd):
+        cmd.extend(['-x', action.lang])
     cmd.extend(config.analyzer_extra_arguments)
     cmd.extend(action.analyzer_options)
     if output:
@@ -36,8 +44,8 @@ def get_compile_command(action, config, source='', output=''):
     if source:
         cmd.append(source)
 
-    if all(not opt.startswith('-std=') for opt in action.analyzer_options):
-        cmd.append(action.compiler_standard)
+    if not has_flag('-std', cmd) and not has_flag('--std', cmd):
+        cmd.append(action.compiler_standard[compile_lang])
 
     return cmd
 
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/statistics_collector.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/statistics_collector.py
@@ -23,6 +23,8 @@ from codechecker_common.logger import get_logger
 
 from codechecker_analyzer.env import extend_analyzer_cmd_with_resource_dir
 
+from ..flag import has_flag
+
 LOG = get_logger('analyzer')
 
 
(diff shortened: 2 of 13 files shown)


analyzer/codechecker_analyzer/analyzers/clangsa/statistics_collector.py
```diff
@@ -65,11 +65,25 @@ def build_stat_coll_cmd(action, config, source):
         cmd.extend(['-Xclang',
                     '-analyzer-checker=' + coll_check])
 
-    if action.target != "":
-        cmd.append("--target=" + action.target)
+    def has_flag(flag):
```
martong: This is defined also in `ctu_triple_arch.py`. Would it be possible to define only once?

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_triple_arch.py
```diff
@@ -22,11 +22,22 @@ def get_compile_command(action, config, source='', output=''):
     for other operations. """
 
     cmd = [config.analyzer_binary]
-    if action.target != "":
-        cmd.append("--target=" + action.target)
+
+    def has_flag(flag):
```
martong: It is a small function, but we could have a unittest for it, or a comment about an example param and return value.

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -443,47 +443,103 @@ def get_compiler_standard(compiler, language):
...
+                ICI.compiler_info[compiler][cpp_lang]['includes'] = \
+                    ICI.get_compiler_includes(compiler, cpp_lang,
+                                              details['analyzer_options'])
+                ICI.compiler_info[compiler][cpp_lang]['target'] = \
+                    ICI.get_compiler_target(compiler)
+                ICI.compiler_info[compiler][cpp_lang]['default_standard'] = \
+                    ICI.get_compiler_standard(compiler, cpp_lang)
+
+        # Set the collected information for the build action.
+        details['compiler_includes'][c_lang] = \
```
martong: It seems quite easy to make an error here, perhaps using a function which takes three params would make it easier to see if there is an error.
Something like
`set('compiler_includes',c_lang,'includes')`

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -443,47 +443,103 @@ def get_compiler_standard(compiler, language):
         return standard
 
     @staticmethod
-    def load_compiler_info(filename, compiler):
+    def load_compiler_info(filename, compiler, c_lang, cpp_lang):
```
bruntib: What is the reason of these two parameters? We could hard-code "c" and "c++" strings in this function.

gyorb: I do not wanted to hardcode it here and in the `ICI.set()` method too. With that there should not be any language key mismatch whet loading information into ICI (from the compiler) or loading the information from the compile info json file.

analyzer/codechecker_analyzer/buildlog/build_action.py
```diff
@@ -73,7 +73,7 @@ def __hash__(self):
         hash_content = []
         hash_content.extend(self.analyzer_options)
         hash_content.append(str(self.analyzer_type))
-        hash_content.append(self.target)
+        hash_content.append(self.target[self.lang])
```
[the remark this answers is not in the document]
gyorb: Because a BuildAction object is build from a dictionary (in the constructor) and `__slots__` magic I did not found any easy way without further hacking to used properties.

analyzer/tests/functional/analyze/test_analyze.py
```diff
@@ -181,22 +182,29 @@ def test_compiler_info_file_is_loaded(self):
             source.write(simple_file_content)
 
         with open(compiler_info_file, 'w') as source:
-            source.write('{"clang++": {"default_standard": "-std=FAKE_STD", '
-                         '"target": "FAKE_TARGET", "includes": ["-isystem '
-                         '/FAKE_INCLUDE_DIR"]}}')
+            source.write('''{
```
martong: Would we be able to handle existing compiler_info_files? Seems like the structure of the json is changed.
This is a problem for us, because we have existing report dirs with the old format in the CI jobs.
Either we can handle the old format, or could you please adapt https://github.com/Ericsson/codechecker/blob/master/scripts/debug_tools/renew_info_files.py to support this yet newer format?

analyzer/tests/functional/analyze/test_analyze.py
```diff
@@ -181,22 +182,29 @@ def test_compiler_info_file_is_loaded(self):
...
                        "--analyzers", "clangsa", "--verbose", "debug",
                        "-o", reports_dir]
-
         # Run analyze.
         process = subprocess.Popen(
             analyze_cmd, stdout=subprocess.PIPE,
             stderr=subprocess.PIPE, cwd=self.test_dir)
         out, _ = process.communicate()
-
+        print(out)
```
martong: left here accidentally? Or log?

analyzer/tests/unit/test_flag.py
```diff
@@ -0,0 +1,28 @@
...
+from codechecker_analyzer.analyzers import flag
+
+
+class Flag(unittest.TestCase):
+    """Compiler flag related tests."""
+
+    def test_has_flag(self):
+        """Test if flag is found in a command."""
+        cmd = ["g++", "--target", "x86_64"]
+        self.assertTrue(flag.has_flag("--target", cmd))
```
martong: self.assert???(flag.has_flag("x86_64", cmd))

analyzer/tests/functional/analyze/test_analyze.py
gyorb: I've kept it for debugging.

analyzer/tests/functional/analyze/test_analyze.py
gyorb: I've updated the renew_info_files script.

analyzer/tests/unit/test_flag.py
gyorb: I've extended the test.

PR #2206, opened 2019-07-03
Stats fix

dkrupp:
statistics collector checkers trash the normal report output. This patch prevents this by switching off these checkers at normal analzysis.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzer.py
@@ -171,6 +171,15 @@ def perform_analysis(args, skip_handler, context, actions, metadata):
         config_map[ClangSA.ANALYZER_NAME].set_checker_enabled(
             ReturnValueCollector.checker_analyze)
 
+    # Statistics collector checkers must be explicitly disabled
+    # as they trash the output.
+    if "clangsa" in analyzers:
+        config_map[ClangSA.ANALYZER_NAME].set_checker_enabled(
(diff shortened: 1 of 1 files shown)


analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -171,6 +171,15 @@ def perform_analysis(args, skip_handler, context, actions, metadata):
         config_map[ClangSA.ANALYZER_NAME].set_checker_enabled(
             ReturnValueCollector.checker_analyze)
 
+    # Statistics collector checkers must be explicitly disabled
+    # as they trash the output.
+    if "clangsa" in analyzers:
```
gyorb: If I enable the clangsa analyzer with the `--stats` flag the checkers will be disabled even if they were enabled previously.
I think we should check here if stats was enabled.
`if 'stats_enabled' in args and args.stats_enabled` we should leave them enabled.

dkrupp: Collector checkers will be be enabled even if --stats are given, because they are explicitly added in pre_analaysis_manager: https://github.com/Ericsson/codechecker/blob/master/analyzer/codechecker_analyzer/pre_analysis_manager.py#L38

I tried it out and also the tests pass. So I think this is fine.

gyorb: ok, I did not remember that we explicitly enable the collector checkers later.

PR #2212, opened 2019-07-08
ctu display progress flag changed in clang8

gyorb:
The ctu-display-progress flag did change in clang8.
The argument is turned on in case of ctu failure to
collect the source files used by ctu for analysis.

Not just the flag but the analyzer output format did
changed in clang8 too.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
@@ -205,9 +205,9 @@ class ClangSA(analyzer_base.SourceAnalyzer):
                      'experimental-enable-naive-ctu-analysis=true',
                      '-Xclang', '-analyzer-config', '-Xclang',
                      'ctu-dir=' + self.get_ctu_dir()])
-                if config.ctu_has_analyzer_display_ctu_progress:
-                    analyzer_cmd.extend(['-Xclang',
-                                         '-analyzer-display-ctu-progress'])
+                ctu_display_progress = config.ctu_capability.display_progress
+                if ctu_display_progress:
+                    analyzer_cmd.extend(ctu_display_progress)
 
             def has_flag(flag):
                 return bool(next((x for x in analyzer_cmd if
@@ -261,7 +261,7 @@ class ClangSA(analyzer_base.SourceAnalyzer):
             return set()
 
         regex_for_ctu_ast_load = re.compile(
-            r"ANALYZE \(CTU loaded AST for source file\): (.*)")
+            r"CTU loaded AST file: (.*).ast")
 
         paths = set()
 
@@ -337,11 +337,6 @@ class ClangSA(analyzer_base.SourceAnalyzer):
             handler.ctu_dir = os.path.join(args.output_path,
                                            args.ctu_dir)
 
-            handler.ctu_has_analyzer_display_ctu_progress = \
-                host_check.has_analyzer_feature(
-                    context.analyzer_binaries.get(cls.ANALYZER_NAME),
-                    '-analyzer-display-ctu-progress',
-                    environ)
             handler.log_file = args.logfile
             handler.path_env_extra = context.path_env_extra
             handler.ld_lib_path_extra = context.ld_lib_path_extra
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/ctu_autodetection.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/ctu_autodetection.py
@@ -16,6 +16,7 @@ import re
 import subprocess
 
 from codechecker_common.logger import get_logger
+from codechecker_analyzer import host_check
 
 LOG = get_logger('analyzer.clangsa')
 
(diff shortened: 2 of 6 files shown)


analyzer/codechecker_analyzer/analyzers/clangsa/ctu_autodetection.py
```diff
@@ -246,6 +247,33 @@ def mapping_tool_path(self):
 
         return False
 
+    @property
+    def display_progress(self):
+        """ """
```
csordasmarton: Function docstring is missing.

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_autodetection.py
```diff
@@ -246,6 +247,33 @@ def mapping_tool_path(self):
...
+                return None
+            return ['-Xclang', '-analyzer-config',
+                    '-Xclang', 'display-ctu-progress']
+        else:
+            ok = host_check.has_analyzer_feature(
+                self.__analyzer_binary, ['-analyzer-display-ctu-progress'],
+                self.environ)
+            if not ok:
+                return None
+            else:
```
csordasmarton: There is no need to write the `else` keyword (like above), just use return.

Xazax-hun:
How important it is to support the old flag?
I think the old flag was only present in our fork, it was never upstream. Internal users usually get a clang implementation from our fork with the CodeChecker package. 
So unless someone is actually building an out to date version of our fork the old flag will never be used. I don't think we want to support that scenario.

gyorb:
I've removed the support for the old flag.

martong:
> Codacy Here is an overview of what got changed by this pull request:
> 
> ```diff
> Issues
> ======
> - Added 1
>            
> ```
> 
> See the complete overview on [Codacy](https://app.codacy.com/app/CodeChecker/codechecker/pullRequest?prid=3822072&bid=13363914)

This Codacy bot generates so many trash into the comments.

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -261,7 +261,7 @@ def get_analyzer_mentioned_files(self, output):
             return set()
 
         regex_for_ctu_ast_load = re.compile(
-            r"ANALYZE \(CTU loaded AST for source file\): (.*)")
+            r"CTU loaded AST file: (.*).ast")
```
martong: I think we must change the code after the regex match too.
The Clang7 regex matched the source file (with extension .c, .cpp, etc).
In Clang8, the regex matches the AST file, with the .ast extension.
Probably we need something like this:
```
path = match.group(1)[:-4] # strip the .ast extension (Do we always have a .ast extension?)
```

Anyway, we have a test which checks for the existence of the .c files in the failure zip: `test_ctu_failure_zip(self)`.
How could that test pass?

gyorb: We always add the `.ast` extension to the file name:
https://github.com/Ericsson/codechecker/blob/8eb31361d642ead646b3a0b54ab70ba42aa20382/analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py#L99-L100

I prefer extending the regex which is a bit more verbose and easier to understand I think compared to `[:4]` which will always strip the last 4 characters no matter what that is.

Actually the problem was that in the clang7 fork the `.ast` extension was removed by clang when the loaded files were printed out.

martong: > I prefer extending the regex which is a bit more verbose and easier to understand I think compared to `[:4]` which will always strip the last 4 characters no matter what that is.

But, actually, we always want to remove the last 4 character no matter what don't we? The extension will always be ".ast".
If we change the regex then we have to express in the regex to not match ".ast" from the end of the string which is going to be an ugly regex IMHO (I consider all regex ugly BTW). In contrast to the very simple [:-4] annotated with a comment.

martong: And consider the edge cases like `banana.ast.cpp.ast`

analyzer/.noserc
```diff
@@ -10,4 +10,4 @@ detailed-errors=1
 stop=1
 
 # do not capture stdout
-#nocapture=1
+nocapture=1
```
csordasmarton: Do not forget to comment this when you are done.

analyzer/tests/functional/ctu_failure/test_ctu_failure.py
```diff
@@ -16,6 +16,7 @@
 import zipfile
 
 from codechecker_analyzer import host_check
+from codechecker_analyzer.analyzers.clangsa import ctu_autodetection
```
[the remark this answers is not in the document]
csordasmarton: Please fix this Codacy error.

analyzer/tests/functional/ctu_failure/test_ctu_failure.py
```diff
@@ -54,8 +55,13 @@ def __set_up_test_dir(self, project_path):
...
             host_check.has_analyzer_feature(self.__getClangSaPath(),
-                                            '-analyzer-display-ctu-progress')
-        print("Has -analyzer-display-ctu-progress? " +
+                                            ['-Xclang',
+                                             '-analyzer-config',
+                                             '-Xclang',
+                                             'display-ctu-progress=true'],
+                                            self.env)
+
+        print("Has display-ctu-progress=true? " +
```
csordasmarton: Please remove this debug message when you are done.

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
gyorb: The regex will only match and return this part `banana.ast.cpp` only.

analyzer/tests/functional/ctu_failure/test_ctu_failure.py
gyorb: This was previously there too, I've just renamed the config value which changed.

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
martong: OK.

PR #2226, opened 2019-07-15
Extend CTU autodetection to handle monorepo builds

gamesh411:
Monorepo builds of Clang do not have 2 repository urls in their version
string. Another parsing strategy is added to CTU autodetection to handle
this case.


gamesh411:
This PR is replaced by #2236.

PR #2237, opened 2019-07-19
Distinguish reports of different checkers

bruntib:
When a checker is alias of another then it is possible that two
checkers are reporting the same message. We wouldn't like to unique
them because it was random which checker won the possibility of
reporting. Now both checkers are emitting the report and the user can
explicitly choose to switch off any of them.

This commit also contains some improvement on checking "skipping". The
operator "|=" is not lazy and it is not always necessary to evaluate
the right hand side.

### DIFF
--- a/codechecker_common/report.py
+++ b/codechecker_common/report.py
@@ -43,6 +43,10 @@ class Report(object):
         return self.__main['issue_hash_content_of_line_in_context']
 
     @property
+    def check_name(self):
+        return self.__main['check_name']
+
+    @property
     def bug_path(self):
         return self.__bug_path
 
--- a/analyzer/codechecker_analyzer/cmd/parse.py
+++ b/analyzer/codechecker_analyzer/cmd/parse.py
@@ -151,7 +151,7 @@ class PlistToPlaintextFormatter(object):
                                     key=lambda r: r.main['location']['line'])
 
             for report in sorted_reports:
-                path_hash = get_report_path_hash(report.bug_path, report.files)
+                path_hash = get_report_path_hash(report)
                 if path_hash in self._processed_path_hashes:
                     LOG.debug("Not showing report because it is a "
                               "deduplication of an already processed report!")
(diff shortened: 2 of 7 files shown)


codechecker_common/report.py
```diff
@@ -251,7 +251,7 @@ def get_report_path_hash(report, files):
         col = str(event['location']['col']) if 'location' in event else 0
 
         report_path_hash += line + '|' + col + '|' + event['message'] + \
-            file_name
+            file_name + report.check_name
```
gyorb: Do we want to add the checker name for each event in the path? Is it not enough to add it only once to the hash content?

Please extend the function documentation about the change.

bruntib: Thanks for the good catch, it was not intentional. I didn't notice the loop few lines above.

bruntib:
I also added WIP tag to this PR, because there is a Phabricator patch to clang-tidy (https://reviews.llvm.org/D65065) for emitting duplicate reports. Only in this case will this improvement make sense.
Maybe there will be a command line option for clang-tidy to force these duplicate reports, however, this is under discussion yet (see the linked URL).

gyorb:
@bruntib any updates on this issue?

bruntib:
@gyorb I don't really understand your request about the extension of the commit message. I wrote that there are checker name aliases in ClangTidy.

And the answer for the first question is: yes. This PR is important, or to be precise it will be important in Clang10 where ClangTidy will emit the reports of alias checkers under each alias names. Otherwise CodeChecker is not able to distinguish the reports of alias checkers.

dkrupp:
this will be a clang 10 feature, so moved to a later codechecker release.

bruntib:
@gyorb I don't think this should depend on Clang version. That would only add an unnecessary complexity. I think those features should depend on Clang version which wouldn't work at all otherwise. For example this was the searching of clang-func-mapping or some compiler command line argument handling, etc.

gyorb:
@bruntib can we merge this the clang10 release is out, please fix the conflict.

PR #2249, opened 2019-07-26
Refactoring log_parser to an external tool

bruntib:


gyorb:
try out the converted compile command json with clang-tidy and clangd

gyorb:
@bruntib the work in #2632 is the same you started here rigth?

gamesh411:
> 
> 
> @bruntib the work in #2632 is the same you started here right?

It definitely seems that both PR-s are heading in the same direction.

bruntib:
Log parser will be an external tool made by @gamesh411: #2632.

PR #2280, opened 2019-08-16
-fstack-reuse is not supported by clang

dkrupp:


analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -38,6 +38,7 @@
 # will not be included in the output Clang command.
 IGNORED_OPTIONS = [
     # --- UNKNOWN BY CLANG --- #
+    '-fstack-reuse',
```
bruntib: The flags in this list are in alphabetical order, but apart from that it looks good to me.

PR #2285, opened 2019-08-22
Upgrade thrift from 0.9.1/2 to 0.11.0. Clean rebuild required.

rpavlik:
Fixes failure to build/run on Debian Buster. (See my note on #1827)

### DIFF
--- a/web/requirements.txt
+++ b/web/requirements.txt
@@ -3,4 +3,4 @@ sqlalchemy==1.3.5
 alembic==0.9.2
 portalocker==1.1.0
 psutil==5.2.2
-thrift==0.9.1
+thrift==0.11.0
--- a/web/requirements_py/db_pg8000/requirements.txt
+++ b/web/requirements_py/db_pg8000/requirements.txt
@@ -2,6 +2,6 @@ lxml==4.3.4
 sqlalchemy==1.3.5
 alembic==0.9.2
 pg8000==1.10.2
-thrift==0.9.1
+thrift==0.11.0
 psutil==5.2.2
 portalocker==1.1.0
--- a/web/requirements_py/db_psycopg2/requirements.txt
+++ b/web/requirements_py/db_psycopg2/requirements.txt
@@ -2,6 +2,6 @@ lxml==4.3.4
 sqlalchemy==1.3.5
 alembic==0.9.2
 psycopg2-binary==2.7.7
-thrift==0.9.1
+thrift==0.11.0
 psutil==5.2.2
 portalocker==1.1.0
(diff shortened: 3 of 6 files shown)


Xazax-hun:
Thanks! We will need to move away from thrift 0.9 soon, as Python 2 is approaching its EOL and that thrift version does not support Python 3.

gyorb:
Thanks for the PR!
Sorry but we can not merge this PR yet.

Our minimal required Linux version is the latest Ubuntu LTS and on Ubuntu no newer thrift-compiler is available yet (eoan has the newer version but it is not released yet).

Another problem is that in travis-ci to run our tests only Ubuntu LTS can be selected with this changes the tests will fail.

We were thinking to resolve this issue by generating the python and js files manually and put them in the repository so no thrift-compiler would be required to build the package and if the api changes we update the generated files, but this is not done yet. (For this we might need to rename shared.thrift to generate a different python module name (codechecker_api_shared) because it needs to be in the PYTHONPATH to be able to be imported by the generated files (can not be under some other module) and the current shared might not be the best name for that.)

rpavlik:
OK, that's fine. Just figured I'd offer up what I had to do to run it on the current Debian stable release, especially since the bug describing my problem was closed without any apparent action taken.  Good luck with the ubuntu LTS - as someone working in VR, I'm well aware of the struggles that can pose :D

gyorb:
Thanks for the PR it is great that you tested it on debian and only a few changes are needed!

We plan to move to newer thrift and I hope we can do it soon. For python3 we will need to change the thrift version because 0.9.2 can not generate python3 stubs.

hellow554:
Thrifts current version is 0.13.0, so this is somewhat obsolete.

Is there a way to detect the host thrift version and use the libraries for that instead?

gyorb:
With the changes introduced in #2570 we do not depend on the local thrift compiler to generate the stubs.
The generated thrift stubs are published to pypi and npmjs and used as a dependency.
The pypi packages are [codechecker-api](https://pypi.org/project/codechecker-api/) and [codechecker-api-shared](https://pypi.org/project/codechecker-api-shared/) and the npm packages are [codechecker-api](https://www.npmjs.com/package/codechecker-api)  for node.js and  [codechecker-api-js](https://www.npmjs.com/package/codechecker-api-js)  for jquery.
Right now the api stubs were generated with thrift v0.9.2 (python2 only) but they will be updated.
We need the newer stubs for python3.

On the master branch you should just update or re-create your virtualenv to download the dependencies.

Let us know if you run into any problems.

PR #2286, opened 2019-08-22
Omit extra newlines while logging build commands

bruntib:


analyzer/codechecker_analyzer/buildlog/build_manager.py
```diff
@@ -44,7 +44,7 @@ def execute_buildcmd(command, silent=False, env=None, cwd=None):
         if not line and proc.poll() is not None:
             break
         if not silent:
-            print(line)
+            sys.stdout.write(line)
```
gyorb: Should we check if the line is empty?

bruntib: I don't think so. If the line was empty in the original build output and we skip it, that would change the look of the original build.

PR #2292, opened 2019-08-28
api modeling checkers should be enabled in all profiles

gyorb:
api modelling checker do some background modelling
which information is required for the other checkers
they should be enabled


Szelethus:
Are we sure that the current "disable everything that isn't enabled" isn't redundant with `--analyzer-no-default-checks`? The change is fine, but that flag still bugs me.

gamesh411:
I have checked the current implementation of --analyzer-no-default-checks, and if its boolean value is false, then API-modelling alongside with platform-specific checks are enabled. My personal preference would be to NOT provide the flag to Clang when using CodeChecker (thus making the option value false, which would mean enabling the core, API, and platform checks), however, I think making such a change would require throughout testing, as it could affect user expericence. Otherwise LGTM.

dkrupp:
This pull request is the replacement of this
https://github.com/Ericsson/codechecker/pull/2306

gamesh411:
We have come to the conclusion, that we should merge this change into both master, and 6.10.1 branch, and the new selection of default checkers should only be released in 6.11.

PR #2293, opened 2019-08-29
gcc compiler flags against Spectre unknown by clang

gyorb:
gcc uses different flags to harden the code against Spectre
-mindirect-branch=
-mfunction-return=
-mindirect-branch-register

clang v6 introduced these compiler flags:
-mretpoline
-mretpoline-external-thunk


bruntib:
We have a `REPLACE_OPTIONS_MAP` for replacing GCC compiler flags to their Clang corresponding. If these flags are not necessary for the analysis then we can omit them of course.

gyorb:
It influences the assembly created by the compiler I do not think skipping these arguments should affect the analysis.

PR #2298, opened 2019-09-03
match for multiple digit clang versions in tests

gyorb:
in case the current latest clang master branch (v10.0.0) is used
during the tests the regex did not match the double digit major version
and failed to run the tests


gyorb:
already fixed

PR #2300, opened 2019-09-04
Crash in splitting output of "clang -### ..."

bruntib:
The output of "clang -### ..." command can't always be split with
shlex.split() so we now use the simple split().

The output of "clang -### ..." command may contain the invocation of
the linker if the original build command didn't have "-c" flag. This
last line doesn't contain "-triple" flag but the previous one does.
So now we try to find "-triple" in the whole output instead of using
only its last line.


dkrupp:
please create a pull request to master branch too

PR #2303, opened 2019-09-10
the stdlib flag by clang modifies the implicit include paths

gyorb:
The -stdlib= flag used by clang modifies the implicit include paths.
If the flag is in the compile command it should be used when
the implicit include paths are detected.


gyorb:
duplicate of #2294 for the release branch

PR #2304, opened 2019-09-10
gcc compiler flags against Spectre unknown by clang

gyorb:
gcc uses different flags to harden the code against Spectre
-mindirect-branch=
-mfunction-return=
-mindirect-branch-register

clang v6 introduced these compiler flags:
-mretpoline
-mretpoline-external-thunk


gyorb:
duplicate of #2293 for the release branch

PR #2305, opened 2019-09-10
api modeling checkers should be enabled in all profiles

gyorb:
api modelling checker do some background modelling
which information is required for the other checkers
they should be enabled


gyorb:
duplicate of #2292 for the release branch

PR #2306, opened 2019-09-12
[WIP] Use the default checkers suggested by ClangSA

gamesh411:
The `--analyzer-no-default-checks` option is no longer passed to Clang,
so the default-enabled checkers are picked. Also no checkers are explicitly
disabled by CodeChecker.


Szelethus:
I'll add myself if you don't mind, wouldn't like to miss out on analyzer related changes :)

gamesh411:
The explicit modelling of default state this way leads to code complexity increase and spreads uncertainty in the later stages of result handling (eg. suppose the user wants to know why a result was different between diffs: is it because the code was fixed, or the checker was disabled? this cannot be determined when the state of the checker is the default. Was the checker enabled or disabled by default?... all these questions lead to reduced usability. An alternative approach will be implemented.

PR #2317, opened 2019-09-17
ld_logger improvement

### LINKED ISSUE #2315: make it possible to give absolute path in the CC_GCC_LOGGER_LIKE environment variable
if the compiler wrapper script and the real compiler has the same name, both will be logged in the compile_commands.json and there is not possibility to skip the wrapper script.

I propose the following solution:

if the user sets a path in CC_LOGGER_GCC_LIKE like:
CC_LOGGER_GCC_LIKE=DT_DXCC_SLES12/x86_64-dx/4.8/4.8.5/bin/x86_64-dx-linux-gnu-g++

codechecker should match this string from the end of the full path of the compilation command that is invoked.

bruntib:
In case a compiler in CC_LOGGER_?_LIKE environment variable contains
slash (/) character, then we assume that this is a postfix of a
compiler's path. Otherwise we check if it is the infix of the compiler
name like in case of the original behavior.

Fixes #2315

### DIFF
--- a/analyzer/tools/build-logger/src/ldlogger-tool.c
+++ b/analyzer/tools/build-logger/src/ldlogger-tool.c
@@ -10,6 +10,7 @@
 
 #include <stdlib.h>
 #include <assert.h>
+#include <string.h>
 
 #define PROG_LIST_SEPARATOR ":"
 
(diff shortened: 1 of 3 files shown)


dkrupp:
closes #2315

PR #2337, opened 2019-09-27
[WIP] Handle enabled checker per profile for ClangSA

gamesh411:
Currently, the checkers are listed one by one, and the profiles they belong to are listed for each of them. This patch reverses the mapping, and groups the checkers into packages where applicable.


config/config.json
```diff
@@ -17,219 +17,18 @@
...
...ugprone-misplaced -e bugprone-move -e bugprone-sizeof -e bugprone-string-compare -e bugprone-string-constructor -e bugprone-string-literal -e bugprone-suspicious-enum -e bugprone-suspicious-memset -e bugprone-suspicious-missing -e bugprone-suspicious-semicolon -e bugprone-swapped -e bugprone-terminating -e bugprone-throw -e bugprone-too -e bugprone-undefined -e bugprone-undelegated -e bugprone-unused -e bugprone-use -e bugprone-virtual -e cert-dcl54 -e cert-dcl58 -e cert-dcl59 -e cert-err09 -e cert-fio38 -e cert-oop11 -e cppcoreguidelines-special -e google-build-namespaces -e google-global -e misc-assert -e misc-bool -e misc-dangling -e misc-definitions -e misc-fold -e misc-forward -e misc-forwarding -e misc-inaccurate -e misc-incorrect -e misc-inefficient -e misc-lambda -e misc-macro -e misc-misplaced -e misc-move -e misc-redundant -e misc-sizeof -e misc-string-compare -e misc-string-constructor -e misc-string-literal -e misc-suspicious-enum -e misc-suspicious-missing -e misc-suspicious-semicolon -e misc-swapped -e misc-unconventional -e misc-uniqueptr -e misc-unused-raii -e misc-use -e misc-virtual -e performance-inefficient-algorithm -e performance-move -e performance-noexcept",
```
gyorb: I'm thinking about a small change to convert this to a list:
```json
   "default": [
    "-e bugprone-assert",
    "-e bugprone-bool",
    "-e bugprone-copy",
    "-e bugprone-dangling",
    "-e bugprone-fold"
  ]
```
I think reading and parsing it would be easier, the same applies for the other profiles.

gamesh411: I have created an enable-only list, and I have updated the PR to reflect the most recent changes.

gamesh411:
> Please add to the commit the information that a checker grouping was done too.
> 
> @gamesh411 was the plan to do the conversion without grouping and than a separate PR with the grouping?

There is no consensus on whether this should be 2 commits or one. I will create a two-commit version if that would be more informative.

gamesh411:
I have uploaded a 2 commit version. Please make sure to merge them without squashing, so the diff of the grouping can be checked.

PR #2342, opened 2019-09-27
Do not apply skip list to CTU pre-analysis

### LINKED ISSUE #2299: Skip list should not be applied to the CTU pre-analysis step
otherwise CTU cannot be applied in incremental analysis

dkrupp:
> Closes #2299 

Don't apply the skip list to CTU and statistical pre-analysis
by default, because it makes the analysis broken.
Skip list can be applied explicitly for these phases
if analyze is called with --ctu-collect or --stats-collect

skip list will not be used in the pre-analysis phase if only --ctu or --statistics parameter is given.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzer.py
@@ -247,11 +247,19 @@ def perform_analysis(args, skip_handler, context, actions, metadata):
 
         pre_analyze = [a for a in actions
                        if a.analyzer_type == ClangSA.ANALYZER_NAME]
+        pre_anal_skip_handler = None
+
+        # Skip list is applied only in pre-analysis
+        # if --ctu-collect or --stats-collect  was called explicitly
+        if ((ctu_collect and not ctu_analyze)
+                or ("stats_output" in args and args.stats_outptut)):
+            pre_anal_skip_handler = skip_handler
+
         pre_analysis_manager.run_pre_analysis(pre_analyze,
                                               context,
                                               config_map,
                                               args.jobs,
-                                              skip_handler,
+                                              pre_anal_skip_handler,
                                               ctu_data,
                                               statistics_data,
                                               manager)
--- a/analyzer/codechecker_analyzer/cmd/analyze.py
+++ b/analyzer/codechecker_analyzer/cmd/analyze.py
@@ -592,7 +592,6 @@ def main(args):
             load_json_or_empty(log_file),
             report_dir,
             args.compile_uniqueing,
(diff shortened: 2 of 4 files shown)


analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -247,11 +247,18 @@ def perform_analysis(args, skip_handler, context, actions, metadata):
 
         pre_analyze = [a for a in actions
                        if a.analyzer_type == ClangSA.ANALYZER_NAME]
+        pre_anal_skip_handler = None
+
+        # Skip list is applied only in pre-analysis
+        # if --ctu-collect or --stats-collect  was called explicitly
+        if (ctu_collect and not ctu_analyze) or "stats_output" in args:
```
gyorb: Shouldn't we check if a value is set for stats_output as we do at other places?
https://github.com/Ericsson/codechecker/blob/9c5dc8202576fb2af98428946f7f1624842cf5a6/analyzer/codechecker_analyzer/analyzer.py#L109
If the flag is activated it will be in the args Namespace even if the user did not set it.

analyzer/codechecker_analyzer/pre_analysis_manager.py
```diff
@@ -94,6 +94,7 @@ def pre_analyze(params):
 
     if skip_handler and skip_handler.should_skip(action.source):
         return
+
```
gyorb: Do we need this extra line here?

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -592,7 +592,7 @@ def main(args):
             load_json_or_empty(log_file),
             report_dir,
             args.compile_uniqueing,
-            skip_handler,
+            None,
```
gyorb: We could remove the skip_handler from the `parse_unique_log` function if it is not used. I did not found any other invocation where we would `call parse_unique_log` with a `skip_handler`.

gyorb:
please create a new PR for the release branch too.

analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -247,11 +247,19 @@ def perform_analysis(args, skip_handler, context, actions, metadata):
 
         pre_analyze = [a for a in actions
                        if a.analyzer_type == ClangSA.ANALYZER_NAME]
+        pre_anal_skip_handler = None
```
irishrover: Not the best name for a variable :-)

PR #2344, opened 2019-09-27
-nostdinc[++] affects implicit include paths too

bruntib:


analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -209,12 +209,12 @@ def filter_compiler_includes_extra_args(compiler_flags):
...
     compiler_flags -- A list of compiler flags which may affect the list
                       of implicit compiler include paths, like -std=,
-                      --sysroot=, -m32, -m64 or -stdlib=.
+                      --sysroot=, -m32, -m64, -nostdinc or -stdlib=.
     """
     # If these options are present in the original build command, they must
     # be forwarded to get_compiler_includes and get_compiler_defines so the
     # resulting includes point to the target that was used in the build.
-    pattern = re.compile('-m(32|64)|-std=|-stdlib=')
+    pattern = re.compile('-m(32|64)|-std=|-stdlib=|-nostdinc')
```
gyorb: will it match for `-nostdinc++` too, we should keep that too right? See: https://gcc.gnu.org/onlinedocs/cpp/Invocation.html

bruntib: Yes, it will, because we use `re.match()` which does a prefix match.

gyorb:
@irishrover could you verify that after this change you are still able to analyze chromium, instead of the `include-after` change introduced in #2308?

Thanks!

irishrover:
> @irishrover could you verify that after this change you are still able to analyze chromium, instead of the `include-after` change introduced in #2308?
> 
> Thanks!

Hi, yes it helps the Chromium to be analyzed, see my comment https://github.com/Ericsson/codechecker/pull/2308#issuecomment-536058667

gyorb:
Sorry, I've missed that comment, thanks for checking it!

gyorb:
@bruntib I would recommend to revert the commit from introduced in #2308 so we can see the reverted changes more explicitly, this commit message does not mention why the include-after flags were changed only the `nostdinc` changes. After the revert this commit should only contain the `nostdinc` related changed.

irishrover:
> @bruntib I would recommend to revert the commit from introduced in #2308 so we can see the reverted changes more explicitly, this commit message does not mention why the include-after flags were changed only the `nostdinc` changes. After the revert this commit should only contain the `nostdinc` related changed.

I've created a reverting commit: https://github.com/Ericsson/codechecker/pull/2353

docs/analyzer/user_guide.md
```diff
@@ -638,7 +638,7 @@ instead of the auto-detection you can pass that to the
 
 There are some implicit include paths (for example those which contain
 `include-fixed` directory) which are used only by GCC. By default CodeChecker
-doesn't collect them if `--skip-gcc-fix-include` flag is given. For
+doesn't collect them unless `--keep-gcc-include-fixed` flag is given. For
```
gyorb: What do you think about rephrasing these sentences a bit? I think it can be misunderstood, that the implicit include files will be only collected unless the  `--keep-gcc-include-fixed` flag is used. We should highlight a bit more that the include paths will be collected only the `include-fixed` directories will be skipped/removed from the list of the include paths.

PR #2347, opened 2019-10-01
'-' was missing from the iterator checkers argument

gyorb:


analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -204,7 +204,7 @@ def construct_analyzer_cmd(self, result_handler):
             analyzer_cmd.extend(['-Xclang',
                                  '-analyzer-config',
                                  '-Xclang',
-                                 'aggressive-binary-operation-simplification'
+                                 '-aggressive-binary-operation-simplification'
                                  '=true'])
```
bruntib: Is it correct that `'=true'` is a separate command line argument? The command will look like this:
```.sh
clang ... -Xclang -aggressive-binary-operation-simplification =true ...
```
Doesn't that extra space cause an error?

baloghadamsoftware:
- was intentionally missing. Adding an - is an error.

PR #2368, opened 2019-10-04
Removing MallocOverflow throws too many FPs

dkrupp:
So removing it from the sensitive profile.


gyorb:
duplicate of #2366 for the release branch

PR #2389, opened 2019-10-16
enable improved loop execution modeling

gyorb:
More details can be found here:
https://summerofcode.withgoogle.com/archive/2017/projects/6071606019358720/
https://reviews.llvm.org/D12358


gyorb:
Since clang 10 (I did not check previous versions) there are some new flags which could be used, maybe it is easier to configure clang analyzer like this:
- ` -funroll-loops          Turn on loop unroller`
- `-freroll-loops          Turn on loop reroller`

gyorb:
I'm closing this @martong please create a new PR if there will be a decision if the loop related analyzer features should be enabled or disabled by default.

PR #2403, opened 2019-10-24
[Statistics] Replace analyzer option `api-metadata-path` with package option `APIMetadataPath`

baloghadamsoftware:
Clang 9 enables package options beside analyzer and checker options.
This allows us to use package option instead of analyzer option for
passing the path of the YAML files needed by the statistics-based
checkers. The advantage of this approach is that the plugin now works
with the official Clang 9 release, no patching is needed. Since the
statistics-based checkers are in the same package, the option needs only
to be passed once for the whole analysis.

This patch adjust CodeChecker to work correctly with the new plugin.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/statistics_collector.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/statistics_collector.py
@@ -137,7 +137,8 @@ class SpecialReturnValueCollector(object):
             return []
         else:
             return ['-Xclang', '-analyzer-config',
-                    '-Xclang', 'api-metadata-path=' + path]
+                    '-Xclang',
+                    'alpha.ericsson.statisticsbased:APIMetadataPath=' + path]
 
     def total(self):
         return self.stats.get('total')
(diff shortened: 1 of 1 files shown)


analyzer/codechecker_analyzer/analyzers/clangsa/statistics_collector.py
```diff
@@ -137,7 +137,8 @@ def checker_analyze_cfg(path):
             return []
         else:
             return ['-Xclang', '-analyzer-config',
-                    '-Xclang', 'api-metadata-path=' + path]
+                    '-Xclang',
+                    'alpha.ericsson.statisticsbased:APIMetadataPath=' + path]
```
gyorb: Shouldn't this be clang version dependent?

We are doing it in [clang_options.py](https://github.com/Ericsson/codechecker/blob/9c2ab20fa1ca818fa9c935fad3a7e832d2640bfa/analyzer/codechecker_analyzer/analyzers/clangsa/clang_options.py)

dkrupp: since this checker is only available within ericsson, which will be released with clang9 anyway, this is ok.

PR #2434, opened 2019-11-19
create labeler github workflow action

gyorb:
Add github labeler workflow action to automatically label pull request.
The config file for the workflow action is in #2433.


gyorb:
I've put the config file for the action into the other PR #2433.

PR #2442, opened 2019-11-26
Skip file analysing if any of skip_handlers returns True, not both.

irishrover:


irishrover:
IMHO there's a typo in the current code.

irishrover:
This patch dramatically reduces the amount of memory consumed and the processing time when 

> -i SKIP_FILE

is used.

irishrover:
Gentle ping)

irishrover:
> Do you use CTU for the analysis?

Mostly not. May be we can rewrite the condition as 

`if ... OR ctu_skip_handler_is_on AND (ctu_skip_handler and ctu_skip_handler.skip)`
?

Note `ctu_skip_handler_is_on`

gyorb:
I've made a PR #2462 with some changes I'm thinking on but I did not test it yet properly.
It is a bit more change but makes the different skipping methods more explicit.

PR #2449, opened 2019-11-26
do not convert to absolute path the files after -include

### LINKED ISSUE #2440: Incorrect full path reconstruction for `-include`
```
$ mkdir /tmp/sa-test && cd /tmp/sa-test
$ touch input.c
$ mkdir header
$ echo zz > header/input.c
$ /usr/local/bin/clang -I/tmp/sa-test/header/ -include input.h -o /tmp/sa-test/input.o -c /tmp/sa-test/input.c
In file included from <built-in>:1:
/tmp/sa-test/header/input.h:1:1: error: unknown type name 'zz'
zz
^
/tmp/sa-test/input.c:1:1: error: expected identifier or '('
2 errors generated.
$ cat compile_commands.json 
[
{
  "directory": "/tmp/sa-test",
  "command": "/usr/local/bin/clang -I/tmp/sa-test/header/ -include input.h -o /tmp/sa-test/input.o -c /tmp/sa-test/input.c",
  "file": "/tmp/sa-test/input.c"
}
]
$ CodeChecker analyze --ctu /tmp/sa-test/compile_commands.json
[INFO 2019-11-23 12:30] - Pre-analysis started.
[INFO 2019-11-23 12:30] - Collecting data for ctu analysis.
[INFO 2019-11-23 12:30] - [1/1] input.c
[ERROR 2019-11-23 12:30] - Error generating AST.

command:

/usr/lib/llvm-10/bin/clang --target=x86_64-pc-linux-gnu -isystem /usr/local/include -isystem /usr/include/x86_64-linux-gnu -isystem /usr/include -c -x c -I/tmp/sa-test/header -include /tmp/sa-test/input.h /tmp/sa-test/input.c -std=gnu11 -emit-ast -D__clang_analyzer__ -w -o /tmp/clang-sa-reports/ctu-dir/x86_64/ast/tmp/sa-test/input.c.ast

stderr:

<built-in>:1:10: fatal error: '/tmp/sa-test/input.h' file not found
#include "/tmp/sa-test/input.h"
         ^~~~~~~~~~~~~~~~~~~~~~
1 error generated.

[ERROR 2019-11-23 12:30] - Error generating function map.

command:

/usr/lib/llvm-10/bin/clang-extdef-mapping /tmp/sa-test/input.c -- --target=x86_64-pc-linux-gnu -isystem /usr/local/include -isystem /usr/include/x86_64-linux-gnu -isystem /usr/include -c -x c -I/tmp/sa-test/header -include /tmp/sa-test/input.h -std=gnu11

stderr:

<built-in>:1:10: fatal error: '/tmp/sa-test/input.h' file not found
#include "/tmp/sa-test/input.h"
         ^~~~~~~~~~~~~~~~~~~~~~
1 error generated.
Error while processing /tmp/sa-test/input.c.
```

So upon seeing incomplete file path in `-include`, it just prepends `"directory"` field.
While clang proper has no issues finding the appropriate header.

gyorb:
File path after the -include flag should not be converted to absolute path.
Only the flags with include path directories should be converted.
If every other include directory flag is set properly the
file set by -include should be found too.

Resolves #2440

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -672,8 +672,8 @@ def __collect_compile_opts(flag_iterator, details):
         # TODO: If Clang will be extended with an extra analyzer option in
         # order to print these absolute paths natively, this conversion will
         # not be necessary.
-        flags_with_path = ['-I', '-idirafter', '-imacros', '-imultilib',
-                           '-include', '-iquote', '-isysroot', '-isystem',
+        flags_with_path = ['-I', '-idirafter', '-imultilib',
+                           '-iquote', '-isysroot', '-isystem',
                            '-iwithprefix', '-iwithprefixbefore', '-sysroot',
                            '--sysroot']
         if flag in flags_with_path:


analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -671,7 +671,7 @@ def __collect_compile_opts(flag_iterator, details):
         # order to print these absolute paths natively, this conversion will
         # not be necessary.
         flags_with_path = ['-I', '-idirafter', '-imacros', '-imultilib',
```
bruntib: `-imacros` should be removed too.

Xazax-hun:
Why do we ever convert something to absolute path in the first place? If we always set the working directory properly, relative paths should never be the problem.

Xazax-hun:
Oh, so the plist files section might contain duplicated if the same file is passed both as absolute and as relative path to the compiler invocation. So we are working around a bug in plists. I think we should fix plists at some point rather than trying to work this around.

gyorb:
Yes, we do it because of the plists, but just converting every path in the plist to absolute is not always the best solution, sometimes you need the relative paths.
The analyzer tools (cppcheck, clang) convert the file path to absolute if the compilation command contains absolute path and if it contains relative path they keep it relative (which is not good if we want to store them)

One option would be to introduce a flag in each analyzer to generate the reports with absolute paths even if the compilation command contained relative, but for this we would have to modify all the supported and planned analyzers. For now it is easier to do the conversion in CodeChecker.

PR #2398, opened 2019-10-22
keep all flags if the analyzer is the same as the compiler

gyorb:
If clang compiler is used to build and to analyze a project
do not filter the compiler options, keep them all.

The compiler installed dir and major versions are compared
to detect if the same clang is used for compilation and
analysis.

clang and clang++ are handled as they were the same binaries.

if the compiler is the same as the analyzer only a minimal
compiler option filtering is done.

- source, output and other files are removed from the compile command
- include paths are converted to absolute path required for the reports
- preprocessor flags are filtered out (the same way as for gcc/g++)

The list of filtered preprocessor flags is extended to remove -MD, -MG
options.

__collect_clang_compile_opts was introduced to
collect all the compile options should be run at
last everything should be filtered out earlier

__collect_compile_opts was split up
so the include path modification and the option filtering
is done separately, because the include path modification
is required for clang but the filtering not

resolves #2382

PR #2453, opened 2019-11-27
fix handling '--sysroot=' flag format

gyorb:
--sysroot flag format can be --sysroot=/path/to/include
In this case before the normalization the '=' sign must be removed.
We put back the original
--sysroot=/path/to/include as
--sysroot /path/to/include
which is a valid format too.


gyorb:
fix moved to #2398

PR #2461, opened 2019-11-29
Logger absolute include paths

### LINKED ISSUE #2447: log compile commands with absolute paths
Introduce environment variable to to convert the relative file paths to absolute during logging.
Source file, include paths ...

If the compile command database contains relative file paths the generated reports (clang, cppcheck) will contain relative file paths too.
After the analysis is done the CodeChecker store command will not know where the files (with relative path) are.

With this change the logged compilation database could be used with other tools to generate the reports with absolute file paths.

Cmake with the `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON ` converts the include paths and source files to absolute (if set correctly in the cmake file)

bruntib:
Some paths are converted to absolute path in the build command by
the logger if CC_LOGGER_ABS_PATH=1 given.

resolves #2447

### DIFF
--- a/analyzer/tools/build-logger/src/ldlogger-tool-gcc.c
+++ b/analyzer/tools/build-logger/src/ldlogger-tool-gcc.c
@@ -186,7 +186,7 @@ void getPathsFromEnvVar(
     if (length > PATH_MAX - 1){
       // If the string is too long, skip it.
       from = to + 1;
-      to = strchr(from, ":");
+      to = strchr(from, ':');
       continue;
     }
     strncpy(token, from, length);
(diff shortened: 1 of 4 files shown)


analyzer/tools/build-logger/src/ldlogger-tool-gcc.c
```diff
@@ -243,6 +243,61 @@ int isObjectFile(const char* filename_)
   return 0;
 }
 
+void transformSomePathsAbsolute(LoggerVector* args_)
+{
+  /* TODO: The argument of -isysroot (and maybe some other of these flags) may
```
gyorb: Does the `=` sign only modify the behavior when it can be found after the `-idirafter` flag?
```sh
-idirafter dir
       Add the directory dir to the list of directories to be searched for header files during preprocessing.
If dir begins with =, then the = is replaced by the sysroot prefix; see --sysroot and
... 
```

PR #2524, opened 2020-01-21
[tu-collector] Recognize CCache in build command

bruntib:
When the build command compiles with CCache, the tu-collector tries to
use it as preprocessor in order to gather the files of the translation
unit. This fails the TU collection, so in this commit we find the
original compiler.

### DIFF
--- a/tools/tu_collector/tu_collector/tu_collector.py
+++ b/tools/tu_collector/tu_collector/tu_collector.py
@@ -27,6 +27,7 @@ import string
 import subprocess
 import sys
 import zipfile
+from distutils.spawn import find_executable
 
 
 def __random_string(l):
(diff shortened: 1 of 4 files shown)


tools/tu_collector/tu_collector/tu_collector.py
```diff
@@ -51,6 +52,40 @@ def __get_toolchain_compiler(command):
                                 'g++' if is_cpp else 'gcc')
 
 
+def __determine_compiler(gcc_command):
```
gyorb: If you add `find_executable` function as a function argument to the `__determine_compiler` you can easily unit test this function.
In the test the `find_executable_mock` would just return some value or None depending on the test.

PR #2530, opened 2020-01-24
Fix error when query limit is larger than max limit

Vodorok:


web/server/codechecker_server/api/report_server.py
```diff
@@ -95,7 +95,10 @@ def verify_limit_range(limit):
     max_query_limit = constants.MAX_QUERY_SIZE
     if limit > max_query_limit:
         LOG.warning('Query limit %d was larger than max query limit %d, '
-                    'setting limit to %d', limit, max_query_limit, max)
```
Vodorok: max was wrong here. As its  buitlt-in, not a number, and a TypeError was thrown.

gyorb:
@Vodorok please create a pull request for the `release-v6.11.1` branch too about this fix, thanks!

PR #2533, opened 2020-01-27
Fix error when query limit is larger than max limit

Vodorok:


Vodorok:
PR created here also, as @gyorb reguested in #2530.

PR #2565, opened 2020-02-12
[analyzer] add -fno-keep-static-consts flag to gcc ingore list

gyorb:


analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -82,6 +82,7 @@
     '-fno-aggressive-loop-optimizations',
     '-fno-delete-null-pointer-checks',
     '-fno-jump-table',
+    '-fno-keep-static-consts'
```
bruntib: Comma is missing at the end of line.

PR #2586, opened 2020-02-28
check the limit value for every api endpoint

gyorb:
For some api endpoints no limit checking was done wich should be done.
The value of the limit was not checked if not set it could be None,
set the value to max query limit if no limit was set.


gyorb:
I moved this commit back to PR #2585 to make the test run successfully

PR #2652, opened 2020-03-26
[analyzer] Capture compiler errors as reports

bruntib:
There are some constructions which GCC compiles but Clang gives a
compilation error. In case such a compilation error happens, then the
whole translation unit is not analyzed and the users may not be informed
about this. By this change compiler errors also become CodeChecker
reports.

### DIFF
--- a/analyzer/codechecker_analyzer/analysis_manager.py
+++ b/analyzer/codechecker_analyzer/analysis_manager.py
@@ -335,6 +335,11 @@ def handle_failure(source_analyzer, rh, zip_file, result_base, actions_map):
 
     LOG.debug("ZIP file written at '%s'", zip_file)
 
+    # In case of compiler errors the error message still needs to be collected
+    # from the standard output by this postprocess phase so we can present them
+    # as CodeChecker reports.
+    rh.postprocess_result()
+
     # Remove files that successfully analyzed earlier on.
     plist_file = result_base + ".plist"
     if os.path.exists(plist_file):
--- a/analyzer/codechecker_analyzer/analyzer_context.py
+++ b/analyzer/codechecker_analyzer/analyzer_context.py
@@ -26,17 +26,21 @@ LOG = logger.get_logger('system')
 class SeverityMap(Mapping):
     """
     A dictionary which maps checker names to severity levels.
-    If a key is not found in the map and the checker name is a compiler warning
-    it will return severity level of MEDIUM.
+    If a key is not found in the map then it will return MEDIUM severity in
+    case of compiler warnings and CRITICAL in case of compiler errors.
     """
 
     def __init__(self, *args, **kwargs):
         self.store = dict(*args, **kwargs)
 
     def __getitem__(self, key):
-        # Key is not specified in the store and it is a compiler warning.
-        if key not in self.store and key.startswith('clang-diagnostic-'):
-            return "MEDIUM"
+        # Key is not specified in the store and it is a compiler warning
+        # or error.
+        if key not in self.store:
+            if key == 'clang-diagnostic-error':
+                return "CRITICAL"
+            elif key.startswith('clang-diagnostic-'):
+                return "MEDIUM"
 
         return self.store[key] if key in self.store else 'UNSPECIFIED'
 
--- a/analyzer/tests/functional/analyze_and_parse/test_files/compiler_error.output
+++ b/analyzer/tests/functional/analyze_and_parse/test_files/compiler_error.output
@@ -1,0 +1,37 @@
+NORMAL#CodeChecker log --output $LOGFILE$ --build "make compiler_error" --quiet
+NORMAL#CodeChecker analyze $LOGFILE$ --output $OUTPUT$ --analyzers clang-tidy --quiet
+NORMAL#CodeChecker parse $OUTPUT$
+--------------------------------------------------------------------------------
+[] - Starting static analysis ...
+[] - Analyzing compiler_error.cpp with clang-tidy  failed!
+[] - ----==== Summary ====----
+[] - Failed to analyze
+[] -   clang-tidy: 1
+[] - Total analyzed compilation commands: 1
+[] - ----=================----
+[] - Analysis finished.
+[] - To view results in the terminal use the "CodeChecker parse" command.
+[] - To store results use the "CodeChecker store" command.
+[] - See --help and the user guide for further options about parsing and storing the reports.
+[] - ----=================----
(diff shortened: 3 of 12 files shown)


analyzer/tests/functional/analyze_and_parse/test_files/compiler_error.output
```diff
@@ -0,0 +1,37 @@
...
+[] - Failed to analyze
+[] -   clang-tidy: 1
+[] - Total analyzed compilation commands: 1
+[] - ----=================----
+[] - Analysis finished.
+[] - To view results in the terminal use the "CodeChecker parse" command.
+[] - To store results use the "CodeChecker store" command.
+[] - See --help and the user guide for further options about parsing and storing the reports.
+[] - ----=================----
+[MEDIUM] compiler_error.cpp:3:11: expected ';' after return statement [clang-diagnostic-error]
```
csordasmarton: We set the severity levels of compiler warnings (clang-diagnostic-<warning>) by default to `MEDIUM`. I think it would be better to set these errors to a much higher severity level such as `HIGH`. What do you think?

You can handle diagnostic errors here:
https://github.com/Ericsson/codechecker/blob/9e33f21258f67f6426dcd30cf426e9bdeb6187b1/web/codechecker_web/shared/webserver_context.py#L33-L34

bruntib: Thanks for the feedback. However, I chose "critical" severity instead in case of compiler errors even if its severity could be equivalent with a warning's severity in GCC, indicating that further reports are missing from this file.

gyorb:


```diff

Complexity increasing per file
==============================
- web/codechecker_web/shared/webserver_context.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5251341&bid=17208316)

gyorb:


```diff

Complexity increasing per file
==============================
- web/codechecker_web/shared/webserver_context.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5251341&bid=17208316)

gyorb:


```diff

Complexity increasing per file
==============================
- web/codechecker_web/shared/webserver_context.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5251341&bid=17208316)

analyzer/codechecker_analyzer/analyzer_context.py
```diff
@@ -35,8 +35,11 @@ def __init__(self, *args, **kwargs):
 
     def __getitem__(self, key):
         # Key is not specified in the store and it is a compiler warning.
-        if key not in self.store and key.startswith('clang-diagnostic-'):
-            return "MEDIUM"
+        if key not in self.store:
+            if key == 'clang-diagnostic-error':
+                return "CRITICAL"
```
csordasmarton: Please update the documentation of this class (SeverityMap) above.

web/codechecker_web/shared/webserver_context.py
```diff
@@ -29,9 +29,13 @@ def __init__(self, *args, **kwargs):
...
 
     def __getitem__(self, key):
-        # Key is not specified in the store and it is a compiler warning.
-        if key not in self.store and key.startswith('clang-diagnostic-'):
-            return "MEDIUM"
+        # Key is not specified in the store and it is a compiler warning
+        # or error.
+        if key not in self.store:
+            if key == 'clang-diagnostic-error':
+                return "CRITICAL"
```
csordasmarton: Same as above, please update the docstring of this class.

analyzer/codechecker_analyzer/analysis_manager.py
```diff
@@ -335,6 +335,8 @@ def handle_failure(source_analyzer, rh, zip_file, result_base, actions_map):
 
     LOG.debug("ZIP file written at '%s'", zip_file)
 
+    rh.postprocess_result()
```
csordasmarton: Maybe it would be good to leave a comment here which will remember us why do we need to call this function on analysis failure too.

gyorb:


```diff

Complexity increasing per file
==============================
- web/codechecker_web/shared/webserver_context.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5251341&bid=17208316)

bruntib:
This is not a checker. Not even a warning which could be turned off. We are talking about a compilation error, like missing a semicolon, or something. Or do I misunderstand something in your request? Thanks.

csordasmarton:
@bruntib If I analyze a project, I can disable or enable checkers and warnings. This will be also a checker with the `clang-diagnostic-error` name and it will be enabled by default. It can be annoying to the user who wants to disable this checker but he can't. He will always see these results. If we do not want to allow to the user to disable these checkers we should document it somewhere. For example here: https://github.com/Ericsson/codechecker/blob/master/docs/analyzer/user_guide.md#toggling-checkers or in the help message of the CodeChecker analyze/check commands.

What do you think @gyorb ?

bruntib:
It is clear that all checkers and warnings can be disabled in a uniform manner, but compilation errors are different. We can instruct `clang-tidy` not to emit the messages of a specific checker or a specific warning (with `-w<warning_type>` flag). However, we can't instruct the compiler to hide a compilation error, and pretend the translation unit being compiled.

If we want to let the users disabling this kind of report, then we need to implement this logic separately in CodeChecker so it works as it did so far. In this case we have to explain the users why (s)he can't disable this report with `-d Werror` (since this flag means something else), but with another format which we should invent.
The other option as you mentioned is that this type of report can't be disabled. This would have the advantage that the user is aware of the fact that this TU is not compiled, so the whole analysis is missing on this file. Actually this is the reason of this feature's introduction. The disadvantage of this option is that reports from compiler errors become an exception in the handling of "checker disable" feature.

From this point I think it is a philosophical question, so let's discuss it :)

bruntib:
According to our discussion I extended the documentation with a description which says the `clang-diagnostic-error`s can't be disabled.

gyorb:


```diff

Complexity increasing per file
==============================
- web/codechecker_web/shared/webserver_context.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5251341&bid=17208316)

PR #2711, opened 2020-05-05
[analyzer][config] do not enable alpha iterator checkers in the profiles

gyorb:
Some iterator checkers can conflict and may cause an analyzer crash
they are not enabled by default in the sensitive and extreme profiles.


dkrupp:
I guess this patch can be abandoned because this will be the final solution
https://github.com/Ericsson/llvm-project/pull/4131

PR #2756, opened 2020-05-21
[fix] Bugfix in HTML generator in parse command

bruntib:
The checker name is needed so the hash function can compute a hash for the
report. If this checker name is not the member of the Report object then hash
generation crashes.


bruntib:
It is used here:
https://github.com/Ericsson/codechecker/blob/9ad0c68f13f43c50ffc36e6a1cc1cd10a6feb6a3/tools/codechecker_report_hash/codechecker_report_hash/hash.py#L291

This function is called in the next line after the modified line.

PR #2772, opened 2020-06-04
[fix][cmd] Faster diff command + utf8 bugfix

bruntib:
In case of plain text output the diff command queries the file contents from
the server. However the source code lines were queried to all reports again and
again. Of course always the same content was returned.

This commit also contains a minor bugfix: if a source file contains a special
character then the utf8 decoder function crashed.

### DIFF
--- a/web/server/codechecker_server/api/report_server.py
+++ b/web/server/codechecker_server/api/report_server.py
@@ -1722,7 +1722,7 @@ class ThriftRequestHandler(object):
                 sourcefile = session.query(File).get(lines_in_file.fileId)
                 cont = session.query(FileContent).get(sourcefile.content_hash)
                 lines = zlib.decompress(
-                    cont.content).decode('utf-8').split('\n')
+                    cont.content).decode('utf-8', 'ignore').split('\n')
                 for line in lines_in_file.lines:
                     content = '' if len(lines) < line else lines[line - 1]
                     if encoding == Encoding.BASE64:
--- a/web/client/codechecker_client/cmd_line_client.py
+++ b/web/client/codechecker_client/cmd_line_client.py
@@ -504,7 +504,7 @@ def handle_diff_results(args):
 
     context = webserver_context.get_context()
 
-    lines_in_files_requested = []
+    source_line_contents = {}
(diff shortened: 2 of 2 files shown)


web/server/codechecker_server/api/report_server.py
```diff
@@ -1722,7 +1722,7 @@ def getLinesInSourceFileContents(self, lines_in_files_requested, encoding):
                 sourcefile = session.query(File).get(lines_in_file.fileId)
                 cont = session.query(FileContent).get(sourcefile.content_hash)
                 lines = zlib.decompress(
-                    cont.content).decode('utf-8').split('\n')
+                    cont.content).decode('utf-8', 'replace').split('\n')
```
csordasmarton, line 1725: We use `decode` function in multiple places in this file. Don't we need to pass this extra second parameter (`'replace'`) everywhere?

bruntib: To be honest, I'm a little afraid of changing it at all occurrences. Even in this case I don't know if `'ignore'` should be better, because we use that at the other parts of the program. And sometimes it might be meaningful how to handle special characters, for example when we compute a bug hash from the file content at these lines.

gyorb: Yes it can cause hash mismatch if the hash is generated from this string after storage where `replace` is used compared to where it is usually done after analysis, I think ther the `get_line` function is used 
https://github.com/Ericsson/codechecker/blob/4e55e4c0398ca213bb9f3e3d40052addaeb60107/codechecker_common/util.py#L36 where characters with errors are ignored by default.

We talked about to use `replace` so at least the user can see if there are some non utf-8 valid characters.
If we want to follow that we can use `ignore` here too, but I do not think this should be used for hash generation.

gyorb:


```diff

Complexity increasing per file
==============================
- web/client/codechecker_client/cmd_line_client.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5639864&bid=18468839)

PR #2774, opened 2020-06-05
[server] Improvement in a slow query preventing timeout

bruntib:
Earlier the "CodeChecker cmd diff..." command assembled an SQL query which
contained case insensitive pattern matching on the bug hashes with potential
joker characters. The number of these pattern matches is proportional to the
number of reports. This query could cause a timeout in the underlying SQL
server. The solution is checking equality instead of pattern matching.

### DIFF
--- a/web/server/codechecker_server/api/report_server.py
+++ b/web/server/codechecker_server/api/report_server.py
@@ -211,8 +211,18 @@ def process_report_filter(session, report_filter):
         AND.append(or_(*OR))
 
     if report_filter.reportHash:
-        OR = [Report.bug_id.ilike(conv(rh))
-              for rh in report_filter.reportHash]
+        OR = []
+        no_joker = []
+
+        for rh in report_filter.reportHash:
+            if '*' in rh:
+                OR.append(Report.bug_id.ilike(conv(rh)))
+            else:
+                no_joker.append(rh)
+
+        if no_joker:
+            OR.append(Report.bug_id.in_(no_joker))
+
         AND.append(or_(*OR))
 
     if report_filter.severity:


gyorb:


```diff

Issues
======
- Added 2
           

Complexity decreasing per file
==============================
+ web/server/codechecker_server/api/report_server.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5646651&bid=18492184)

web/server/codechecker_server/api/report_server.py
```diff
@@ -211,9 +211,7 @@ def process_report_filter(session, report_filter):
         AND.append(or_(*OR))
 
     if report_filter.reportHash:
-        OR = [Report.bug_id.ilike(conv(rh))
-              for rh in report_filter.reportHash]
-        AND.append(or_(*OR))
+        AND.append(Report.bug_id.in_(report_filter.reportHash))
```
csordasmarton, line 214: @bruntib @gyorb Maybe we can extend the Thrift API ReportFilter data structure with a new field (e.g.: reportHashRgx) to search for bug ids by using regex expressions too, so on the UI we would be able to search not only full size hashes but substrings too. In case of diff we would use the exact match filter field. So reportHash field would do exact match and reportHashRgx would do regex match. What do you think?

gyorb: The full hash string is not printed on the UI right? It can be problematic to get the full hash only from the web UI.
Keeping the old behavior in some cases can be useful (Maybe in the CLI filters too?).

bruntib: What use case would a regex solve in terms of a bug hash (or any hash)? In case somebody is looking for a hash, then the whole hash is in his hands. Ok, sometimes it is more comfortable if the tool provides prefix search for a hash (for example git commands accept commit hash prefixes or docker commands accept image and container hash prefixes). But couldn't we delay this development until an explicit user request arrives?

bruntib:
@gyorb the hash length checking has been revoked from the GUI.

gyorb:


```diff

Complexity decreasing per file
==============================
+ web/server/codechecker_server/api/report_server.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5646651&bid=18492184)

web/server/codechecker_server/api/report_server.py
csordasmarton: For example in the GUI I do not see the whole hash by default. It will show only a substring of it (~10 characters):

image

I have to increase the column size to get the full hash.

For me this patch contains a backward compatibility change. We don't know how many user relies on this feature and how many users use this without a full hash. Maybe we should think about it to keep the old behavior as @gyorb said.

I am thinking on the following solution: if the first item in the `report_filter.reportHash` contains a `*` quantifier we will work as before (query with LIKE) otherwise we will query with IN. What do you think @bruntib?

bruntib: This is a good idea, I like it.

gyorb: I think the solution @csordasmarton mentioned should be ok @bruntib where it is checked if the hash starts with (maybe contains) a `*` or not.

gyorb:


```diff

Complexity decreasing per file
==============================
+ web/server/codechecker_server/api/report_server.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5646651&bid=18492184)

gyorb:


```diff

Complexity increasing per file
==============================
- web/server/codechecker_server/api/report_server.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5646651&bid=18492184)

gyorb:


```diff

Complexity increasing per file
==============================
- web/server/codechecker_server/api/report_server.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5646651&bid=18492184)

PR #2849, opened 2020-07-20
[travis] Allow macOS jobs to fail

whisperity:
To speed the build up as it seems something is lastingly broken with the script.
See https://docs.travis-ci.com/user/customizing-the-build#jobs-that-are-allowed-to-fail and https://docs.travis-ci.com/user/customizing-the-build#fast-finishing for the two options I added.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ services:
   - postgresql
 
 jobs:
+  fast_finish: true
   include:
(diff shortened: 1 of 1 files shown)


whisperity:
Cancel... `jobs.include` is **not** supported by `jobs.allow_failures`. Damn! Why is this so complicated to do...

PR #2872, opened 2020-07-29
[analyzer] collect implicit include paths in case of clang mismatch

gyorb:
If different clang binaries are used for compilation and analyisis
collect the implicit include paths from the clang doing the
compilation with can be used by the clang doing the analysis.

It is possible that the clang doing the compilation was compiled
differently (using std lib headers from a non standard place)
compared to the clang which is doing the analysis.
In this case the clang executing the analysus tries to use
the system headers from the host which might not be compatible
with the project which was built.

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -899,7 +899,8 @@ def parse_options(compilation_db_entry,
                   keep_gcc_include_fixed=False,
                   keep_gcc_intrin=False,
                   get_clangsa_version_func=None,
-                  env=None):
+                  env=None,
+                  analyzer_clang_version=None):
     """
     This function parses a GCC compilation action and returns a BuildAction
     object which can be the input of Clang analyzer tools.
@@ -925,6 +926,8 @@ def parse_options(compilation_db_entry,
                             get_clangsa_version_func(compiler_binary, env)
                             Should return false for a non clang compiler.
     env -- Is the environment where a subprocess call should be executed.
+    analyzer_clang_version -- version information about the clang which is
+                              used to execute the analysis
     """
     details = {
         'analyzer_options': [],
@@ -1003,11 +1006,17 @@ def parse_options(compilation_db_entry,
     ImplicitCompilerInfo.compiler_versions[details['compiler']] \
         = compiler_version_info
 
-    using_clang_to_compile_and_analyze = False
-    if ImplicitCompilerInfo.compiler_versions[details['compiler']]:
+    using_same_clang_to_compile_and_analyze = False
+    compiler_clang = \
+        ImplicitCompilerInfo.compiler_versions.get(details['compiler'])
+
+    if compiler_clang:
         # Based on the version information the compiler is clang.
-        using_clang_to_compile_and_analyze = True
         flag_processors = clang_flag_collectors
+        if compiler_clang.installed_dir == \
+                analyzer_clang_version.installed_dir:
+            LOG.debug("Same clang is used for compilation and analysis.")
+            using_same_clang_to_compile_and_analyze = True
 
     for it in OptionIterator(gcc_command[1:]):
         for flag_processor in flag_processors:
@@ -1054,7 +1063,7 @@ def parse_options(compilation_db_entry,
     # Store the compiler built in include paths and defines.
     # If clang compiler is used for compilation and analysis,
     # do not collect the implicit include paths.
-    if (not toolchain and not using_clang_to_compile_and_analyze) or \
+    if (not toolchain and not using_same_clang_to_compile_and_analyze) or \
             (compiler_info_file and os.path.exists(compiler_info_file)):
         ImplicitCompilerInfo.set(details, compiler_info_file)
 
@@ -1181,7 +1190,8 @@ def parse_unique_log(compilation_database,
                      analysis_skip_handler=None,
                      pre_analysis_skip_handler=None,
                      ctu_or_stats_enabled=False,
-                     env=None):
+                     env=None,
+                     analyzer_clang_version=None):
     """
     This function reads up the compilation_database
     and returns with a list of build actions that is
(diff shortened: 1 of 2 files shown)


gyorb:


```diff

Complexity increasing per file
==============================
- analyzer/codechecker_analyzer/buildlog/log_parser.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5929579&bid=19428670)

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -899,7 +899,8 @@ def parse_options(compilation_db_entry,
                   keep_gcc_include_fixed=False,
                   keep_gcc_intrin=False,
                   get_clangsa_version_func=None,
-                  env=None):
+                  env=None,
+                  analyzer_clang=None):
```
csordasmarton, line 903: Maybe `analyzer_clang_version` or simply `analyzer_version` would be a better name for this.

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -1181,7 +1189,8 @@ def parse_unique_log(compilation_database,
                      analysis_skip_handler=None,
                      pre_analysis_skip_handler=None,
                      ctu_or_stats_enabled=False,
-                     env=None):
+                     env=None,
+                     analyzer_clang=None):
```
csordasmarton, line 1193: Same as above.

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -867,6 +867,11 @@ def main(args):
     LOG.debug("args: %s", str(args))
     LOG.debug("Output will be stored to: '%s'", args.output_path)
 
+    clang_binary = \
+        context.analyzer_binaries.get(
+            clangsa.analyzer.ClangSA.ANALYZER_NAME)
+    analyzer_clang = clangsa.version.get(clang_binary, analyzer_env)
```
csordasmarton, line 873: Same as above: `analyzer_clang_version` or `analyze_version`.

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -1003,11 +1006,16 @@ def parse_options(compilation_db_entry,
...
+    using_same_clang_to_compile_and_analyze = False
+    compiler_clang = \
+        ImplicitCompilerInfo.compiler_versions.get(details['compiler'],
+                                                   False)
+    if compiler_clang:
         # Based on the version information the compiler is clang.
-        using_clang_to_compile_and_analyze = True
         flag_processors = clang_flag_collectors
+        if compiler_clang.installed_dir == analyzer_clang.installed_dir:
+            LOG.debug("Same clang is used for compilation and analysis.")
```
csordasmarton, line 1017: I logged my dummy project with `/usr/bin/clang` and analyzed with a different clang `...CodeChecker/cc_bin/clang` but I still see this debug message in my output.

csordasmarton:
There is some typo in the commit message:
- `doing the compilation with can` vs `doing the compilation which can`
- `executing the analysus` vs `executing the analysis`

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -1003,11 +1006,16 @@ def parse_options(compilation_db_entry,
     ImplicitCompilerInfo.compiler_versions[details['compiler']] \
         = compiler_version_info
 
-    using_clang_to_compile_and_analyze = False
-    if ImplicitCompilerInfo.compiler_versions[details['compiler']]:
+    using_same_clang_to_compile_and_analyze = False
+    compiler_clang = \
+        ImplicitCompilerInfo.compiler_versions.get(details['compiler'],
+                                                   False)
```
bruntib, line 1012: Function `get()` defaults to `None` if no second parameter given.

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -867,6 +867,11 @@ def main(args):
     LOG.debug("args: %s", str(args))
     LOG.debug("Output will be stored to: '%s'", args.output_path)
 
+    clang_binary = \
+        context.analyzer_binaries.get(
+            clangsa.analyzer.ClangSA.ANALYZER_NAME)
```
bruntib, line 872: Which `clang` specifically? What about `clang-tidy`? Do we assume that they are always the same version in the same directory?

gyorb: this version information is from the clang version which will be used for analysis (based on the config file values, package relative or from the PATH ...)
if clang-tidy has some problem during analysis because of the version mismatch (compiler clang, analyzer clang-tidy) probably there will be a problem with the analyzer clang too, this change just extends the collected information from the compiler clang

analyzer/codechecker_analyzer/buildlog/log_parser.py
gyorb: are the installDir values the same for both of the clang, is it possible that the built package still used the clang from the PATH, I did not experience this, but I will double check.

gyorb:


```diff

Complexity increasing per file
==============================
- analyzer/codechecker_analyzer/buildlog/log_parser.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=5929579&bid=19428670)

analyzer/codechecker_analyzer/cmd/analyze.py
whisperity: Probably, but it's not a requirement to run Clang SA-based analysis, or even have the `clang` binary installed.

whisperity:
Uh, this is a bit convoluted to wrap my head around, but I get what we are going for. Can we add a test for this? Even if inside the test, `clang` is a very mock shell script that just emits some appropriate outputs to some command-line options (using a faux version and a dummy hardcoded plist output?).

analyzer/codechecker_analyzer/cmd/analyze.py
gyorb: yes, but if clang is not installed it will not be logged and so the implicit include paths will not be collected for it, they will be collected for the used compiler and used if clang-tidy is executed

gyorb:
Yes, testing this is not that easy because a fake clang would be needed which returns a valid version information, fake implicit include paths, some supported and valid compilation target, and the supported c/c++ language versions.

csordasmarton (approved):
LGTM!

PR #2906, opened 2020-09-01
[test] Added simple tu_collector test.

balazske:
The test analyzes a project and collects the files with tu_collector.
Then these files are analyzed again in a new directory.

The goal is that the first step of analysis is not needed, tu_collector should generate a compiler_info.json file. Currently it is really not needed but if the compiler_info.json would be included then yes. Additionally, a next step is to make the test work with "failed" zip files too that are created during the analysis.

### DIFF
--- a/scripts/test/run_debug_tools_test.py
+++ b/scripts/test/run_debug_tools_test.py
@@ -1,0 +1,178 @@
+# -------------------------------------------------------------------------
+#
+#  Part of the CodeChecker project, under the Apache License v2.0 with
+#  LLVM Exceptions. See LICENSE for license information.
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# -------------------------------------------------------------------------
+"""
+Test the scripts in debug_tools.
+"""
+
+
+import argparse
+import tempfile
+import os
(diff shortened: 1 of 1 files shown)


gyorb:


```diff

Complexity increasing per file
==============================
- scripts/test/run_debug_tools_test.py  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=6105533&bid=19985380)

bruntib (changes requested):
I think it is a good idea extending the test suites of this tool. However, I have some concerns with this patch:
- This test is not in the nosetest infrastructure, and no "make" target belongs to it, so this  script is not run in any of our CI systems.
- The test is too special: the input of the test is a GitHub repository URL or a compressed source code that can be downloaded with `wget`. The build system is also hard-coded with the simplest cmake or autoconf build system calls. I think some test files should be committed which this can analyze instead.
- The test assumes that CodeChecker binary is available in the environment with all of its dependencies (a CTU-capable clang is among them).
- The test checks the number of error logs during ctu-collect and ctu-analyze phases. This is not in the scope of tu_collector.

PR #2927, opened 2020-09-09
[fix] handle case when the analyzer version is not set

gyorb:
If for some reason the analyzer version is not set
assume that the compiler used to compile the project
and the analyzer used for analysis are not the same.
If the analyzer version could not be collected for
some reason there could be problems
later during the analysis too.


csordasmarton (approved):
LGTM!

PR #2932, opened 2020-09-11
Revert "remove info logs because they can mess up the json and other outputs"

gyorb:
Reverts Ericsson/codechecker#2931
There should be no problem if these log messages are correctly sent to the stderr and do not mess up
the json and other output formats.


csordasmarton (approved):
LGTM!

PR #2947, opened 2020-09-20
Fix the anchor to fb infer

sylvestre:
https://github.com/Ericsson/codechecker/blob/master/tools/report-converter/README.md#facebook-infer
is the right link


csordasmarton (approved):
LGTM!

PR #2948, opened 2020-09-20
Add two new severities

sylvestre:
For altera-struct-pack-align, I reused the same severity as: 
"optin.performance.Padding":                                  "LOW",


csordasmarton (approved):
LGTM!

PR #2954, opened 2020-09-28
Set severity for cppcoreguidelines-prefer-member-initializer

sylvestre:
Reusing "modernize-use-default-member-init" severity but we could move it to LOW
as it has an impact on performance


csordasmarton (approved):
LGTM!

PR #2965, opened 2020-10-09
[refactor] reorder how the log levels are printed

### LINKED ISSUE #2922: Misleading bug levels
We display in our help output that the debug levels are:
```
```
This suggests the following relation between the levels:
```
info > debug > debug_analyzer
```
But the above relation does not hold because we define `debug_analyzer` to be 15, however, info is 20 and debug is 10 (https://docs.python.org/3/library/logging.html)
```
info:20 debug:10 debug_analyzer:15
```

So, users tend to send the most verbose debug output by setting `--verbose=debug_analyzer`, but that simply skips all `debug` logs.

Proposed solution: set debug_analyzer to 5.

**CodeChecker version**
v6.13.0

gyorb:
The log level verbosity goes from:
info -> debug_analyzer -> debug

The printed order should reflect that.

resolves #2922

### DIFF
--- a/codechecker_common/logger.py
+++ b/codechecker_common/logger.py
@@ -25,7 +25,7 @@ ERROR = logging.ERROR
 CRITICAL = logging.CRITICAL
 NOTSET = logging.NOTSET
 
-CMDLINE_LOG_LEVELS = ['info', 'debug', 'debug_analyzer']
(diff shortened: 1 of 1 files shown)


csordasmarton (approved):
LGTM!

PR #2970, opened 2020-10-13
[user_guide.md]: Add note about new bitbake integration option

dl9pf:
Add note about meta-codechecker layer with an integration as OpenEmbedded compatible layer.


gamesh411 (approved):
IMHO integration alternatives are always welcome. Thanks!

gyorb (approved):

PR #2971, opened 2020-10-13
[fix] Fix concurrency error by sharing config dict

gamesh411:
Wrap analyzer options dictionary in multiprocessing manager proxy object to
ensure safe sharing.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzer.py
@@ -294,6 +294,7 @@ def perform_analysis(args, skip_handler, context, actions, metadata_tool,
     manager = SyncManager()
     manager.start(__mgr_init)
 
+    config_map = manager.dict(config_map)
     actions_map = create_actions_map(actions, manager)
 
     # Setting to not None value will enable statistical analysis features.
--- a/analyzer/codechecker_analyzer/analysis_manager.py
+++ b/analyzer/codechecker_analyzer/analysis_manager.py
@@ -743,6 +743,11 @@ def start_workers(actions_map, actions, context, analyzer_config_map,
             # while map or map_async function is running.
             # It is a python bug, this does not happen if a timeout is
             # specified, then receive the interrupt immediately.
+
+            # FIXME: Ensure all shared data structures are wrapped in manager
+            #        proxy objects before passing them to other processes via
+            #        map_async.
+            #        Note that even deep-copying is known to be insufficient.
             pool.map_async(check,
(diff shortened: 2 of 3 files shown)


gamesh411:
This solution is basically a hack (and n order to remind ourselves to the fact, i have placed a FIXME comment), nevertheless this seems to solve a problem with massively parallel (-j100) runs involving statistics-based checkers.

The proper way to pass a data structure to map_async will have to be enforced by some solution, which does not seem straightforward to me ATM. I'll open an issue for the discussion.

analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -167,6 +167,16 @@ def perform_analysis(args, skip_handler, context, actions, metadata_tool,
     Additionally, insert statistical information into the metadata dict.
     """
 
+    # FIXME: Ensure all mutable data structures are wrapped in manager proxy
+    #        objects before passing the data structure to other processes via
+    #        map_async.
+    #        Note that (only) top-level immutability is not sufficient.
```
gyorb, line 173: Did you found any description about this?
For the config_map we could use the frozen version you mentioned because they should not be changed.

gyorb (changes requested):

analyzer/codechecker_analyzer/analyzer.py
gamesh411: We have tried the deep-copy approach, however it failed to remove the buggy behaviour. This could still mean that something else is also improperly shared, but for now I'll just reorganize the comments, and leave the dict-sharing method as is.

gamesh411: Also I have found no indication as to why it is wrong to simply shared non-wrapped data structures if they are read-only. The documentation indicates that changes are not reflected in the shared data structure (https://docs.python.org/2/library/multiprocessing.html#sharing-state-between-processes), mentions that locking could be important as well. Still, the semantics of read-only state sharing is murky. From experience, the whole 3.8.x branch seems to produce the undesired behaviour.

gyorb (commented):
The change LGTM but, we should not forget to check all the data shared between the processes.

gyorb (approved):

gyorb (changes requested):
@gamesh411 could you check why the tests failed

gamesh411:
Linux tests pass, I believe OSX ones fail because of a Travis limitation.

gyorb (approved):

PR #2983, opened 2020-10-26
[analyzer] Use maximum CPU resources by default during analysis

bruntib:


csordasmarton (changes requested):
Please check why analyzer related test cases are failed on travis.

csordasmarton (changes requested):
@bruntib As far as I see you didn't modified anything just rebased your pull request. I checked the previous [failure](https://travis-ci.org/github/Ericsson/codechecker/jobs/738982489) and it failed because the output of one test was the following:
```
[] - [1/2] clangsa analyzed simple1.cpp successfully.
[] - [1/2] clangsa analyzed simple1.cpp successfully.
```
instead of this:
https://github.com/Ericsson/codechecker/blob/cd94f9e3797ac0d13d493c7ebbacffcfc026e47b/analyzer/tests/functional/analyze_and_parse/test_files/simple1.deduplication.output#L9-L10
So the first value in the brackets was 1 in both cases.

I think we were lucky that the test cases are not failed this time but in some cases it will still fail. So we either need to run this test case with single thread or we should fix this bug to increment this value properly:
https://github.com/Ericsson/codechecker/blob/cd94f9e3797ac0d13d493c7ebbacffcfc026e47b/analyzer/codechecker_analyzer/analysis_manager.py#L644

bruntib:
I rebased the commit to master and Travis is green for me now except for GUI and OSX tests. I also checked the reason of previous failure and it is strange that numbering of build actions is bad. The purpose of this patch is using multiple CPUs for analysis by default and it really can cause bad numbering if that counter variable is incremented in different threads. It is even worse that in this case it would be a non-deterministic behavior, just like the order of build actions. So it is a miracle that functional tests are passing. Or does it mean that there is only 1 CPU in Travis environment for running tests? It would be a reasonable restriction of resources. Otherwise `-j1` must be given to every functional test. I'll check the number of CPUs on Travis and if it is 1 then this should be deterministically green.

gyorb (approved):

Issue #2802, opened 2020-06-29
format python code with black


gyorb:
Manual code formatting to make the code pep8 compliant can take a lot of time.
[black](https://github.com/psf/black) could be introduced to format automatically the python source code,
it can be configured easily and there is a support for a lot of [editors](https://github.com/psf/black/blob/master/docs/editor_integration.md) to automatically format the source code while writing it.

If the whole code was formatted in one commit, with a newer git the formatting commit can be left out from the git blame see the description [here](https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame).

Can be integrated into the travis-ci too to check if the source code was formatted right for a PR.

gamesh411:
I am all for automatic code formatting. In clang, I use clang-format manually and with various tools like arcanist, git-hooks. I find it a major convenience factor for me. I use a vim-plugin CoC which suggests the following auto-formatting solutions right now:
- autopep8
- black
- yapf

I am not for or against any of these in particular, but I know that autopep8 refuses to format certain cases, therefore it not a fully automatic solution. I haven't tried the other ones.
The ignore-rev-file option seems really handy 😉

bruntib:
Currently we use `pycodestyle` for checking formatting. The automatic formatting is not in the scope currently, but `pycodestyle` is integrated to the gating.

PR #2988, opened 2020-10-28
[web][client] handle exceptions from the server at the client side

gyorb:
If there is some problem on the server side or the
user does not have the proper access rights, handle
the exception returned from the server in the client,
and print some useful error messages.

### DIFF
--- a/web/client/codechecker_client/cmd_line_client.py
+++ b/web/client/codechecker_client/cmd_line_client.py
@@ -443,9 +443,8 @@ def check_deprecated_arg_usage(args):
 
 def get_run_data(client, run_filter, sort_mode=None,
                  limit=constants.MAX_QUERY_SIZE):
-    """
-    Get all runs based on the given run filter.
-    """
+    """Get all runs based on the given run filter."""
+
     all_runs = []
 
     offset = 0
@@ -458,6 +457,37 @@ def get_run_data(client, run_filter, sort_mode=None,
             break
 
     return all_runs
+
+
+def get_run_results(client,
+                    run_ids,
+                    limit,
+                    offset,
+                    sort_type,
+                    report_filter,
+                    compare_data,
+                    query_report_details):
+    """Get all the results with multiple api request.
+
+    In each api request get the limit ammount of reports.
+    Collect and return all the reports based on the filters.
+    """
+
+    all_results = []
+    while True:
+        results = client.getRunResults(run_ids,
+                                       limit,
+                                       offset,
+                                       sort_type,
+                                       report_filter,
+                                       compare_data,
+                                       query_report_details)
+        all_results.extend(results)
+        offset += limit
+        if len(results) < limit:
+            break
+
+    return all_results
 
 
 def validate_filter_values(user_values, valid_values, value_type):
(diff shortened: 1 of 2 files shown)


gyorb:


```diff

Complexity increasing per file
==============================
- web/client/codechecker_client/cmd_line_client.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/CodeChecker/codechecker/pullRequest?prid=6416584&bid=20789288)

web/client/codechecker_client/cmd_line_client.py
```diff
@@ -403,7 +403,13 @@ def check_run_names(client, check_names):
     run_filter.names = check_names
     run_filter.exactMatch = check_names is not None
 
-    run_info = {run.name: run for run in get_run_data(client, run_filter)}
+    run_info = {}
```
csordasmarton, line 406: I don't really like this change. We are handling exceptions which come from the server side in the [thrift_call.py](https://github.com/Ericsson/codechecker/blob/master/web/client/codechecker_client/thrift_call.py) file globally.

So in this file for example we will catch the `RequestFailed` exceptions, we print the message and raise the exception again, and we catch these exception again where we called the API functions.
Maybe it would be a better approach that after we print the error message we exit with an error code.
If an error happens, for example I am not able to fetch the runs from the server, I want to know it and if I am running these command for example in a Jenins job than the job will fail. And this is the expected behavior instead of working with false data.

In this patch sometimes we will exit with an error code and sometimes we will not (for example in this place if getting the run data is failed we will print an error message but the will not stop the program). I think this is not a good approach.

If the error messages are not expressive, we should fix these messages on the server side where we throw these exceptions and in the thrift_call.py file we should print these messages nicely and exit with an error code.

csordasmarton (changes requested):

web/client/codechecker_client/cmd_line_client.py
gyorb: I've updated the PR with the changes we discussed.

web/client/codechecker_client/thrift_call.py
```diff
@@ -58,14 +59,15 @@ def wrapper(self, *args, **kwargs):
                     codechecker_api_shared.ttypes.ErrorCode.UNAUTHORIZED:
                 LOG.error('Unauthorized to access\n %s',
                           str(reqfailure.message))
+                LOG.error('Ask the product admin for '
+                          'additional access rights.')
```
csordasmarton, line 63: Use as long as lines as it is possible.
```suggestion
                LOG.error('Ask the product admin for additional access '
                          'rights.')
```

csordasmarton (changes requested):
Just some tiny comments otherwise LGTM!

web/client/codechecker_client/cmd_line_client.py
```diff
@@ -1304,10 +1318,17 @@ def print_reports(client,
...
         os.makedirs(output_dir)
 
+    client = None
+    # We set up the client if we are not comparing two local report directory.
+    if basename_run_names or newname_run_names:
+        client = setup_client(args.product_url)
+
     if basename_local_dirs and newname_local_dirs:
-        reports = get_diff_local_dirs(basename_local_dirs, newname_local_dirs)
+        reports = get_diff_local_dirs(basename_local_dirs,
```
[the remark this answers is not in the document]
gyorb: We definitely should introduce some automatic formatting to not waste time with this reformatting tasks.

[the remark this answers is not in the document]
csordasmarton: Yes, I support it. You already created an issue for that #2802.

csordasmarton (approved):
LGTM!

PR #3028, opened 2020-11-17
[analyzer][fix] Unique lines when collecting statistics

baloghadamsoftware:
If a header file is included into multiple source files in a C code,
its lines are counted multiple times when processing the statistics.
This leads to false positives if e.g. the return value of a function
is used in a header but not in the source files. This patch fixes it
by collecting all lines to a set first and then processing the set.


csordasmarton (approved):
It's LGTM!

PR #3039, opened 2020-11-20
[analyzer][fix] Invocation list YAML config no longer break lines

gamesh411:
The config file 'invocation-list.yml' may contain line breaks inside
list items. However, this is not properly supported by the analyzer YAML
parser. This fix sets the line width parameter to very high value.

### DIFF
--- a/analyzer/tests/functional/ctu/test_ctu.py
+++ b/analyzer/tests/functional/ctu/test_ctu.py
@@ -10,10 +10,11 @@
 
 
 import glob
-import json
 import os
 import shutil
 import unittest
+
+from typing import IO
 
 from libtest import env
 from libtest.codechecker import call_command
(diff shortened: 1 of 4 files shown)


gamesh411:
Apart from fixing lint errors, I have a question:
Should the upgrade to `pathlib` be a separate patch (and that would then upgrade every use of `os.path` to `pathlib.Path` alternatives) or just do it here for only this test case (that would mean 1 patch, but 2 commits, one is where I upgrade this test file to `pathlib`, and another where I implement the new test case)?

gamesh411:
Updated with the bare version.

martong:
> Apart from fixing lint errors, I have a question:
> Should the upgrade to `pathlib` be a separate patch (and that would then upgrade every use of `os.path` to `pathlib.Path` alternatives) or just do it here for only this test case (that would mean 1 patch, but 2 commits, one is where I upgrade this test file to `pathlib`, and another where I implement the new test case)?

I am okay to have all this in one PR.

martong (approved):
LG! And thanks for the new test!

analyzer/tests/functional/ctu/test_ctu.py
```diff
@@ -228,3 +246,45 @@ def test_ctu_makefile_generation(self):
...
+    def test_ctu_ondemand_yaml_format(self):
+        """ Test the generated YAML used in CTU on-demand mode.
+        The YAML file should not contain newlines in individual entries in the
+        generated textual format. """
+
+        cmd = [self._codechecker_cmd, 'analyze',
+               '-o', self.report_dir,
+               '--analyzers', 'clangsa',
+               '--ctu-collect',  # ctu-directory is needed, and it remains
+                                 # intactonly if a single ctu-phase is
```
gyorb, line 261:
```suggestion
                                 # intact only if a single ctu-phase is
```

bruntib (approved):

bruntib:
Tests need to be fixed but apart from that it looks good.

analyzer/tests/functional/ctu/test_ctu.py
```diff
@@ -58,21 +61,36 @@ def setUp(self):
...
+        self.complex_buildlog = os.path.join(
+            self.test_workspace, 'complex_buildlog.json')
+
         # Fix the "template" build JSONs to contain a proper directory
         # so the tests work.
-        raw_buildlog = os.path.join(self.test_dir, 'buildlog.json')
-        with open(raw_buildlog,
-                  encoding="utf-8", errors="ignore") as log_file:
-            build_json = json.load(log_file)
+        def adjust_buildlog(buildlog_filename: str):
```
gyorb, line 70: I think this `adjust_buildlog` method could be useful for other tests too, or is this adjustment really specific for this test? Only the source file paths are updated.
There are already some helpers for the test environment here `codechecker/analyzer/tests/libtest/`

gyorb (changes requested):
Some small suggestions otherwise LGTM.

analyzer/tests/functional/ctu/test_ctu.py
gamesh411: Good Point!
I have put the utility into `codechecker/analyzer/tests/libtest/env.py`, it should be usable by other tests now. Another round will be necessary to check where this can be used.

analyzer/tests/functional/ctu/test_ctu.py
gamesh411: Thanks, fixed the typo ✍️ 🔨

gamesh411:
Just fixed typos removed the unused YAML import line, and extended the documentation text a bit in the last few pushes.

csordasmarton (approved):
LGTM!

PR #3058, opened 2020-11-26
[feat] Enable sending reports from windows

rasjani:
This patch allows me running CodeChecker store  from windows to  server running on  linux.

### DIFF
--- a/bin/CodeChecker
+++ b/bin/CodeChecker
@@ -39,7 +39,6 @@ def run_codechecker(checker_env, subcommand=None):
     common_lib = os.path.join(package_root,
                               'lib',
                               'python3')
-
     checker_env['PYTHONPATH'] = common_lib
 
     checker_env['CC_PACKAGE_ROOT'] = \
@@ -113,7 +112,8 @@ def main(subcommand=None):
         if proc_pid:
             os.kill(proc_pid, signal.SIGHUP)
 
-    signal.signal(signal.SIGHUP, signal_reload_handler)
+    if not sys.platform == "win32":
+        signal.signal(signal.SIGHUP, signal_reload_handler)
 
     try:
         run_codechecker(checker_env, subcommand)
--- a/codechecker_common/util.py
+++ b/codechecker_common/util.py
@@ -12,6 +12,8 @@ Util module.
 
 import json
 import os
+import ntpath
+import sys
 
 import portalocker
 
@@ -135,9 +137,13 @@ def trim_path_prefixes(path, prefixes):
     # Find the longest matching prefix in the path.
     longest_matching_prefix = None
     for prefix in prefixes:
-        if not prefix.endswith('/'):
-            prefix += '/'
-
+        if ntpath.sep in prefix:
+            prefix = prefix.replace("\\\\", "\\")
+            if not prefix.endswith(ntpath.sep):
+                prefix += ntpath.sep
+        else:
+            if not prefix.endswith('/'):
+                prefix += '/'
         if path.startswith(prefix) and (not longest_matching_prefix or
                                         longest_matching_prefix < prefix):
             longest_matching_prefix = prefix
@@ -148,3 +154,16 @@ def trim_path_prefixes(path, prefixes):
         return path
 
     return path[len(longest_matching_prefix):]
+
+
+def strip_drive_letter(full_path):
+    """Removes drive letter from path if it has one """
+    _, tail = ntpath.splitdrive(full_path)
+
+    return tail
+
+
+def convert_windows_sep_to_posix(full_path):
+    """Converts \\  to /"""
+
+    return full_path.replace(ntpath.sep, "/")
--- a/web/client/codechecker_client/cmd/store.py
+++ b/web/client/codechecker_client/cmd/store.py
@@ -439,12 +439,26 @@ def scan_for_review_comment(job):
     return comments
 
 
+def get_source_file_with_comments(jobs, zip_iter=map):
+    """
+    Get source files where there is any codechecker review comment at the main
+    report positions.
+    """
+    files_with_comment = set()
+
+    for job, comments in zip(jobs,
+                             zip_iter(scan_for_review_comment, jobs)):
+        file_path, _ = job
+        if comments:
+            files_with_comment.add(file_path)
+
+    return files_with_comment
+
+
 def filter_source_files_with_comments(source_file_info, main_report_positions):
     """Collect the source files where there is any codechecker review
     comment at the main report positions.
     """
-
-    files_with_comment = set()
     jobs = []
     for file_path, v in source_file_info.items():
         if not bool(v):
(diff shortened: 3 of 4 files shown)


codechecker_common/util.py
```diff
@@ -148,3 +155,18 @@ def trim_path_prefixes(path, prefixes):
         return path
 
     return path[len(longest_matching_prefix):]
+
+
+def strip_driveletter(full_path):
```
csordasmarton, line 160:
```suggestion
def strip_drive_letter(full_path):
```

csordasmarton (changes requested):

bin/CodeChecker
```diff
@@ -24,6 +24,10 @@ import tempfile
 proc_pid = None
 
 
+def is_windows():
```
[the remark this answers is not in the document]
rasjani: codechecker_common is not in search path at this stage, doing that would require moving `common_lib` (and few other vars) to be moved to outside of `run_codechecker()` and then adding `common_lib` to sys.path.. Thats ok ?

web/client/codechecker_client/cmd/store.py
```diff
@@ -559,6 +578,24 @@ def parse_report_files(report_files):
...
+                if report_file_info.store_it:
+                    files_to_compress.add(report_f)
+
+                source_file_info.update(source_in_reports.source_info)
+                changed_files = \
+                    changed_files | source_in_reports.changed_since_report_gen
+                main_report_positions.extend(
+                    report_file_info.main_report_positions)
+                missing_source_files = \
+                    missing_source_files | source_in_reports.missing
```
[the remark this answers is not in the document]
rasjani: Is usage of `nonlocal` allowed here ?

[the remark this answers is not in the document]
rasjani: i did submit a fix for this but, i'd assume things would be cleaner if i could use nonlocal in nested function to avoid passing and returning the variables back and forth to loop body implementation..

rasjani:
So, no comments ?

web/client/codechecker_client/cmd/store.py
```diff
@@ -455,12 +455,29 @@ def filter_source_files_with_comments(source_file_info, main_report_positions):
...
+
+    # Currently ProcessPoolExecutor fails completely in windows
+    # Reason is most likely combination of venv and fork() not
+    # being present in windows, so stuff like setting up
+    # PYTHONPATH in parent CodeChecker before store is executed
+    # are lost. Maybe
+    if util.is_windows():
+        for job, comments in zip(jobs, map(scan_for_review_comment, jobs)):
+            files_with_comment = \
+                    _populate_files(job, comments, files_with_comment)
```
csordasmarton, line 473: We are still duplication a couple of thing because of the `map vs executor.map` functions. Also we don't extend the set here, we reassign a new value in every iteration.
What do you think about the following approach:
```py
def get_source_file_with_comments(jobs, zip_iter=map):
    """
    Get source files where there is any codechecker review comment at the main
    report positions.
    """
    files_with_comment = set()

    for job, comments in zip(jobs,
                             zip_iter(scan_for_review_comment, jobs)):
        file_path, _ = job
        if comments:
            files_with_comment.add(file_path)

    return files_with_comment


def filter_source_files_with_comments(source_file_info, main_report_positions):
    """Collect the source files where there is any codechecker review
    comment at the main report positions.
    """
    jobs = []
    for file_path, v in source_file_info.items():
        if not bool(v):
            # missing file
            continue
        lines = [rep.line for rep in main_report_positions
                 if rep.filepath == file_path]

        jobs.append((file_path, lines))

    # Currently ProcessPoolExecutor fails completely in windows.
    # Reason is most likely combination of venv and fork() not
    # being present in windows, so stuff like setting up
    # PYTHONPATH in parent CodeChecker before store is executed
    # are lost.
    if sys.platform == "win32":
        return get_source_file_with_comments(jobs)
    else:
        with concurrent.futures.ProcessPoolExecutor() as executor:
            return get_source_file_with_comments(jobs, executor.map)
```

web/client/codechecker_client/cmd/store.py
```diff
@@ -542,23 +559,62 @@ def parse_report_files(report_files):
...
+                main_report_positions,
+                changed_files,
+                missing_source_files)
+
+    # Currently ProcessPoolExecutor fails completely in windows
+    # Reason is most likely combination of venv and fork() not
+    # being present in windows, so stuff like setting up
+    # PYTHONPATH in parent CodeChecker before store is executed
+    # are lost. Maybe
+    if util.is_windows():
```
csordasmarton, line 588: We can use the similar approach what I mentioned above:
```py
def parse_report_files(report_files, zip_iter=map):
    """Parse and collect source code information mentioned in a report file.

    Collect any mentioned source files wich are missing or changed
    since the report generation. If there are missing or changed files
    the report will not be stored.
    """

    files_to_compress = set()
    source_file_info = {}
    main_report_positions = []
    changed_files = set()
    missing_source_files = set()

    for report_f, v in zip(report_files,
                           zip_iter(parse_collect_plist_info,
                                    report_files)):

        report_file_info, source_in_reports = v

        if report_file_info.store_it:
            files_to_compress.add(report_f)

        source_file_info.update(source_in_reports.source_info)
        changed_files = \
            changed_files | source_in_reports.changed_since_report_gen
        main_report_positions.extend(
            report_file_info.main_report_positions)
        missing_source_files = \
            missing_source_files | source_in_reports.missing

    return (source_file_info,
            main_report_positions,
            files_to_compress,
            changed_files,
            missing_source_files)
```
On the call side:
```py
    if sys.platform == "win32":
        (source_file_info,
         main_report_positions,
         files_to_compress,
         changed_files,
         missing_source_files) = parse_report_files(report_files)
    else:
        with concurrent.futures.ProcessPoolExecutor() as executor:
            (source_file_info,
             main_report_positions,
             files_to_compress,
             changed_files,
             missing_source_files) = parse_report_files(report_files,
                                                        executor.map)
```

web/server/codechecker_server/api/report_server.py
```diff
@@ -2568,12 +2568,14 @@ def __store_source_files(self, source_root, filename_to_hash,
         file_path_to_id = {}
 
         for file_name, file_hash in filename_to_hash.items():
+            trimmed_file_path = util.trim_path_prefixes(file_name,
```
csordasmarton, line 2571: Why did you move trimming the file path up in the code? This way the `source_file_name` variable will refer to a file which doesn't exist.

csordasmarton (changes requested):
Please also check why travis test cases are failed.

rasjani:
> Please also check why travis test cases are failed.

Fixed pycodestyle target to work locally, however, travis still complains and that puzzles me:

```client/codechecker_client/cmd/store.py:965:34: E261 at least two spaces before inline comment

client/codechecker_client/cmd/store.py:965:80: E501 line too long (82 > 79 characters)
```

The file in my patch doesn't even have that many lines  .. something im missing ?

web/server/codechecker_server/api/report_server.py
rasjani: considering that drive is part of the absolute path, most people would assume that path prefix that should be trimmed also includes drive .. Say, i have my jenkins checks out the code to f:\jenkins\workspace\jobname\my_repo so moving the trim here made the generated zip file work and my expectation that im stripping from full path, not from "path that might reside on any drive on the box  .. did it cause things to fail somewhere ? at least things worked perfectly on windows - have about 200 runs on jenkins with source files uploaded properly..

I did change the last patch thought  but im pretty sure if thats how its needed - i need to fix the same usecase somewhere else ..

csordasmarton (changes requested):

csordasmarton:
@rasjani Unfortunately the test cases are failed. I also have some tiny comment to separate the logic on windows and on unix. If you fix those comments and the tests are passed we can merge this PR.

We really appreciate your time and that you are helping us to make CodeChecker better :blush:

rasjani:
Still not thrown this away -- i'll fix things eventually when i get some time allocated to this..

csordasmarton:
@rasjani For now we have pypi packages from CodeChecker and in the support PR I handled most of the things from this PR (#3251). You can now install CodeChecker on Windows system by using `pip install codechecker` command. Could you please verify it whether it works for you?

csordasmarton:
Closing this pull request as there have been no activity on it in a long time. Also most of the part of this code are in the CodeChecker already (see my previous comment).

PR #3169, opened 2021-01-28
[fix] Opt-in fix escape in logger in case of backslash, CR, LF

steakhal:
The build logger did not escape '\\', '\r' and '\n' correctly.
This patch fixes this issue, while preserving the old version.

By default the old version will be used, unless the
`CC_LOGGER_NEW_ESCAPING` environment variable is defined.

It is mentioned in the README.md.

I test both versions in the test files.
Also adds a few tips how to debug such cases with confidence.

### DIFF
--- a/analyzer/tools/build-logger/README.md
+++ b/analyzer/tools/build-logger/README.md
@@ -88,3 +88,12 @@ are provided to the compiler then the complete build action will be thrown
 away. This means that build actions which only perform linking will not be
 captured. We consider a file as object file if its extension is `.o`, `.so` or
 `.a`.
+
+### `CC_LOGGER_NEW_ESCAPING`
+If the environment variable is defined, we will use the *new* shell-escape mechanism
+when we generate the `compile_commands.json`.
+Use this option **if** without it the `CodeChecker log` command fails to produce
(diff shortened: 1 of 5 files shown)


analyzer/tools/build-logger/README.md
```diff
@@ -88,3 +88,10 @@ are provided to the compiler then the complete build action will be thrown
 away. This means that build actions which only perform linking will not be
 captured. We consider a file as object file if its extension is `.o`, `.so` or
 `.a`.
+
+### `CC_LOGGER_FALLBACK`
```
dkrupp, line 92: Please make the old mode the default and allow to switch the new mode with an env var.
It would be too risky to release this change with so short testing period and testing on production code.

In CodeChecker 6.16.0 we can make this new mode the default if it works without any issues and also solves the probelms of the users.

dkrupp (changes requested):

analyzer/tools/build-logger/README.md
steakhal: Updated accordingly.

bruntib (approved):

PR #3232, opened 2021-03-18
[cmd] Change exit codes

bruntib:
It is not possible to compute exit code 2 at "CodeChecker analyze"
command efficiently taking suppression into account. So according to the
new exit codes we indicate only the success and failure of analysis by
zero and non-zero exit codes respectively.

Moreover "diff" and "parse" commands have an exit status depending on
whether there is a resulting report.

### DIFF
--- a/analyzer/codechecker_analyzer/cmd/parse.py
+++ b/analyzer/codechecker_analyzer/cmd/parse.py
@@ -55,6 +55,12 @@ epilog_env_var = f"""
                          Default: {_severity_map_file}
 """
 
+epilog_exit_status = """
+0 - No report
+1 - CodeChecker error
+2 - At least one report emitted by an analyzer
+"""
+
 
 class PlistToPlaintextFormatter(object):
     """
@@ -392,6 +398,10 @@ printed by the `parse` command.""",
 Environment variables
 ------------------------------------------------
 {epilog_env_var}
+
+Exit status
+------------------------------------------------
+{epilog_exit_status}
 """,
 
         # Help is shown when the "parent" CodeChecker command lists the
@@ -905,3 +915,6 @@ def main(args):
                     "reports!", changed_files)
 
     os.chdir(original_cwd)
+
+    if report_count != 0:
+        sys.exit(2)
--- a/docs/analyzer/user_guide.md
+++ b/docs/analyzer/user_guide.md
@@ -506,9 +506,8 @@ https://github.com/Ericsson/codechecker/blob/master/docs/analyzer/report_identif
 
 Exit status
 ------------------------------------------------
-0 - Successful analysis and no new reports
+0 - Successful analysis
 1 - CodeChecker error
-2 - At least one report emitted by an analyzer and there is no analyzer failure
 3 - Analysis of at least one translation unit failed
 128+signum - Terminating on a fatal signal whose number is signum
 
@@ -1650,6 +1649,12 @@ Environment variables
                          generating gerrit output.
   CC_SEVERITY_MAP_FILE   Path of the checker-severity mapping config file.
                          Default: <package>/config/checker_severity_map.json
+
+Exit status
+------------------------------------------------
+0 - No report
+1 - CodeChecker error
+2 - At least one report emitted by an analyzer
 ```
 </details>
 
--- a/web/client/codechecker_client/cmd/cmd.py
+++ b/web/client/codechecker_client/cmd/cmd.py
@@ -1324,6 +1324,12 @@ envionment variables:
   CC_CHANGED_FILES    Path of changed files json from Gerrit. Use it when
                       generating gerrit output.
 
+Exit status
+------------------------------------------------
+0 - No difference between baseline and newrun
+1 - CodeChecker error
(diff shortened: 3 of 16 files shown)


analyzer/codechecker_analyzer/cmd/parse.py
```diff
@@ -55,6 +55,11 @@
                          Default: {_severity_map_file}
 """
 
+epilog_exit_status = """
+0 - Successful analysis
```
csordasmarton, line 59: Why the error code 1 is not here? The parse command can exit with this code multiple places:
```suggestion
0 - Successful analysis
1 - CodeChecker error
```
See: https://github.com/Ericsson/codechecker/blob/e02baa734ab58c696170c7289477b2ebee1d5b06/analyzer/codechecker_analyzer/cmd/parse.py#L651-L660

Please don't forget to update the docs below.

web/tests/functional/diff_local_remote/test_diff_local_remote.py
```diff
@@ -405,14 +405,9 @@ def test_set_env_diff_gerrit_output(self):
...
 
-        with self.assertRaises(subprocess.CalledProcessError):
-            out = get_diff_results([self._run_names[0]], [self._local_reports],
-                                   '--unresolved', 'gerrit',
-                                   ["--url", self._url, "-e", export_dir])
-
-            self.assertIn("'CC_REPO_DIR'", out)
-            self.assertIn("'CC_CHANGED_FILES'", out)
-            self.assertIn("needs to be set", out)
+        get_diff_results([self._run_names[0]], [self._local_reports],
```
csordasmarton, line 408: Why it is changed. It should raise an exception if the environment variables are not set. See: 
https://github.com/Ericsson/codechecker/blob/e02baa734ab58c696170c7289477b2ebee1d5b06/web/client/codechecker_client/cmd_line_client.py#L817-L819

zomen2 (approved):

web/tests/libtest/codechecker.py
```diff
@@ -96,8 +96,9 @@ def get_diff_results(basenames, newnames, diff_type, format_type=None,
     if extra_args:
         diff_cmd.extend(extra_args)
 
-    out = subprocess.check_output(
-        diff_cmd, encoding="utf-8", errors="ignore", env=cc_env)
+    out, _ = subprocess.Popen(
```
csordasmarton, line 99: Why do we have to change the original code here?

csordasmarton (changes requested):

web/tests/libtest/codechecker.py
bruntib: We had to change this, because `check_output()` throws an exception in case of non-zero exit code. In this commit we changed exit code to 2 if `CodeChecker cmd diff` results some difference in reports. However, even in this case we'd like to check the standard output and standard error.

csordasmarton (changes requested):
Test cases related to the web part are failed. Please check it why.

web/client/codechecker_client/cmd/cmd.py
```diff
@@ -1324,6 +1324,11 @@ def add_arguments_to_parser(parser):
   CC_CHANGED_FILES    Path of changed files json from Gerrit. Use it when
                       generating gerrit output.
 
+Exit status
+------------------------------------------------
+0 - Successful analysis
```
csordasmarton, line 1329: Why the 1 error code is missing from here?
The diff command can exit with 1 error code in case of CodeChecker error:
https://github.com/Ericsson/codechecker/blob/1617b22cedcc76f0d5bf6e2a3fcc5cd58d1ca56f/web/client/codechecker_client/cmd_line_client.py#L815
https://github.com/Ericsson/codechecker/blob/1617b22cedcc76f0d5bf6e2a3fcc5cd58d1ca56f/web/client/codechecker_client/cmd_line_client.py#L819
https://github.com/Ericsson/codechecker/blob/1617b22cedcc76f0d5bf6e2a3fcc5cd58d1ca56f/web/client/codechecker_client/cmd_line_client.py#L1313

Don't forget to update the userguide too.

csordasmarton (changes requested):

csordasmarton (approved):
LGTM!

docs/analyzer/user_guide.md
```diff
@@ -1650,6 +1649,12 @@ Environment variables
                          generating gerrit output.
   CC_SEVERITY_MAP_FILE   Path of the checker-severity mapping config file.
                          Default: <package>/config/checker_severity_map.json
+
+Exit status
+------------------------------------------------
+0 - Successful analysis
```
dkrupp, line 1655: Shouldn't we say instead?:
0: No report.

web/client/codechecker_client/cmd/cmd.py
```diff
@@ -1324,6 +1324,12 @@ def add_arguments_to_parser(parser):
...
```
dkrupp, line 1329: Shouldnt we say here?
0: no difference between baseline and newrun

dkrupp (changes requested):
Small comments to the help text.

PR #3233, opened 2021-03-19
Checker label

bruntib:


analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -255,15 +266,60 @@ def format_guideline(guideline):
...
+            header = ['Label']
+            rows = [[label] for label in cl.labels()]
+
+        if args.output_format in ['csv', 'json']:
+            header = list(map(uglify, header))
+
+        print(twodim.to_str(args.output_format, header, rows))
+        return
+
+    # TODO: --profile is a deprecated flag. This section should be removed in
```
csordasmarton, line 313: We should write a warning message if this option is still used to identify that it will be removed in the next release.

analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -214,31 +214,23 @@ def initialize_checkers(self,
...
-            if identifier.startswith('guideline:'):
-                guideline_name = identifier[len('guideline:'):]
-                if guideline_name not in guideline_map.available_guidelines():
-                    LOG.error('No such guideline: %s', guideline_name)
-                    sys.exit(1)
-                for checker in guideline_map.by_guideline(guideline_name):
-                    self.set_checker_enabled(checker, enabled)
-            elif identifier in profile_map.available_profiles():
+            elif identifier in checker_labels.get_constraint(
+                    'profile', 'choice'):
```
csordasmarton, line 221: We are in a for loop, so this `get_constraint` function call will be called multiple times but it will return the same value every time. So we should call it only once.

analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -214,31 +214,23 @@ def initialize_checkers(self,
...
                 if identifier in reserved_names:
                     LOG.warning("Profile name '%s' conflicts with a "
                                 "checker(-group) name.", identifier)
-                for checker in profile_map.by_profile(identifier):
+                for checker in checker_labels.checkers_by_labels(
+                        ['profile:' + identifier]):
```
csordasmarton, line 226:
```suggestion
                        [f'profile:{identifier}']):
```

analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -214,31 +214,23 @@ def initialize_checkers(self,
...
                     self.set_checker_enabled(checker, enabled)
-            elif identifier in guideline_map.available_guidelines():
+            elif identifier in checker_labels.occurring_values('guideline'):
```
csordasmarton, line 228: Same as above, it will be called multiple times.

analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -214,31 +214,23 @@ def initialize_checkers(self,
...
                 if identifier in reserved_names:
                     LOG.warning("Guideline name '%s' conflicts with a "
                                 "checker(-group) name.", identifier)
-                for checker in guideline_map.by_guideline(identifier):
+                for checker in checker_labels.checkers_by_labels(
+                        ['guideline:' + identifier]):
```
csordasmarton, line 233:
```suggestion
                        [f'guideline:{identifier}']):
```

codechecker_common/checker_labels.py
```diff
@@ -0,0 +1,219 @@
...
+        Return the list of available labels.
+        """
+        return self.__data['constraints'].keys()
+
+    def occurring_values(self, label):
+        """
+        Return the list of values belonging to the given label which were used
+        for at least one checker.
+        """
+        result = set()
```
csordasmarton, line 212:
```suggestion
        values = set()
```

codechecker_common/checker_labels.py
```diff
@@ -0,0 +1,219 @@
...
+            lambda lab: lab[0] == label,
+            self.labels_of_checker(checker))
+
+        if self.__data['constraints'].get(label, {}).get('unique'):
+            try:
+                return next(labels)[1]
+            except Exception:
+                return self.__data['constraints'].get(label, {}).get('default')
+        else:
+            return map(lambda _, value: value, labels)
```
csordasmarton, line 167:
```suggestion
        return map(lambda _, value: value, labels)
```

codechecker_common/checker_labels.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def checkers_by_labels(self, filter_labels):
+        """
+        Returns a list of checkers that have at least one of the specified
+        labels.
+
+        filter_labels -- A string list which contains labels with specified
+                         values. E.g. ['profile:default', 'severity:high'].
+        """
+        result = []
```
csordasmarton, line 135:
```suggestion
        checkers = []
```

csordasmarton (changes requested):

analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -94,20 +94,14 @@ def get_argparser_ctor_args():
...
-  CC_GUIDELINE_MAP_FILE  Path of the checker-guideline mapping config file.
-                         Default: '{}'
-  CC_PROFILE_MAP_FILE    Path of the checker-profile mapping config file.
-                         Default: '{}'
-""".format(os.path.join(package_root, 'config', 'checker_profile_map.json'),
-           os.path.join(package_root, 'config', 'checker_severity_map.json'),
-           os.path.join(package_root, 'config', 'checker_guideline_map.json'),
-           os.path.join(package_root, 'config', 'checker_profile_map.json')),
+""".format(os.path.join(package_root, 'config', 'checker_labels.json'),
+           os.path.join(package_root, 'config', 'checker_labels.json')),
```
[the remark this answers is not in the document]
bruntib: This has to be substituted to two places, but the comment is valid. I'll use `{0}` as a placeholder, thanks.

config/checker_labels.json
```diff
@@ -0,0 +1,2682 @@
...
+      }
+    },
+    "guideline": {
+      "choice": {
+        "sei-cert": "https://wiki.sei.cmu.edu/confluence/display/seccode"
+      }
+    },
+    "sei-cert": {
+
+    },
```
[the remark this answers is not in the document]
bruntib: My plan was to make it mandatory to list available labels in "constraints" section even if they're empty. This helps against the typos that you mentioned in another comment.

config/checker_labels.json
```diff
@@ -0,0 +1,2682 @@
...
+      },
+      "unique": true,
+      "default": "UNSPECIFIED"
+    }
+  },
+
+  "labels": {
+    "abseil-duration-addition": [
+
+    ],
```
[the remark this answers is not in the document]
bruntib: Yes, it would be good :)

docs/analyzer/user_guide.md
```diff
@@ -1,40 +1,47 @@
...
-    * [`parse`](#parse)
-        * [Exporting source code suppression to suppress file](#suppress-file)
-    * [`fixit`](#fixit)
-    * [`checkers`](#checkers)
-    * [`analyzers`](#analyzers)
- * [`Configuring Clang version`](#clang_version)
- * [Source code comments (review status)](#source-code-comments)
-    * [Supported formats](#supported-formats)
+- [Table of Contents](#table-of-contents)
+- [Easy analysis wrappers <a name="easy-analysis-wrappers"></a>](#easy-analysis-wrappers-)
```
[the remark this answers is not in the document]
bruntib: Sorry, this was an automatic change of my editor. I'll undo it.

codechecker_common/checker_labels.py
```diff
@@ -0,0 +1,219 @@
...
+
+        for checker, labels in self.__data['labels'].items():
+            labels = set(map(self.__get_label_key_value, labels))
+
+            if labels.intersection(filter_labels):
+                result.append(checker)
+
+        return result
+
+    def label_of_checker(self, checker, label):
```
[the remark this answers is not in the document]
bruntib: I'm not so sure about this. Labels are rather untyped things. It is just a historical thing that we use severity and profile for a long type. As for avoiding typos, listing labels in the "constraints" section is mandatory, so it could be checked if a valid label name was given here.

config/checker_labels.json
```diff
@@ -0,0 +1,2682 @@
...
+    "severity": {
+      "choice": {
+        "STYLE": "A true positive indicates that the source code is against a specific coding guideline or could improve readability.",
+        "LOW": "A true positive indicates that the source code is hard to read/understand or could be easily optimized.",
+        "MEDIUM": "A true positive indicates that the source code that may not cause a run-time error (yet), but against intuition and hence prone to error.",
+        "HIGH": "A true positive indicates that the source code will cause a run-time error.",
+        "CRITICAL": "Currently used for indicating compilation errors.",
+        "UNSPECIFIED": "Checker severity is not specified for a checker."
+      },
+      "unique": true,
```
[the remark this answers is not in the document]
bruntib: Because the severity of a checker must be unique. If this constraint is violated, then CodeChecker throws an error in the very beginning. By the way you can use "required" constraint too. That would be useful if you want to force providing severities for each checker.

analyzer/codechecker_analyzer/analyzers/config_handler.py
bruntib: This is proportional to the number of `--enable`/`--disable` flags and this is a constant operation. I don't think if there is any measurable difference, but if you insist, I can factor it out.

codechecker_common/checker_labels.py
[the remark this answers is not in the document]
bruntib: Done

config/checker_labels.json
```diff
@@ -0,0 +1,2682 @@
...
+      "default": "UNSPECIFIED"
+    }
+  },
+
+  "labels": {
```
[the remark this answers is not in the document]
bruntib: @dkrupp suggested another format: inserting analyzer names before checker names:
```.json
"labels": {
  "clang-tidy:abseil-string-find-startswith": [
    "severity:STYLE"
  ]
}
```
Which one should we use?

csordasmarton:
@bruntib don't forget to update the new config file with these changes: #3256.

config/checker_labels.json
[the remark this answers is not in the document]
dkrupp: I agrued with prefixed checker names because it can prevent name clashes in the command line too such as

CodeChecker analyze -e clangsa:security-checker -e clang-tidy:security-chekcer
we could use short names if there is no clashing
CodeChecker analyze -e unique-tidy-checker -e unique-clangsa-checker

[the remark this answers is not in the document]
dkrupp: but actually it could be defined in this file as Marci suggested, because it is shorter and enable the naming prefixing in the command line if the checker names would be clashing

codechecker_common/tests/unit/test_gerrit_converter.py
```diff
@@ -21,7 +21,21 @@
...
+                    return 'LOW'
+                else:
+                    # This assertion warns when a new test-case in the future
+                    # intends to query the severity of another checker. The
+                    # original behavior of this function is to return the
+                    # defult 'UNSPECIFIED' value by defult when the severity is
+                    # not provided in the config file.
+                    assert False, \
+                        'Currently no test-case quieries other labels for ' \
+                        'other checkers.'
```
csordasmarton, line 36: We can save an indentation here:
```suggestion
                if checker == 'my_checker':
                    return 'LOW'

                # This assertion warns when a new test-case in the future
                # intends to query the severity of another checker. The
                # original behavior of this function is to return the
                # defult 'UNSPECIFIED' value by defult when the severity is
                # not provided in the config file.
                assert False, \
                    'Currently no test-case quieries other labels for ' \
                    'other checkers.'
```

config/config.md
```diff
@@ -1,12 +1,20 @@
 
 # Package configuration
 
-### Checker severity map
-checker_severity_map.json file contains a mapping between a
-checker name and a severity level. Severity levels can be found in the
-codechecker_api_shared.thrift file.
+### Checker labels
+`checker_labels.json` file contains a labels to the checkers which describe
```
csordasmarton, line 5: This file no longer exists.

docs/analyzer/user_guide.md
```diff
@@ -348,13 +348,13 @@ checker configuration:
...
+  Checker labels
   ------------------------------------------------
-  In CodeCheckers there is a manual grouping of checkers. These groups are
-  called profiles. The collection of profiles is found in
-  config/checker_profile_map.json file. The goal of these profile is that you
-  can enable or disable checkers by these profiles. See the output of
-  "CodeChecker checkers --profile list" command.
+  In CodeChecker there is a manual grouping of checkers. These groups are
+  determined by labels. The collection of labels is found in
+  config/checker_labels.json file. The goal of these labels is that you can
```
csordasmarton, line 355: This file is removed, so the documentation need to be updated here. Please search this file name, I think it occures more places in the code.

tools/plist_to_html/plist_to_html/PlistToHtml.py
```diff
@@ -213,7 +213,9 @@ def create(self, output_path: str, report_data: ReportData):
         # Add severity levels for reports.
         for report in report_data['reports']:
             checker = report['checkerName']
-            report['severity'] = self._severity_map.get(checker, 'UNSPECIFIED')
+            report['severity'] = 'UNSPECIFIED'
+            if self._checker_labels:
+                report['severity'] = self._checker_labels.severity(checker)
```
csordasmarton, line 218: Lets create a function to get severity level for a cheker in this class and use it in both places:
```py
def get_severity(self, checker):
    """ TODO """
    if not self._checker_labels:
        return 'UNSPECIFIED'

    return self._checker_labels.severity(checker)

or

def get_severity(self, checker):
    """ TODO """
    return self._checker_labels.severity(checker) \
        if self._checker_labels else 'UNSPECIFIED'
```

web/client/codechecker_client/cmd/cmd.py
```diff
@@ -1492,6 +1507,17 @@ def add_arguments_to_parser(parser):
     importer.set_defaults(func=cmd_line_client.handle_import)
     __add_common_arguments(importer)
 
+    labels = subcommands.add_parser(
+        'labels',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        description="Checkers can be assigned properties by labels. This "
+                    "command lists these and store them to the server so "
+                    "filtering is possible based on these labels.",
```
csordasmarton, line 1515: Are you sure that this is true? I think now we can't store these labels to the server and we can't filter on them.

web/client/tests/unit/test_report_converter.py
```diff
@@ -41,8 +41,23 @@ def test_Report_to_ReportData(self):
...
+                    return 'LOW'
+                else:
+                    # This assertion warns when a new test-case in the future
+                    # intends to query the severity of another checker. The
+                    # original behavior of this function is to return the
+                    # defult 'UNSPECIFIED' value by defult when the severity is
+                    # not provided in the config file.
+                    assert False, \
+                        'Currently no test-case quieries other labels for ' \
+                        'other checkers.'
```
csordasmarton, line 56:
```suggestion
                if checker == check_name:
                    return 'LOW'

                # This assertion warns when a new test-case in the future
                # intends to query the severity of another checker. The
                # original behavior of this function is to return the
                # defult 'UNSPECIFIED' value by defult when the severity is
                # not provided in the config file.
                assert False, \
                    'Currently no test-case quieries other labels for ' \
                    'other checkers.'
```

config/labels/label_descriptions.json
```diff
@@ -0,0 +1,20 @@
+{
```
csordasmarton, line 1: I think the `label` and `checker_labels` keywords are a little bit redundant for me because we are in the `labels` directory. What do you think on the following structure:
```sh
|- labels
  |- analyzers
    |- clangsa.json
    |- clang-tidy.json
  |- descriptions.json
```

web/client/codechecker_client/cmd_line_client.py
```diff
@@ -1720,3 +1720,32 @@ def handle_import(args):
         sys.exit(1)
 
     LOG.info("Import successful!")
+
+
+def handle_labels(args):
+    if 'list' in args:
+        checker_labels = CheckerLabels(
+            os.path.join(os.environ['CC_PACKAGE_ROOT'], 'config', 'labels'))
```
csordasmarton, line 1728: I am getting an exception when I run the following command: `CodeChecker cmd labels --list`
```
  File "~/codechecker/build/CodeChecker/lib/python3/codechecker_common/cli.py", line 175, in main
    sys.exit(args.func(args))
  File "~/codechecker/build/CodeChecker/lib/python3/codechecker_client/cmd_line_client.py", line 1728, in handle_labels
    os.path.join(os.environ['CC_PACKAGE_ROOT'], 'config', 'labels'))
  File "/usr/lib/python3.6/os.py", line 669, in __getitem__
    raise KeyError(key) from None
KeyError: 'CC_PACKAGE_ROOT
```

```suggestion
            os.path.join(os.environ['CC_DATA_FILES_DIR'], 'config', 'labels'))
```

After fixing it I will get another exception:
```
  File "~/codechecker/build/CodeChecker/lib/python3/codechecker_common/cli.py", line 175, in main
    sys.exit(args.func(args))
  File "~/codechecker/build/CodeChecker/lib/python3/codechecker_client/cmd_line_client.py", line 1746, in handle_labels
    rows = [[label] for label in checker_labels.labels()]
AttributeError: 'CheckerLabels' object has no attribute 'labels'
```

It would be good to add a cmdline test case which calls this command and test whether it returns with a correct output or not.

csordasmarton (changes requested):

codechecker_common/checker_labels.py
```diff
@@ -0,0 +1,207 @@
...
+                    c = next(filter(
+                        lambda c: checker.startswith(c),
+                        iter(checkers.keys())), None)
+
+                labels.extend(
+                    map(self.__get_label_key_value, checkers.get(c, [])))
+
+        return labels
+
+    def get_description(self, label):
```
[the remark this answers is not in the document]
bruntib: I was thinking of this, if it's just a question of style or does it have any pros/cons? I found that by transforming it to a property leaks the data representation, so the user of that property needs to know the structure of the underlying multilevel dictionary. On the other hand if we remove the function's parameter, it becomes less verbose and carries less information about what key needs to be given to the function. So I found these reasons supporting the current implementation. I can't see the advantages of a property yet, but if you are aware of any, please, share it. I think it would be a relevant question in the future too, because we already use a lot of properties at other places and I'm not sure if they're reasonable. It would be interesting and useful to define some rule of thumb about when to use properties.

web/client/codechecker_client/cmd/cmd.py
```diff
@@ -1255,6 +1255,21 @@ def __register_del(parser):
     __add_common_arguments(del_t, needs_product_url=False)
 
 
+def __register_labels(parser):
```
csordasmarton, line 1258: If I print the help message of this command it will have a --url option:
```common arguments:
  --url PRODUCT_URL     The URL of the product which will be accessed by the
                        client, in the format of
                        '[http[s]://]host:port/Endpoint'. (default:
                        localhost:8001/Default)
```
Why do we need this? For now I think we don't use it, so it can be removed. Later if a subcommand will require it, we will introduce it.

csordasmarton (changes requested):

analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -255,15 +252,37 @@ def format_guideline(guideline):
...
+        for label in labels:
+            if label[0] in guidelines:
+                result[label[0]].append(label[1])
+
+        return result
+
+    cl = context.checker_labels
+
+    # TODO: --profile is a deprecated flag. This section should be removed in
+    # the next release.
```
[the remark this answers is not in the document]
bruntib: This comment refers to the code lines under it, not the `--profile` flag. I would leave this TODO message here, because removal of a deprecated flag doesn't trigger the modification of this part.

codechecker_common/checker_labels.py
```diff
@@ -0,0 +1,261 @@
+import os
+from collections import defaultdict
+from itertools import chain
+from typing import Dict, Iterable, List, Tuple, Union
+from codechecker_common.util import load_json_or_empty
+
+
+class CheckerLabels:
+    # These labels should be unique by each analyzer. If this uniqueness is
```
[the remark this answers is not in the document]
bruntib: I didn't mean a verb here, because not the analyzers make them unique. I really wanted to write that the labels should be unique. Maybe the continuation is bad: "should be unique per analyzer"?

codechecker_common/checker_labels.py
```diff
@@ -0,0 +1,261 @@
...
+        label_json_files = map(
+            lambda f: os.path.join(checker_labels_dir, 'analyzers', f),
+            label_json_files)
+
+        self.__data = self.__union_label_files(label_json_files)
+        self.__check_json_format(self.__data)
+
+    def __union_label_files(
+            self,
+            label_files: Iterable[str]) -> Dict[str, Dict[str, List[str]]]:
```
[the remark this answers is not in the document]
bruntib: Sure, I like it.

codechecker_common/checker_labels.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        A label has a value separated by colon (:) character, e.g:
+        "severity:high". This function returns this key and value as a tuple.
+        Optional whitespaces around (:) and at the two ends of this string are
+        not taken into account.
+        """
+        pos = key_value.index(':')
+        return key_value[:pos].strip(), key_value[pos + 1:].strip()
+
+    def __check_json_format(self, data: dict) -> bool:
```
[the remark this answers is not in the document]
bruntib: This was intentional, because `valuetype` would be too complex and the goal of this function is to check exactly this format in its whole depth.

analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -154,20 +154,19 @@ def initialize_checkers(self,
...
 
-        profile_map = analyzer_context.profile_map
-        guideline_map = analyzer_context.guideline_map
+        checker_labels = analyzer_context.checker_labels
 
         if 'profile:list' in map(itemgetter(0), cmdline_enable):
             LOG.error("'list' is a reserved profile keyword. ")
             LOG.error("Please choose another profile name in "
-                      "%s/config/checker_profile_map.json and rebuild.",
+                      "%s/config/checker_labels.json and rebuild.",
```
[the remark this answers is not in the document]
csordasmarton: Please also check other places as well:
```suggestion
                      "%s/config/labels/descriptions.json and rebuild.",
```

analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -154,20 +154,19 @@ def initialize_checkers(self,
...
                       analyzer_context.data_files_dir_path)
             sys.exit(1)
 
         if 'guideline:list' in map(itemgetter(0), cmdline_enable):
             LOG.error("'list' is a reserved guideline keyword. ")
             LOG.error("Please choose another guideline name in "
-                      "%s/config/checker_guideline_map.json and rebuild.",
+                      "%s/config/checker_labels.json and rebuild.",
```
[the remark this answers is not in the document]
csordasmarton:
```suggestion
                      "%s/config/labels/descriptions.json and rebuild.",
```

analyzer/codechecker_analyzer/analyzers/config_handler.py
csordasmarton: My comment is still marked as resolved but it is not resolved.

analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -202,217 +206,273 @@ def add_arguments_to_parser(parser):
...
-                                                        analyzer_environment)
-            if not configs:
-                analyzer_failures.append(analyzer)
-                continue
+        if 'label' in args:
+            profile_checkers = [(label, True) for label in args.label]
 
-            rows.extend((':'.join((analyzer, c[0])), c[1]) if 'details' in args
-                        else (':'.join((analyzer, c[0])),) for c in configs)
+        # TODO: Guideline
```
csordasmarton, line 272: Do not we want to implement it?

analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -202,217 +206,273 @@ def add_arguments_to_parser(parser):
...
+            result[label[0]].append(label[1])
+
+    return result
 
-    # If the given output format is not 'table', redirect logger's output to
-    # the stderr.
-    logger.setup_logger(args.verbose if 'verbose' in args else None,
-                        None if args.output_format == 'table' else 'stderr')
 
+def __get_detailed_checker_info(args, cl):
```
csordasmarton, line 234: Docstrings are missing for newly introduced functions and also add type hints where it is possible in this file. Please also do this everywhere in this file where you introduced a new function.

analyzer/tests/unit/test_checker_handling.py
```diff
@@ -21,35 +21,31 @@
...
-            return 'profile1'
-
-        def by_profile(self, profile):
-            if profile == 'default':
+    class CheckerLabels:
+        def checkers_by_labels(self, labels):
+            if labels[0] == 'profile:default':
                 return ['core', 'deadcode', 'security.FloatLoopCounter']
-            elif profile == 'security':
+            elif labels[0] == 'profile:security':
```
csordasmarton, line 28: Can you please add a test case for severity levels too?

codechecker_common/checker_labels.py
```diff
@@ -0,0 +1,262 @@
...
+                if labels.intersection(filter_labels):
+                    collection.append(checker)
+
+        return collection
+
+    def label_of_checker(
+            self,
+            checker: str,
+            label: str,
+            analyzers: Iterable[str] = None) -> Union[str, Iterable[str]]:
```
csordasmarton, line 158: I think this one is more readable:
```suggestion
    def label_of_checker(
        self,
        checker: str,
        label: str,
        analyzers: Iterable[str] = None
    ) -> Union[str, Iterable[str]]:
```
We use this form in other places, so we should use it everywhere. Please change other places too in this file.

config/labels/descriptions.json
```diff
@@ -0,0 +1,20 @@
...
+    "portability": "Checks that aim to detect code issues emerging from platform differences (eg. between 32-bit and 64-bit architectures).",
+    "extreme": "Sensitive checks + more comprehensive checks with a manageable false positive rate."
+  },
+  "severity": {
+    "STYLE": "A true positive indicates that the source code is against a specific coding guideline or could improve readability.",
+    "LOW": "A true positive indicates that the source code is hard to read/understand or could be easily optimized.",
+    "MEDIUM": "A true positive indicates that the source code that may not cause a run-time error (yet), but against intuition and hence prone to error.",
+    "HIGH": "A true positive indicates that the source code will cause a run-time error.",
+    "CRITICAL": "Currently used for indicating compilation errors.",
+    "UNSPECIFIED": "Checker severity is not specified for a checker."
```
csordasmarton, line 15: Can we change the order of the keys to this: CRITICAL, HIGH, MEDIUM, LOW, STYLE, UNSPECIFIED.

codechecker_common/checker_labels.py
```diff
@@ -0,0 +1,262 @@
...
+        """
+        Shorthand for the following call:
+        checker_labels.label_of_checker(checker, 'severity', analyzers)
+        """
+        return self.label_of_checker(checker, 'severity', analyzers)
+
+    def labels_of_checker(
+            self,
+            checker: str,
+            analyzers: Iterable[str] = None) -> List[str]:
```
csordasmarton, line 194: What happens if no analyzer name is specified but the given checker name can be found in multiple label files. Can you please add a test for this use case too?

analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -202,217 +206,273 @@ def add_arguments_to_parser(parser):
...
+    else:
+        state = '+' if row[0] == CheckerState.enabled else \
+            '-' if row[0] == CheckerState.disabled else \
+            '?'
+        guideline = __format_guideline(row[4])
+
+    return (state, row[1], row[2], row[3], guideline, row[5])
+
+
+def __print_checkers(args, cl):
```
csordasmarton, line 354: I tried to run the following command and I got an exception:
```
$ CodeChecker checkers --label profile
  ...
  File "~/codechecker/build/CodeChecker/lib/python3/codechecker_common/checker_labels.py", line 75, in __get_label_key_value
    pos = key_value.index(':')
ValueError: substring not found
```

This should be handle nicely and print an error message if I am using wrongly.

analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -202,217 +206,273 @@ def add_arguments_to_parser(parser):
...
-            if args.guideline is not None:
-                if not match_guideline(warning, args.guideline):
-                    continue
+    cl = analyzer_context.get_context().checker_labels
 
-            guideline = context.guideline_map.get(warning, {})
-            if args.output_format != 'json':
-                guideline = format_guideline(guideline)
+    # TODO: --profile is a deprecated flag. This section should be removed in
+    # the next release.
```
csordasmarton, line 464: If this flag is depreated, please identify it in the help message too, that this flag is deprecated. Also please print a warning message if this flag is used and tell the user that this flag is deprecated and which command needs to be used instead of this.

By the way if we deprecate this option how will you able to list the profiles (CodeChecker checkers --profile list)?

The following command doesn't work:
```
$ CodeChecker checkers --label profile:list
[ERROR 2021-08-23 13:22] - 'list' is a reserved profile keyword. 
[ERROR 2021-08-23 13:22] - Please choose another profile name in ~/codechecker/build/CodeChecker/config/checker_labels.json and rebuild.
```

analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -147,6 +136,13 @@ def add_arguments_to_parser(parser):
                         help="Show details about the checker, such as "
                              "description, if available.")
 
+    parser.add_argument('--label',
+                        nargs='+',
+                        required=False,
+                        default=argparse.SUPPRESS,
+                        help="Filter checkers that are attached the given "
```
csordasmarton, line 143: Can you please add an example section to this and to the check command  when I print the help message of these command similar what we print when running `CodeChecker cmd results -h` (Example scenario:). For example how to get checkers which have severity level of HIGH, or checkers which are in extreme profile etc.

Also in the help message it would be good to mention what kind of profiles do we have and how we can list it. By the way can we list the labels beside the checkers?

csordasmarton (changes requested):

analyzer/codechecker_analyzer/analyzer_context.py
```diff
@@ -257,16 +36,20 @@ def __init__(self):
         self._lib_dir_path = os.environ.get('CC_LIB_DIR', '')
         self._data_files_dir_path = os.environ.get('CC_DATA_FILES_DIR', '')
 
+        # Use this environment variable for testing purposes only. This
+        # variable helps to configure which labels to use in this context.
+        labels_dir = os.path.join(self._data_files_dir_path,
+                                  'config', 'labels')
+        if 'CC_TEST_LABELS_DIR' in os.environ:
+            labels_dir = os.environ['CC_TEST_LABELS_DIR']
+
```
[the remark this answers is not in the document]
bruntib: It is for testing purposes if we need to change labels in a test. It is used in `test_db_cleanup.py`: https://github.com/Ericsson/codechecker/pull/3233/files#diff-1fba94ea7bc654043e003618855cca0b3975e6a7ab45266b1a1c38df3276ed01R54

analyzer/codechecker_analyzer/cmd/checkers.py
bruntib: Yes, I forgot it, but now it's done.

codechecker_common/checker_labels.py
bruntib: I missed your question here, sorry. In that case the checkers of all analyzers are taken into account. I'll write a test on it too soon.

analyzer/codechecker_analyzer/cmd/checkers.py
bruntib: I removed this comment because we'll not deprecate this flag. I also fixed that command.

analyzer/codechecker_analyzer/cmd/checkers.py
bruntib: Yes, I forgot it, but now it's done.

analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -179,240 +208,409 @@ def add_arguments_to_parser(parser):
...
+        elif checker[0] == CheckerState.disabled:
+            status = '-'
         else:
-            rows = [(g, ', '.join(sorted(r))) for g, r in result.items()]
+            status = '?'
+
+        labels = cl.labels_of_checker(checker[1], checker[2])
+
+        print(checker[1])
+        print('  Enabled:', status)
```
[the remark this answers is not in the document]
bruntib: Ok, I used label `Status` here, however, in JSON format I still used `enabled` because its value is boolean. `enabeld: true` seems more intuitive than `status: true`. I think this inconsistency is acceptable regarding that not only labels but the structure itself differs in JSON format.

analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -179,240 +208,409 @@ def add_arguments_to_parser(parser):
...
+            status = '+'
+        elif checker[0] == CheckerState.disabled:
+            status = '-'
         else:
-            rows = [(g, ', '.join(sorted(r))) for g, r in result.items()]
+            status = '?'
+
+        labels = cl.labels_of_checker(checker[1], checker[2])
+
+        print(checker[1])
```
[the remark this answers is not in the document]
bruntib: I changed JSON format a little. Could you please check if this new solution is acceptable?

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -657,13 +651,13 @@ def add_arguments_to_parser(parser):
...
 
-Profiles
+Checker labels
 ------------------------------------------------
-In CodeCheckers there is a manual grouping of checkers. These groups are called
-profiles. The collection of profiles is found in
-config/checker_profile_map.json file. The goal of these profile is that you can
-enable or disable checkers by these profiles. See the output of "CodeChecker
-checkers --profile list" command.
+In CodeChecker there is a manual grouping of checkers. These groups are
```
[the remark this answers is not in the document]
bruntib: I didn't write command hints because `CodeChecker check` command doesn't have this section yet. Instead I wrote an example to `-e` and `-d` flags.

analyzer/codechecker_analyzer/cmd/checkers.py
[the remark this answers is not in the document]
csordasmarton: Now the json format looks like this:
```json
{
        "enabled": true,
        "name": "performance-move-constructor-init",
        "analyzer": "clang-tidy",
        "description": "",
        "labels": [
            [
                "profile",
                "default"
            ],
            [
                "profile",
                "extreme"
            ],
            [
                "profile",
                "sensitive"
            ],
            [
                "severity",
                "MEDIUM"
            ]
        ]
    },
```
Two things:
- I think we use a simpe list for labels: ["profile:default", "severity:Medium", ...]. There is no need to create a list of list, by the way we are doing the same in different output format.
- Instead of enabled we can use the same status keyword which can be a string with values `unknown`, `enabled`, `disabled`.

This way the output format in json will be similar with the default output.

docs/analyzer/user_guide.md
```diff
@@ -1778,41 +1779,66 @@ optional arguments:
...
-                        a rule. Without additional parameter, the available
-                        guidelines and their corresponding rules will be
-                        listed.
+  --details             Show details about the checker, such as status,
+                        checker name, analyzer name, severity, guidelines and
+                        description. Status shows if the checker is enabled
+                        besides the given labels. If the labels don't trigger
+                        a checker then the status is determined by the
+                        analyzer tool.
+  --label [LABEL]       Filter checkers that are attached the given label. If
```
csordasmarton, line 1788: Please identify the format of the label (e.g.: `<label name>:<label value>`) and give a hint how the we can list the label values for a label (`CodeChecker checkers --label <label name>`), and in brackets add a command how we can list the available labels: `CodeChecker checkers --label`.

analyzer/codechecker_analyzer/cmd/analyze.py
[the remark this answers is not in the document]
csordasmarton: Okay, it will be good, thanks :)

csordasmarton (changes requested):

docs/analyzer/user_guide.md
bruntib: I included label format, but I think a usage example here is not necessary. I think it has no additional value to give an example to every label about its own usage, if it is correctly described. And actually we have that at the hints at the bottom of the page. We have examples on other sub-commands as cross referencing, because their description is found in the help page of another command.

csordasmarton (approved):
LGTM!

PR #3247, opened 2021-03-29
[report-converter][fix] SpotBugs Report Converter Hotfix

baloghadamsoftware:
If a bug report does not have a line number and neither of its events
have one then an exception is thrown. This patch fixes the issue.


csordasmarton (approved):
LGTM!

PR #3262, opened 2021-04-09
[report-converter][fix] Fix for SpotBugs Report Conveter with Plugins

baloghadamsoftware:
The SpotBugs report converter tries to handle every element in the
<Project> section as a file path. However, if a plugin is used, such as
find-sec-bugs an additional element with <Plugin> tag is also included
in this section with empty text part, which crashes the converter because
it expects a non-empty text. This patch fixes this issue by only
considering tags <Jar>, <AuxClasspathEntry> and <SrcDir>.


csordasmarton (approved):
LGTM!

PR #3269, opened 2021-04-14
[analizer] Refactor CTU autodetection class to not use lazy evaluation

zomen2:
The problem is that CodeChecker calls underlying clang analizer four times for every analyzed files to detect CTU capabilities of underlying analyzer.
The analyzer main function allocates "analizator actions" during its run and runs them in parallel subprocesses. These subprocesses allocates their own CTUAutodetection objects. In consequence, all subprocess should detect clangsa capabilities independently to each other.
This refactoring does not resolve the whole issue, but one step toward.
The main goal in this commit was that the object properties to not use lazy evaluation of capability check. Instead of that, in construction time all of checks will be performed. The subsequent property calls (availability check function calls) returns already stored values from object member variables.
The final solution should allocate only one CTUAutodetection object before the analizator child processes started.
- This modification decreases the clang calls by one.
- The typing information added also.
- The docstring are updated also.


analyzer/codechecker_analyzer/analyzers/clangsa/ctu_autodetection.py
```diff
@@ -57,133 +29,230 @@ class CTUAutodetection(object):
...
     """
 
-    def __init__(self, analyzer_binary, environ):
-        self.__analyzer_binary = analyzer_binary
-        self.environ = environ
-        self.__analyzer_version_info = None
+    @staticmethod
+    def _invoke_binary_checked(
+            binary_path: str, args: Optional[List[str]] = None,
+            environ: Optional[Dict[str, str]] = None) -> str:
```
martong, line 35: :1st_place_medal:

martong (approved):
Looks good to me! Can't wait to see the next steps.  However, I don't have much confidence in my review, so I suggest to wait for another reviewer as well.

PR #3283, opened 2021-04-26
[test] Compile test project with c++11 explicitly

bruntib:
If an old GCC compiler is installed on a machine which doesn't use C++11
by default, the tests may fail. There are some checkers (e.g.
misc-definitions-in-headers) which report only from C++11. If these
reports are missing with an old compiler then the number of reports will
not be as expected.


csordasmarton (approved):
LGTM!

PR #3298, opened 2021-04-30
Add semicolon to web Makefile

OmniTroid:
Running make package on a zsh shell currently yields a parse error.
It seems to be because zsh expects a semicolon or a newline, however
the newline is escaped and there is no semicolon.
Adding one makes it work.


csordasmarton (approved):

PR #3307, opened 2021-05-11
[test] Refactor process runner function

zomen2:
It returns the underlying process return code too.


csordasmarton (approved):
LGTM!

Just a side comment: later for variable names you can use the short version instead of the long version: `result` vs `res`.

PR #3310, opened 2021-05-13
Enable the build logger fix for CR and LF by default

steakhal:


steakhal:
The diff looks ugly, but the change is simple.

csordasmarton (approved):
LGTM!

PR #3345, opened 2021-06-07
[web] Downgrade python-ldap

bruntib:
python-ldap Python dependency seems to be too new for handling the
currently installed OpenLDAP library.


zomen2 (approved):

csordasmarton (changes requested):
We are still investigating it that we need to downgrade this dependency. So for now I request a change on this PR to block the merge.

bruntib:
The root cause of the error was independent from python-ldap version, so I close this PR.

PR #3352, opened 2021-06-10
Don't hardcode GCC in build-logger Makefile

LebedevRI:
There may not be a `gcc` program, and `CC` may be pointing
at the right one, but this would completely override it,
failing the build.


csordasmarton (approved):

LebedevRI:
Thank you!

PR #3360, opened 2021-06-15
[server] Comment date collision

bruntib:
It is possible that "comments" table contains two comments with the
same date field. It can occur when importing comments from a file which
contains a comment that has the same date as an existing comment.
The list of comments is returned ordered by dates, however, the order
of comments with the same date is not deterministic. This made the
tests fail sometimes.


csordasmarton (approved):
LGTM!

PR #3363, opened 2021-06-22
[fix] Add explicit wheel dependency for virtualenv

gamesh411:
Python package wheel provides the bdist_wheel command required for
building wheels for codechecker-api and codechecker-api-shared
packages. This modification ensures the wheel dependency for each
virtualenv.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,7 @@ venv:
 	# Create a virtual environment which can be used to run the build package.
 	python3 -m venv venv --prompt="CodeChecker venv" && \
 		$(ACTIVATE_RUNTIME_VENV) && \
+		pip3 install wheel && \
 		cd $(CC_ANALYZER) && pip3 install -r requirements.txt && \
 		cd $(CC_WEB) && pip3 install -r $(CC_WEB)/requirements.txt
 
@@ -105,6 +106,7 @@ venv_osx:
 	# Create a virtual environment which can be used to run the build package.
 	python3 -m venv venv --prompt="CodeChecker venv" && \
 		$(ACTIVATE_RUNTIME_VENV) && \
+		pip3 install wheel && \
 		cd $(CC_ANALYZER) && pip3 install -r requirements_py/osx/requirements.txt && \
 		cd $(CC_WEB) && pip3 install -r requirements_py/osx/requirements.txt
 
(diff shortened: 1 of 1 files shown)


gamesh411:
I had a problem on Ubuntu 20.04 (inside WSL2) with making a new virtualenv (make venv). bdist_wheel was needed but not found when building codechecker-api and its -shared package's wheels.

gamesh411:
I could have injected the dependency into the requirements.txt files and/or could have specified an exact version number for wheel, but my reasoning behind not doing so is this:
The wheel package is only used for building the other two packages, and thus will not contribute to the runtime code directly, so there are no API contracts to uphold/break. So at build time, an up-to-date version of the package seems like a good idea to me.

whisperity:
On a real Ubuntu 18.04 I was able to do `pip install codechecker-api codechecker-api-shared`, however, `pip install codechecker` failed:

~~~~
Building wheels for collected packages: codechecker
  Running setup.py bdist_wheel for codechecker ... error
  Complete output from command /home/w/TEMPORARY_CRAP_ENV/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-x9i4yi32/codechecker/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/tmpg1_qjztfpip-wheel- --python-tag cp36:
  /usr/lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'long_description_content_type'
    warnings.warn(msg)
  usage: -c [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
     or: -c --help [cmd1 cmd2 ...]
     or: -c --help-commands
     or: -c cmd --help

  error: invalid command 'bdist_wheel'

  ----------------------------------------
  Failed building wheel for codechecker
  Running setup.py clean for codechecker
Failed to build codechecker
~~~~

tomrittervg:
I hit errors with bdist_wheel not being found and this patch resolved them for me.

whisperity:
Right now, `pip install codechecker` seems to work on all major systems. @gamesh411 Can you check in let's say a Docker env whether most of the LTS Linuxes (Ubuntu 20, 22, and maybe latest Debian, and Alpine Linux as well?) and their **default** Python versions can do the `pip install`. If so, this change is actually not needed? If not, then we should move forward with reviewing this patch.

Makefile
```diff
@@ -98,13 +98,15 @@ venv:
 	# Create a virtual environment which can be used to run the build package.
 	python3 -m venv venv --prompt="CodeChecker venv" && \
 		$(ACTIVATE_RUNTIME_VENV) && \
+		pip3 install wheel && \
```
whisperity, line 101: Why is this better than having the package specified in the `requirements.txt`?

bruntib: I believe, @gamesh411 had this answer for this question from earlier:

> The wheel package is only used for building the other two packages, and thus will not contribute to the runtime code directly, so there are no API contracts to uphold/break. So at build time, an up-to-date version of the package seems like a good idea to me.

However, I would also agree to put dependencies in the `requirements.txt` of the `venv_dev` target. This target is dedicated for dependencies during development and not during usage.

PR #3367, opened 2021-06-28
[CI] Add a job that checks coverage of checker labelling

whisperity:
A CI job which we can use to check if the latest LLVM version (as available in the official PPA) has some checkers added which we haven't set a severity for. The job fails if this list is not empty. :slightly_smiling_face: I added the option to filter checkers, groups, etc. we don't seem to care about.

~~In addition, the job also lists all the checkers we know a severity for, but the checker no longer exists. While these are reported, the CI isn't set up to fail if this is the case.~~ This option exists in the script, but is disabled by default.

### DIFF
--- a/.github/workflows/config_coverage.yml
+++ b/.github/workflows/config_coverage.yml
@@ -1,0 +1,150 @@
+name: checker-config-coverage
+
+on:
+  push:
+    paths:
+      - '.github/workflow/config_coverage.yml'
+      - '.github/workflow/config_label_check.py'
+      - 'config/labels/analyzers/clang-tidy.json'
+      - 'config/labels/analyzers/clangsa.json'
+  pull_request:
+    paths:
+      - '.github/workflow/config_coverage.yml'
+      - '.github/workflow/config_label_check.py'
+      - 'config/labels/analyzers/clang-tidy.json'
+      - 'config/labels/analyzers/clangsa.json'
+  schedule:
+    # Run every Sunday at 21:30 (to latest master at that time).
+    - cron: '30 21 * * SUN'
+  # Allow running this job manually from either API or GitHub UI.
+  workflow_dispatch:
+
+jobs:
+  checker-config-coverage:
+    name: "Config coverage of checkers"
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - name: "Install dependencies"
+        run: |
+          sudo apt-get -qy update
+          sudo apt-get -y --no-install-recommends install \
+            build-essential \
+            curl            \
+            gcc-multilib    \
+            python3-dev     \
+            python3-venv
+      - name: "Get latest LLVM binary package from the community PPA"
+        run: |
+          export DISTRO_FANCYNAME="$(lsb_release -c | awk '{ print $2 }')"
+
+          echo "::group::Setup LLVM PPA"
+          curl -sL http://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+          sudo add-apt-repository -y "deb http://apt.llvm.org/$DISTRO_FANCYNAME/ llvm-toolchain-$DISTRO_FANCYNAME main"
+          echo "::endgroup::"
+
+          # Get the largest Clang package number available.
+          export LLVM_VER="$(apt-cache search --full 'clang-[[:digit:]]*$' | grep '^Package: clang' | cut -d ' ' -f 2 | sort -V | tail -n 1 | sed 's/clang-//')"
+          echo "::group::Install Clang and Clang-Tidy version ${LLVM_VER}"
+          sudo apt-get -y --no-install-recommends install \
+            clang-$LLVM_VER      \
+            clang-tidy-$LLVM_VER
+          sudo update-alternatives --install                   \
+            /usr/bin/clang clang /usr/bin/clang-$LLVM_VER 1000 \
+            --slave                                            \
+              /usr/bin/clang-tidy clang-tidy /usr/bin/clang-tidy-$LLVM_VER
+          echo "::endgroup::"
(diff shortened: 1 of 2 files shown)


.github/workflows/config_coverage.yml
```diff
@@ -0,0 +1,87 @@
...
+            --existing-ignore-suffix \
+              "Sanitizer" \
+            --new-ignore \
+              "alpha." \
+              "apiModeling." \
+              "debug." \
+              "optin.osx." \
+              "osx." \
+              "darwin-" \
+              "objc-"
```
whisperity, line 77: These are just things I came up with by running the command locally, and the actual values are up for discussion! :innocent:

whisperity:
Should this patch move forward or be tried again with the new config syntax? I'm more familiar with GHA scripts now, and realised how I can emit soft errors (notices, warnings). I think it would still be useful to have at least a proper list of *"new checks that we SHOULD categorise because they will come in the next LLVM release"* kind of way...

@dkrupp

whisperity:
I've rebased and aligned the test with the new layout of the config files. Brand new workflows do not trigger from pull requests, it seems. You can see some of the recent runs here: http://github.com/whisperity/CodeChecker/actions/workflows/config_coverage.yml

I've also added a line in the job to run scheduled, once a week. Originally, I wanted to do "once every other week" to kind-a sync it up with our release schedule, but unfortunately crontabs do not support such complex notions, and I did not want to complicate matters.

whisperity:
@bruntib Do we expect changes to the config file format? If not, could we please discuss this patch and merge it, so it doesn't have to be refactored over and over again?

.github/workflows/config_coverage.yml
```diff
@@ -0,0 +1,150 @@
...
+      - 'config/labels/analyzers/clangsa.json'
+  pull_request:
+    paths:
+      - '.github/workflow/config_coverage.yml'
+      - '.github/workflow/config_label_check.py'
+      - 'config/labels/analyzers/clang-tidy.json'
+      - 'config/labels/analyzers/clangsa.json'
+  schedule:
+    # Run every Sunday at 21:30 (to latest master at that time).
+    - cron: '30 21 * * SUN'
```
bruntib, line 18: What is the purpose of this scheduling?

.github/workflows/config_coverage.yml
```diff
@@ -0,0 +1,150 @@
...
+            --existing-filter "clang-diagnostic-" \
+            --new-filter "clang-diagnostic-"
+          EXIT_STATUS=$?
+          echo "Coverage check returned: $EXIT_STATUS."
+          # Explicitly check if the bit for "8" is set in the result,
+          # indicating new checkers without severity set.
+          if [[ $(($EXIT_STATUS & 8)) -eq 8 ]]
+          then
+            echo "::warning title=New unconfigured diagnostics::The checker label config files lack some new diagnostic report (\"warning\") kinds."
+            exit 0
```
bruntib, line 110: How does this GitHub action indicate when some modification in our "checker label file" is needed?

.github/workflows/config_coverage.yml
```diff
@@ -0,0 +1,150 @@
...
+        run: |
+          export PATH="${{ steps.codechecker.outputs.CODECHECKER_PATH }}:$PATH"
+          CodeChecker analyzers --details
+          CodeChecker checkers \
+            --analyzer clangsa clang-tidy \
+            --output rows \
+              > checker_list_normal.txt
+          CodeChecker checkers \
+            --analyzer clangsa clang-tidy \
+            --warnings \
```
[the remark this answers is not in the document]
whisperity: Yes. However, we do not have a **specific** label in the config for *every warning*. So if I used the `--warnings` flag for the only list, every new warning would also trigger a failure, something I anticipate we do not intend to deal with. Thus, warnings are considered separate, and a new warning only triggers a... warning on the CI, but with the job still passing.

.github/workflows/config_coverage.yml
```diff
@@ -0,0 +1,150 @@
...
+              "osx." \
+              "darwin-" \
+              "objc-"
+          EXIT_STATUS=$?
+          echo "Coverage check returned: $EXIT_STATUS."
+          # Explicitly check if the bit for "8" is set in the result,
+          # indicating new checkers without severity set.
+          if [[ $(($EXIT_STATUS & 8)) -eq 8 ]]
+          then
+            exit 8
```
whisperity, line 142: @bruntib This is the execution step which makes the entire job fail if there are new **checkers** that are added, without a label.

whisperity: The job will run scheduled and it will trigger a failure. Once you go into the detailed log of the job, the standard output of the execution will contain the details on which checks are lacking a label:

```
Checkers ADDED to upstream, without labels in CodeChecker:
 + abseil-cleanup-ctad
 + altera-kernel-name-restriction
 + altera-single-work-item-barrier
 + bugprone-shared-ptr-array-mismatch
 + bugprone-stringview-nullptr
 + linuxkernel-must-check-errs
 + misc-misleading-bidirectional
 + misc-misleading-identifier
 + readability-container-contains
 + webkit.UncountedLambdaCapturesChecker
Coverage check returned: 8.
Error: Process completed with exit code 8.
```

whisperity:
The "magic" exit statuses `4` and `8` are used because the script itself is written in a way that it is capable of dealing with checks that **disappeared** and trigger a failing result on that too. However, it seems the CodeChecker project tries to be compatible with multiple versions of LLVM, so the CI does not execute the logic that checks for *removed* checks.

(Exit `0` is normal, `1` is execution error, `2` is when `argparse` reported an issue, and then comes `4` and `8` as the next bits in a flag.)

In case we ever decide we no longer want to support version X of LLVM anymore, we will have the script handy to run against that version and see what checks no longer appear in that version, just by passing the appropriate flag.

bruntib (approved):

PR #3369, opened 2021-06-29
Add core.StackAddrEscapeBase checker as HIGH.

AlexisTM:
This adds the `core.StackAddrEscapeBase` checker severity as `HIGH`.

The error is currently `Unspecified` and catches dangling references:

> Address of stack memory associated with local variable "Thing" is still referred to by the global variable "thing" upon returning to the caller. This will be a dangling reference.

### DIFF
--- a/config/checker_severity_map.json
+++ b/config/checker_severity_map.json
@@ -154,6 +154,7 @@
   "core.DynamicTypePropagation":                                "MEDIUM",
   "core.NonNullParamChecker":                                   "HIGH",
   "core.NullDereference":                                       "HIGH",
+  "core.StackAddrEscapeBase":                                   "HIGH",
   "core.StackAddressEscape":                                    "HIGH",
   "core.UndefinedBinaryOperatorResult":                         "HIGH",
   "core.uninitialized.ArraySubscript":                          "HIGH",


dkrupp (changes requested):
core.StackAddrEscapeBase is just a modeling checker for core.StackAddrEscape checker: https://clang.llvm.org/docs/analyzer/checkers.html#core-stackaddressescape-c

In theory it should not emit any warning under this checkername so there should be no need to specify a severity for it. Did you see any direct report from this checker?

AlexisTM:
I indeed got this error. 

image

The code triggering it was like:

something.c
```
static Type* variable; 

void updateVariable() {
    Type data[N];
    variable = &data;
}
```

whisperity:
@Szelethus Do you have an idea why this might be happening?

Szelethus:
> @Szelethus Do you have an idea why this might be happening?

This bug should've been fixed a long while ago by https://reviews.llvm.org/D78101. Indeed, with a recent version of clang, this error is correctly emitted under core.StackAddressEscape.

dkrupp:
> > @Szelethus Do you have an idea why this might be happening?
> 
> This bug should've been fixed a long while ago by https://reviews.llvm.org/D78101. Indeed, with a recent version of clang, this error is correctly emitted under core.StackAddressEscape.

Then we don't need this patch, right? Can it be abandoned?

AlexisTM:
I closed it as I must have used clang 9 instead of clang 12 causing the issue.

PR #3421, opened 2021-09-02
[test] Test workspace is not necessarily under HOME

bruntib:
On some systems the test workspace directory is not under HOME
directory so it is not a good choice for testing "trim path prefix".


csordasmarton (approved):
LGTM!

PR #3424, opened 2021-09-06
[db] Add sqlite config to alembic.ini

Vodorok:
Extended both the config, and the default product database file.


csordasmarton (approved):
LGTM!

PR #3428, opened 2021-09-08
[analyzer] Use `-imacros` flag instead of `-macros`

whisperity:
This was seemingly last touched in Ericsson/codechecker#2449 (Ericsson/codechecker@7d93b32edcbe885c391cb59d4e13c3aa4163463a) but it seems like a typo. There's no such flag as `-macros` in neither Clang nor GCC.

> via @steakhal

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -220,7 +220,7 @@ COMPILE_OPTIONS_MERGED = [
     '-[DIUF]',
     '-idirafter',
     '-isystem',
-    '-macros',
+    '-imacros',
     '-isysroot',
     '-iprefix',
     '-iwithprefix',


csordasmarton (approved):
LGTM!

bruntib:
I wouldn't be so brave to do this modification in the last minute before the release. It is possible that we met this flag in a non-standard compiler. If the addition of -imacros is needed, then add it instead of change the previous one.

whisperity:
> If the addition of -imacros is needed, then add it instead of change the previous one.

 @bruntib Yeah that's why I wanted you to look at the patch too... I couldn't figure out the reasoning behind the changes in the previous patch.

The addition of `-imacros` is definitely needed, the patch is coming from field experience with a user whose build broke specifically because of this argument (their build system used it like **3 times**) was torn out from the analysis invocation.

I'll change the patch ASAP!

csordasmarton:
I think it was a typo during code refactoring when this `macros` flag was skipped instead of `imacros` in the following commit: https://github.com/Ericsson/codechecker/commit/1a5863416d06b39bdb7264572451e5fbaa2e7a94

Previous imacros was used:
macros

steakhal:
> I think it was a typo during code refactoring when this `macros` flag was skipped instead of `imacros` in the following commit: [1a58634](https://github.com/Ericsson/codechecker/commit/1a5863416d06b39bdb7264572451e5fbaa2e7a94)
> 
> Previous imacros was used:
> macros

This seems like a clear indication of a typo. IMO the patch is good as-is.

whisperity:
(Side note: we should really put some comments up what these big lists of flags mean.)

I have the updated patch ready if you want that, but I am fine with keeping the current patch.

steakhal:
I guess this missed the 6.17.0 release.

whisperity:
> I guess this missed the 6.17.0 release.

@steakhal Yeah, seems like so, I just asked.

PR #3431, opened 2021-09-10
[test] Add script to test store times of a Codechecker binary

Vodorok:


scripts/store_test.py
```diff
@@ -0,0 +1,120 @@
...
+            r"(Base package version\D*)(\d+.\d+.\d+)",
+            version_p.stdout.decode("utf8")
+        )
+        version = m.group(2) if m.group(2) else binary
+
+        timesheet[version] = []
+        for count in range(args.count):
+            with tempfile.TemporaryDirectory() as tmpdir:
+                server = subprocess.Popen([binary, "server", "-w", tmpdir])
+                time.sleep(3)  # wait for server spin up.
```
whisperity, line 46: This is brittle. Could you instead repeatedly poll the output of `CodeChecker cmd runs` or something?

PR #3469, opened 2021-10-28
[gui] Fix for filter product in gui test

### LINKED ISSUE #3435: Gui tests fail often on github actions
**Describe the bug**
Gui tests often fail on GitHub actions, even when a pull request has no connection to any frontend code.

The relevant part of the error msg:
 {"status":-1,"code":"","state":"element click intercepted",..."errorStatus":64,"error":"Element <input id=\"input-28\" type=\"text\"> is not clickable at point (246,104) because another element <div class=\"v-overlay__scrim\"> obscures it","httpStatusCode":500}

**CodeChecker version**
Not applicable.

**To Reproduce**
Run Github actions a few times, and the error may disappear.

**Expected behaviour**
Gui tests for pull requests that did not modify frontend code, should pass flawlessly.

**Additional context**
Add any other context about the problem.

vodorok:
Wait until the div with class v-overlay__srim disappears from the font
of the search bar.

Should fix #3435

### DIFF
--- a/web/server/vue-cli/e2e/pages/product.js
+++ b/web/server/vue-cli/e2e/pages/product.js
@@ -9,8 +9,8 @@ const commands = {
       .to.be.visible.before(5000);
 
     this.section.editAnnouncementDialog
-      .clearAndSetValue("@input", value, this.section.editAnnouncementDialog)
-      .click("@confirmBtn");
+      .clearAndSetValue("@input", value, this.section.editAnnouncementDialog);
+    this.section.editAnnouncementDialog.click("@confirmBtn");
 
     this.expect.section("@editAnnouncementDialog")
       .to.not.be.present.before(5000);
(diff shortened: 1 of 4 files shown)


web/server/vue-cli/e2e/pages/product.js
```diff
@@ -210,7 +211,8 @@ module.exports = {
     searchInput: ".v-toolbar__content input[type='text']",
     progressBar: ".v-data-table__progress",
     editBtn: "tr .edit-btn",
-    removeBtn: "tr .remove-btn"
+    removeBtn: "tr .remove-btn",
+    overlayScrim: "div .v-overlay__scrim"
```
csordasmarton, line 215: When you open a vuetify dialog window it will create two separate elements in the application root element:
- `v-overlay`
- `v-dialog`

image

When you close a dialog, both of them will be dissapeared.

You class `v-overlay__scrim` is the child of the `v-overlay` element, and the `v-overlay` is connected to a `v-dialog` element.

So instead of this we have to find out which dialog is still in the DOM and have to wait for it while it is dissapeared instead of relying on an inner representation of a component.

csordasmarton (changes requested):

bruntib (approved):

PR #3475, opened 2021-11-02
[test][cfg] Use clang-13 in the CI, uplift tests accordingly

steakhal:
The `-Wunused-but-set-variable` and `-Wunused-but-set-parameter` warnings were introduced by https://reviews.llvm.org/D100581, so had to suppress these warnings in the test file to make them forward compatible.

Let's see if the build bots are happy.

### DIFF
--- a/.github/workflows/install-deps.sh
+++ b/.github/workflows/install-deps.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
-sudo add-apt-repository 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-11 main' -y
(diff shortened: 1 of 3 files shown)


csordasmarton (changes requested):
Can you please change the version number in this file too:
https://github.com/Ericsson/codechecker/blob/4be1ecad9249470171a623feeacf17da83858fce/scripts/test/check_clang.py#L20-L21

steakhal:
> Can you please change the version number in this file too:
> 
> https://github.com/Ericsson/codechecker/blob/4be1ecad9249470171a623feeacf17da83858fce/scripts/test/check_clang.py#L20-L21

The tests seem to run even on clang-11.

steakhal:
From the log, it's not clear to me what is failing. Any guidance?

csordasmarton (approved):
LGTM!

PR #3494, opened 2021-11-12
Removing cppcoreguidelines-virtual-class-destructor from the profiles

dkrupp:
cppcoreguidelines-virtual-class-destructor is unstable,
so removing it from the profiles


csordasmarton (approved):
LGTM!

whisperity:
@steakhal, have you landed the fix about the crashes, macros, etc. involved in this check? In that case, perhaps it could be re-added to the config?

steakhal:
Oh sure. They landed on this Monday. I'll do something about it tomorrow.

PR #3503, opened 2021-11-16
Update the Usage Guide

dkrupp:
-Describe what the user should do in case the analysis failed.
-Description about the structure of the failed zip added.
-The local analysis workflow should be part of Step-8 and not an additional step
 as it is a server-less alternative


csordasmarton (approved):
LGTM!

PR #3517, opened 2021-11-23
Mention more details in the build instructions

jimis:
Forgetting to install requirements for LDAP or PAM authentication is quite
a subtle problem: the build completes, the server runs fine,
but users can't login. So mention this in the instructions.

Also mention the possibility of building a standalone_package.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -281,7 +281,7 @@ or the UI code is changed. If you wouldn't like to build the UI code you can
 set the `BUILD_UI_DIST` environment variable to `NO` before the package build:
 `BUILD_UI_DIST=NO make package`.
 - Use `make standalone_package` instead of `make package` to avoid
-manually activating the environment before running CodeChecker.
+having to manually activate the environment before running CodeChecker.
 
 ### Upgrading environment after system or Python upgrade
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -258,8 +258,9 @@ cd ~/codechecker
 make venv
 source $PWD/venv/bin/activate
 
-# Optional step in case you want LDAP or PAM authentication support.
-#pip3 install -r web/requirements_py/auth/requirements.txt
+# [Optional] If you want to use external authentication methods (LDAP / PAM)
+# follow the instructions in
+# docs/web/authentication.md#external-authentication-methods
 
 # Build and install a CodeChecker package.
 make package
--- a/docs/README.md
+++ b/docs/README.md
@@ -258,6 +258,9 @@ cd ~/codechecker
 make venv
 source $PWD/venv/bin/activate
 
+# Optional step in case you want LDAP or PAM authentication support.
+#pip3 install -r web/requirements_py/auth/requirements.txt
+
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -258,6 +258,9 @@ cd ~/codechecker
 make venv
 source $PWD/venv/bin/activate
 
+# Optional step in case you want LDAP or PAM authentication support.
+#pip3 install -r web/requirements_py/auth/requirements.txt
+
```
whisperity, line 263: Very strong disagree here. These additional requirements usually depend on system packages which may or may not be available on the system, especially not if the user is building a container! We should not have first-time users think too much about what to do in the build process. Lack of system-level packages, like the necessary library headers, will result in building and extracting the pip package not succeed.

The documentation (https://github.com/Ericsson/codechecker/blob/master/docs/web/authentication.md#external-authentication-methods-) which you need to read to set LDAP up anyway, discusses the need for additional **system-level** packages.

docs/README.md
```diff
@@ -277,6 +280,8 @@ environment variable to `YES` before the package build:
 or the UI code is changed. If you wouldn't like to build the UI code you can
 set the `BUILD_UI_DIST` environment variable to `NO` before the package build:
 `BUILD_UI_DIST=NO make package`.
+- Use `make standalone_package` instead of `make package` to avoid
+manually activating the environment before running CodeChecker.
```
whisperity, line 284:
```suggestion
- Use `make standalone_package` instead of `make package` to avoid
having to manually activate the environment before running CodeChecker.
```

Previous wording would indicate that running CodeChecker would itself activate the environment, and this is something the user does not want to happen, whereas what you are trying to say is that the user could just not have this action-dependency.

whisperity (dismissed):

docs/README.md
jimis: It's a real problem however. I am the same person that at some point enabled LDAP auth, building CodeChecker with the needed requirements, and at some other point needed to upgrade CodeChecker so I just followed the build instructions on the "install" page, and then people couldn't login. The problem is solved for me (by following my notes, instead of the project's install instructions) but it was subtle so I created this pull request.

I could suggest other alternatives:
* Add a bullet point noting that you might have to install more packages depending the features you want to enable, with links to the relevant pages (which are not easy to find BTW). Are there more optional features that require certain packages? It would be nice to collect this info on the installation page.
* Fail to build the server if the build configuration has LDAP enabled. I am not sure how this would be possible in CodeChecker, but other projects would do that during `./configure --enable-ldap`, for example.
* Fail to run the server if the runtime configuration has LDAP enabled.

jimis:
Thanks for the suggestions, I accepted them as separate commits. If the pull request gets merged, I'll squash everything into one commit.

csordasmarton (approved):
LGTM!

docs/README.md
jimis: The pull request has changed. Do we "resolve" this conversation?

PR #3532, opened 2021-11-30
Include cppcoreguidelines-virtual-class-destructor in profiles

steakhal:
This patch partially reverts #3494

The https://github.com/llvm/llvm-project/commit/0685e83534ef8917f277b394da2927cabff8129f fixed the issue about the `cppcoreguidelines-virtual-base-class-destructor` checker, so we can safely include this in the intended profiles.

Special thanks @whisperity for reminding me.


csordasmarton (approved):
LGTM!

PR #3547, opened 2021-12-10
[fix] No rebuild on satisfied requirements

bruntib:
The command "make package" fetches and installs JS dependencies in case
the version of some dependencies changed. CodeChecker API is such a
dependency deployed locally in this repo as .tar.gz files. The verison
checking on these files emit slightly different outputs, so npm runs
unnecessarily even if no change happened.
The old output format was valid while we deployed CodeChecker API in
npm.


csordasmarton (approved):
Nice catch! LGTM!

PR #3633, opened 2022-03-18
[bugfix] Determine stats capability in pre-analysis

bruntib:
get_analyzer_checkers() should be used for querying checker list because
this can be used for getting modeling checkers too. Pre-analysis depends
on the existence of some modeling checkers.


csordasmarton (approved):
LGTM!

PR #3640, opened 2022-03-30
Pin the number of jobs in some Make targets

steakhal:
If someone runs `make -j16 test_unit test_analyzer test_web_sqlite`,
then the command would spuriously fail due to a filesystem race-condition.

The `cp ldlogger build/bin` would basically try to copy a not-yet-existing file,
causing a failure immediately, or later in the test pipeline, when it
exercises the logger. Causing an error similar to this:
```
make[1]: Leaving directory `codechecker/analyzer'
make: *** [test_analyzer] Error 2
make: *** Waiting for unfinished jobs....
```

Let's work around the issue by omitting the `all` target from the `test*` targets.
This way if it's already built, tests should just pass.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,6 +43,7 @@ jobs:
         working-directory: analyzer/tools/build-logger
         run: |
           pip install -r requirements_py/dev/requirements.txt
+          make all
           make test
 
       - name: Run merge-clang-extdef-mappings tests
--- a/analyzer/tools/build-logger/tests/Makefile
+++ b/analyzer/tools/build-logger/tests/Makefile
@@ -5,9 +5,9 @@ REPO_ROOT ?= REPO_ROOT=$(ROOT)
 # Nose test runner configuration options.
 NOSECFG = --config .noserc
 
-test: all pycodestyle pylint test_unit
+test: pycodestyle pylint test_unit
 
-test_in_env: all pycodestyle_in_env pylint_in_env test_unit_in_env
+test_in_env: pycodestyle_in_env pylint_in_env test_unit_in_env
 
 PYCODESTYLE_TEST_CMD = pycodestyle tests/unit
 


Makefile
```diff
@@ -111,8 +111,8 @@ venv_osx:
 clean_venv:
 	rm -rf venv
 
-PIP_DEV_DEPS_CMD = $(MAKE) -C $(CC_ANALYZER) pip_dev_deps && \
-  $(MAKE) -C $(CC_WEB) pip_dev_deps && \
+PIP_DEV_DEPS_CMD = $(MAKE) -C $(CC_ANALYZER) -j1 pip_dev_deps && \
+  $(MAKE) -C $(CC_WEB) -j1 pip_dev_deps && \
```
csordasmarton, line 115: Why do we need `j1` here? The `pip_dev_deps` target in both cases will call a single pip install command:

https://github.com/Ericsson/codechecker/blob/f11eb1a09c796a083a8fb9ea7d94946a262fe8da/analyzer/Makefile#L29-L30
https://github.com/Ericsson/codechecker/blob/f11eb1a09c796a083a8fb9ea7d94946a262fe8da/web/Makefile#L50-L51

analyzer/tests/Makefile
```diff
@@ -66,7 +66,7 @@ test_functional_in_env: venv_dev
 		$(ACTIVATE_DEV_VENV) && $(FUNCTIONAL_TEST_CMD)
 
 test_build_logger:
-	$(MAKE) -C tools/build-logger test
+	$(MAKE) -C tools/build-logger -j1 test
```
csordasmarton, line 69: The problem with the `test` target is that it depends on the `all` target:
https://github.com/Ericsson/codechecker/blob/f11eb1a09c796a083a8fb9ea7d94946a262fe8da/analyzer/tools/build-logger/tests/Makefile#L8

And what `all` target  will do is to build the package:
https://github.com/Ericsson/codechecker/blob/f11eb1a09c796a083a8fb9ea7d94946a262fe8da/analyzer/tools/build-logger/Makefile#L76

This is why your make command failes and gives you an error when you run it on multiple threads.

Can't we remove this dependency from the `test` target and assume you already built a package from ldlogger? That would solve the race condition.

csordasmarton (changes requested):

analyzer/tests/Makefile
steakhal: Possible, but why should we assume it's already done?

Makefile
steakhal: I think I tried this, but let me try again.

analyzer/tests/Makefile
csordasmarton: If we do not assume that and if you run the for example the `all` target with multiple threads it is not guaranteed that the dependencies will be run in the correct order. For example the `all` target depends on the `pack32bit` target:
https://github.com/Ericsson/codechecker/blob/f11eb1a09c796a083a8fb9ea7d94946a262fe8da/analyzer/tools/build-logger/Makefile#L78

This target depends on two other targets: `32bit` and `packbin`. The order is important here because the `packbin` target will assume that `32bit` already run. If you run it with multiple threads this is not guaranteed.

steakhal: These should not depend on the order. If that's the case then `packbin` should explicitly depend on `32bit`.

steakhal: Anyway, I updated the patch to simply omit the `all` target from the `test` target deps.

csordasmarton (changes requested):
The test cases related to analyzer tools are failing. With your changes now you have to build a package from ldlogger before running the test cases:
https://github.com/Ericsson/codechecker/blob/f11eb1a09c796a083a8fb9ea7d94946a262fe8da/.github/workflows/test.yml#L42-L46
I think that would solve the problem.

csordasmarton (approved):
LGTM!

PR #3647, opened 2022-04-06
[web] Fix exceptions during blame information storage

csordasmarton:
The python git library may throw not only `TypeError` exceptions but
`ValueError` exceptions as well. For example when the `HEAD` file exists
in the `.git` directory but the user who is running the `CodeChecker store`
command doesn't have permission to this file.

To solve this problem we will catch all the exceptions when getting the
tracking branch and remote url.

### DIFF
--- a/web/client/codechecker_client/blame_info.py
+++ b/web/client/codechecker_client/blame_info.py
@@ -16,6 +16,27 @@ LOG = get_logger('system')
 FileBlameInfo = Dict[str, Optional[Dict]]
 
 
+def __get_tracking_branch(repo: Repo) -> Optional[str]:
+    """
+    Get the tracking branch name or the current commit hash from the given
+    repository.
+    """
+    try:
+        # If a commit is checked out, accessing the active_branch member will
+        # throw an error.
+        return str(repo.active_branch.tracking_branch())
+    except Exception:
+        pass
+
+    try:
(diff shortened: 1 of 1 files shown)


web/client/codechecker_client/blame_info.py
```diff
@@ -16,26 +16,41 @@
...
+    except Exception:
+        pass
+
+    try:
+        # Use the current commit hash if it's available.
+        return repo.head.commit.hexsha
+    except Exception:
+        pass
+
+    return None
```
bruntib, line 37: Returning `None` is the default behavior of a Python function so this line is not needed.

bruntib (approved):

web/client/codechecker_client/blame_info.py
csordasmarton: Yes, I know it, but because of `mypy` we need to write out explicitly.

Let's try out the following example:
```py
from typing import Optional

def foo() -> Optional[str]:
  try:
    return "Hello"
  except Exception:
    pass
```

If you analyze it with mypy you will get the following error:
```
$ mypy test.py 
test.py:3: error: Missing return statement
```
Also if you write only a return statement but without a value you will get the following error:
```
$ mypy test.py 
test.py:9: error: Return value expected
```

PR #3653, opened 2022-04-20
Fix ctu extdef mapping file with space problem

martong:
@Szelethus Could you please commandeer this patch? I guess we'd need some tests for the input files and some unit tests for a common function that is used in both places.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
@@ -155,9 +155,16 @@ def func_map_list_src_to_ast(func_src_list, ctu_on_demand):
 
     func_ast_list = []
     for fn_src_txt in func_src_list:
-        dpos = fn_src_txt.find(" ")
-        mangled_name = fn_src_txt[0:dpos]
-        path = fn_src_txt[dpos + 1:]
+        if fn_src_txt[0].isdigit():
+            length_str, _ = fn_src_txt.split(':', 1)
+            length = int(length_str)
+            sep_pos = len(length_str) + 1 + length
+            mangled_name = fn_src_txt[0: sep_pos]
+            path = fn_src_txt[sep_pos + 1:]  # Skipping the ' ' separator
+        else:
+            dpos = fn_src_txt.find(" ")
+            mangled_name = fn_src_txt[0:dpos]
+            path = fn_src_txt[dpos + 1:]
 
         # On-demand analysis does not require any preprocessing on the source
         # file paths, contrary to AST-dump based.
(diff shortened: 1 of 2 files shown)


analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
```diff
@@ -155,9 +155,16 @@ def func_map_list_src_to_ast(func_src_list, ctu_on_demand):
...
     func_ast_list = []
     for fn_src_txt in func_src_list:
-        dpos = fn_src_txt.find(" ")
-        mangled_name = fn_src_txt[0:dpos]
-        path = fn_src_txt[dpos + 1:]
+        if fn_src_txt[0].isdigit():
+            length_str, _ = fn_src_txt.split(':', 1)
+            length = int(length_str)
+            sep_pos = len(length_str) + 1 + length
+            mangled_name = fn_src_txt[0: sep_pos]
```
bruntib, line 162: Are you sure that the mangled name starts at the beginning of the line? Is the length part of the mangled name?

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
```diff
@@ -155,9 +155,16 @@ def func_map_list_src_to_ast(func_src_list, ctu_on_demand):
 
     func_ast_list = []
     for fn_src_txt in func_src_list:
-        dpos = fn_src_txt.find(" ")
-        mangled_name = fn_src_txt[0:dpos]
-        path = fn_src_txt[dpos + 1:]
+        if fn_src_txt[0].isdigit():
+            length_str, _ = fn_src_txt.split(':', 1)
```
bruntib, line 159: I'm sure it's quite unlikely, but can we make sure that the line contains exactly one colon? I mean can the path be in Windows format? (e.g. C:\blabla)

bruntib (changes requested):
These comments are

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
martong: The format is `<Length>:<USR> <File-Path>`. My understanding is that `split(':', 1)` would create a tuple with two elements. The first element will be the string until the **first** colon. In that sense it doesn't matter if we have a colon in the path, because that is placed in the second element of the tuple.

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
martong: Well, it seems misleading that we handle the length as part of the mangled_name. However, it doesn't really matter because at the end we just concatenate the two and append it as a string:
```
func_ast_list.append(mangled_name + " " + mapped_path)
```

whisperity:
In general, wouldn't it be more expressive (albeit perhaps a little bit slower?) to use a regular expression for the capturing of the individual bits of the input lines, instead of manual juggling of lengths and offsets?

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
whisperity: But if you're processing the data into a space-separated list, can the consumer still handle it...? How would or why wouldn't the same issue resurface in a later step of the pipeline? My guess is that this list is now only for CodeChecker to handle internally. Could we pass a list of two-tuples?

dkrupp (approved):

bruntib (approved):

martong:
Related Clang patch https://reviews.llvm.org/D102669

PR #3655, opened 2022-04-21
[config][CLI] Merge, and don't override when multiple --analyzer-configs are specified

Szelethus:
Previous to this patch, multiple analyzer configs could be specified
like this:

CodeChecker analyze ./compile_commands.json -o ./reports -c --verbose debug \
  --analyzers clang-tidy \
  --analyzer-config clang-tidy:HeaderFilterRegex=".*" clang-tidy:AnalyzeTemporaryDtors="true"

However, if you specifiy --analyzer-config more than once, the only the
last one counts, like many other flags.

The problem is that --analyzer-config may be specified from both the CLI
and a CodeChecker config file, and even in this case, only one of these
would count.

Also, the -analyzer-config flag from the Clang Static Analyzer works
differently, as multiple -analyzer-configs in the invocation and
appended, not overwritten. Coming from that space, that philosophy feels
more intuitive to me, and I think other users might be just as surprised
by this behaviour.

This patch merges --analyzer-config (we append the list of configs)
values instead of overwriting them.

### DIFF
--- a/analyzer/codechecker_analyzer/arg.py
+++ b/analyzer/codechecker_analyzer/arg.py
@@ -36,6 +36,38 @@ class OrderedCheckersAction(argparse.Action):
         if 'ordered_checkers' not in namespace:
             namespace.ordered_checkers = []
         ordered_checkers = namespace.ordered_checkers
+        # Map each checker to whether its enabled or not.
         ordered_checkers.append((value, self.dest == 'enable'))
 
         namespace.ordered_checkers = ordered_checkers
+
+
+class OrderedConfigAction(argparse.Action):
+    """
+    Action to store --analyzer-config and --checker-config values. These may
+    come from many sources, including the CLI, the CodeChecker config file,
+    the saargs file and the tidyargs file, or some other places.
+    """
+
+    def __init__(self, option_strings, dest, nargs=None, **kwargs):
+        if nargs is not '*':
+            raise ValueError("nargs must be '*' for backward compatibility "
+                             "reasons!")
+        super(OrderedConfigAction, self).__init__(option_strings, dest,
+                                                  nargs, **kwargs)
+
+    def __call__(self, parser, namespace, value, option_string=None):
+
+        assert isinstance(value, list), \
+               f"--analyzer-config or --checker-config value ({value}) is " \
+               "not a list, but should be if nargs is not None!"
+
+        if not hasattr(namespace, self.dest):
+            setattr(namespace, self.dest, [])
+
+        dest = getattr(namespace, self.dest)
+
+        for flag in value:
+            if flag in dest:
+                dest.remove(flag)
+            dest.append(flag)
--- a/analyzer/tests/functional/config/test_config.py
+++ b/analyzer/tests/functional/config/test_config.py
@@ -118,6 +118,129 @@ class TestConfig(unittest.TestCase):
         self.assertIn("clangsa analyzed simple.cpp", out)
         self.assertNotIn("clang-tidy analyzed simple.cpp", out)
 
+    def test_config_file_multiple_analyzer_config_resolution(self):
+        """
+        Test whether multiple --analyzer-config arguments from a CodeChecker
+        config file are merged, and don't overwrite one another.
+        """
+        with open(self.config_file_json, 'w+',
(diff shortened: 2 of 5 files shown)


analyzer/codechecker_analyzer/arg.py
```diff
@@ -36,6 +36,36 @@ def __call__(self, parser, namespace, value, option_string=None):
...
+
+    def __call__(self, parser, namespace, value, option_string=None):
+
+        if 'analyzer_config' not in namespace:
+            namespace.analyzer_config = []
+        analyzer_config = namespace.analyzer_config
+
+        assert isinstance(value, list), \
+               "--analyzer-config value  (" + str(value) + ") is not a list," \
+               " but should be if nargs is not None!"
```
bruntib, line 67: Might be useful to know that strings can be assembled simpler:
```suggestion
               f"--analyzer-config value  ({value}) is not a list," \
               " but should be if nargs is not None!"
```

bruntib (changes requested):
Just two minor comments, otherwise it looks good.

Szelethus:
It appears that currently --config overrides CLI arguments. I need to dig tothe bottom of that before merging.

analyzer/tests/functional/config/test_config.py
```diff
@@ -118,6 +118,102 @@ def test_only_clangsa_config(self):
...
+
+        self.assertNotEqual(returncode, 1)
+        self.assertIn("track-conditions=false", out)
+        self.assertIn("{\"HeaderFilterRegex\": \".*\"}", out)
+
+    def test_cmd_overrides_config_file(self):
+        """
+        Test the override of multiple --analyzer-config arguments from *both* a
+        CodeChecker config file and the CLI are merged, but if they specify the
+        very same config, CLI arguments override the config file.
```
bruntib, line 193: I think we need to rephrase it. A command line option doesn't override the ones In config file, but appends it to a later position in the analyzer command.

Szelethus:
When this lands, CLI arguments will be present twice in `analyzer_config` after all arguments are parsed. This is becaused the CLU arguments are parsed to check the value of `--config`, that file is parsed and prepended to the CLI argument list, and that whole thing is parsed again.

This isn't an immediate concern (the latest argument will come to effect, which will be the CLI one), but isn't too pretty either. We should likely not parse all command line arguments twice anyways.

bruntib:
Based on a discussion in our team, these to things should also be included in this pull request:
- Eliminate duplication when using both config file and command line flag: #3665
- Duplicate analyzer/checker config options should also be eliminated if a specific analyzer/checker occurs both in config file and command line.

analyzer/codechecker_analyzer/arg.py
```diff
@@ -36,6 +36,38 @@ def __call__(self, parser, namespace, value, option_string=None):
...
+            raise ValueError("nargs must be '*' for backward compatibility "
+                             "reasons!")
+        super(OrderedConfigAction, self).__init__(option_strings, dest,
+                                                  nargs, **kwargs)
+
+    def __call__(self, parser, namespace, value, option_string=None):
+
+        assert isinstance(value, list), \
+               f"--analyzer-config value ({value}) is not a list," \
+               " but should be if nargs is not None!"
```
bruntib, line 63:
```suggestion
               f"--analyzer-config or --checker-config value ({value}) is not a list," \
               " but should be if nargs is not None!"
```

analyzer/codechecker_analyzer/arg.py
```diff
@@ -36,6 +36,38 @@ def __call__(self, parser, namespace, value, option_string=None):
...
         ordered_checkers = namespace.ordered_checkers
+        # Map each checker to whether its enabled or not.
         ordered_checkers.append((value, self.dest == 'enable'))
 
         namespace.ordered_checkers = ordered_checkers
+
+
+class OrderedConfigAction(argparse.Action):
+    """
+    Action to store --analyzer-config values. These may come from many sources,
```
bruntib, line 47:
```suggestion
    Action to store --analyzer-config and --checker-config values. These may come from many sources,
```

bruntib (approved):

PR #3663, opened 2022-04-27
Find unused port for codechecker server

### LINKED ISSUE #3657: CodeChecker server - implement listening on 'port 0' behavior
CodeChecker server uses port number 8001 by default. If this port is used already then another port needs to be given. Using port 0 could mean finding a free port automatically: https://stackoverflow.com/questions/1075399/how-to-bind-to-any-available-port/1077305

tal66:
Hi, this closes #3657. 
if default port is not available, and the user didn't ask for a specific one - let the os find a port.

### DIFF
--- a/web/server/codechecker_server/cmd/server.py
+++ b/web/server/codechecker_server/cmd/server.py
@@ -43,6 +43,7 @@ from codechecker_web.shared import webserver_context, database_status, \
     host_check, env
 
 LOG = logger.get_logger('server')
+DEFAULT_SERVER_PORT = 8001
 
 
 def get_argparser_ctor_args():
@@ -113,7 +114,7 @@ def add_arguments_to_parser(parser):
                         type=int,
                         dest="view_port",
                         metavar='PORT',
-                        default=8001,
+                        default=DEFAULT_SERVER_PORT,
                         required=False,
                         help="The port which will be used as listen port for "
                              "the server.")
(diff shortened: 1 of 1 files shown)


bruntib (changes requested):
Thanks for the improvement, it's a very nice job!
I think, however, that finding a free port shouldn't be an automatic thing even in case of using the default port. If the default port (8001) is not free then it still should be an error. We should only get a free port from `os` only if the user explicitly provides port 0.

tal66:
so the user can just use `-p 0` in that case

bruntib:
Oh, sorry, the specification in the original ticket is not accurate. You implemented it according to the specification, but I think we shouldn't consider port 8001 as a special case. We should choose one of these alternatives:

- If the user provides a port number then we use that. If it's not free then an error message should be printed. If the user doesn't provide any port number then we use 8001 which also can be occupied in which case an error message is printed. If user provides 0 as port number explicitly then then os should find a free one.
- If the user doesn't provide a port number then CodeChecker should work as if 0 was given. With other words: 0 could be the default port number.

In my previous comment I suggested the first option, because many other CodeChecker commands are assuming this port number. For example `CodeChecker store` stores to port 8001 by default, so that's why I's vote on the first alternative.

tal66:
without my commit these are the scenarios:
`-p 0` - os finds a port
`-p <num>` - use *num* or error
` `  (no port request) - use 8001 or error
which match your first alternative.

bruntib:
Yes, that message would be a good solution. I've checked the original version of this ticket in an internal ticket system and this task belonged to an old CodeChecker version. Port 0 really works as expected by now, sorry, I haven't checked it before creating this ticket. My bad :(

tal66:
no problem. I canceled the changes, and just changed the message

bruntib (approved):
Thank you!

PR #3667, opened 2022-04-29
Fix typo

crazyBaboon:


crazyBaboon:
@bruntib 
Voilá!

bruntib:
Thanks so much, it's a nice improvement in the docs!
I just opened the .md file with LibreOffice in linux and checked the red underlines. But in linux almost all text editors are using the same dictionary file, so it can be opened with anything. My dream is to try out Grammerly one day :) It checks not only typos but real grammar issues which I believe can be even closer to native English.

bruntib (approved):

PR #3669, opened 2022-05-06
[cmd] Get failed files with "CodeChecker cmd runs"

bruntib:
CodeChecker cmd runs is added `--details` flag which extends JSON
output with the list of files that are failed to analyze.


web/server/codechecker_server/api/report_server.py
```diff
@@ -3107,7 +3107,7 @@ def getAnalysisStatistics(self, run_id, run_history_id):
 
             for stat, run_id in query:
                 failed_files = zlib.decompress(stat.failed_files).decode(
-                    'utf-8').split('\n') if stat.failed_files else None
+                    'utf-8').split('\n') if stat.failed_files else []
```
vodorok, line 3110: Whats the reason of the empty list?

bruntib: This is assigned to the variable `failed_files` which is the part of the JSON dumped to the command line interface. On the user side it will be easier to handle this JSON member without the extra check on whether this is `None`.

PR #3674, opened 2022-05-25
[CLI] Allow Unix shell-style wildcards in --trim-path-prefix

Szelethus:
Exactly what is says on the tin!

### DIFF
--- a/tools/report-converter/codechecker_report_converter/util.py
+++ b/tools/report-converter/codechecker_report_converter/util.py
@@ -11,6 +11,8 @@ import logging
 import os
 import portalocker
 import sys
+import fnmatch
+import re
 
 from typing import Dict, List, Optional, TextIO
 
@@ -76,9 +78,19 @@ def trim_path_prefixes(path: str, prefixes: Optional[List[str]]) -> str:
         if not prefix.endswith('/'):
             prefix += '/'
 
-        if path.startswith(prefix) and (not longest_matching_prefix or
-                                        longest_matching_prefix < prefix):
-            longest_matching_prefix = prefix
+        regex_str = fnmatch.translate(prefix)
+        assert regex_str[-2:] == '\\Z', \
+               r'fnmatch.translate should leave \\Z at the end of the matcher!'
+
+        prefix_matcher = re.compile(regex_str[:-2])
+
+        matches = prefix_matcher.match(path)
+        if matches:
+            matching_prefix = matches[0]
+            if not longest_matching_prefix or \
+                    len(longest_matching_prefix) < len(matching_prefix):
+
+                longest_matching_prefix = matching_prefix
 
     # If no prefix found or the longest prefix is the root do not trim the
     # path.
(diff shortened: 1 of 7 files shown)


tools/report-converter/codechecker_report_converter/util.py
```diff
@@ -76,8 +78,16 @@ def trim_path_prefixes(path: str, prefixes: Optional[List[str]]) -> str:
...
+        regex_str = fnmatch.translate(prefix)
+        if (regex_str[-2:] != '\\Z'):
+            print(regex_str[-2:])
+        assert regex_str[-2:] == '\\Z', \
+               r'fnmatch.translate should leave \\Z at the end of the matcher!'
+
+        prefix_matcher = re.compile(regex_str[:-2])
+
+        if prefix_matcher.match(path) and \
+           (not longest_matching_prefix or longest_matching_prefix < prefix):
```
bruntib, line 90:
```suggestion
           (not longest_matching_prefix or len(longest_matching_prefix) < len(prefix)):
```
It is interesting that earlier we checked alphabetical order on matching path prefixes. According to the specification we should check their lengths. Would you agree with this fix, or am I missing something?

bruntib (changes requested):

tools/report-converter/tests/unit/util/test_trim_path_prefix.py
```diff
@@ -57,3 +57,12 @@ def test_prefix_in_filename(self):
                          trim_path_prefixes(test_path, ['/a/b/common']))
         self.assertEqual('common.txt',
                          trim_path_prefixes(test_path, ['/a/b/']))
+
+    def test_prefix_blob(self):
+        test_path = '/a/b/common.txt'
+        self.assertEqual('common.txt', trim_path_prefixes(test_path, ['/*/b']))
+        self.assertEqual('common.txt', trim_path_prefixes(test_path, ['/a/*']))
+        self.assertEqual('b/common.txt',
+                         trim_path_prefixes(test_path, ['/*/']))
```
bruntib, line 66: This test case doesn't pass, because `trim_path_prefixes()` function returns `common.txt` on these inputs. The test fails, but it is not presented probably due to another style checker:
```
codechecker_report_converter/util.py:91:17: E125 continuation line with same indent as next logical line
```
By the way I think it is correct that this function returns 'common.txt' on these parameters, because we should mimic shell behavior where `*` substitutes multiple characters, including `/`.

Please, extend the documentation of `trim_path_prefixes()` with the behavior of joker characters (`?`. `*`), and also mention this in the command-line documentations:
https://github.com/Ericsson/codechecker/blob/6499dfdca5e13ca356081d368fe4cd38a93f5d19/docs/analyzer/user_guide.md?plain=1#L1626
https://github.com/Ericsson/codechecker/blob/6499dfdca5e13ca356081d368fe4cd38a93f5d19/docs/web/user_guide.md?plain=1#L457
https://github.com/Ericsson/codechecker/blob/6499dfdca5e13ca356081d368fe4cd38a93f5d19/analyzer/codechecker_analyzer/cmd/check.py#L707
https://github.com/Ericsson/codechecker/blob/6499dfdca5e13ca356081d368fe4cd38a93f5d19/analyzer/codechecker_analyzer/cmd/parse.py#L182
https://github.com/Ericsson/codechecker/blob/6499dfdca5e13ca356081d368fe4cd38a93f5d19/web/client/codechecker_client/cmd/store.py#L240

bruntib (changes requested):

bruntib (approved):

PR #3677, opened 2022-06-02
[bugfix] Minor fix in a string compare operation

bruntib:


steakhal:
Why do the github actions fail on this one?

bruntib:
Seems like an environment issue: something can't be downloaded. I'll try rerunning it later.

whisperity (approved):
It seems this mistake was made because Python is optimising away copies if strings are the same, but of course `!=` is safer because that does not rely on optimisations.

```py
>>> nargs = input()
*
>>> nargs
'*'
>>> id(nargs)
140219812695040
>>> id('*')
140219812695040
```

PR #3691, opened 2022-06-23
[fix] Fix remote diff behavior

bruntib:
When two runs are compared then reports should be considered as closed
even if their review status is false positive or intentional. Generally
any event that sets "fixed_at" date of a report makes it closed. This
should be taken into account when comparing two runs remotely.


dkrupp (approved):
LGTM

PR #3696, opened 2022-06-27
[doc] Updating docs about diff command.

bruntib:


dkrupp (changes requested):
Please rephrase the "we" sentences into passive.
Explain differences a bit better.

bruntib:
Thanks for the review! I fixed it based on your comments and I also extended the GUI user guide about the behavior of diff multiplicity.

docs/usage.md
```diff
@@ -497,6 +497,87 @@ CodeChecker cmd diff --basename ./reports --newname ./reports-ctu --new
  You can also use JSON format output of `CodeChecker cmd diff` command if
  you want to use it for further processing by an other program.
 
+### Run comparison
```
dkrupp, line 500: The content of this whole section could be moved under https://github.com/Ericsson/codechecker/blob/607c27991aa8173a21e3bc440e1adabb6c038a7b/docs/usage.md#diffs-between-runs

because it is about the same topic, but more detailed and with example.

And you could refer to this section at the end of https://github.com/Ericsson/codechecker/blob/607c27991aa8173a21e3bc440e1adabb6c038a7b/docs/usage.md#using-diff

dkrupp (changes requested):
Please update the TOC (regenerate it with vscode).

There is a similar section already about calculating diffs (How diffs between runs are calculated). I suggest moving the new content under there.

dkrupp (approved):
LGTM

PR #3698, opened 2022-06-28
[CLI] Don't ignore compiler warnings, even if clangtidy:take-config-from-directory=true is specified

Szelethus:
clangtidy:take-config-from-directory is an analyzer config that makes
clang-tidy get its arguments from a .clang-tidy file, and only from that
file. What this implies, is that all other options on the command line
for clang tidy will be ignored.

The problem is that this also ignores compiler warnings. Now, compiler
warnings can be emitted through clang-tidy's bug report generation
infrastructure, but are NOT clang-tidy checks. They an option to the the
*compiler*, not *clang-tidy*. For this reason, we should not ignore
them, however we invoke clang-tidy.


analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -203,6 +179,28 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
             elif state == CheckerState.disabled:
                 checkers.append('-' + checker_name)
 
+        # -checks=-clang-analyzer-* option is added to the analyzer command by
+        # default except when all analyzer config options come from .clang-tidy
+        # file. The content of this file overrides every other custom config.
```
bruntib, line 184: This comment is no longer valid. Do we want this to be valid, or should we remove this comment?

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -203,6 +179,28 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
...
+        if config.analyzer_config.get('take-config-from-directory') == 'true':
```
bruntib, line 185: We should initiate a discussion about the necessity of this `take-config-from-directory` option. It was introduced because the default value of `--analyzer-config` flag is `['clang-HeaderFilterRegex=.*']`. This option enables clang-tidy reports in header files too. However, this option should be added where clang-tidy options are handled under `analyzers` directory. Merging clang-tidy configs from command line, config file, etc. should be simple, because this is just yaml data.

bruntib (changes requested):

analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -37,6 +37,18 @@ class CheckerState(Enum):
...
+def get_compiler_warning_name(checker_name):
+    """
+    Removes 'W' or 'Wno' from the compiler warning name, if this is a
+    compiler warning. Returns None otherwise.
+    """
+    # Checker name is a compiler warning.
+    if checker_name.startswith('W'):
+        return checker_name[4:] if \
+            checker_name.startswith('Wno-') else checker_name[1:]
+    return None
```
bruntib, line 49: Functions in Python return `None` by default, conventionally we don't make this explicit.

bruntib (approved):

PR #3704, opened 2022-07-06
Document misc-confusable-identifiers & modernize-macro-to-enum

sylvestre:


sylvestre:
@csordasmarton @Xazax-hun any issue with this PR?
thanks

Xazax-hun (approved):

Xazax-hun:
The PR looks good to me, but I no longer have rights to merge it. @bruntib  could you take a look?

whisperity:
In general this should be okay (and I could actually merge it once I'm at a computer, I think), but I would prefer if we did a swoop and documented all new checkers (LLVM 15.0 just got tagged, anyway...) in one go.

Also, I think we've a new regulation on release cadence that we need to sometimes do a bugfix-only release that does **not** change how the analysers work or are driven, so this might be a blocker wrt. merging this on master. It could be that the current `master` is for `X.Y.Z` instead of `X.(Y+1).0` until the CodeChecker (aka Python-only) release is done.

So I hope @bruntib will enlighten us. ☺

bruntib:
Feel free to merge anything to master. When we deploy a bugfix release then we create a branch from the last main release and cherry-pick the necessary commits.
Another related patch: #3709

sylvestre:
can it be merged please ? :)

config/labels/analyzers/clang-tidy.json
```diff
@@ -5078,6 +5078,13 @@
       "profile:sensitive",
       "severity:LOW"
     ],
+    "misc-confusable-identifiers": [
+      "doc_url:https://clang.llvm.org/extra//clang-tidy/checks/misc/confusable-identifiers.html",
```
whisperity, line 5082: There is an extra `/` here that does not appear in other links.

config/labels/analyzers/clang-tidy.json
```diff
@@ -5078,6 +5078,13 @@
...
+      "profile:default",
+      "profile:extreme",
+      "profile:sensitive",
+      "severity:LOW"
```
whisperity, line 5086: Should something that detects a CVE be of *LOW* severity (and not included in the security profile)?

sylvestre: It has a CVE but it can be argued about the risk....
I don't mind changing it if you prefer :)
I don't have strong opinion about it

whisperity: Meh, even the CVE site linked in the Tidy docs starts with `DISPUTED`. It's a strange attack vector from the get-go, e.g., for me the following two lines render **completely** differently:

```cpp
int fo;
int 𝐟o;
```

Visual render example

Think this will be okay as is.

whisperity (approved):

config/labels/analyzers/clang-tidy.json
sylvestre: If you are curious to know more :
https://opensourcesecuritypodcast.libsyn.com/episode-296-is-trojan-source-a-vulnerability
;)

PR #3706, opened 2022-07-11
[bugfix] Headers with same name in different dirs

bruntib:
On server side "path hash" prevents multiple storage of a report in a
header when the header is included in multiple translation units.
Otherwise a report would be stored as many times as many TU contains
it. Earlier this "path hash" contained only the header file's name, so
the reports of another header with the same name but different
directory weren't stored.


dkrupp (approved):
LGTM

PR #3707, opened 2022-07-13
Add support for pch paramteres of clang-extdef-mapping

martong:


analyzer/tests/libtest/codechecker.py
```diff
@@ -142,6 +142,20 @@ def check_force_ctu_capable(is_capable):
...
 
 
+def check_force_extdef_mapping_can_read_pch():
+    """
+    Returns True if extdef mapping is force enabled by the
+    'CC_TEST_FORCE_EXTDEF_MAPPING_CAN_READ_PCH' environment variable.
+    """
+    try:
+        return bool(util.strtobool(
+            os.environ['CC_TEST_FORCE_EXTDEF_MAPPING_CAN_READ_PCH']))
```
martong, line 152: @bruntib TODO set a JIRA ticket to turn this variable on before our next internal release, when we rebase Clang.

martong:
The overall CSA runtime (with the default profile) improved roughly with 1-2 % in case of C++, and with 0-1% in case of C.
image

bruntib (changes requested):
Technically it looks good to me, though, it's not clear why relative/absolute paths matter. Maybe it would be worth for a comment about its reasons.

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_autodetection.py
```diff
@@ -171,6 +171,23 @@ def is_ctu_capable(self):
...
+        """
+
+        tool_path = self.mapping_tool_path
+
+        if not tool_path:
+            return False
+
+        help = invoke_binary_checked(tool_path, ['-help'], self.environ)
+        LOG.debug("output of clang-extdef-mapping\n %s" % help)
+        if help is False:
```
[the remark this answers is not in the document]
martong: I've got this from pylint `E712: comparison to False should be 'if cond is False:' or 'if not cond:'` so I've rather change it to `if not help`.

analyzer/codechecker_analyzer/analyzers/clangsa/config_handler.py
```diff
@@ -63,3 +66,7 @@ def analyzer_plugins(self):
     @property
     def ctu_capability(self):
         return CTUAutodetection(self.analyzer_binary, self.environ)
+
+    def detect_extdef_mapping_can_read_pch(self):
```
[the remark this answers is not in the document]
martong: Okay, I've changed it to update the member variable.

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_autodetection.py
```diff
@@ -171,6 +171,23 @@ def is_ctu_capable(self):
...
+            return False
+        return 'ast files' in help
```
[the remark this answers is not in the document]
martong: This is the content.
```
OVERVIEW:
This tool collects the USR name and location of external definitions in the source files (excluding headers).
Input can be either source files that are compiled with compile database or .ast files that are created from clang's -emit-ast option.

USAGE: clang-extdef-mapping [options] <source0> [... <sourceN>]

OPTIONS:

Generic Options:

  --help                      - Display available options (--help-hidden for more)
  --help-list                 - Display list of available options (--help-list-hidden for more)
  --version                   - Display the version of this program

clang-extdefmapgen options:

  --extra-arg=<string>        - Additional argument to append to the compiler command line
  --extra-arg-before=<string> - Additional argument to prepend to the compiler command line
  -p=<string>                 - Build path

-p <build-path> is used to read a compile command database.

        For example, it can be a CMake build directory in which a file named
        compile_commands.json exists (use -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
        CMake option to get this output). When no build path is specified,
        a search for compile_commands.json will be attempted through all
        parent paths of the first input file . See:
        https://clang.llvm.org/docs/HowToSetupToolingForLLVM.html for an
        example of setting up Clang Tooling on a source tree.

<source0> ... specify the paths of source files. These paths are
        looked up in the compile command database. If the path of a file is
        absolute, it needs to point into CMake's source tree. If the path is
        relative, the current working directory needs to be in the CMake
        source tree and the file must be in a subdirectory of the current
        working directory. "./" prefixes in the relative files will be
        automatically removed, but the rest of a relative path must be a
        suffix of a path in the compile command database.
```

> Is this two-word string quite reliable?
I think it is, these two words are included only in the new version that is capable of handling pch files.

martong:
> Technically it looks good to me, though, it's not clear why relative/absolute paths matter. Maybe it would be worth for a comment about its reasons.

Thanks for the review!  Relative/absolute paths matter because the "on-demand" and the "pch" based CTU uses the paths differently. This is documented in `ctu_manager.py` at the function `func_map_list_src_to_ast`:
```
    """ Turns textual function map list with source files into a
    mapping from mangled names to mapped paths, which can be absolute paths to
    the original source files if ctu_on_demand is True, or relative path
    segments to AST-dump files that reside in CTU-DIR directory otherwise. """
```

martong:
Ping

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
```diff
@@ -134,20 +138,33 @@ def generate_ast(triple_arch, action, source, config, env):
...
 
 
-def ast_dump_path(source_path):
+def ast_dump_path(path, config, triple_arch):
     """ AST-dump based analysis uses preprocessed paths, here the path prefix
     'ast' and the filename suffix '.ast' is added to the name of the original
     source file. """
 
+    # If clang-extdef-mapping can read a pch then the file entry is an absolute
+    # path to the ast file. Let's transform that to a relative path from the
```
whisperity, line 147: This is only true, and only needed, for Clang PCHs, right? (I am concerned about accidentally running into GCC PCHs.)

martong: Right.

martong:
Ping @Szelethus

Szelethus (changes requested):
This absolutely cannot go in without a proper PR description and commit message. Please make the review, and the subsequent archeology easier.

Does this have a corresponding LLVM commit and/or phabricator revision? Is there an issue or something for this? Why do we need this? Whats to gain?

martong:
> This patch is about supporting https://reviews.llvm.org/D128704 from Clang.
> 
> When doing CTU analysis setup we parse the .cpp to get an .ast and then we run clang-extdef-mapping on the .cpp file later. This is sub-optimal since we have to re-parse the file again.
> 
> With this patch we can now run clang-extdef-mapping directly on the .ast file. That saves some time.

This patch is about supporting https://reviews.llvm.org/D128704 from Clang.

When doing CTU analysis setup we parse the .cpp to get an .ast and then
we run clang-extdef-mapping on the .cpp file later. This is
sub-optimal since we have to re-parse the file again.

With this patch we can now run clang-extdef-mapping directly on
the .ast file. That saves some time.

martong:
> This absolutely cannot go in without a proper PR description and commit message. Please make the review, and the subsequent archeology easier.
> 
> Does this have a corresponding LLVM commit and/or phabricator revision? Is there an issue or something for this? Why do we need this? Whats to gain?

This patch is about supporting https://reviews.llvm.org/D128704 from Clang.

When doing CTU analysis setup we parse the .cpp to get an .ast and then
we run clang-extdef-mapping on the .cpp file later. This is
sub-optimal since we have to re-parse the file again.

With this patch we can now run clang-extdef-mapping directly on
the .ast file. That saves some time.

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_autodetection.py
```diff
@@ -171,6 +171,23 @@ def is_ctu_capable(self):
...
+
+        tool_path = self.mapping_tool_path
+
+        if not tool_path:
+            return False
+
+        help = invoke_binary_checked(tool_path, ['-help'], self.environ)
+        LOG.debug("output of clang-extdef-mapping\n %s" % help)
+        if not help:
+            return False
```
Szelethus, line 188: This is not a hill I intend to die on, but this kind of implies that the binary is broken, right? Shouldn't this warrant an error, or raising an exception?

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
```diff
@@ -134,20 +138,33 @@ def generate_ast(triple_arch, action, source, config, env):
                   cmdstr, err)
 
 
-def ast_dump_path(source_path):
+def ast_dump_path(path, config, triple_arch):
     """ AST-dump based analysis uses preprocessed paths, here the path prefix
     'ast' and the filename suffix '.ast' is added to the name of the original
     source file. """
 
+    # If clang-extdef-mapping can read a pch then the file entry is an absolute
```
Szelethus, line 146:
```suggestion
    # If clang-extdef-mapping can read a pch then parameter 'path' is an absolute
```

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
```diff
@@ -134,20 +138,33 @@ def generate_ast(triple_arch, action, source, config, env):
...
+                                            triple_arch, 'ast'))
+
     # Normalize path on Windows OS.
-    path = os.path.splitdrive(source_path)[1]
+    path = os.path.splitdrive(path)[1]
     # Make relative path out of absolute.
     path = path[1:] if path[0] == os.sep else path
+
+    # If clang-extdef-mapping can read a pch then the file entry already has
+    # the .ast suffix.
```
Szelethus, line 160: Is this something worth `assert`in then?

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
```diff
@@ -168,18 +185,24 @@ def func_map_list_src_to_ast(func_src_list, ctu_on_demand):
...
-
+        mapped_path = path if config.ctu_on_demand else \
+            ast_dump_path(path, config, triple_arch)
         func_ast_list.append(mangled_name + " " + mapped_path)
     return func_ast_list
 
 
-def get_extdef_mapping_cmd(action, config, source, func_map_cmd):
+def get_extdef_mapping_cmd(
+        action, config, source, func_map_cmd, triple_arch, makefile):
```
Szelethus, line 195: So `makefile` is supposed to be a boolean parameter, not a path to a Makefile? What is its purpose then? Can you explain a bit more? Can we rename it to something that sounds a bit more boolean-like? (e.g. `has_makefile`)

analyzer/codechecker_analyzer/makefile.py
```diff
@@ -138,8 +138,12 @@ def __get_ctu_pre_analysis_cmds(self, action):
         cmds.append(' '.join(cmd))
 
         # Get command to create CTU index file.
+        # FIXME Even if clang-extdef-mapping supports pch files as parameters,
+        # feed the source files rather. This way, the below 'sed' command is
+        # still meaningful.
```
Szelethus, line 143: You lost me here a little. What is the context?

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
martong: Right.

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
martong: No, because there is another case we want to support: when the clang-extdef-mapping cannot read a pch (i.e. older Clang).

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
martong: Ok, `has_makefile` or `for_makefile_generation` is really better.

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
martong: I don't understand. In this line, we add a the ".ast" suffix for the path, but only when necessary.

analyzer/codechecker_analyzer/makefile.py
martong: > The FIXME states the current hack, or what the fix should be? I'm not in the clear. 
`makefile.py` had been architectured with a technical debt.

> Shouldn't the FIXME state that we should use the ast_dump_path method? 
Ok, I am updating the comment with that.

> What is preventing us from doing it? Why does this divergence exist?
That would require further changes in `makefile.py` which is out of the scope of this patch, in my opinion. Addressing the mentioned technical debt should be done in a follow up patch.

> Also, what you wrote here could be in the code as well, right?
Could be, yes.

analyzer/codechecker_analyzer/analyzers/clangsa/ctu_manager.py
```diff
@@ -171,18 +188,25 @@ def func_map_list_src_to_ast(func_src_list, ctu_on_demand):
...
+        mapped_path = path if config.ctu_on_demand else \
+            ast_dump_path(path, config, triple_arch)
         func_ast_list.append(mangled_name + " " + mapped_path)
     return func_ast_list
 
 
-def get_extdef_mapping_cmd(action, config, source, func_map_cmd):
+def get_extdef_mapping_cmd(
+        action, config, source, func_map_cmd, triple_arch,
+        for_makefile_generation):
```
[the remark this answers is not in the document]
martong: Ok, added. Plus I've added some more explanatory comments to the command assembly.

analyzer/tests/libtest/codechecker.py
Szelethus: This should be done before the PR is merged.

Szelethus (approved):
LGTM! Very nice! :smiling_face_with_tear:

PR #3713, opened 2022-07-27
[bugfix] Review status count error fix

bruntib:
The SQL statement querying review status counts when filtering reports
in GUI with uniqueing was invalid.


vodorok (changes requested):
I would rephrase the commit message:
The SQL statement querying review status counts was invalid, when filtering reports in rhe GUI with uniqueing enabled.

Otherwise LGTM. Good catch!  :rocket:

dkrupp (approved):

vodorok (approved):

PR #3714, opened 2022-08-07
[config] Document `bugprone-assignment-in-if-condition` and `misc-const-correctness`

sylvestre:


config/labels/analyzers/clang-tidy.json
```diff
@@ -138,6 +138,13 @@
       "sei-cert:pre31-c",
       "severity:MEDIUM"
     ],
+    " bugprone-assignment-in-if-condition": [
```
bruntib, line 141: Extra space in the checker name.

config/labels/analyzers/clang-tidy.json
```diff
@@ -5085,6 +5092,12 @@
       "profile:sensitive",
       "severity:LOW"
     ],
+    "misc-const-correctness": [
+      "doc_url:https://clang.llvm.org/extra/clang-tidy/checks/misc/const-correctness.html",
+      "profile:extreme",
```
whisperity, line 5097: I'm not sure whether we should add this to any profile by default, but the severity looks okay.

sylvestre:
As you wish :)
I don't have any opinion on this!

config/labels/analyzers/clang-tidy.json
dkrupp: I think it is okay in the extreme profile

dkrupp (approved):
LGTM. thanks for the patch.

PR #3702, opened 2022-06-30
[doc] Clang-tidy doc URLs changed

bruntib:
The location of clang-tidy documentations changed. Until now they were
at the same directory level, but they have been organized to separate
directories based on their main group names.

Fixes #3701

PR #3715, opened 2022-08-07
[config] Fix the 404 in the Clang-Tidy docs

sylvestre:
Most of them moved in:
https://reviews.llvm.org/D128072


bruntib:
Hi,
Changing URLs is done in this PR: #3702. We should choose one and check which is more complete. Some additional checkers have been added in the meantime.

bruntib:
#3702 Has been merged.

PR #3717, opened 2022-08-18
[feat] Find by files anywhere on bugpath

bruntib:
In the GUI the set of reports can be filtered by filename or source
component. However, these filters are concerning the last bug point,
i.e. one can list the set of reports ending in a specific file.

In this patch a new filter option has been introduced which returns all
reports where the file is involved at any part of the bug path.


dkrupp:
Clicking on a report in report-detail view, I get this error
image

dkrupp:
report counts don't add up
image

dkrupp (approved):
This works as intended.

Please create a followup issue for this bug:
When the "anywhere on path" in checked, the file list and report counts are not updated:
image

PR #3726, opened 2022-08-22
[fix] File filter count for files with 0 reports

bruntib:
When filtering for reports that occur anywhere on the bugpath (i.e.
not just the last point is in a given file) then searching for files
containing no reports is necessary.


dkrupp (approved):
LGTM

PR #3727, opened 2022-08-22
[CI][analyzer] Run coverage check for Cppcheck checkers

### LINKED ISSUE #3723: [CI] Ensure that Cppcheck is also checked by the checker label coverage job

whisperity:
Enables the weekly "checker coverage" CI job for Cppcheck labels.

> Closes #3723.


whisperity:
Here you can see a version of the job, how it reports missing checkers. If Cppcheck cannot be built locally, the script falls back to the official *apt* source, which in the case of Ubuntu is a much older Cppcheck that has some checkers (`constArgument` and `operatorEq`) in it that are no longer available, and thus were reported:

https://github.com/whisperity/CodeChecker/runs/7950902750

image

bruntib (approved):

PR #3728, opened 2022-08-25
[ctu] Don't silently ignore exceptions thrown in pre-analysis

Szelethus:
Exceptions thrown in a thread pool by default stay in the thread pool
(won't be re-raised at the call to map_async or similar functions).
However, you can query whether an exception occured with the help of
AsyncResult.successful().

This patch is simple -- don't pretend that pre-analysis exceptions can
be ignored, just because they run asyncronously, exceptions there are
just as severe as elsewhere.

### DIFF
--- a/analyzer/codechecker_analyzer/pre_analysis_manager.py
+++ b/analyzer/codechecker_analyzer/pre_analysis_manager.py
@@ -206,13 +206,18 @@ def run_pre_analysis(actions, context, clangsa_config,
         #        proxy objects before passing them to other processes via
         #        map_async.
         #        Note that even deep-copying is known to be insufficient.
-        pool.map_async(pre_analyze, collect_actions)
+        result = pool.map_async(pre_analyze, collect_actions)
         pool.close()
     except Exception:
         pool.terminate()
         raise
     finally:
         pool.join()
+        # Return whether the call completed without raising an exception.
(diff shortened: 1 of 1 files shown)


analyzer/codechecker_analyzer/pre_analysis_manager.py
```diff
@@ -206,13 +206,14 @@ def signal_handler(signum, frame):
...
         #        Note that even deep-copying is known to be insufficient.
-        pool.map_async(pre_analyze, collect_actions)
+        result = pool.map_async(pre_analyze, collect_actions)
         pool.close()
     except Exception:
         pool.terminate()
         raise
     finally:
         pool.join()
+        assert result.successful()
```
whisperity, line 216: Python assertions are non-binding and depending on environment conditions and the running interpreter, might not be used. The most easy solution is running an "optimised build" of Python, i.e. `python -O foo.py`. This means that the user has a very broad and dangerously accidental control over whether this check is performed or not. Asserts in Python (and in Java) are non-idiomatic because the raised exception type (if *any* is created by the interpreter!) is too broad.

If we do not want to continue execution because of an issue, especially for debugging purposes, the information should be presented to the user so they know what happened and whether they can do to fix their environment.

bruntib (approved):

PR #3729, opened 2022-08-26
README.md: add `python3-setuptools` dependency

itzurabhi:
While running build steps inside `ubuntu:22.04` container, `python3-setuptools` is also required.


Szelethus (approved):
LGTM

PR #3731, opened 2022-08-30
[bugfix] Invalid SQL query when filter for file and source component.

bruntib:
The error happened when getFileCounts() function got a filter where both
"filepath" and "componentNames" fields were used.

### DIFF
--- a/web/tests/functional/report_viewer_api/test_report_counting.py
+++ b/web/tests/functional/report_viewer_api/test_report_counting.py
@@ -271,6 +271,30 @@ class TestReportFilter(unittest.TestCase):
         self.assertEqual(len(res), len(self.run1_files))
         self.assertDictEqual(res, self.run1_files)
 
+    def test_filter_by_file_and_source_component(self):
+        """
+        File and source component filter in getFileCounts().
+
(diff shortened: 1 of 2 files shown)


web/tests/functional/report_viewer_api/test_report_counting.py
```diff
@@ -271,6 +271,30 @@ def test_run1_all_file(self):
...
+        """
+        File and source component filter in getFileCounts().
+
+        Earlier this function resulted an SQL error due to an invalid SQL
+        statement (File table was ambiguously used, because it was joined
+        multiple times).
+
+        On the other hand we test here that getFileCounts() returns the number
+        of reports in all files regardless the filter fields. The reason is
+        that it wouldn't be possible on the GUI to display the options of file
```
Szelethus, line 284:
```suggestion
        that it wouldn't be possible on the GUI to display the options of the file
```

Szelethus (approved):
Please amend the commit message to describe that the specific error happened due to the mentioned querry returning two "File" objects, but `remove` only removed the first occurance. If that is done, please merge at your leisure.

PR #3735, opened 2022-09-02
[bugfix] Fix system include path order

bruntib:
The implicit include paths should be added after the user-defined flags
and include paths even in the pre-ctu phase, otherwise the build commands
will not be equivalent with the original ones.

### DIFF
--- a/analyzer/tests/functional/ctu/test_ctu.py
+++ b/analyzer/tests/functional/ctu/test_ctu.py
@@ -9,6 +9,7 @@
 """ CTU function test."""
 
 
+import subprocess
 import glob
 import os
 import shutil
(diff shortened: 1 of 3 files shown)


vodorok (approved):
Looks good to me, @martong could you please also take a look?

analyzer/tests/functional/ctu/test_ctu.py
```diff
@@ -91,6 +92,25 @@ def test_ctu_loading_mode_requires_ctu_mode(self):
         self.assertEqual(1,
                          run(cmd, cwd=self.test_dir, env=self.env).returncode)
 
+    @skipUnlessCTUCapable
+    def test_system_headers_last(self):
+        """
+        Test if implicit system include paths are appended after all
+        user-defined flags and include paths in the pre-ctu phase.
```
Szelethus, line 99: You describe the "what", but not the "why". Why is this important?

Szelethus (changes requested):
Could you expand on why is this needed? I find the summary lacking a bit.

Szelethus (changes requested):
Please update the commit message with the bit you wrote in the test file. You used the phrase "The two build commands will not be equivalent", but I don't think that captures what the problem really is. The problem is that we will analyze the files with the wrong set of build commands, because we messed up processing the build command json. When you mention equivalence, I immediately think of storing elements in a set, trying to query a specific build from a database, or something algebraic like that.

Szelethus (approved):
Please add that FYI comment to `analyzer/codechecker_analyzer/analyzers/clangsa/ctu_triple_arch.py` as well. Otherwise LGTM, merge at your leisure.

PR #3736, opened 2022-09-02
[fix] Replace Binary with LargeBinary

bachp:
Binary is deprecated in SQLAlchemy and was removed in 1.4.

It was just an alias for LargeBinary


vodorok (changes requested):
Hi @bachp! I apologize for neglecting this PR for this long. Great work so far, but we also have schema migrations under `web/server/codechecker_server/migrations/report/versions`. Could you please check, what happens in the case of a db migration in which we want to migrate an existing database with Binary columns to a database with LargeBinary columns?

whisperity:
Dear @bachp,

2519a2fffd64e18152ea749affd5eb012b979bee was merged in #4089. :clap:

PR #3737, opened 2022-09-05
[fix] Fix compiler params with space for Cppcheck

vodorok:
Compiler paramters in format of -I </path/to/file> were handled
incorrectly in Cppcheck. Only the -I part were included in the analyzer
command resulting in a failing analysis.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
@@ -84,9 +84,25 @@ class Cppcheck(analyzer_base.SourceAnalyzer):
         pass
 
     def parse_analyzer_config(self):
-        "Parses a white listed compiler flags"
+        """
+        Parses a set of a white listed compiler flags.
+        Cppcheck can only use a subset of the parametes
+        found in compilation commands.
+        These are:
+        * -I: flag for specifing include directories
+        * -D: for build time defines
+        * -U: for undefining names
+        * -std: The languange standard
+        Any other parameter different from the above list will be dropped.
+        """
         params = []
-        interesting_option = re.compile("^-[I|U|D].*$")
+        interesting_option = re.compile("-[I|U|D].*")
+        # the std flag is different. the following are all valid flags:
+        # * --std c99
+        # * -std=c99
+        # * --std=c99
+        # BUT NOT: -std c99
+        std_regex = re.compile("-?-std.*")
         for i, analyzer_option in enumerate(self.buildaction.analyzer_options):
             if interesting_option.match(analyzer_option):
                 params.extend([analyzer_option])
@@ -99,11 +115,18 @@ class Cppcheck(analyzer_base.SourceAnalyzer):
                     params.extend(
                         [self.buildaction.analyzer_options[i+1]]
                     )
-            if analyzer_option.startswith("-std"):
-                standard = analyzer_option.split("=")[-1] \
-                    .lower().replace("gnu", "c")
+            if std_regex.match(analyzer_option):
+                standard = ""
+                if "=" in analyzer_option:
+                    standard = analyzer_option.split("=")[-1]
+                # Handle space separated parameter
+                # The else clause is never executed until a log parser
+                # limitation is addressed, as only this "-std=xxx" version
+                # of the paramter is forwareded in the analyzer_option list.
+                else:
+                    standard = self.buildaction.analyzer_options[i+1]
+                standard = standard.lower().replace("gnu", "c")
                 params.extend(["--std=" + standard])
-        LOG.debug(params)
         return params
 
     def construct_analyzer_cmd(self, result_handler):
--- a/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
@@ -83,6 +83,29 @@ class Cppcheck(analyzer_base.SourceAnalyzer):
         """
         pass
 
+    def parse_analyzer_config(self):
+        "Parses a white listed compiler flags"
+        params = []
+        interesting_option = re.compile("^-[I|U|D].*$")
+        for i, analyzer_option in enumerate(self.buildaction.analyzer_options):
+            if interesting_option.match(analyzer_option):
+                params.extend([analyzer_option])
+                # The above extend() won't properly insert the analyzer_option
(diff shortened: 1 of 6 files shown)


analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
```diff
@@ -83,6 +83,29 @@ def get_analyzer_mentioned_files(self, output):
         """
         pass
 
+    def parse_analyzer_config(self):
+        "Parses a white listed compiler flags"
```
Szelethus, line 87: ...a __set of__* white listed...

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
```diff
@@ -83,6 +83,29 @@ def get_analyzer_mentioned_files(self, output):
...
+        params = []
+        interesting_option = re.compile("^-[I|U|D].*$")
```
Szelethus, line 89: Is this the set? Can we do something more formal, like for each interesting option, something like this:

```python
# The above extend() won't properly insert the analyzer_option
# in case of the following format -I <path/to/include>.
# The below check will add the next item in the
# analyzer_options list if the parameter is specified with a
# space, as that should be actual path to the include.
interesting_options.append("-I.*");
# The bloopity blooper needs special handling.
interesting_options.append("-bLoOoOp.*");
```

Szelethus (changes requested):
Your title and summary mentions a bug fix, but your patch adds, or at the very least, refactors a feature, the way to whitelist compiler flags. I find that confusing.

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
Szelethus: Make a judgement call on this, I don't insist if its very incovenient.

bruntib, line 89:
```suggestion
        interesting_option = re.compile("-[I|U|D]")
```
Maybe it's just a style issue, so feel free to ignore it, but this is supposed to be the same. `re.match()` finds the pattern at the beginning of the string by default.

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
```diff
@@ -83,6 +83,29 @@ def get_analyzer_mentioned_files(self, output):
...
+                # The above extend() won't properly insert the analyzer_option
+                # in case of the following format -I <path/to/include>.
+                # The below check will add the next item in the
+                # analyzer_options list if the parameter is specified with a
+                # space, as that should be actual path to the include.
+                if interesting_option.match(analyzer_option).span() == (0, 2):
+                    params.extend(
+                        [self.buildaction.analyzer_options[i+1]]
+                    )
+            if analyzer_option.startswith("-std"):
```
bruntib, line 102: I think `--std` can also occur in build commands.

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
```diff
@@ -83,6 +83,29 @@ def get_analyzer_mentioned_files(self, output):
...
+                standard = analyzer_option.split("=")[-1] \
+                    .lower().replace("gnu", "c")
+                params.extend(["--std=" + standard])
+        LOG.debug(params)
```
bruntib, line 106: I think we should add some label to it so we know what is printed here. I also think that we should print debug logs only if we can gain information from it when some debugging is needed with the users. I think we should eliminate debug logs that are only useful during development. The assembled build command can be seen even now in debug logs.

bruntib (changes requested):

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
vodorok: I think the part ".*" that matches to any sequence is needed, before the dollar sign, because on line 98 the matched span length is used to determine if it is paramater that is defined with a space. A span of (0, 2) means it is -I /abc/ and -I/abc/ will be matched as a span of (0, 5). Maybe the dot wildcard without the kleene star is enough. I'll check.

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
vodorok: Added explanation.

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
vodorok: True, this is a leftover development debug print.

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
vodorok: With the added explanation, I think it will suffice.

bruntib: Oh, you're right :)

bruntib (approved):

vodorok:
Don't merge this yet, I need to verify some things about the --std flag.

analyzer/tests/functional/analyze_and_parse/test_files/Makefile
```diff
@@ -45,3 +45,7 @@ cppcheck_args:
 	$(CXX) -w cppcheck_args.cpp -o /dev/null
 cppcheck_undef_args:
 	$(CXX) -w cppcheck_args.cpp -UHAVE_NULL_DEREFERENCE -o /dev/null
+cppcheck_include:
+	$(CXX) -w cppcheck_include.cpp -I ./includes/ -o /dev/null
+cppcheck_undef_include:
+	$(CXX) -w cppcheck_include.cpp -I./includes/ -U HAVE_NULL_DEREFERENCE -o /dev/null
```
dkrupp, line 51: Please add and example for -D, --std=

dkrupp (changes requested):
add some more tests.

analyzer/tests/functional/analyze_and_parse/test_files/Makefile
dkrupp: will be added in a a followup patch

dkrupp (approved):

PR #3741, opened 2022-09-14
[analyzer] Tests for Cppcheck standard handling

vodorok:
Analyzer tests were missing for the cppcheck standard forwarding. The standard translation is also tested in case of a gnu** standard is given.
Also added a small change in the business logic for handling the standard parameter. This should make it more robust against future refactoring.

### DIFF
--- a/analyzer/tests/functional/analyze/test_analyze.py
+++ b/analyzer/tests/functional/analyze/test_analyze.py
@@ -938,6 +938,57 @@ class TestAnalyze(unittest.TestCase):
 
         self.assertEqual(errcode, 0)
 
+    def test_cppcheck_standard(self):
+        """
+        Testing the standard ("--std") compiler paramater translation
+        to cppcheck input parameter.
+        Cppcheck can only understand a subset of the available options.
+        """
+        build_json = os.path.join(self.test_workspace, "cppcheck_std.json")
+        analyze_cmd = [self._codechecker_cmd, "analyze",
+                       build_json,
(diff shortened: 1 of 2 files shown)


analyzer/tests/functional/analyze/test_analyze.py
```diff
@@ -938,6 +938,57 @@ def test_multiple_invalid_checker_names(self):
...
+
+        out = subprocess.run(analyze_cmd,
+                             cwd=self.test_dir,
+                             # env=self.env,
+                             stdout=subprocess.PIPE).stdout.decode()
+
+        # Test correct handover.
+        self.assertTrue("--std=c99" in out)
+
+        # Cppcheck does not support qnu variants of the standards,
```
Szelethus, line 972:
```suggestion
        # Cppcheck does not support gnu variants of the standards,
```

Szelethus (commented):
How does that change make it "more robust in the future"? Also, I'm not sure wether we have a tradition for marking patches NFC in the title when no changed are intended, but maybe we should ;)

I realize I'm nitpicking, but please state that "standard" in the commit message and the summary refers to `--std` (as you have stated in the comments in the code).

analyzer/tests/functional/analyze/test_analyze.py
vodorok: Fixed, thanks

vodorok:
Updated the title & message.

Szelethus (approved):
LGTM

PR #3746, opened 2022-09-19
[bugfix] Old client has different behavior with new server

bruntib:
This getDiffResultsHash() function is returning a set of reports based on what are they compared to in a "CodeChecker cmd diff" command. Earlier this function didn't consider false positive and intentional reports as closed reports. The client's behavior also changed from CodeChecker 6.20.0 and this behavior is adapted to the new server behavior. The problem is that the old client works correcly only with the old server. For this reason we are branching based on the client's version.

### DIFF
--- a/web/server/codechecker_server/api/report_server.py
+++ b/web/server/codechecker_server/api/report_server.py
@@ -1205,6 +1205,7 @@ class ThriftRequestHandler:
                  auth_session,
                  config_database,
                  package_version,
+                 client_version,
                  context):
 
         if not product:
@@ -1216,6 +1217,7 @@ class ThriftRequestHandler:
         self._auth_session = auth_session
         self._config_database = config_database
         self.__package_version = package_version
+        self.__client_version = client_version
         self._Session = Session
         self._context = context
         self.__permission_args = {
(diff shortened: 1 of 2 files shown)


web/server/codechecker_server/api/report_server.py
```diff
@@ -1599,10 +1615,15 @@ def getDiffResultsHash(self, run_ids, report_hashes, diff_type,
                     return []
 
                 base_hashes = session.query(Report.bug_id.label('bug_id')) \
-                    .outerjoin(File, Report.file_id == File.id) \
-                    .filter(
+                    .outerjoin(File, Report.file_id == File.id)
+
+                if minor >= 50:
```
dkrupp: Also please check for the major version too.
Should be >=6

web/server/codechecker_server/api/report_server.py
```diff
@@ -1630,9 +1651,15 @@ def getDiffResultsHash(self, run_ids, report_hashes, diff_type,
 
                     return new_hashes
             elif diff_type == DiffType.RESOLVED:
-                results = session.query(Report.bug_id) \
-                    .filter(or_(Report.bug_id.notin_(report_hashes),
-                                Report.fixed_at.isnot(None)))
+                results = session.query(Report.bug_id)
+
+                if minor >= 50:
```
dkrupp: check for major version too

web/server/codechecker_server/api/report_server.py
```diff
@@ -1641,10 +1668,16 @@ def getDiffResultsHash(self, run_ids, report_hashes, diff_type,
 
             elif diff_type == DiffType.UNRESOLVED:
                 results = session.query(Report.bug_id) \
-                    .filter(Report.bug_id.in_(report_hashes)) \
-                    .filter(Report.detection_status.notin_(
-                        skip_statuses_str)) \
-                    .filter(Report.fixed_at.is_(None))
+                    .filter(Report.bug_id.in_(report_hashes))
+
+                if minor >= 50:
```
dkrupp: check for major version too

dkrupp (approved):
LGTM

PR #3747, opened 2022-09-21
[bugfix] Different behavior of diff when run tags are provided

bruntib:
In a previous commit (f6d0fedaf14b583df7bd26078a8a22b557be57c6) we fixed a bug, which was about getting the difference of false positive reports in a baseline to a local directory. For further details see the message of that commit. However, another bug still remained in the system, when the diff command compared a report directory to a specific run tag. This case is fixed by this commit.


vodorok (approved):
LGTM

PR #3749, opened 2022-09-26
[bugfix] Don't update review status date

bruntib:
When a review status is set in the GUI then a new entry is inserted to review_statuses table. Every time the same report is stored, its review status date was updated. This date shouldn't be updated, because not the storage sets a review status rule's date.

### DIFF
--- a/web/tests/functional/review_status/test_review_status.py
+++ b/web/tests/functional/review_status/test_review_status.py
@@ -543,13 +543,28 @@ class TestReviewStatus(unittest.TestCase):
             "This is false positive.")
         self.assertIsNone(multi_unreviewed_report)
 
+        rule_filter = ReviewStatusRuleFilter(
+            reportHashes=[UNCOMMENTED_BUG_HASH])
+        review_status_rule_before = self._cc_client.getReviewStatusRules(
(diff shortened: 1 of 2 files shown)


web/tests/functional/review_status/test_review_status.py
```diff
@@ -543,14 +543,29 @@ def test_review_status_changes(self):
...
         # to the database with the same bug hashes. The review status of
         # reports without source code comment should get the default review
-        # status set earlier.
+        # status set earlier. In the meantime, the review status rule date must
+        # not change.
 
         test_project_name2 = 'review_status_change2'
         codechecker.store(codechecker_cfg, test_project_name2)
 
+        rule_filter = ReviewStatusRuleFilter(
```
vodorok, line 560: Rule filter could be reused from before?
Maybe give it a more distinct name like uncommented_rule_filter or something.

vodorok (approved):
LGTM

web/tests/functional/review_status/test_review_status.py
vodorok: As discussed ofline, leaving out the second might cause it to change meaning if a third on would be introduced.

Issue #3617, opened 2022-03-02
report-converter does not work with leak sanitizer when more than one leaks have been detected


jimis:
If more than one leaks have been detected, then report-converter prints:
```
[INFO] - No 'lsan' results can be found in the given code analyzer output.
```
To reproduce the issue see new files in #3613.

PR #3750, opened 2022-09-28
report-converter: Parse all leaks reported by LeakSanitizer

hannesweisbach:
Previously, report-converter was able to parse output by LeakSanitizer only, if exactly one leak was found. This issue was reported as #3617.

By changing the regex to find the beginning of a single leak - instead of finding the beginning of the entire LeakSanitizer report - report-converter is now able to parse and convert LeakSanitizer output with multiple leaks


hannesweisbach:
Just fixed the pylint warning of a too long line. I'm pretty sure that it makes the code actually /less/ readable and there is no technical reason for there to be only 79 columns, since we don't develop on VT100 (actual hardware) terminals anymore.
But your codebase, your rules ...

whisperity:
(The logic behind 80-col in today's world is that if you develop with a vertical split (either because you have to co-edit multiple files, or your non-active split auto-jumps to definition or documentation), you can fit two files properly on the screen. This would otherwise not be possible without forcing a local line wrap, which breaks consistency as such configuration isn't versioned... especially in a language like Python where indentation matters!)

hannesweisbach:
> That being said, the reason why Pylint would hard require 79 exclusive instead of 80 inclusive is beyond me.

It's probably counting from 0 😂

I fixed the broken tests (output changed). And I didn't realize some code was shared between sanitizer output parsers.

Anyway I just saw #3613. Maybe I should rebase onto that to take advantage of the new test cases?

Locally running the tests still fail on: "Get report data for plist which contains macro expansion. ... ERROR". But this is also the case on a clean master?

bruntib:
Thanks for the improvement, it's a nice solution!

bruntib (approved):

PR #3751, opened 2022-09-28
(FIPS) Added usedforsecurity=False to hashlib.md5(...)

tristanvdb:
Required on UBI-8 when host is FIPS compliant. It prevents:
> ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS

It might not be portable. I use yum's python38 package on UBI-8.


whisperity:
Definitely not portable. This flag is only available starting with Python 3.9 officially, so we might need to instead polyfill this around by hoisting the call to a comaptibility library that deals with catching the expression. Alternatively, it was suggested that using `.hash` and passing the flag to the constructor of the hash context object as a kwarg might not throw if the existence of the algorithm is not supported.

whisperity (changes requested):
@tristanvdb Thank you for noticing this and the patch! We do not have FIPS in the EU so this was all new to me that this might even be a problem.

Unfortunately, the official Python 3.8 version on an Ubuntu 20.04 LTS system would be broken by this change:

```
❯ python3
Python 3.8.10 (default, Jun 22 2022, 20:18:18)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import hashlib
>>> hashlib.md5("hello".encode('utf-8'), usedforsecurity=False)
  File "<stdin>", line 1, in <module>
TypeError: openssl_md5() takes at most 1 argument (2 given)
```

(And to the best of my knowledge we also have to target Ubuntu 18.04 (Python 3.6) for a few systems, but this might be outdated info. Targeting 20.04 is definitely a must have.)

There are two ways to remediate this.

Either we have to explicitly create a new hash context object with the `hashlib.new()` which does accept (and most likely ignores) the `usedforsecurity` parameter (in its `**kwargs`), and then explicitly `.update()` and `.hexdigest()` the contents:

```
❯ python3
Python 3.8.10 (default, Jun 22 2022, 20:18:18)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import hashlib
>>> Ctx = hashlib.new('md5', usedforsecurity=False)
>>> Ctx.update("hello".encode('utf-8'))
>>> Ctx.hexdigest()
'5d41402abc4b2a76b9719d911017c592'
```

Unfortunately, according to the documentation:

> The named constructors are much faster than [new()](https://docs.python.org/3.8/library/hashlib.html#hashlib.new) and should be preferred.

Due to this, the better approach would be to create a compatibility function either in `codechecker_common/util.py` or make a new `codechecker_common/hash.py` and create an md5 wrapper that deals with catching the exception (and possibly caching the fact that the system is running with a FIPS kernel(?)) and running the appropriate routine.

tristanvdb:
What about something like below for codechecker_common/hash.py:
`
import hashlib
hashlib_md5_kwargs={'usedforsecurity':False}
def md5(utf8_str):
    global hashlib_md5_kwargs
    try:
        return hashlib.md5(utf8_str, **hashlib_md5_kwargs).hexdigest()
   except TypeError as e:
        # some log?
        hashlib_md5_kwargs={}
        return md5(utf8_str)
`
Should it include the call to `.encode('utf-8')`?
What would be a proper log command?

whisperity:
To keep it in line with the existing API, I think it's fine if the user is expected to `.encode()` first. Also, is a logging for that error really needed? It'd just litter the output for everyone running Python 3.8 and stock LTS Ubuntu/Debian versions. The entire reason for this function is to provide a compatibility jump point.

I do have fears about that tail recursion, though. If any other sort of `TypeError` happens, e.g. someone passes an `int` instead of string or something, it will just infinitely recurse...

tristanvdb:
I agree with the recursion. I'll simply duplicate the call to hashlib.
I'll make the changes and update the PR when I get some time.

tristanvdb:
I am having trouble building CodeChecker from source in a UBI-8 container.
Particularly, `make standalone_package` hangs on:
```
if ! pip3 install -r ./requirements.txt 2>&1 | grep -q "Requirement already satisfied (use --upgrade to upgrade): codechecker-api"; then \
        if [ "YES" = "YES" ]; then \
          cd server/vue-cli && npm install && rm -rf /codechecker/web/server//vue-cli/dist; \
        fi; \
fi
```
I installed node with `dnf module install -y nodejs:16`

codechecker_common/hash.py
```diff
@@ -0,0 +1,12 @@
+
+import hashlib
+
+hashlib_md5_kwargs={'usedforsecurity':False}
```
bruntib, line 4: I'm not too familiar with this `usedforsecurity` flag. Do I understand correctly, that it changes the hash algorithm, but the resulting hash value will always be the same?
Could you add some comments about the motivation of this flag?
Is this relevant only for md5 or some other hash algorithms too?
This might be a theoretical question, but what happens if someone isn't aware of the existence of this utility function? Can we force its usage over `hashlib.md5`?

bruntib (approved):
Thanks for the contribution to CodeChecker. I posted some questions, but most of them are based on the fact that I'm not too familiar with this extra argument :)

Issue #3748, opened 2022-09-23
CodeChecker parse produces invalid HTML (Uncaught SyntaxError: "" literal not terminated before end of script)


dilyanpalauzov:
**Describe the bug**
CodeChecker parse --export html produces, among others, this file: [parser_html.zip](https://github.com/Ericsson/codechecker/files/9632470/parser_html.zip)

When loaded in firefox it reports on line 733: Uncaught SyntaxError: "" literal not terminated before end of script

Chromium reports on the same line “Uncaught SyntaxError: Invalid or unexpected token”.

**CodeChecker version**
```
[INFO 2022-09-23 12:35] - CodeChecker analyzer version:
---------------------------------------------------------------
Kind                 | Version                                 
---------------------------------------------------------------
Base package version | 6.20.0                                  
Package build date   | 2022-09-16T12:36                        
Git commit ID (hash) | 7d65d89c2a3fed84f395016d4c9815bab860d0b3
Git tag information  | 6.20                                    
---------------------------------------------------------------

[INFO 2022-09-23 12:35] - CodeChecker web version:
------------------------------------------------------------------------------
Kind                                | Version                                 
------------------------------------------------------------------------------
Base package version                | 6.20.0                                  
Package build date                  | 2022-09-16T12:36                        
Git commit ID (hash)                | 7d65d89c2a3fed84f395016d4c9815bab860d0b3
Git tag information                 | 6.20                                    
Server supported Thrift API version | 6.51                                    
Client Thrift API version           | 6.51                                    
------------------------------------------------------------------------------
```

dilyanpalauzov:
This HTML:
```html
<!DOCTYPE html><htlm>
  <head>
  </head>
  <body>
    Hello,
  <script>
    var f = "<script>var a ='bcd'</script><noscript>XXX</noscript>";
    console.log(f);
  </script>
Peter!
</body>
</html>
```
is invalid in both Firefox (Error message «Uncaught SyntaxError: "" literal not terminated before end of script») and Chromium (error message «Uncaught SyntaxError: Invalid or unexpected token»).

When the C-input contains `printf("</script>\n");` that input is literally inserted into `var data=…` by report.output.html, generating invalid HTML.

PR #3753, opened 2022-10-03
report output html: replace in var data=… </script> with &lt;/script&gt; to produce valid HTML

dilyanpalauzov:
Close https://github.com/Ericsson/codechecker/issues/3748

### DIFF
--- a/tools/report-converter/codechecker_report_converter/report/output/html/html.py
+++ b/tools/report-converter/codechecker_report_converter/report/output/html/html.py
@@ -6,6 +6,7 @@
 #
 # -------------------------------------------------------------------------
 
+import html
 import io
 import json
 import logging
@@ -171,7 +172,9 @@ class HtmlBuilder:
             file_content = InvalidFileContentMsg
 
         self.files[file.id] = {
-            'id': file.id, 'filePath': file.path, 'content': file_content}
+            'id': file.id, 'filePath': file.path,
+            'content': html.escape(file_content)
+        }
 
         return self.files[file.id]
 
--- a/tools/report-converter/codechecker_report_converter/report/output/html/static/js/bugviewer.js
+++ b/tools/report-converter/codechecker_report_converter/report/output/html/static/js/bugviewer.js
@@ -195,7 +195,9 @@ var BugViewer = {
 
     this._sourceFileData = file;
     this._filepath.innerHTML = file.filePath;
-    this._codeMirror.doc.setValue(file.content);
+    let e = document.createElement('div');
+    e.innerHTML = file.content;
+    this._codeMirror.doc.setValue(e.innerText);
     this._refresh();
   },
 


dilyanpalauzov:
This produces outputs of the form `#include &quot;acl.h&quot;` so is not best choice.

dilyanpalauzov:
Is this a CodeMirror 5 bug, when the html.escape above is not applied?

dilyanpalauzov:
The last change replaces `</script>` from input C-files with `&lt;/script&gt;`.  The literal `&lt;/script&gt;` does appear in the output, despite not being present in the input.

I think this is a good approach, compared to generating invalid HTML, as shown at https://github.com/Ericsson/codechecker/issues/3748#issuecomment-1266805179.

tools/report-converter/codechecker_report_converter/report/output/html/html.py
```diff
@@ -171,7 +173,9 @@ def _add_source_file(self, file: File) -> FileSource:
             file_content = InvalidFileContentMsg
 
         self.files[file.id] = {
-            'id': file.id, 'filePath': file.path, 'content': file_content}
+            'id': file.id, 'filePath': file.path,
+            'content': replace_script_tag.sub(r"&lt;/\1&gt;", file_content)
```
bruntib, line 177: Hi,
Thanks for the patch, this is a very nice catch.
Python has a module named `html` which provides escaping functionality for this situation. Would you think, this could be a more general solution to this issue?
```suggestion
            'content': html.escape(file_content)
```
Of course `html` needs to be imported instead of `re`;

Thanks again!

dilyanpalauzov:
I tried this in the first version of the patch: https://github.com/Ericsson/codechecker/commit/90af5ffc9a94295775909c778f1f2907b7b903a8

The result was, that from the C-input `#include <stdio.h>;`  was shown literally in the output as “#include &tl;stdio.h&gt;;` and I do not know how to tell CodeMirror to render `&tl;` from `var data={…}` as `<` instead of `&lt;`.

bruntib:
Hi,

The encoded HTML string can be decoded in the browser so they appear properly.

https://github.com/Ericsson/codechecker/blob/master/tools/report-converter/codechecker_report_converter/report/output/html/static/js/bugviewer.js#L198

This is the line which sets the content of CodeMirror. Something like this could be inserted before this line:

```js
var e = document.createElement('div');
e.innerHTML= file.content;
this._codemirror.doc.setValue(e.innerText);
```

This works together with the first version of your patch. What do you think?

dilyanpalauzov:
If it works, then apply it please.

bruntib:
It works for me. Will you implement it together with going back to `html.escape()` in this PR? Thank you!

bruntib:
No worries, the GUI tests are failing for a while. This is a known issue and we'll fix it. For this reason GUI tests are not set as required.
Thanks for your work, I merge it.

bruntib (approved):

PR #3760, opened 2022-10-14
[fix] Fix chunks function not working on certain iterables

vodorok:
When called with a set, the chunks function gave back as many chunks, as the length of the list.


vodorok:
Should I fix this on all of the calling sites?

bruntib (approved):

PR #3761, opened 2022-10-14
[test] Update to Ubuntu 20.04 in tests

bruntib:
18.04 has been deprecated in GitHub actions.

### DIFF
--- a/.github/workflows/install-deps.sh
+++ b/.github/workflows/install-deps.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 
 wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
-sudo add-apt-repository 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-14 main' -y
+sudo add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-14 main' -y
 
 sudo apt-get update -q
 
 sudo apt-get install \
-  g++-6 \
+  g++-9 \
   gcc-multilib \
   libc6-dev-i386 \
   libpq-dev \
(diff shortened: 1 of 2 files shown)


.github/workflows/install-deps.sh
```diff
@@ -1,12 +1,12 @@
...
 
 wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
-sudo add-apt-repository 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-14 main' -y
+sudo add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-14 main' -y
 
 sudo apt-get update -q
 
 sudo apt-get install \
-  g++-6 \
+  g++-9 \
```
whisperity, line 9: Note that Ubuntu images by GitHub Actions come with some pre-packaged compilers and such installed in `/usr/local/bin` which takes precedence over the version installed by apt!

This means that tests can subtly pass (due to a newer compiler) in the CI than what you would be able to have on a **real** stock Ubuntu machine!

(I've run into this problem multiple times with wanting to ensure old CMake support in my projects, etc.)

vodorok: A subsequent update-alternatives call where we set the newly installed g++ as default, should be do the trick?

bruntib: I think the goal of this installation is to have any gcc on the machine so we don't need to rely on the content of the GitHub image. Here I just upgraded the version number, because g++-6 is not available in Ubuntu 20.04.

vodorok (approved):
LGTM

.github/workflows/install-deps.sh
whisperity: @vodorok update-alternatives doesn't concern the /usr/local structure either. 😔

PR #3755, opened 2022-10-07
Document 'bugprone-suspicious-realloc-usage'

sylvestre:

PR #3769, opened 2022-10-18
[analyzer][clang][clang-tidy] Assign new check profiles

gamesh411:
Clang-16 introduced new ClangSA and Clang-Tidy checks. These are assigned to profiles.


gamesh411:
for ClangSA checkers, that have results on OS projects, here are the results:

https://codechecker.gic.ericsson.se/clangsa_checkers/reports?review-status=Unreviewed&review-status=Confirmed%20bug&detection-status=New&detection-status=Reopened&detection-status=Unresolved&run=%2a_rc_ctu&is-unique=off&diff-type=New&checker-name=alpha.unix.Errno

https://codechecker.gic.ericsson.se/clangsa_checkers/reports?review-status=Unreviewed&review-status=Confirmed%20bug&detection-status=New&detection-status=Reopened&detection-status=Unresolved&run=%2a_rc_ctu&is-unique=off&diff-type=New&checker-name=alpha.unix.cstring.UninitializedRead

dkrupp (approved):
LGTM

config/labels/analyzers/clang-tidy.json
```diff
@@ -5121,8 +5138,17 @@
       "doc_url:https://clang.llvm.org/extra/clang-tidy/checks/misc/misleading-bidirectional.html",
       "profile:default",
       "profile:extreme",
+      "profile:security",
       "profile:sensitive",
-      "severity:LOW"
+      "severity:HIGH"
+    ],
+    "misc-misleading-identifiers": [
+      "doc_url:https://clang.llvm.org/extra/clang-tidy/checks/misc/misleading-identifiers.html",
```
whisperity, line 5146: This is typod to be in the plural form, while the name of the check is just `misc-misleading-identifier`. Refer to the [CI job](https://github.com/Ericsson/codechecker/actions/runs/3271768229/jobs/5381949211) for the list of checkers still not having a label after this patch! :wink:

whisperity (changes requested):

gamesh411:
Please don't merge yet.
I'll fix the coverage test.

gamesh411:
Fixed the coverage issue, now every check is accounted for (one way or another in case of cppcheck).

config/labels/analyzers/clang-tidy.json
```diff
@@ -406,6 +420,13 @@
       "sei-cert:mem35-c",
       "severity:HIGH"
     ],
+    "bugprone-suspicious-realloc-usage": [
+      "doc_url:https://clang.llvm.org/extra/clang-tidy/checks/bugprone/suspicious-realloc-usage.html",
+      "profile:default",
+      "profile:extreme",
+      "profile:sensitive",
+      "severity:MEDIUM"
+    ],
```
whisperity, line 429: This was categorised as `severity:HIGH` in #3755.

whisperity (changes requested):
I've went over the proposed file manually and cross-referenced them with the documentation to suggest at least severities for checks in **Clang-Tidy**

# Abseil

## STYLE

These are all checks from the paper *Hyrum Wright: Incremental type migration using time algebra* where they move chrono-like stuff onto strong types.

 - abseil-duration-addition
 - abseil-duration-comparison
 - abseil-duration-conversion-cast
 - abseil-duration-division
 - abseil-duration-factory-float
 - abseil-duration-factory-scale
 - abseil-duration-subtraction
 - abseil-duration-unnecessary-conversion
 - abseil-time-comparison
 - abseil-time-subtraction
 - abseil-upgrade-duration-conversions

Project-specific generic style recommendations:

 - abseil-cleanup-ctad
 - abseil-no-internal-dependencies

## LOW

Checks that find potential performance issues but are not dangerous if not fixed.

 - abseil-redundant-strcat-calls
 - abseil-str-cat-append
 - abseil-faster-strsplit-delimiter
 - abseil-string-find-startswith
 - abseil-string-find-str-contains

## MEDIUM

A style-specific check that warns third-party code not to open a namespace that the Abseil library regulates as strictly as the Standard does with `std`.

 - abseil-no-namespace


# Android 
 
## HIGH

These are all checks that find obsolete POSIX calls that are **unsafe**, in almost all case due to leaking file descriptors on an `exec()`.

 - android-cloexec-accept
 - android-cloexec-accept4
 - android-cloexec-dup
 - android-cloexec-epoll-create1
 - android-cloexec-fopen
 - android-cloexec-inotify-init1
 - android-cloexec-memfd-create
 - android-cloexec-open
 - android-cloexec-pipe
 - android-cloexec-pipe2
 - android-cloexec-socket

## MEDIUM

This check finds a specific macro from kernel code that reruns syscalls until deterministically failed. If you do a comparison in the macro, it will not figure out the failure reason internally, so the macro is useless and the code doesn't do what you intended it to do.

 - android-comparison-in-temp-failure-retry

## LOW

Warning on an outdated API due to security reasons, but **this** check does not fix the security issue. It suggests a rewrite to `epoll_create1` which has another checked (aforementioned in the **HIGH** section!) which will suggest using close-on-exec to fix security issues.


 - android-cloexec-epoll-create
 - android-cloexec-creat 
 - android-cloexec-inotify-init


# Zircon

## STYLE

 - zircon-temporary-objects

gamesh411:
Raised the severity of the realloc check to HIGH
Applied the workflow fix

gamesh411:
I have lowered the severity of "android-cloexec-creat" from medium to low to match the suggestion.
All other severity values are new or are stricter than previously specified.

whisperity (commented):
The Clang-Tidy part is good to go, I believe.

whisperity (approved):
(Forgot to select :heavy_check_mark:...)

config/labels/analyzers/clangsa.json
```diff
@@ -326,6 +327,13 @@
       "doc_url:https://clang.llvm.org/docs/analyzer/checkers.html#alpha-webkit-uncountedlocalvarschecker",
       "profile:extreme"
     ],
+    "apiModeling.Errno": [
+      "profile:default",
+      "profile:extreme",
+      "profile:portability",
+      "profile:sensitive",
+      "severity:LOW"
+    ],
```
Szelethus, line 336: We should never explicitly mention apiModeling checkers, it might lead to them being disabled!

PR #3771, opened 2022-10-18
[fix] Speed up resolved diffing

vodorok:
This commit tries to address the everlasting diff runtime, when the report count is large (~60000) and the ReviewStatus count is also substantial.
Two main edits were made, one is in getLinesInSourceFileContents the other is in getReviewStatusRules api endpoints. In getLinesInSourceFileContents my goal was to combine the requests into chunks.
In getReviewStatusRules the goal was also the same, but here a monolithic request was chopped into chunks.

### DIFF
--- a/web/server/codechecker_server/api/report_server.py
+++ b/web/server/codechecker_server/api/report_server.py
@@ -19,6 +19,7 @@ import stat
 import time
 import zlib
 
+from copy import deepcopy
 from collections import defaultdict
 from datetime import datetime, timedelta
 from typing import Dict, List, Optional, Set, Tuple
@@ -2157,31 +2158,49 @@ class ThriftRequestHandler:
     @timeit
     def getReviewStatusRules(self, rule_filter, sort_mode, limit, offset):
         self.__require_view()
+        if not sort_mode:
+            sort_mode = ReviewStatusRuleSortMode(
+                type=ReviewStatusRuleSortType.DATE,
+                ord=Order.DESC)
 
         result = []
-        with DBSession(self._Session) as session:
-            if not sort_mode:
-                sort_mode = ReviewStatusRuleSortMode(
-                    type=ReviewStatusRuleSortType.DATE,
-                    ord=Order.DESC)
 
-            q = get_rs_rule_query(session, rule_filter, sort_mode)
+        # To avoid modifiying the collection due to chunking
+        rule_filter_copy = deepcopy(rule_filter)
+
+        def getRules(reportHashes=None):
+            if rule_filter and reportHashes:
+                rule_filter_copy.reportHashes = reportHashes
+            q = get_rs_rule_query(session, rule_filter_copy, sort_mode)
 
             if limit:
                 q = q.limit(limit).offset(offset)
 
-        for review_status, associated_report_count in q:
-            result.append(ReviewStatusRule(
-                reportHash=review_status.bug_hash,
-                reviewData=create_review_data(
-                    review_status.status,
-                    review_status.message,
-                    review_status.author,
-                    review_status.date,
-                    False),
-                associatedReportCount=associated_report_count))
+            for review_status, associated_report_count in q:
+                result.append(ReviewStatusRule(
+                    reportHash=review_status.bug_hash,
+                    reviewData=create_review_data(
+                        review_status.status,
+                        review_status.message,
+                        review_status.author,
+                        review_status.date,
+                        False),
+                    associatedReportCount=associated_report_count))
+            return result
 
-        return result
+        with DBSession(self._Session) as session:
+            if rule_filter and rule_filter.reportHashes:
+                # Diffing with a large ammount of report hashes passed in the
+                # filter (60K) caused a hanging report diffing.
+                # The probable cause of this is the ILIKE matching in the
+                # underlying logic. Chunking the request solves this issue.
+                for hash_chunk in util.chunks(
+                        rule_filter.reportHashes,
+                        SQLITE_MAX_VARIABLE_NUMBER):
+                    getRules(hash_chunk)
+                return result
+            else:
+                return getRules()
 
     @exc_to_thrift_reqfail
     @timeit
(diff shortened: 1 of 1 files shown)


web/server/codechecker_server/api/report_server.py
```diff
@@ -2534,20 +2557,44 @@ def getLinesInSourceFileContents(self, lines_in_files_requested, encoding):
...
-            for lines_in_file in lines_in_files_requested:
-                if lines_in_file.fileId is None:
-                    LOG.warning("File content requested without a fileId.")
-                    LOG.warning(lines_in_file)
-                    continue
-                sourcefile = session.query(File).get(lines_in_file.fileId)
-                cont = session.query(FileContent).get(sourcefile.content_hash)
+
+            contents = []
+            for chunk in util.chunks(
```
bruntib, line 2562: This loop could be merged with the other one in line 2585 and that way only `contents_to_file_id` would be needed and `contents` could be eliminated. It would half the memory usage.
By the way shouldn't it be `file_id_to_contents`?

web/server/codechecker_server/api/report_server.py
```diff
@@ -2534,20 +2557,44 @@ def getLinesInSourceFileContents(self, lines_in_files_requested, encoding):
...
+                                        f"File content "
+                                        "requested without fileId {l}")
+                                    for line in chunk]))
+                    .all())
+
+            lines_in_files_requested.sort(key=lambda line: line.fileId)
+            contents.sort(key=lambda c: c.id)
+
+            for file in lines_in_files_requested:
+                res[file.fileId]
```
bruntib, line 2581: You were right, this really inserts the given key with a default constructed value. However, this loop is still unnecessary, because the insertion happens in line 2596 anyways.

web/server/codechecker_server/api/report_server.py
```diff
@@ -2534,20 +2557,44 @@ def getLinesInSourceFileContents(self, lines_in_files_requested, encoding):
...
-                if lines_in_file.fileId is None:
-                    LOG.warning("File content requested without a fileId.")
-                    LOG.warning(lines_in_file)
-                    continue
-                sourcefile = session.query(File).get(lines_in_file.fileId)
-                cont = session.query(FileContent).get(sourcefile.content_hash)
+
+            contents = []
+            for chunk in util.chunks(
+                    lines_in_files_requested, SQLITE_MAX_VARIABLE_NUMBER):
```
bruntib, line 2563: Please, add a short comment which tells that the goal of this chunking is not for performance, but for being compatible with SQLite. However, I think it is a relevant question in many other queries where SQLite is used. We should find some general solution for handling SQLite limitations without complicating the code in case of PostgreSQL.

I'm also thinking if we really need to support large data in case of SQLite. We are proposing that DBMS for simplified CodeChecker usage with less amount of data.

bruntib (changes requested):

web/server/codechecker_server/api/report_server.py
```diff
@@ -2157,29 +2158,51 @@ def changeReviewStatus(self, report_id, status, message):
...
-            q = get_rs_rule_query(session, rule_filter, sort_mode)
+        # To avoid modifiying the collection due to chunking
+        rule_filter_copy = deepcopy(rule_filter)
+
+        def getRules(reportHashes=None):
+            nonlocal rule_filter
+            nonlocal sort_mode
+            nonlocal limit
+            nonlocal offset
+            nonlocal result
```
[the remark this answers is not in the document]
vodorok: True, removed these.

web/server/codechecker_server/api/report_server.py
```diff
@@ -2157,29 +2158,51 @@ def changeReviewStatus(self, report_id, status, message):
...
+                        review_status.message,
+                        review_status.author,
+                        review_status.date,
+                        False),
+                    associatedReportCount=associated_report_count))
+            return result
+
+        with DBSession(self._Session) as session:
+            if rule_filter and rule_filter.reportHashes:
+                for hash_chunk in util.chunks(
```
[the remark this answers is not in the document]
vodorok: Added a comment

web/server/codechecker_server/api/report_server.py
```diff
@@ -2157,29 +2158,51 @@ def changeReviewStatus(self, report_id, status, message):
...
+                        iter(rule_filter.reportHashes),
+                        SQLITE_MAX_VARIABLE_NUMBER):
+                    getRules(hash_chunk)
+                return result
+            else:
+                return getRules()
 
         return result
```
vodorok, line 2207: Will remove this return result

web/server/codechecker_server/api/report_server.py
vodorok: True, merged it in with the chunk loop/

vodorok: Also changed the variable name, I always struggle with these, thanks.

web/server/codechecker_server/api/report_server.py
vodorok: Removed it,

web/server/codechecker_server/api/report_server.py
vodorok: Added a comment.

I agrree with that, we should find a better way to handle these queries. Maybe with an enchanced ORM/Model layer, with proper methods to work with data could facade these from the API endpoint code.
Regarding your SQLite comment, isn't there a way to configure  SQLite to work with larger variable sizes in query? Maybe enfoce it somehow through CodeChecker. Or maybe give a warning about an unsupported version, which cannot handle large data?

web/server/codechecker_server/api/report_server.py
```diff
@@ -2157,31 +2158,49 @@ def changeReviewStatus(self, report_id, status, message):
...
 
-        return result
+        with DBSession(self._Session) as session:
+            if rule_filter and rule_filter.reportHashes:
+                # Diffing with a large ammount of report hashes passed in the
+                # filter (60K) caused a hanging report diffing.
+                # The probable cause of this is the ILIKE matching in the
+                # underlying logic. Chunking the request solves this issue.
+                for hash_chunk in util.chunks(
+                        iter(rule_filter.reportHashes),
```
bruntib, line 2198: No need to use `iter()` here, because it is applied in `util.chunk()`. I think we shouldn't expect the caller to know what kind of iterable can be given to `util.chunk()`.

web/server/codechecker_server/api/report_server.py
```diff
@@ -2534,20 +2553,38 @@ def getLinesInSourceFileContents(self, lines_in_files_requested, encoding):
...
-                    LOG.warning(lines_in_file)
-                    continue
-                sourcefile = session.query(File).get(lines_in_file.fileId)
-                cont = session.query(FileContent).get(sourcefile.content_hash)
-                lines = zlib.decompress(
-                    cont.content).decode('utf-8', 'ignore').split('\n')
-                for line in lines_in_file.lines:
+
+            # This will contain all the lines for the given fileId
+            contents_to_file_id = defaultdict(lambda: list())
```
bruntib, line 2558:
```suggestion
            contents_to_file_id = defaultdict(list)
```
`list` in itself is a callable.

bruntib (changes requested):

web/server/codechecker_server/api/report_server.py
vodorok: True, this modification was around the time we did the chunks fix. Will remove.

web/server/codechecker_server/api/report_server.py
vodorok: Yep, forgot this one.

bruntib (approved):

PR #3776, opened 2022-10-22
Typos a/an

dilyanpalauzov:
Replace SSL → TLS.

### DIFF
--- a/analyzer/tools/build-logger/src/ldlogger-logger.c
+++ b/analyzer/tools/build-logger/src/ldlogger-logger.c
@@ -128,7 +128,7 @@ static void writeActions(FILE* stream_, char const* wd_, const LoggerVector* act
 
   fprintf(stream_, "]");
 
-  /* An fclose also closes the file descriptor, so wo only call an fflush
+  /* fclose() also closes the file descriptor, so we only call fflush()
    * here */
   fflush(stream_);
 }
--- a/docs/web/user_guide.md
+++ b/docs/web/user_guide.md
@@ -13,7 +13,7 @@
         - [Server Configuration (Authentication and Server Limits)](#server-configuration-authentication-and-server-limits)
         - [Database Configuration](#database-configuration)
       - [Master superuser and authentication forcing](#master-superuser-and-authentication-forcing)
-      - [Enfore secure socket (SSL)](#enfore-secure-socket-ssl)
+      - [Enfore secure socket (TLS/SSL)](#enfore-secure-socket-ssl)
       - [Managing running servers](#managing-running-servers)
       - [Manage server database upgrades](#manage-server-database-upgrades)
     - [`store`](#store)
@@ -171,7 +171,7 @@ optional arguments:
   -f CONFIG_DIRECTORY, --config-directory CONFIG_DIRECTORY
                         Directory where CodeChecker server should read server-
                         specific configuration (such as authentication
-                        settings, and SSL certificates) from.
+                        settings, and TLS/SSL certificates) from.
                         (default: /home/<username>/.codechecker)
   --host LISTEN_ADDRESS
                         The IP address or hostname of the server on which it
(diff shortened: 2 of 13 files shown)


analyzer/tools/build-logger/src/ldlogger-logger.c
```diff
@@ -128,7 +128,7 @@ static void writeActions(FILE* stream_, char const* wd_, const LoggerVector* act
 
   fprintf(stream_, "]");
 
-  /* An fclose also closes the file descriptor, so wo only call an fflush
+  /* fclose() also closes the file descriptor, so we only call fflush()
```
dilyanpalauzov, line 131: > fclose() also closes the file descriptor, so we only call fflush() here

Here it might be meant, that fclose() also FLUSHES the file descriptor.

dilyanpalauzov: No, this is not meant here.

docs/web/user_guide.md
```diff
@@ -332,17 +332,17 @@ root account arguments:
                         not require authentication otherwise.
 ```
 
-#### Enfore secure socket (SSL)
+#### Enfore secure socket (TLS)
```
whisperity, line 335: I'd not **replace** SSL with TLS but if we want to mention TLS then let's mention both. The terminology common to many people involved in running servers and web connection is still "SSL", and not saying "TLS", no matter how wrong this is. If people are just searching the documentation how to use SSL, they will no longer have findings related to this (to some very important) feature.

dilyanpalauzov:
I have replaced SSL with TLS/SSL.

There is no such thing like SSL certificate or TLS certificate, but PEM or DER certificate.  The documentation says now TLS certificate.

whisperity:
> There is no such thing like SSL certificate or TLS certificate

While this might be true if you are talking on a highly technical level, it's unfortunately not true for people who are administrators or users of web technologies. In fact, [DigiCert](http://www.digicert.com/what-is-an-ssl-certificate), [Cloudflare](http://www.cloudflare.com/en-gb/learning/ssl/what-is-an-ssl-certificate/), [Go Daddy](http://www.godaddy.com/en-uk/web-security/ssl-certificate) and [VeriSign](http://www.verisign.com/en_US/website-presence/online/ssl-certificates/index.xhtml) all call the entity we are talking about `SSL Certificate`, and in fact when you go as much down (in terms of search engine results, after having bypassed the ads) to reach [*Let's Encrypt!*](https://letsencrypt.org/), they are the ones who say `TLS Certificate` on their site.

bruntib (approved):
@whisperity feel free to merge this PR if you think this SSL/TLS description is correct. Thanks for the patch and the review!

dilyanpalauzov:
> Have you found any more accidental typos?

No

docs/web/user_guide.md
```diff
@@ -171,7 +171,7 @@ optional arguments:
   -f CONFIG_DIRECTORY, --config-directory CONFIG_DIRECTORY
                         Directory where CodeChecker server should read server-
                         specific configuration (such as authentication
-                        settings, and SSL certificates) from.
+                        settings, and TLS/SSL certificates) from.
                         (default: /home/<username>/.codechecker)
```
whisperity, line 175: These texts should be in synch with the contents of the Python script that generates the help when the user says `--help`. Please add the same change to

https://github.com/Ericsson/codechecker/blob/01a042ce6442f66d007fbca750951f6f904d350c/web/server/codechecker_server/cmd/server.py#L89-L97

Then we can merge! :smile:

whisperity (changes requested):
&nbsp;

whisperity (approved):

whisperity:
@Szelethus Perhaps this would be worthwhile to cherry-pick to **6.21**. No need for a release notes entry.

PR #3777, opened 2022-10-24
[refactor] Change the regex search to string search

vodorok:
The regex search() which intent to find codechecker source code comments are swappable with a much faster string.find() method. This will greatly improve performance of retrieving source code comments from the source files.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/source_code_comment_handler.py
+++ b/tools/report-converter/codechecker_report_converter/source_code_comment_handler.py
@@ -34,7 +34,8 @@ def contains_codechecker_comment(fp):
     pos_before_read = fp.tell()
     if pos_before_read != 0:
         fp.seek(0)
-    match = re.search(".*codechecker_.*", fp.read())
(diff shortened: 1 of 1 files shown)


vodorok:
Created some profiling of the massStoreRun API endpoint on the serve before and after the patch.
These are the stats of the stored report folder:
```
--------------------------------------------------------
Number of processed analyzer result files        | 698  
Number of analyzer reports                       | 49916
Number of source files                           | 645  
Number of source files with source code comments | 645  
Number of blame information files                | 0    
--------------------------------------------------------
----=================----
[INFO 2022-10-25 00:20:15] - Compressing report zip file...
[INFO 2022-10-25 00:20:17] - Compressing report zip file done (89.9MiB / 4.7MiB).
```
Before patch 71.6s:
image


After patch 30,6s:
image
The runtime of the modified function can be seen in detail here:
image

bruntib:
Wow, this is a great improvement! :astonished:

bruntib (approved):

Issue #2289, opened 2019-08-25
CodeChecker disables apiModeling by default *CRITICAL*


Szelethus:
**Describe the bug**
CodeChecker runs the analyzer with the flag `--analyzer-no-default-checks`, resulting in apiModeling being disabled by default.

**CodeChecker version**
```
---------------------------------------------------------------
Kind                 | Version                                 
---------------------------------------------------------------
Base package version | 6.10.1                                  
Package build date   | 2019-08-22T21:20                        
Git commit ID (hash) | 65d42ed762814d942546a27121b8fc4fefdd655f
Git tag information  | 6.10.1                                  
---------------------------------------------------------------

CodeChecker web version:
------------------------------------------------------------------------
Kind                          | Version                                 
------------------------------------------------------------------------
Base package version          | 6.10.1                                  
Package build date            | 2019-08-22T21:20                        
Git commit ID (hash)          | 65d42ed762814d942546a27121b8fc4fefdd655f
Git tag information           | 6.10.1                                  
Configuration schema          | v6.0                                    
Database schema               | v6.0                                    
Server supported API (Thrift) | 6.20                                    
Client API (Thrift)           | 6.20                                    
------------------------------------------------------------------------
```

**Expected behavior**
The way CodeChecker invokes the analyzer is totally out of whack. There is no sane explanation for what we're doing right now. I find the addition of `--analyzer-no-default-checks` in particular extremely concerning.

I expect CodeChecker to lean on the analyzer configuring itself a lot more. Don't use `--analyzer-no-default-checks`. Don't disable anything purely because it wasn't explicitly enabled.

Szelethus:
I pinned this because I'm legitimately worried. apiModeling is a super impactful package, and it seems like we've been missing out on it for a long while. This also implies that we haven't tested an important part of the analyzer well enough in our own playground. I also feel uncomfortable with our nonsensical invocation -- I see how it was needed back then due to how dependencies work (not that it solved the issue, but I feel for the desire to do something about it), but it's well overdue to get this right.

Szelethus:
I've been digging deeper into this and I'm fairly positive that considering the **current** state of the analyzer, the way we enable/disable checkers is fundamentally flawed, but I'm none the wiser as to how we should go about this.

[...]

PR #3778, opened 2022-10-25
[analyzer][cfg] Don't acknowledge apiModeling checkers

Szelethus:
As I complained in
https://github.com/Ericsson/codechecker/issues/2289 back in the day, apiModeling checkers should NEVER be explicitly disabled or enabled. CodeChecker seems to somewhat agree, because all apiModeling checkers we do have records on are present in all profiles. This is still incorrect though.

ApiModeling checkers are similar to core checkers, namely that if you disable them, you cannot expect the analyzer to function properly. However, we have even fewer excuses to disable them, because they are incapable of emitting reports (there is an assert that fires if they would).

However, if they are a part of profile, they can be unintentionally disabled through them.

Removing them from this file wouldn't prevent powerusers from disabling them however, just prevents unintentional misuse.

### DIFF
--- a/config/labels/analyzers/clangsa.json
+++ b/config/labels/analyzers/clangsa.json
@@ -327,51 +327,6 @@
       "doc_url:https://clang.llvm.org/docs/analyzer/checkers.html#alpha-webkit-uncountedlocalvarschecker",
       "profile:extreme"
     ],
-    "apiModeling.Errno": [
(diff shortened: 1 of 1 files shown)


bruntib:
Are these checkers implicitly enabled by clang?

bruntib (approved):

PR #3780, opened 2022-10-27
Issue a warning about this release being only an RC

Szelethus:
This is the output when this commit is rebased on branch release-v6.21.0-rc1.

image

### DIFF
--- a/analyzer/codechecker_analyzer/cmd/analyzer_version.py
+++ b/analyzer/codechecker_analyzer/cmd/analyzer_version.py
@@ -32,6 +32,9 @@ class Version:
         self.git_hash = context.package_git_hash
         self.git_tag = context.package_git_tag
 
+    def is_release_candidate(self):
+        return 'rc' in self.version
+
     def to_dict(self) -> Dict[str, str]:
         """ Get version information in dictionary format. """
         return {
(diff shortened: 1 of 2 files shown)


Szelethus:
The flush is needed because I suspect that LOG.warning outputs to stderr, which resulted in stdout being flushed AFTER the warning was issues, which looked dumb.

analyzer/codechecker_analyzer/cmd/analyzer_version.py
```diff
@@ -32,6 +32,9 @@ def __init__(self):
         self.git_hash = context.package_git_hash
         self.git_tag = context.package_git_tag
 
+    def is_release_candidate(self):
+        return 'rc' in self.version
```
bruntib, line 36: Shouldn't it be
```suggestion
        return 'rc' in self.version and self.version['rc']
```
?

And this function could go under some codechecker_common/util module, I think.

bruntib (changes requested):
I have a minor comment, and please, check out the failing style tests. Thanks.

analyzer/codechecker_analyzer/cmd/analyzer_version.py
bruntib: Oh, this is not a dict, but a string. Sorry, then this comment is irrelevant.

Szelethus: self.version is a string here, I believe

Szelethus: This would indeed be better in codechecker_common, in fact, this entire version chicanery could be, but that's outside the scope of this patch :)

bruntib (approved):

PR #3781, opened 2022-11-02
[analyzer] Change the log level to debug when overwriting plist files

Szelethus:
When CodeChecker does an analysis, it collects the results in a folder in plist files. However, when a new analysis is run where the result folder is the same, FOR EACH RESULT FILE a warning is issued about it being overwritten.

This is completely overkill, especially when this is the expected behaviour, like in the case of incremental analysis.

I changed these to debug level. I didn't change the 'info' level for when the '-c' flag is used (that also issues a log message, but in a lower severity), because that feels appropriate.


bruntib (approved):

bruntib:
Please, check the liters' results, after fixing the formatting issues, it can be merged.

PR #3783, opened 2022-11-05
[gui] Set "anywhere on path" in URL

bruntib:
File filter and source component filters have an "anywhere on report path" option in the web UI. The state of this option is now saved in the URL.

### DIFF
--- a/web/server/vue-cli/src/components/Report/ReportFilter/Filters/BaseSelectOptionFilter.mixin.js
+++ b/web/server/vue-cli/src/components/Report/ReportFilter/Filters/BaseSelectOptionFilter.mixin.js
@@ -46,7 +46,7 @@ export default {
 
     getIconClass() {},
 
-    initByUrl() {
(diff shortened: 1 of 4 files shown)


web/server/vue-cli/src/components/Report/ReportFilter/Filters/BaseSelectOptionFilter.mixin.js
```diff
@@ -46,7 +46,7 @@ export default {
 
     getIconClass() {},
 
-    initByUrl() {
+    initOptionsByUrl() {
```
Szelethus, line 49: More like `initCheckboxOptionsByUrl`?

Szelethus (approved):
LGTM! I had a single inline comment, I leave it to you whether you think that should be changed or not.

Szelethus (approved):
LGTM

vodorok (approved):
LGTM

PR #3786, opened 2022-11-08
[analyzer] Ignore flag '-mword-relocations'

Szelethus:
Exactly what it says on the tin! This flag, as I understand it, only affects the backend of the compiler.


dkrupp (approved):
LGTM

PR #3791, opened 2022-11-11
[test] Fix fixit test

bruntib:
In some environments/Python versions the standard input given to subprocess.Popen is handled differently. This fix uses the standard input parameter in a way that it works in every environment.


vodorok (approved):
LGTM, might be worth to try subprocess.DEVNULL in the future, when the stdin is ignored.

PR #3796, opened 2022-11-21
Further enhancements to speed up the store procedure

vodorok:
This pull request tries to optimize the store process by trimming down on the python access layer of SQLAlchemy.
Consists of two parts:
1. Only get the review status rules from the source code in the initial time of the report storage. After all the reports were added to the database, in a separate steps first get all applying review status rules from the server, then do a "bulk update" of all the newly stored reports, where a GUI based review status rule applies.
2. Cache the severity of the reports. Before this patch, a relatively expensive search of the checker labels were done, for every report being stored. Due to the nature of the analysis results, the number of reports greatly outweigh the number of type of reports (checkers), by caching this severity level can improve the store time.

Profiled the storage time of CodeChecker (master) before and after the modifications:
The reports folder was made with master clang + master cppcheck with master CodeChecker, with the --enable-all flag:
`Number of processed analyzer result files  | 1047 `
`Number of analyzer reports                        | 184471`

Reports were stored into empty databases.
On the labels, the colors are preserved for the sets. (in case the labels are hidden on the overview labels.)

Baseline measurements:
Mass storage time was 1823 secs.
image
The above label shows the full profile of the entire API call.
The below label shows the `__add_report` method. Please not the `checker_labels.py(severity)` call batch of 145 seconds.
image
The following label shows the detailed view of the get_review_status call batch:
image
Here the magenta bar represents the parsing of the source files (parse_codechecker_review_comment), the runtime of that was: 140 sec. It is faster than the previous release, due to fixes applied: #3777
The other call batches are python (and SQLA) overhead of the per report severity query.

Measurements after modification:
Mass storage time was 1313 secs
image
Please note the absence of the get review status rules call batch on the above label.
The orange section after the report_file.py is the remnant of the slow review status retrieval logic. Parsing the file is still needed, (parse_codechecker_review_comment) 
image

The last labels shows the impact of severity levels cache:
This is 1000x times faster (on large report counts)
image
image

All in all I've managed to shave off ~28% percent of the run time. (1823s -> 1313s), Which can be pretty significant when the store times are scratching the one hour mark.

The above figures clearly show, that the same improvement could be done for the `__add_report` method. (Cache the new reports, and do a bulk update after the setup. With this approach the bulk update step of this patch could be incorporated into the bulk insert.)

An interesting (and needed in my opinion) measurement would be to do the same measurements with a database heavily populated with review status rules.

### DIFF
--- a/web/server/codechecker_server/api/mass_store_run.py
+++ b/web/server/codechecker_server/api/mass_store_run.py
@@ -862,8 +862,7 @@ class MassStoreRun:
                 report: Report) -> SourceReviewStatus:
             """
             Return the review status set in the source code belonging to
-            the given report whether the review status comes from
-            a source code comment.
+            the given report.
 
             - Return the review status if it is set in the source code.
             - If the review status is ambiguous (i.e. multiple source code
--- a/web/server/codechecker_server/api/mass_store_run.py
+++ b/web/server/codechecker_server/api/mass_store_run.py
@@ -828,6 +828,10 @@ class MassStoreRun:
                         macro.message, file_path_to_id[macro.file.path],
                         db_report.id, data_type))
 
+                if report.annotations:
+                    self.__validate_and_add_report_annotations(
+                        session, db_report.id, report.annotations)
+
             session.flush()
 
         except Exception as ex:
@@ -982,9 +986,6 @@ class MassStoreRun:
                 rs_from_source, detection_status, detected_at,
                 run_history_time, analysis_info, analyzer_name, fixed_at)
 
-            if report.annotations:
-                self.__validate_and_add_report_annotations(
-                    session, report_id, report.annotations)
             self.__new_report_hashes[report.report_hash] = \
                 rs_from_source.status
             self.__already_added_report_hashes.add(report_path_hash)
@@ -1037,8 +1038,6 @@ class MassStoreRun:
                     f"'{value}' has wrong format. '{key}' annotations must be "
                     f"'{ALLOWED_ANNOTATIONS[key]['display']}'."
                 )
-
-        session.flush()
 
     def __store_reports(
         self,
--- a/web/server/codechecker_server/api/mass_store_run.py
+++ b/web/server/codechecker_server/api/mass_store_run.py
@@ -39,8 +39,8 @@ from ..database.config_db_model import Product
 from ..database.database import DBSession
 from ..database.run_db_model import AnalysisInfo, AnalyzerStatistic, \
     BugPathEvent, BugReportPoint, ReportAnnotations, ExtendedReportData, \
-    File, FileContent, Report as DBReport, ReviewStatus, Run, RunHistory, \
-    RunLock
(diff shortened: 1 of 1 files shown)


vodorok:
Please give me some time to fix all the failing tests before diving into the details of the modifications.

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -979,22 +977,29 @@ def get_skip_handler(
...
         enabled_checkers: Set[str] = set()
         disabled_checkers: Set[str] = set()
 
         # Processing analyzer result files.
         processed_result_file_count = 0
+
+        for root_dir_path, _, report_file_paths in os.walk(report_dir):
+            for f in report_file_paths:
+                if not report_file.is_supported(f):
+                    continue
```
bruntib, line 1001: What is the purpose of this part?

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -1014,9 +1019,40 @@ def get_skip_handler(
                 self.__process_report_file(
                     report_file_path, session, source_root, run_id,
                     file_path_to_id, run_history_time,
-                    skip_handler, hash_map_reports)
+                    skip_handler, report_to_report_id)
                 processed_result_file_count += 1
 
+        # Get all relevant review_statuses for the newly stored reports
+        # TODO Call self.getReviewStatusRules instead of the beloew query
```
bruntib, line 1026:
```suggestion
        # TODO Call self.getReviewStatusRules instead of the query below
```

bruntib (changes requested):

dkrupp (changes requested):
Please make sure that you test this change extensively on an open source project.

I suggest the following.

1. Analyze version x of xerces and store it to the server
2. Change the review status of 10 reports on the GUI
3. Change the review status of 5 report in the source code
4. Clone version x+y of xerces and apply 3 source code suppressions to the new versoion
5. analyze and store it to the same run

Repeat this experiment with the baseline version of codechecker.

The detection & review  status of the issues must be the same.

Create a jenkins job of this. We could use this as a test for further store changes.

vodorok:
This modification will affect the `idle in transaction` timeout settings of the PostgreSQL config. For larger report folders I would advise 15-30 min settings.

bruntib (approved):

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -808,28 +853,18 @@ def __process_report_file(
...
 
-        def get_review_status(report: Report) -> Tuple[ReviewStatus, bool]:
+        def get_review_status_from_source(
+                report: Report) -> SourceReviewStatus:
             """
-            Return the review status belonging to the given report and a
-            boolean value which indicates whether the review status comes from
+            Return the review status set in the source code belonging to
+            the given report whether the review status comes from
             a source code comment.
```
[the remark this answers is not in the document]
vodorok: Fixed

bruntib (approved):

dkrupp (approved):
LGTM

PR #3801, opened 2022-12-07
[cmd] Gracefully exit instead of crashing when cmd diff is missing a param

Szelethus:
When CodeChecker cmd diff is invoked like this:

CodeChecker cmd diff -n new_run

it will crash instead of gracefully exiting and warning about the
missing -b flag. This patch fixes that.


bruntib:
Wouldn't it be a better solution to make `-b` and `-n` required fields? This could be simply forced by the `argparse` module. I can't come up with a use-case where either `-b` or `-n` is needed only.

bruntib (approved):

PR #3803, opened 2022-12-12
Document the new checker misc-use-anonymous-namespace

sylvestre:


whisperity:
This checker is interesting because AFAIK it goes against LLVM's own convention. 😬 (It's been a while since I coded C++ 😦 )

whisperity:
I do appreciate the vigilance from Sylvestre, but I would prefer if we did not litter the history with merge committing one checker at a time. Would it be bad if we only imported the new checker labels **once** per (LLVM|CodeChecker) release? Do we have a functionality that allows users to user their own labelling config file from their home, without having to build a package themselves, @vodorok? (If we don't, can we have a ticket about it at least?) Being too bleeding edge means that we'd be defaulting on checkers that could get removed before the next LLVM release hits the market.

sylvestre:
@whisperity as you wish, you are upstream, you decide :)
However, I am not sure to see what is the issue ? (commits are cheap)
Doing it incrementally when they arrive is easy and spreads the work

whisperity:
Indeed there's merit to both ways. What I'm trying to prevent is accidentally default-enabling (or sensitive, which is also used by plenty of our clients) a checker that's not up to spec for some reason (e.g. bugprone-unchecked-optional-access which crashes, misc-const-correctness which is slow). LLVM's master is not a stable thing so it's not a good idea to add something in the window where it might be reverted, even.

Nevertheless, having a user-specific profile option that doesn't need to write system directories (in the case of the PyPI package) would have benefits because that way some of the people who use codechecker to drive a custom clang (that has checkers our upstream cfg doesn't know about) could set it without having to manually rebuild the package every now and then.

And diffing between two tags makes it relatively easy to see how many new checks were labelled between CCh releases.

config/labels/analyzers/clang-tidy.json
```diff
@@ -5488,6 +5488,13 @@
       "sei-cert:exp63-cpp",
       "severity:HIGH"
     ],
+    "misc-use-anonymous-namespace": [
+      "doc_url:https://clang.llvm.org/extra/clang-tidy/checks/misc/use-anonymous-namespace.html",
+      "profile:default",
+      "profile:extreme",
+      "profile:sensitive",
+      "severity:HIGH"
```
bruntib, line 5496: I would grade it as a style or low severity checker, but definitely not high.

bruntib (approved):

PR #3804, opened 2022-12-16
Fix three bugs and a couple of style issues

geeksmith:
Remove three bugs:
- $MASTER_EXISTS vs. $PREVIOUS_EXISTS bug fixed with simpler test
- Using quotes around $ANALYZER_OPTIONS when calling CodeChecker results in an error
- Fix escaping and indentation of $DIFF_CMD
  - we could use <<- and a real tab to indent END but the web formatting doesn't preserve the tab

Fix style issues:
- Use a variable instead of a file to store $WAS_OUTPUT
- Remove superfluous curly braces.

### DIFF
--- a/docs/script_update.md
+++ b/docs/script_update.md
@@ -26,14 +26,14 @@ HTML_DIR="${WORKSPACE}/html"
 # -- Execute CodeChecker --
 
 # 1. Build your project with CodeChecker
-pushd ${SOURCE_DIR}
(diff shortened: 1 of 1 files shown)


geeksmith:
One more bug I forgot to mention -- - using quotes around $ANALYZER_OPTIONS when calling CodeChecker results in an error

bruntib (approved):

PR #3807, opened 2023-01-03
[refactor] Analyzer context as singleton

bruntib:
`analyzer_context.Context` is a singleton class. Anywhere it is constructed, the same object is acquired. So there is no need to pass it as parameter through long function call chains, because this object can be constructed where it's needed. It has no cost, because the same instance is used once it was created.


martong:
LGTM 😉

Szelethus (commented):
Is there anything in `analyzer_context.Context`'s documentation thats worth adjusting? Maybe a line about how to query this object? (maybe its basic python and I'm just oblivious to it)

analyzer/codechecker_analyzer/cmd/checkers.py
```diff
@@ -270,26 +270,6 @@ def __uglify(text: str) -> str:
     return text.lower().replace(' ', '_')
 
 
-def guideline_rules_for_checker(
```
Szelethus, line 273: No longer needed? This seems to be gone for good.

Szelethus (approved):
Otherwise LGTM

analyzer/codechecker_analyzer/cmd/checkers.py
bruntib: Please, double check it, but I haven't found it being used.

analyzer/codechecker_analyzer/analyzers/clangsa/config_handler.py
```diff
@@ -29,8 +29,6 @@ def __init__(self, environ):
         self.ctu_dir = ''
         self.ctu_on_demand = False
         self.log_file = ''
-        self.path_env_extra = ''
```
vodorok, line 32: Why are these removed?

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -420,29 +414,28 @@ def version_compatible(cls, configured_binary, environ):
...
     @classmethod
-    def construct_config_handler(cls, args, context):
+    def construct_config_handler(cls, args):
 
-        environ = env.extend(context.path_env_extra,
-                             context.ld_lib_path_extra)
+        context = analyzer_context.get_context()
+        environ = context.analyzer_env
 
         handler = config_handler.ClangSAConfigHandler(environ)
```
vodorok, line 434: Just a question. Could the env parameter factored out from the Config Handler the same way?

analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -69,41 +68,6 @@ def create_actions_map(actions, manager):
     return result
 
 
-def __get_analyzer_version(context, analyzer_config_map):
```
vodorok, line 72: Where did this go? Was this an unused func?

analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -159,14 +123,16 @@ def __get_ctu_data(config_map, ctu_dir):
...
-def perform_analysis(args, skip_handlers, context, actions, metadata_tool,
+def perform_analysis(args, skip_handlers, actions, metadata_tool,
                      compile_cmd_count):
     """
     Perform static analysis via the given (or if not, all) analyzers,
     in the given analysis context for the supplied build actions.
     Additionally, insert statistical information into the metadata dict.
     """
 
+    context = analyzer_context.get_context()
```
vodorok, line 134: I would move this defintion of context closer to the first usage.

analyzer/codechecker_analyzer/analyzers/analyzer_base.py
```diff
@@ -126,7 +127,7 @@ def post_analyze(self, result_handler):
         pass
 
     @staticmethod
-    def run_proc(command, env=None, cwd=None, proc_callback=None):
+    def run_proc(command, cwd=None, proc_callback=None, env=None):
```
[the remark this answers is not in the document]
vodorok: https://github.com/Ericsson/codechecker/pull/3807#discussion_r1071109465

analyzer/tests/unit/__init__.py
```diff
@@ -16,6 +16,8 @@
 REPO_ROOT = os.path.abspath(os.environ['REPO_ROOT'])
 PKG_ROOT = os.path.join(REPO_ROOT, 'build', 'CodeChecker')
 
+os.environ["CC_DATA_FILES_DIR"] = PKG_ROOT
```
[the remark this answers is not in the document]
bruntib: In many cases the analyzer context singleton object was constructed at a top-level point of the program and it was passed to some inner function through many function calls. Now the context object is constructed in some of the functions that are tested. The constructor needs to find the label config files that are under this path given here.

analyzer/tests/functional/__init__.py
```diff
@@ -18,5 +18,6 @@
 REPO_ROOT = os.path.abspath(os.environ['REPO_ROOT'])
 PKG_ROOT = os.path.join(REPO_ROOT, 'build', 'CodeChecker')
 
+os.environ["CC_DATA_FILES_DIR"] = PKG_ROOT
```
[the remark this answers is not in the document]
bruntib: In many cases the analyzer context singleton object was constructed at a top-level point of the program and it was passed to some inner function through many function calls. Now the context object is constructed in some of the functions that are tested. The constructor needs to find the label config files that are under this path given here.

analyzer/codechecker_analyzer/analyzers/analyzer_types.py
```diff
@@ -81,55 +74,46 @@ def is_statistics_capable(context):
...
 
-def is_ignore_conflict_supported(context):
+def is_ignore_conflict_supported():
     """
     Detects if clang-apply-replacements supports --ignore-insert-conflict flag.
     """
-    analyzer_env = env.extend(context.path_env_extra,
-                              context.ld_lib_path_extra)
-
+    context = analyzer_context.get_context()
```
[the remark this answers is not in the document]
bruntib: In this case it is used at two places: line 113 and 116.

analyzer/codechecker_analyzer/analyzers/analyzer_base.py
```diff
@@ -78,15 +79,15 @@ def get_analyzer_mentioned_files(self, output):
...
+                                 skiplist_handler):
         """
         This method constructs the class that is responsible to handle the
         results of the analysis. The result should be a subclass of
         ResultHandler
         """
         raise NotImplementedError("Subclasses should implement this!")
 
-    def analyze(self, analyzer_cmd, res_handler, env=None, proc_callback=None):
+    def analyze(self, analyzer_cmd, res_handler, proc_callback=None, env=None):
```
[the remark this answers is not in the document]
bruntib: The `env` parameter of this function will be eliminated in the long term. It is placed to the last position so the caller can easily use its default value. This is a first step in the elimination of this parameter.

analyzer/codechecker_analyzer/cmd/checkers.py
bruntib: This function was never really used. It was an alternative implementation which was commented out from the source code: https://github.com/Ericsson/codechecker/blob/1ed457ea8a811ac1b66860d76c46936471cf8e86/analyzer/codechecker_analyzer/cmd/checkers.py#L374

PR #3808, opened 2023-01-06
Fix error using Clang option '-stdlib=libc++'

TotalNormal2:
Using Clang option "-stdlib=libc++" failed with:
cppcheck: error: unrecognized command line option: "--std=libc++"

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
@@ -101,8 +101,10 @@ class Cppcheck(analyzer_base.SourceAnalyzer):
         # * --std c99
         # * -std=c99
         # * --std=c99
-        # BUT NOT: -std c99
-        std_regex = re.compile("-?-std.*")
+        # BUT NOT:
+        # * -std c99
+        # * -stdlib=libc++
+        std_regex = re.compile("-?-std[ |=].*")
         for i, analyzer_option in enumerate(self.buildaction.analyzer_options):
             if interesting_option.match(analyzer_option):
                 params.extend([analyzer_option])
--- a/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
@@ -96,7 +96,7 @@ class Cppcheck(analyzer_base.SourceAnalyzer):
         Any other parameter different from the above list will be dropped.
         """
         params = []
-        interesting_option = re.compile("-[I|U|D].*")
+        interesting_option = re.compile("-[IUD].*")
         # the std flag is different. the following are all valid flags:
         # * --std c99
         # * -std=c99
@@ -104,7 +104,7 @@ class Cppcheck(analyzer_base.SourceAnalyzer):
         # BUT NOT:
         # * -std c99
         # * -stdlib=libc++
-        std_regex = re.compile("-?-std[ |=].*")
+        std_regex = re.compile("-(-std$|-?std=.*)")
         for i, analyzer_option in enumerate(self.buildaction.analyzer_options):
             if interesting_option.match(analyzer_option):
                 params.extend([analyzer_option])


analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
```diff
@@ -101,8 +101,10 @@ def parse_analyzer_config(self):
         # * --std c99
         # * -std=c99
         # * --std=c99
-        # BUT NOT: -std c99
-        std_regex = re.compile("-?-std.*")
+        # BUT NOT:
+        # * -std c99
+        # * -stdlib=libc++
+        std_regex = re.compile("-?-std[ |=].*")
```
bruntib, line 107: Thanks, it's a nice catch!

`[ |=]` matches any of the three characters between square brackets, because `|` is not a separator in this context. Also, the `build_action.analyzer_option` doesn't contain a space character if it starts with `--std`, because command line arguments are already split here.

What do you think of this regex?
```suggestion
        std_regex = re.compile(r"-?-std\b.*")
```

TotalNormal2: Thank you for looking into this. You're right about the split and `|`.

I had to extend my regex knowledge first to understand your nice idea. I did some testing using an online regex evaluator (selected Python) and found:
- the pattern needs not to be at the start (solvable with `^`)
- if a `-` (or any other special char) is following `std` anything afterwards will also be accepted 

Example: `ThisCanBeAnything-std-AlsoHere=AndMatch`

I've tested other solutions like `"^-(-std$)|(-?std=.*)"` or `"^--?std($)|(=.*)"` (the latter still accepts `-std c99`) to avoid any undesired patterns. I compared them to `"^-?-std\b.*"` which requires only up to 10 steps to evaluate a string while the others require up to 30. On the downside it still accepts `-std c99` or `-std-Anything...` but I vote for speed:
```suggestion
        std_regex = re.compile(r"^-?-std\b.*")
```

Looking now into [`interesting_option`](https://github.com/Ericsson/codechecker/blob/master/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py#L99) (got that `|` from there) I'm wondering if it should be `interesting_option = re.compile(r"^-[IUD].*")` ?

bruntib: Thanks, it's a very nice analysis of the issue, it's good that we consider all corner cases. As a next step what do you think of this?
```suggestion
        std_regex = re.compile(r"-?-std[\b=].*")
```
It means that after "std" it is either the end of the word or `=` follows. The `^`is not necessarily needed, because function `match()` in line 109 finds the pattern at the beginning of the string.

Yes, the `interesting_option` should really be fixed to `IUD` instead of `I|U|D`.

TotalNormal2: It seems `[\b]` is matching backspace and `[\b=]` is matching `backspace` or `=` but not word boundary.
You're right about `^` not being required using `match()`.

I've tested the following strings (1-3 are the correct ones) in an online-evaluator...

1. --std
2. -std=xxx
3. --std=xxx
4. -std
5. -stdlib=libc++
6. --stdä=hnlich
7. --std-not
8. -stdio=xxx
9. --stdio=xxx
10. -std-99
11. --std-C99
12. -std-99=xxx
13. --std-C99=xxx

...with these expressions (matching strings mentioned)
1. `-(-std$|-?std=.*)` matching: 1,2,3
2. `-?-std($|=.*)` matching: 1,2,3,4
3. `-?-std(\b|=.*)` matching: 1,2,3,4,(6),7,10,11,12,13 ; (6) had different results on different evaluators
4. `-?-std[$=].*` matching: 2,3
5. `-?-std[\b=].*` matching: 2,3

Only expression 1 has zero false-pos and false-neg.
I prefer avoiding `\b` because using `-` within a name is not complete out of the world but it's not a boundary (see 11-13).

bruntib: Wow, thanks, it's a very nice research and test cases :)
I start the tests and we can merge it. Thank you, well done!

bruntib (approved):

PR #3809, opened 2023-01-06
[refactor] Rename and fix load_json_or_empty

bruntib:
"load_json_or_empty" is a too long name for this function, and the
second part doesn't carry any information.

This function is only used for opening JSON files, so "kind" parameter
is not necessary.

Also move this load_json() function to codechecker_common.util module
because originally it was under report converter tool, but most usages
were in all other parts of CodeChecker. Many times this was the only
dependency towards report_converter.


Szelethus (approved):
Fair enough.

PR #3819, opened 2023-01-17
Updates to setup.py/PyPI configuration

whisperity:
@csordasmarton has left the project but their e-mail address was still present in the PyPI configuration. This is rectified by moving to a new public mailing list, which is just a stub, as we will still prefer issues to be reported here on GitHub.

Also added some tags to the built package to make the project better visible.


gamesh411 (approved):

vodorok (approved):
lgtm

PR #3820, opened 2023-01-18
[analyzer] Promote the missing checker warning to an error

Szelethus:
Exactly what it says on the tin!

### DIFF
--- a/web/tests/functional/diff_local/test_diff_local.py
+++ b/web/tests/functional/diff_local/test_diff_local.py
@@ -34,7 +34,7 @@ class DiffLocal(unittest.TestCase):
         test_class = self.__class__.__name__
         print('Running ' + test_class + ' tests in ' + test_workspace)
 
-        # Get the test configuration from the prepared int the test workspace.
+        # Get the test configuration from the prepared into the test workspace.
         self._test_cfg = env.import_test_cfg(test_workspace)
         self._codechecker_cfg = self._test_cfg['codechecker_cfg']
         # Get the test project configuration from the prepared test workspace.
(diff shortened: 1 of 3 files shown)


whisperity:
How will this work with pre-existing profiles? Let's suppose I do `--enable sensitive` which enables a checker which is only available since LLVM 42.0, but I only have 24.2 which does not have that checker? Will this result in breaking of the execution just because the executed LLVM is older than whatever was last synchronised with CodeChecker?

Szelethus:
This is how we enable checkers when a profile is specified on the command line:

```python
  for checker in default_profile_checkers:
      self.set_checker_enabled(checker)
  # ...
  for checker in checker_labels.checkers_by_labels(
          [f'profile:{identifier}']):
      self.set_checker_enabled(checker, enabled)
```

Since this part of CodeChecker seems to be tested almost exclusively with functional tests, I'm not sure how to explicitly test this specific thing.

bruntib:
Please, check the failing test. Otherwise looks good to me.

web/tests/functional/diff_local/test_diff_local.py
```diff
@@ -222,7 +222,6 @@ def test_suppress_reports(self):
         source code comments.
         """
         cfg = dict(self._codechecker_cfg)
-        cfg['analyzers'] = ['clang-tidy']
```
Szelethus, line 225: Needed to remove this, because the codechecker call beneath enables clangsa checkers as well, which would cause an error.

bruntib (approved):

PR #3822, opened 2023-01-19
[cmd] Remove --tidy-config flag

bruntib:
CodeChecker analyze --tidy-config is a rarely used and redundant flag. ClangTidy options can be given through .clang-tidy file, --analyzer-config, --checker-config and --tidyargs flags.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
@@ -133,8 +133,7 @@ class ClangTidy(analyzer_base.SourceAnalyzer):
         Return a list of checkers and warnings what needs to be enabled during
         analysis.
 
-        If the file specified by the '--tidy-config' option contains a 'Checks'
-        key or the 'Checks' option is specified through the '--analyzer-config'
+        If 'Checks' option is specified through '--analyzer-config'
         the return value will be a tuple of empty lists which means do not turn
         checkers explicitly. "clang-analyzer-*" is an exception, because we
         want to disable these even if analyzer config is given. If we wouldn't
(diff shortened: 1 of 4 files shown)


Szelethus:
I agree, though maybe it'd be more user friendly if we would depracate this flag first.

Szelethus (changes requested):
In my view, this is the worst way to depracate a feature. You are removing the functionality behind the flag, but not the flag itself, leaving those that don't read warning messages (imagine multi-thousand line CI jobs, or integrated environments like vscode) to think that their configuration works, but it doesn't.

Mind that tidy configs are usually low-impact: its easy to not notice that the feature is off.

You should either
* Remove the flag altogether;
* Keep the functionality, but warn that its been deprecated and will be removed in an upcoming release.

bruntib:
Alright, I restored the flag's functionality. It will be removed in the next release.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -133,8 +133,7 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
         Return a list of checkers and warnings what needs to be enabled during
         analysis.
 
-        If the file specified by the '--tidy-config' option contains a 'Checks'
-        key or the 'Checks' option is specified through the '--analyzer-config'
+        If 'Checks' option is specified through the '--analyzer-config'
```
Szelethus, line 136:
```suggestion
        If 'Checks' option is specified through '--analyzer-config'
```

Szelethus (commented):
Please create a warning for when the flag is used, warning about the removal in the next release. Otherwise LGTM.

Szelethus (approved):

PR #3823, opened 2023-01-20
[cmd] Remove some deprecated flags.

bruntib:
Some flags are removed from the code base that are deprecated for a while.
Also, the deprecation of --ctu-reanalyze-on-failure is undone, because it will not be deprecated in the near future, probably.

### DIFF
--- a/analyzer/codechecker_analyzer/cmd/analyze.py
+++ b/analyzer/codechecker_analyzer/cmd/analyze.py
@@ -530,8 +530,7 @@ Cross-TU analysis. By default, no CTU analysis is run when
                               action='store_true',
                               dest='ctu_reanalyze_on_failure',
                               default=argparse.SUPPRESS,
-                              help="DEPRECATED. The flag will be removed. "
-                                   "If Cross-TU analysis is enabled and fails "
+                              help="If Cross-TU analysis is enabled and fails "
                                    "for some reason, try to re analyze the "
                                    "same translation unit without "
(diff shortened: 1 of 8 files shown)


web/client/codechecker_client/cmd_line_client.py
```diff
@@ -325,16 +325,6 @@ def check_run_names(client, check_names):
 
 
 def check_deprecated_arg_usage(args):
-    if 'suppressed' in args:
```
vodorok, line 328: Was this suppressed param already a no-op?

vodorok (changes requested):

web/client/codechecker_client/cmd_line_client.py
bruntib: Yes, it was not used. `--review-status` is used instead.

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -530,7 +530,7 @@ def add_arguments_to_parser(parser):
                               action='store_true',
                               dest='ctu_reanalyze_on_failure',
                               default=argparse.SUPPRESS,
-                              help="DEPRECATED. The flag will be removed. "
+                              help="The flag will be removed. "
```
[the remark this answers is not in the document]
bruntib: Thanks, it's a nice catch. I fixed it.

vodorok (approved):
lgtm

PR #3825, opened 2023-01-24
[server] Add store time logging

vodorok:
This patch introduces the logging of the runtime of the incoming storage requests.
The logging is done into a file as well as to the stdout. The logfile is weekly rotated, and stored for 2 months.

### DIFF
--- a/codechecker_common/logger.py
+++ b/codechecker_common/logger.py
@@ -13,6 +13,7 @@ import argparse
 import json
 import logging
 from logging import config
+from pathlib import Path
 import os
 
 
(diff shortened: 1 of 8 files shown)


vodorok:
WIP until tests are fixed

codechecker_common/logger.py
```diff
@@ -182,4 +182,26 @@ def setup_logger(log_level=None, stream=None):
             if 'stream' in handler:
                 handler['stream'] = stream
 
+    # if workspace is set, we will log to a file in the workspace
+    # This is added dynamically because the declaratice config is not
+    # flexible enough, and will always create a log file in wierd locations,
+    # before we can initialize the filneame attribute, to the workspace
+    # directories
```
bruntib, line 189:
```suggestion
    # If workspace is set, we will log to a file in the workspace.
    # This is added dynamically because the declarative config (config/logger.conf) is not
    # flexible enough, and will always create a log file in weird locations,
    # before we can initialize the filename attribute, to the workspace
    # directories.
```

bruntib (changes requested):

PR #3831, opened 2023-02-02
CodeChecker server hotswapper

vodorok:
The newly added small script watches for filesystem events inside the web folder in this repo. If a suitable event occurs, first it stops the Previous server, the starts a new one.
This feature depends on the dev_package target, it will only work with that kind of CodeChecker build.

### DIFF
--- a/scripts/codechecker_dev_server.py
+++ b/scripts/codechecker_dev_server.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python3
+
+from pathlib import Path
+import threading
+import subprocess
+import time
+import os
+import sys
+
+from subprocess import Popen
+
+from watchdog.observers import Observer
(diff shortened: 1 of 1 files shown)


whisperity (changes requested):
Even though this is only for developers, let's add some documentation to the file's scope or something at least.

bruntib (approved):
I have some remarks, otherwise it can be merged:

- The check_env() could check if CodeChecker binary is in `PATH`, or we should give an intuitive error message if a user forgot adding it to `PATH`.
- There are some source codes commented out. Those could be removed if not necessary.
- As far as I see, the script assumes that it is started from the root directory of CodeChecker, since it observes file modifications in `./web` relatively. A warning/error message could be printed if the script is not called from the root.

None of these are blocker for me, because this is a developers' utility, these are just convenience improvements.

PR #3838, opened 2023-02-10
[fix] Fix a condition about checkers being compiler warnings 

Szelethus:
Seems like I never tested this well enough in
https://github.com/Ericsson/codechecker/pull/3698, because I could
negate a condition in config_handler.py and no tests broke. This broke
in https://github.com/Ericsson/codechecker/pull/3820, however, when
hicpp-* was mistakenly not understood to be a checker, which caused an
error (instead of a warning).


bruntib (approved):
Maybe you want to rebase this first. Otherwise can be merged.

Issue #3832, opened 2023-02-02
[web] Web app requires a full reload if switching between users of dirrent permissions to show/conceal privileged elements


whisperity:
**Describe the bug**
When switching between users of different permission levels (e.g., a `SUPERUSER` and a normal user), the webapp requires a full reload for the relevant, permission-specific GUI elements to be shown or hidden appropriately.

**:warning: Warning!** Due to the web interface's ViewModel behaviour, this allows the leaking of product configuration to supposedly unauthorised users. The unprivileged user is **not** able to *destructively* change information, and some information (such as product permissions) remain hidden still.

This is **not** a *direct* security issue because to have the information appear, you'd need to log in as a privileged user anyway during the process, at which point (assuming you are an attacker with the option to log in as superuser) you would be capable of destructive changes to the server.

**CodeChecker version**
6.21 release

**To Reproduce**
Steps to reproduce the behaviour:

1. Log in as privileged user.
2. Observe the *"Edit global configuration"* and the :pencil2: icons visible.
3. Log out, and log in as unprivileged user.
4. Observe that the icons are still there.
5. Click on a :pencil2: for a product. The configuration will load and will be visible (such as database user, hostname, etc.).
6. Try changing things, the save action will appropriately be rejected.
7. Switch to the *"Permissions"* tab during the edit of a product. It will remain empty, without any warnings (neither visually nor in the browser's console).
8. (Same applies to the "global permissions" view.)

In some cases weird status messages can be triggered by trying to save from an unauthorised view:
![Screenshot_20230202_100334](https://user-images.githubusercontent.com/1969470/216279155-560e4a07-8f0f-40aa-9c3c-56bdb56ba38d.png)

**Expected behaviour**
The GUI elements are shown/hidden appropriately, and automatically. And that the product configuration not needed for unprivileged users is not available to unprivileged users.

**Desktop (please complete the following information)**
 - OS: Ubuntu 20.04 LTS
 - Browser: Firefox
 - Version: 19.0

vodorok:
Thanks, I suspect we have some stale data in our vuex store, or not properly construct the root component, upon login. Will check this.

bruntib:
Fixed by #3840.

PR #3840, opened 2023-02-13
[web] Fix stale permission caching

vodorok:
Due that the keep-alive component cache is used wrapping the router view, the products page was cached after logout.
Now the keep-alive inclusion list is dynamically adjusted according to the user authentication state.
The cache is only kept for one login session.


vodorok:
Fixes #3832

bruntib (approved):

PR #3841, opened 2023-02-14
Bump GitPython version

vodorok:
See the relevant dependabot report for more context.
https://github.com/Ericsson/codechecker/security/dependabot/81
https://github.com/Ericsson/codechecker/security/dependabot/82

### DIFF
--- a/web/requirements.txt
+++ b/web/requirements.txt
@@ -5,7 +5,7 @@ portalocker==2.2.1
 psutil==5.8.0
 mypy_extensions==0.4.3
 thrift==0.13.0
-gitpython==3.1.11
+gitpython==3.1.30
 PyYAML==5.4.1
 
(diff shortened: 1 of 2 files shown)


vodorok:
The new version needs python 3.7:  https://pypi.org/project/GitPython/ That`s the reason the actions fail.
Re-run, and merge after #3842  is implemented/

whisperity:
Dependabot is sending like 5 emails every day about a bazillion of affected related packages. I propose that we do a **single sweep** of upgrading 3rdparties in the project at some reasonable "sequence point" before every release. (Same proposal as with the "checker label configuration".) Little changes that bump a number here or there just pollute the history and the review queue, and we potentially miss conflicts between changing package versions (e.g., upgrading one package will break the API use of another) on the way.

bruntib (approved):

PR #3863, opened 2023-03-30
[NFC] Factor args out of the diff logic for unit tests

Szelethus:
CodeChecker cmd diff is a little out of whack right now, which is discussed in length here: #3884

One of the reasons why it was possible for this problem to fester into this state is that our testing methodology for diffs (really, the entirety of CodeChecker) is really-really poor. Cmdline diff in particular (though not unique in this manner) were done by literally calling CodeChecker, and reparsing its stdout and stderr by string matching whether we see what we want. 

That is weak even for a python project.

Part of the problem is that key functions/classes in CodeChecker take the ominous "args" argument, which contains the set of user arguments, as well as our defaults. We parse a tiny bit of "args" (possibly redundantly), and then pass it onto another function, that does the same. This is dumb -- we should parse args, and simply get on with our lives. Major function should have very concrete arguments that are, in an ideal scenario, contsructible in a unit test environment.

This multi-commit patch does just that: It gradually threads args out of all major functions and does some minimal refactoring with whats left.

### DIFF
--- a/web/client/codechecker_client/cmd_line_client.py
+++ b/web/client/codechecker_client/cmd_line_client.py
@@ -721,26 +721,31 @@ def handle_list_results(args):
 
 
 def handle_diff_results(args):
+    handle_diff_results_impl(args)
+
+
+def handle_diff_results_impl(args):
     # If the given output format is not 'table', redirect logger's output to
     # the stderr.
     stream = None
-    if 'output_format' in args and args.output_format != 'table':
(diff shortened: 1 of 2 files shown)


Szelethus:
Closing this for now.

Szelethus:
Oops, meant to close something else :)

bruntib (approved):

PR #3868, opened 2023-04-06
[test] Add missing tests for cmdline diffing, and display a bug for tag diffs

Szelethus:
We usually test CodeChecker diff by making API calls that one can find
in report_server.py. However, `CodeChecker cmd diff` doesn't invoke
these plainly, but rather wraps it in various functions, for which we
never had proper tests. Until now!

I also made sure to include (almost all) cases discussed here:
https://github.com/Ericsson/codechecker/issues/3884


Szelethus:
Marking this as a draft. What this PR does poorly is testing whether all diffs that should be identical are identical.

bruntib (approved):

PR #3873, opened 2023-04-11
Upgrade actions to ubuntu-20.04

vodorok:


vodorok:
We need to include this in 6.22 to be able to publish on pypi.

bruntib (approved):
Consider searching all `18.04` strings in the code base and eliminate them in the future. I think we can drop its support. Except for the documentation maybe, which describes the installation to this old platform.

PR #3874, opened 2023-04-12
[fix] Treat clang-diagnostic-* checkers as compiler flags

vodorok:
Previously the handling of clang-diagnostics was flawed because tidy cannot enable compiler warnings when specified  as a checker, only suppression of the resulting reports are available:

From the official [clang-tidy docs](https://clang.llvm.org/extra/clang-tidy/):
 > Clang diagnostics are treated in a similar way as check diagnostics. Clang diagnostics are displayed by clang-tidy and can be filtered out using the -checks= option. However, the -checks= option does not affect compilation arguments, so it cannot turn on Clang warnings which are not already turned on in the build configuration. The -warnings-as-errors= option upgrades any warnings emitted under the -checks= flag to errors (but it does not enable any checks itself).

After this patch all of these checkers will be treated as compiler warning flags, this way the compiler warnings can be enabled as clang-diagnostics.
There is an added benefit of this method, as the checker prefix based activation is also working now.

Before the patch:
* Trying to use `-e clang-diagnostic-...` notation did not produce any result because tidy wouldn't insert the corresponding compiler warning to the input compilation command. We thought adding it to the checks list enabled the warning, but it didn't.

After the patch:
* Using `-e W...` notation results in a warning from CodeChecker to use the clang-diagnostic-unused-variable counterpart.
* Using the `-e clang-diagnostic-...` now properly works. and enables the related compiler warnings.
* When using `--enable-all`, instead of adding every compiler warning individually to the build command, `-Weverything` is added.

There are some considerations regarding the old `-W...` notation.
* Using the `-W...` notation is be deprecated, but for the time being still usable.\
* With `--enable-all`  the usage of `-d -W...` notation is ignored, and results in a warning. The `-d compiler-diagnostics-...` is the recommended way to disable a compiler warning.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
@@ -24,7 +24,8 @@ from codechecker_analyzer import analyzer_context, env
 from codechecker_analyzer.analyzers.clangsa.analyzer import ClangSA
 
 from .. import analyzer_base
-from ..config_handler import CheckerState, get_compiler_warning_name
+from ..config_handler import CheckerState, CheckerType, \
+    get_compiler_warning_name_and_type
 from ..flag import has_flag
 from ..flag import prepend_all
 
@@ -228,37 +229,34 @@ class ClangTidy(analyzer_base.SourceAnalyzer):
         for checker_name, value in config.checks().items():
             state, _ = value
 
-            # Checker name is a compiler warning.
-            if checker_name.startswith('W'):
-                LOG.warning("As of CodeChecker v6.22, the following usage"
-                            f"of '{checker_name}' compiler warning as a "
-                            "checker name is deprecated, please use "
-                            f"'clang-diagnostic-{checker_name[1:]}' instead.")
-
-            warning_name = get_compiler_warning_name(checker_name)
+            warning_name, warning_type = \
+                get_compiler_warning_name_and_type(checker_name)
             if warning_name is not None:
                 # -W and clang-diagnostic- are added as compiler warnings.
-                if state == CheckerState.enabled:
-                    compiler_warnings.add('-W' + warning_name)
-                elif state == CheckerState.disabled:
-                    if config.enable_all:
-                        LOG.warning("Disabling compiler warning with compiler"
-                                    f" flag '-d W{warning_name}' is not "
-                                    "supported.")
-                    compiler_warnings.add('-Wno-' + warning_name)
-                continue
+                if warning_type == CheckerType.compiler:
+                    LOG.warning("As of CodeChecker v6.22, the following usage"
+                                f"of '{checker_name}' compiler warning as a "
+                                "checker name is deprecated, please use "
+                                f"'clang-diagnostic-{checker_name[1:]}' "
+                                "instead.")
+                    if state == CheckerState.enabled:
+                        compiler_warnings.add('-W' + warning_name)
+                    elif state == CheckerState.disabled:
+                        if config.enable_all:
+                            LOG.warning("Disabling compiler warning with "
+                                        f"compiler flag '-d W{warning_name}' "
+                                        "is not supported.")
+                        compiler_warnings.add('-Wno-' + warning_name)
+                # If a clang-diagnostic-... is enabled add it as a compiler
+                # warning as -W..., if it is disabled, tidy can suppress when
+                # specified in the -checks parameter list, so we add it there
+                # as -clang-diagnostic-... .
+                elif warning_type == CheckerType.analyzer:
+                    if state == CheckerState.enabled:
+                        compiler_warnings.add('-W' + warning_name)
+                    elif state == CheckerState.disabled:
+                        checkers.append('-' + checker_name)
 
-            # If a clang-diagnostic-... is enabled add it as a compiler
-            # warning as -W..., if it is disabled, tidy can suppress when
-            # specified in the -checks parameter list, so we add it there
-            # as -clang-diagnostic-... .
-            if checker_name.startswith('clang-diagnostic-') and \
-                    state == CheckerState.enabled:
-                warning_name = checker_name[len('clang-diagnostic-'):]
-                if state == CheckerState.enabled:
-                    compiler_warnings.add('-W' + warning_name)
-                elif state == CheckerState.disabled:
-                    checkers.append('-' + checker_name)
                 continue
 
             if state == CheckerState.enabled:
--- a/analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
@@ -207,7 +207,9 @@ class ClangTidy(analyzer_base.SourceAnalyzer):
         some ClangSA bug.
         """
         checkers = []
-        compiler_warnings = []
+        # Usage of a set will remove compiler warnings and clang-diagnostics
+        # which are the same.
+        compiler_warnings = set()
 
         has_checker_config = \
             config.checker_config and config.checker_config != '{}'
@@ -227,16 +229,20 @@ class ClangTidy(analyzer_base.SourceAnalyzer):
             state, _ = value
 
             # Checker name is a compiler warning.
+            if checker_name.startswith('W'):
+                LOG.warning(f"The following usage of {checker_name} "
+                            "compiler warning as a checker name is "
+                            "deprecated, please use "
+                            "clang-diagnostic-<checker-name> instead")
+
             warning_name = get_compiler_warning_name(checker_name)
             if warning_name is not None:
+                # -W and clang-diagnostic- are added as compiler warnings.
                 if state == CheckerState.enabled:
-                    compiler_warnings.append('-W' + warning_name)
+                    compiler_warnings.add('-W' + warning_name)
                 elif state == CheckerState.disabled:
-                    compiler_warnings.append('-Wno-' + warning_name)
-
+                    compiler_warnings.add('-Wno-' + warning_name)
                 continue
-            elif checker_name.startswith('clang-diagnostic'):
-                checker_name += '*'
 
             if state == CheckerState.enabled:
                 checkers.append(checker_name)
@@ -250,7 +256,7 @@ class ClangTidy(analyzer_base.SourceAnalyzer):
         # for the compiler, clang-tidy is just capable to emit them in its own
         # format.
         if config.analyzer_config.get('take-config-from-directory') == 'true':
-            return [], compiler_warnings
+            return [], list(compiler_warnings)
 
         if has_checker_config:
             try:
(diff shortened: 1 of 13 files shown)


analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -227,16 +229,30 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
             state, _ = value
 
             # Checker name is a compiler warning.
+            if checker_name.startswith('W'):
```
Szelethus, line 232: I feel like this branch should be moved to `get_compiler_warning_name`, that function is the authority on what we interpret to be a compiler warning.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -227,16 +229,30 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
...
+                LOG.warning(f"The following usage of {checker_name} "
+                            "compiler warning as a checker name is "
+                            "deprecated, please use "
+                            "clang-diagnostic-<checker-name> instead")
```
Szelethus, line 236: This feels a little sterile. How about "Since 6.22.0, using 'W' prefix for compiler warnings 
 is deprecated. Please change from '`{checker_name}`' to 'clang-diagnostic-`{checker_name[1:]}'`."

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -227,16 +229,30 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
...
-                    compiler_warnings.append('-Wno-' + warning_name)
-
+                    compiler_warnings.add('-Wno-' + warning_name)
                 continue
-            elif checker_name.startswith('clang-diagnostic'):
-                checker_name += '*'
+
+            # If clang-diagnostic is enabled, else add it as a disabled
+            # check.
+            if checker_name.startswith('clang-diagnostic-') and \
```
Szelethus, line 249: We already have the variable `warning_name`, that holds the information on whether we recognized the checker as compiler warning, no? Why are we reinveting the wheel instead of reusing the result of `get_compiler_warning_name`?

Szelethus (changes requested):
The story here, as I understand it is the following. Both the CLI and the GUI could incorrectly let you believe that we handle compiler warnings as checkers, but that is only party true. We can't reliably enable/disable them. Not only that, the checker name they are displayed with start with "compiler-diagnostic-<warning name>", but they must be named as "W<warning name>" on the CLI.

> Previously the handling of clang-diagnostics was flawed, because tidy cannot enable those as a checker, only suppression of the resulting reports are available:

Yes, because compiler warnings are flags to the _compiler_, not the _analyzer_ component. This makes some sense for a compiler developer, but as a user of a static analyzer tool, compiler warnings _should_ be checkers. If clang-tidy won't do it on its own for whetever reason, we need to make it so that CodeChecker acts as if they were.

> https://github.com/kimgr/clang-tools-extra/blob/master/docs/clang-tidy.rst

This is painfully out of date. Back in the day, LLVM didn't have a single monorepo housing all of its subprojects, but all subprojects had their own repository (imagine how painful an LLVM-wide refactoring used to be!). Not only is this repo from that era, it is ~6 years before the merging, and its not even the official mirror either.

This is probably what you are after:
https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics

> After this patch all of these checkers will be treated as compiler warning flags,

Not the other way around?

> this way the compiler warnings can be enabled as clang-diagnostics.

Meaning that the name of the checker on the GUI and the name to be used on the CLI will be identical.

vodorok:
@Szelethus 
> The story here, as I understand it is the following. Both the CLI and the GUI could incorrectly let you believe that we handle compiler warnings as checkers, but that is only party true. We can't reliably enable/disable them. Not only that, the checker name they are displayed with start with "compiler-diagnostic-", but they must be named as "W" on the CLI.

I think you are mistaken here. The intent is to forgo the `W` designation, and from the perspective of the CodeChecker CLI use only the `clang-diagnostic` naming.

The following matrix shows the current state of compiler warning activation

|         a         | -W...           | -Wno...  | clang-diagnostics... |
| ------------- |-------------| -----|--------|
| --enable      | adds -W... flag to the build command | adds -Wno... flag to the build command | adds clang-diagnostic-... to the checks list / not working |
| --disable     | adds -Wno... flag to the build command  | adds -W.. flag to the build command | adds -clang-diagnostic-.. to the checks list / suppresses the compiler warnings based on checker name prefixes |

> > Previously the handling of clang-diagnostics was flawed, because tidy cannot enable those as a checker, only suppression of the resulting reports are available:
> 
> Yes, because compiler warnings are flags to the _compiler_, not the _analyzer_ component. This makes some sense for a compiler developer, but as a user of a static analyzer tool, compiler warnings _should_ be checkers. If clang-tidy won't do it on its own for whetever reason, we need to make it so that CodeChecker acts as if they were.

This is the intention of the patch.

> > https://github.com/kimgr/clang-tools-extra/blob/master/docs/clang-tidy.rst
> 
> This is painfully out of date. Back in the day, LLVM didn't have a single monorepo housing all of its subprojects, but all subprojects had their own repository (imagine how painful an LLVM-wide refactoring used to be!). Not only is this repo from that era, it is ~6 years before the merging, and its not even the official mirror either.

Thanks for pointing out that I included an outdated (and forked) documentation. The official docs still state the same.
https://clang.llvm.org/extra/clang-tidy/
Look below the checker group table


From the official clang-tidy docs:
> Clang diagnostics are treated in a similar way as check diagnostics. Clang diagnostics are displayed by clang-tidy and can be filtered out using the -checks= option. However, the -checks= option does not affect compilation arguments, so it cannot turn on Clang warnings which are not already turned on in the build configuration. The -warnings-as-errors= option upgrades any warnings emitted under the -checks= flag to errors (but it does not enable any checks itself).

> This is probably what you are after: https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics
This was it thanks!

> > After this patch all of these checkers will be treated as compiler warning flags,
> 
> Not the other way around?

No, because of the previously mentioned facts.

> > this way the compiler warnings can be enabled as clang-diagnostics.
> 
> Meaning that the name of the checker on the GUI and the name to be used on the CLI will be identical.
This is true.

Before the patch: 
* You tried to use `-e clang-diagnostic-vla` you did not get any result because tidy wouldn't insert the corresponding compiler warning to the input compilation command.

After the patch:
* You use `-e Wunused-variable` you get a warning from CodeChecker to use the `clang-diagnostic-unused-variable` counterpart.
* You can use the `clang-diagnostic-vla` naming to enable the vla-related compiler warnings.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
vodorok: Good suggestion, thanks.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
vodorok: I thought about this, but `get_compiler_warning_name` is not a private member. In the future somebody could call it in some other context, then it would warning even if it is not related.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -227,16 +229,30 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
...
                 elif state == CheckerState.disabled:
-                    compiler_warnings.append('-Wno-' + warning_name)
-
+                    compiler_warnings.add('-Wno-' + warning_name)
                 continue
-            elif checker_name.startswith('clang-diagnostic'):
-                checker_name += '*'
+
+            # If clang-diagnostic is enabled, else add it as a disabled
+            # check.
```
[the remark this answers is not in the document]
Szelethus: This english is still a little broken.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -324,7 +345,10 @@ def construct_analyzer_cmd(self, result_handler):
                     has_flag('--std', analyzer_cmd):
                 analyzer_cmd.append(self.buildaction.compiler_standard)
 
-            analyzer_cmd.extend(compiler_warnings)
+            if config.enable_all:
+                analyzer_cmd.append("-Weverything")
```
Szelethus, line 349: This deserves some explanation in the code. Possibly in the summary.

bruntib (approved):

web/tests/functional/report_viewer_api/test_get_run_results.py
```diff
@@ -252,9 +252,24 @@ def test_get_run_results_severity_sort(self):
...
+                            # TODO Hacking in progress
+                            # On github actions the lexicographical order
+                            # of filenames are different than on the local
+                            # machine, and fails the gating action.
+                            # This is a temporary solution, to pass
+                            # the tests until it is fixed.
+                            # On local machinie this is the order:
+                            # path_begin.cpp -> path_begin1.cpp
+                            # On gh the order is reversed.
+                            # Apart from this the order looks good.
```
Szelethus, line 271: Do you suspect where this needs to be fixed in the code? Maybe you could mention that ("I suspect that `worst_sorterino` needs to be tinkered with in `analyze.py`").

Szelethus (changes requested):
Tiny things left to attend to.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
vodorok: Yes this is a new development, I will include the explanation in the summary.

web/tests/functional/report_viewer_api/test_get_run_results.py
vodorok: I have some rough ideas, will include these in the comments, unfortunately, I don't have any conclusion yet.
I will create an issue too.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
Szelethus: I still think this is bad design. The code would look more readable if we moved that logic to `get_compiler_warning_name` (where it belongs), and we really need to, check whether this is a disabled clang-diagnostic _after the fact_.

Szelethus (changes requested):
Ugh, sorry :)

The summary is nice.

Szelethus (approved):
LGTM! If you wanna make things a bit simpler, go ahead, I'll leave that to you.

PR #3876, opened 2023-04-13
[cfg] Upgrade vuetify version

bruntib:


Szelethus (commented):
Is there anything specific to gain? Or we just want the latest and greatest because why not?

bruntib:
There is an internal requirement to upgrade the versions of some 3pp dependencies.

vodorok (approved):
LGTM

PR #3888, opened 2023-04-19
6.22.1 highlights

vodorok:


web/server/vue-cli/src/views/NewFeatures.vue
```diff
@@ -2,6 +2,28 @@
...
+              target="_blank"
+              class="white--text"
+            >
+              Highlights of CodeChecker 6.22.1 release
+            </a>
+          </template>
+
+          <new-feature-item>
+            <template v-slot:title>
+              Bugfix, related to CodeChecker server
```
Szelethus, line 19: I don't know, I'm not too excited about this. Maybe there isn't anything to highlight this release -- fixing a bug is important, but maybe this is not the page for its celebration.

vodorok: We already released this highlights page as is in v6.22.1. In a subsequent pull request, we may change it to better represent the patch release.

bruntib (approved):

PR #3890, opened 2023-04-20
[GUI] Add a tooltip about Diff

Szelethus:
image
image


bruntib (approved):
Please, rebase it to master so the tests pass, otherwise looks good.

Issue #3892, opened 2023-04-21
Actions fail due outdated funcparserlib 


vodorok:
Github actions consistently fail due to an error in funcparserlib.
```
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [1 lines of output]
      error in funcparserlib setup command: use_2to3 is invalid.
      [end of output]
```
It is an implicit dependency for Codechecker, `mockldap` uses it.
If upgrading mockldap does not solve the issue, specifying directly could solve the issue.

PR #3894, opened 2023-04-26
[test] Get rid of mockldap

bruntib:
Mockldap dependency is not supported for a long time. It can't be
installed to the latest Ubuntu versions, due to some of its old version
implicit dependencies (funcparserlib). For this reason the usage of
mockldap module has been eliminated and a simplified mock class has been
introduced instead with hardcoded return values.

This error could be reproduced by "make venv_dev" command which emitted
this error message:
"error in funcparserlib setup command: use_2to3 is invalid."

Fixes https://github.com/Ericsson/codechecker/issues/3892

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -164,6 +164,7 @@ jobs:
           export PGPASSFILE=$HOME/.pgpass
 
           make pip_dev_deps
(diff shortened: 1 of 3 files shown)


Szelethus (approved):
LGTM

PR #3897, opened 2023-04-28
[report-converter] Enable multiple input for report-converter

bruntib:
Sometimes dynamic analysis is executed for the code base several times with different parameters. The output of these analyses may go to different log files which are given to report-converter one-by-one. If there are reports for the same source file in different log output files then the resulting .plist files will be overwritten at a subsequent execution of report-converter.

In this patch we make it possible to give multiple log files or directories containing log files to the report-converter. The reports in these files and directories will be collected all and will be converted to .plist without overriding the ones coming from a different log file.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/cli.py
+++ b/tools/report-converter/codechecker_report_converter/cli.py
@@ -9,6 +9,7 @@
 
 
 import argparse
+from collections.abc import Iterable, Sequence
 import glob
 import importlib
 import logging
@@ -16,7 +17,7 @@ import os
 import shutil
 import sys
 
-from typing import Dict, Optional, Tuple
+from typing import Any, Dict, Iterable, Optional, Sequence, Tuple, Union
 
 
 # If we run this script in an environment where 'codechecker_report_converter'
@@ -79,7 +80,7 @@ supported_metadata_keys = ["analyzer_command", "analyzer_version"]
 
 
 def transform_output(
-    analyzer_result: str,
+    analyzer_result: Iterable[str],
     parser_type: str,
     output_dir: str,
     file_name: str,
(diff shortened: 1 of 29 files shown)


tools/report-converter/codechecker_report_converter/cli.py
```diff
@@ -118,15 +118,46 @@ def process_metadata(metadata) -> Tuple[Dict[str, str], Dict[str, str]]:
...
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: str | Sequence[Any] | None,
+        option_string: str | None = None
+    ):
+        # Type of "values" can have any of the indicated types above. But in
+        # argument parser it is given by "nargs='+'", so it will be a list in
+        # this specific case.
+        assert isinstance(values, Sequence)
```
vodorok, line 138: The example about defining custom actions in the python documentation show that checking the number of items can be done in the constructor
```
def __init__(self, option_strings, dest, nargs=None, **kwargs):
        if nargs is None:
            raise ValueError("nargs needed")
        super().__init__(option_strings, dest, **kwargs)
```
https://docs.python.org/3/library/argparse.html#action
https://docs.python.org/3/library/argparse.html#argparse.Action

vodorok (changes requested):
Please see my remarks

tools/report-converter/codechecker_report_converter/cli.py
bruntib: Thanks, I didn't notice that there is a standard place for checking this precondition :)

tools/report-converter/codechecker_report_converter/cli.py
```diff
@@ -118,15 +118,46 @@ def process_metadata(metadata) -> Tuple[Dict[str, str], Dict[str, str]]:
...
+
+        all_files = []
+
+        for path in values:
```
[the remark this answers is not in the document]
bruntib: Alright, we could use that path handler in the future. However, I would say that this shouldn't be done in this PR. The old file handler API is used at many places in this module. Those should be rewritten in a separate PR.

vodorok (approved):
LGTM

PR #3901, opened 2023-04-30
Makefile: package_gerrit_skiplist should depend on package_dir_structure

tuankiet65:
In the Makefile, the `package_gerrit_skiplist` target copies a script to a directory created in the `package_dir_structure` target. However, `package_gerrit_skiplist` does not have a dependency on `package_dir_structure`, and it is possible that `package_gerrit_skiplist` is run before the directory created in `package_dir_structure` actually exists. This can be reproduced by running `make` with multiple jobs (e.g `make -j16`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ package_dir_structure:
 mkdocs_build:
 	mkdocs build
 
-package_gerrit_skiplist:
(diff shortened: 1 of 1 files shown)


bruntib:
Thanks for the fix, nice catch!

PR #3902, opened 2023-05-04
[analyzer] Disable cppcheck-preprocessorErrorDirective explicitly

bruntib:
This checker gives a lot of false positive reports mainly because cppcheck doesn't support the analysis on different target architectures.


Szelethus (approved):
LGTM

PR #3905, opened 2023-05-10
Upgrade nightwatch

vodorok:


Szelethus (changes requested):
Please don't forget the labels and milestones. At the very least, I do find them helpful. Its not immediately obvious to me whether you want this for the bugfix release or not, so I can't always fill it out.

bruntib (approved):

bruntib (changes requested):
Sorry, I've just checked the GUI tests. Now it gives quite different kinds of errors. Is this a good thing, or a bad thing? :)

whisperity:
I'd really like this to work, as dealing with old nightwatch lack of features was troubling when implementing #4202.

However, nightwatch version 3.y is out now as well. Can we do the migration in one step, or should it be separate steps for major version changes?

bruntib:
We should come back to this issue after upgrading CodeChecker to Vue3. I remove this PR from the current milestone, but nightwatch upgrade will definitely be necessary.

PR #3917, opened 2023-05-22
Fix exception when SourceLine has no source_path attribute

Spatrik95:
The `report-converter` fail with an exception when a SpotBugs input has a `<SourceLine>` element without a `source_path` attribute.

---

Example for an element like above explained from a SpotBugs run:

```xml
<Method classname="org.w3c.dom.Element" name="getAttribute" signature="(Ljava/lang/String;)Ljava/lang/String;" isStatic="false" role="METHOD_RETURN_VALUE_OF">
  <SourceLine classname="org.w3c.dom.Element"/>
  <Message>Return value of org.w3c.dom.Element.getAttribute(String) of type String</Message>
</Method>
```

In this case the `report-converter` failed with the following exception:

```text
  File "<personal path>\codechecker\tools\report-converter\codechecker_report_converter\cli.py", line 243, in <module>
    main()
  File "<personal path>\codechecker\tools\report-converter\codechecker_report_converter\cli.py", line 237, in main
    return transform_output(
  File "<personal path>\codechecker\tools\report-converter\codechecker_report_converter\cli.py", line 100, in transform_output
    parser.transform(
  File "<personal path>\codechecker\tools\report-converter\codechecker_report_converter\analyzers\analyzer_result.py", line 57, in transform
    reports = self.get_reports(analyzer_result_file_path)
  File "<personal path>\codechecker\tools\report-converter\codechecker_report_converter\analyzers\spotbugs\analyzer_result.py", line 47, in get_reports
    report = self.__parse_bug(bug)
  File "<personal path>\codechecker\tools\report-converter\codechecker_report_converter\analyzers\spotbugs\analyzer_result.py", line 122, in __parse_bug
    event = self.__event_from_method(element)
  File "<personal path>\codechecker\tools\report-converter\codechecker_report_converter\analyzers\spotbugs\analyzer_result.py", line 177, in __event_from_method
    source_path = self.__get_abs_path(source_path)
  File "<personal path>\codechecker\tools\report-converter\codechecker_report_converter\analyzers\spotbugs\analyzer_result.py", line 59, in __get_abs_path
    if os.path.exists(source_path):
  File "<personal path>\miniconda3\lib\genericpath.py", line 19, in exists
    os.stat(path)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

With this modification the converter won't be interrupted by an exception, but gives a warning message:

```text
[WARNING] - No source path attribute found for class: org.w3c.dom.Element
```

### DIFF
--- a/tools/report-converter/codechecker_report_converter/analyzers/spotbugs/analyzer_result.py
+++ b/tools/report-converter/codechecker_report_converter/analyzers/spotbugs/analyzer_result.py
@@ -105,8 +105,7 @@ class AnalyzerResult(AnalyzerResultBase):
         long_message = bug.find('LongMessage').text
 
         source_line = bug.find('SourceLine')
-        source_path = source_line.attrib.get('sourcepath')
-        source_path = self.__get_abs_path(source_path)
+        source_path = self.__get_source_path(source_line)
         if not source_path:
             return
 
@@ -148,11 +147,7 @@ class AnalyzerResult(AnalyzerResultBase):
         message = element.find('Message').text
 
         source_line = element.find('SourceLine')
-        if source_line is None:
-            return None
-
-        source_path = source_line.attrib.get('sourcepath')
-        source_path = self.__get_abs_path(source_path)
+        source_path = self.__get_source_path(source_line)
         if not source_path:
             return None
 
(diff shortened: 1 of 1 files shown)


bruntib (changes requested):
Thanks for the contribution, it's a nice catch.
I have two comments, though:
- The content of `__get_abs_path()` could be merged to `__get_source_path()` if only it uses it.
- I'll enable the automatic tests which will probably warn about too long lines.

bruntib (approved):

PR #3918, opened 2023-05-22
Implement Code Similarity Detection with jscpd Tool

Ducnguyen147:
…at for integration with CodeChecker


bruntib:
Hi,
This is a great work, thank you!
I have just one usability question: the copy-paste codes are found properly and displayed in the CodeChecker interface. However, it shows only one message per function: "this is a copy-paste". But it doesn't display, where is the other code fragment which is the same. Could this be shown with a bug note?

PR #3919, opened 2023-05-22
[analyzer] Use absolute path to logger.so in LD_PRELOAD

vodorok:
It can happen that a build system overwrites the LD_LIBRARY_PATH environment variable, this could result in the following error:

```
“ERROR: ld.so: object 'ldlogger.so' from LD_PRELOAD cannot be preloaded: ignored”
```


bruntib (approved):

Szelethus (approved):
LGTM

PR #3927, opened 2023-05-31
[feat] Exclude dynamic parts of checker message in hash generation

bruntib:
Some dynamic analyzers, like sanitizers generate dynamic parts in the checker message (e.g. memory addresses). These are changing in every execution of the analyer. Since the checker message is part of the report hash, the hash changes too. This is bad, because the report itself is supposed to be the same.

The report converter can provide a static_message field for the report which is a stable version of the checker message. If this field is given then it will be used for hash generation.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/report/__init__.py
+++ b/tools/report-converter/codechecker_report_converter/report/__init__.py
@@ -292,8 +292,21 @@ class Report:
         bug_path_positions: Optional[List[BugPathPosition]] = None,
         notes: Optional[List[BugPathEvent]] = None,
         macro_expansions: Optional[List[MacroExpansion]] = None,
-        annotations: Optional[Dict[str, str]] = None
+        annotations: Optional[Dict[str, str]] = None,
+        static_message: Optional[str] = None
     ):
+        """
+        This constructor populates the members of the Report object.
+
+        static_message: hash.get_report_hash() function generates a hash value
+            based on this report object. The checker message is usually the
+            part of this hash. However, sometimes a checker message may contain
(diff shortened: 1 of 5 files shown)


tools/report-converter/codechecker_report_converter/report/__init__.py
```diff
@@ -307,6 +320,9 @@ def __init__(
         self.type = type
         self.annotations = annotations
 
+        self.static_message = \
```
Szelethus, line 323: Do we need a new field? Can we not just change `self.message`?

bruntib: No, because we still want to use the original value of `self.message` to display it for the user. The value of this field goes to the GUI and command line output.

tools/report-converter/codechecker_report_converter/report/hash.py
```diff
@@ -87,7 +87,7 @@ def __get_report_hash_path_sensitive(report: Report) -> List[str]:
 
         hash_content = [report.file.name,
                         report.checker_name,
-                        event.message,
+                        report.static_message,
```
Szelethus, line 90: You documented this field for the report class, but you ough to in the function as well. It still talks about "checker message".

Szelethus (approved):
Please address my comment. Otherwise LGTM.

tools/report-converter/codechecker_report_converter/report/hash.py
bruntib: I updated the documentation, thanks for the comment!

PR #3930, opened 2023-06-05
[ldlogger] Recognize linux_spawn alongside exec* calls

Szelethus:
The simple idea behind ldlogger is the following -- create a .so file
that we force to be loaded for each program that is executed by adding
it to the LD_PRELOAD environmental variable. In that .so, we shadow
exec* calls with our impostor exec* calls -- instead of the real thing,
our version is called, which logs the compile command, and then calls
the actual system function.

Up to, but not including GNU Make 4.3, only exec* system calls were
used. However, as of 4.3, posix_spawn is used by default, so this patch
adds shadow functions for posix_spawn and the very similar posix_spawnp.

Related discussion: https://github.com/Ericsson/codechecker/discussions/3906

### DIFF
--- a/analyzer/tools/build-logger/src/ldlogger-hooks.c
+++ b/analyzer/tools/build-logger/src/ldlogger-hooks.c
@@ -17,23 +17,51 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#include <spawn.h>
 #include <stdbool.h>
 
 #include "ldlogger-hooks.h"
+
+// We are done intercepting system calls by pretending that the functions below
+// are the real thing -- time to call the actual system call.
+#define CALL_ORIGINAL_FN(funName_, arglist, ...) \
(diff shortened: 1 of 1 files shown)


analyzer/tools/build-logger/src/ldlogger-hooks.c
```diff
@@ -17,23 +17,51 @@
...
+  if (!fun) \
+  { \
+    return -1; \
+  } \
+  return (*fun)( __VA_ARGS__ );
+
 #define CC_LOGGER_CALL_EXEC(funName_, arglist, ...) \
   tryLog(__VA_ARGS__); \
   { \
     unsetLDPRELOAD(__VA_ARGS__); \
```
vodorok, line 39: Are you confident, that this unsetLDPRELOAD is not needed in the newly implemented spawn wrapper?

vodorok (approved):

PR #3932, opened 2023-06-14
[test][NFC] Change from nose to pytest (web library)

Szelethus:
Continuation of #3931 and #3926.

For your convenience, I split this patch up into several parts -- in light of the previous patches, the first two commits shouldn't offer anything new.

edit: As a note, pytest doesn't support package level setup/teardown, only class level. Packages where only a single class is present, these are equivalent. This is why the 2nd commit isn't particularly exciting. :) 

The 3rd one needs a little explanation:
* `web/tes/funtional/authentication/`:
  * This test creates a temporary server. Previously, this was done for the entire directory -- create a server, run all tests within the directory, and then close it. Since per-package level setup/teardown isn't a thing in pytest, I changed this to per-class. However, it turns out that we didn't tear down the temporary server properly. The server is created by spawning a multiprocessing thread that listens to an event. If the event is set, the thread (and the CodeChecker server inside it) will be terminated, however, that event needs to be cleared -- otherwise it will still be set by the time the server would be reopened, and as a result, would immediately close.
  * The tests depended one another, and messed with each other, as indicated by my comments. These being separated is just a 100% win.
* The others don't stand out. They all connect to a CodeChecker server shared across all tests. Previously, per-package, in the setup, they created a product, and in the teardown, they removed that product. Now that this has to be done per-class, we stumbled across a bug -- if a product is removed, the endline wouldn't be freed up for it. As a result, we can't just create and remove the same product over and over again, and need to create a new product for each class in the package.

The 4th one is something I still need some fixing on. The only reason I haven't quite done it yet, because I didn't break it, it was broken already, so it feels appropriate to address it in a separate patch :)


web/client/tests/unit/__init__.py
```diff
@@ -9,15 +9,17 @@
...
-
 REPO_ROOT = os.path.abspath(os.environ['REPO_ROOT'])
 PKG_ROOT = os.path.join(REPO_ROOT, 'build', 'CodeChecker')
 
+os.environ["CC_DATA_FILES_DIR"] = PKG_ROOT
+
 sys.path.append(REPO_ROOT)
-sys.path.append(os.path.join(REPO_ROOT, 'web'))
+sys.path.append(os.path.join(
+  REPO_ROOT, 'analyzer', 'tools', 'statistics_collector'))
```
bruntib, line 23: Why is this tool here in a web-related makefile? The analyzer and web related parts of CodeChecker are supposed to be hermetically separated.

web/pytest.ini
```diff
@@ -0,0 +1,13 @@
...
+# increase verbosity level
+  --verbose
+
+# stop running tests on first error
+# FIXME: Pretty please comment this horrible setting out, I hate it with a
+# passion
+#  --maxfail=1
+
+# do not capture stdout
+  --capture=fd
```
bruntib, line 13: In analyzer tests this option was `--capture=sys`. Shouldn't they match?

bruntib (changes requested):

vodorok (approved):
LGTM

vodorok:
One note, It seems that the pr contains another commit, `[ldlogger] Recognize linux_spawn alongside exec* calls`, please rebase before merging.

bruntib (approved):

PR #3933, opened 2023-06-14
[test][NFC] Remove every remaining trace of nose in favor of pytest

Szelethus:
Turns out we never really needed nose for the codechecker_common library. I tried to `git grep` and things like that, but it seems like this patch (after the web library PR) should get rid of nose for good.


codechecker_common/requirements_py/dev/requirements.txt
```diff
@@ -1,4 +1,3 @@
-nose==1.3.7
```
bruntib, line 1: I believe, this works because pytest is installed via the `venv_dev` of analysis and web modules. But shouldn't we add pytest in this requirements.txt explicitly?

bruntib (changes requested):

codechecker_common/requirements_py/dev/requirements.txt
Szelethus: `codechecker_common` didn't use `nosetests` at all, and doesn't use `pytest` either.

bruntib (approved):

PR #3934, opened 2023-06-15
[analyzer][cmd] Adapt to new clang-tidy checker options format.

tinloaf:
Starting with LLVM15 / clang-tidy-15, the output of `--dump-config` has changed, and the old parser failed to parse the options in the new format. We need full YAML parsing now, the format looks like this:

```
---
Checks:          'clang-diagnostic-*,clang-analyzer-*,*'
AnalyzeTemporaryDtors: false
FormatStyle:     none
[…]
CheckOptions:
  performance-unnecessary-value-param.IncludeStyle: llvm
  readability-suspicious-call-argument.PrefixSimilarAbove: '30'
[…]
```

This solution implements a simple trial-and-error approach: It first tries the old parsing version, and if that fails, it tries the new one. This seems a lot simpler than first trying to figure out the clang-tidy version by e.g. parsing the output of `clang-tidy --version`.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
@@ -18,6 +18,8 @@ import shlex
 import subprocess
 from typing import List, Tuple
 
+import yaml
+
 from codechecker_common.logger import get_logger
 
 from codechecker_analyzer import analyzer_context, env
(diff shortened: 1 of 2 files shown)


tinloaf:
@bruntib I just added unit tests. I didn't think they fit very well into any of the existing test files, so I added them in a new file `analyzer/tests/unit/test_checker_option_parsing.py`.  

When I run all unit tests, I see *a lot* of failures. I guess that my system is somehow not set up correctly to run the CodeChecker unit tests. However, the two new tests run successfully on my machine.

bruntib:
Thanks for the patch and the tests!

PR #3935, opened 2023-06-15
[test] Allow additional pytest args to be given through make targets

Szelethus:
Previously, unless you wanted to directly invoke pytest (and even before
that, nosetests), you had to edit the corresponding config file in order
to provide additional args to pytest.
Invoking pytest directly is possible, but cumbersome -- you need to
remove the workspace directory, set environmental variables `REPO_ROOT`,
`BUILD_DIR`, `TEST_PROJ` and `TEST_CLANG_VERSION`.

I am not saying this is ideal, maybe we should make direct invocation a
little less painful, but until we do that, I added support for another
environmental variable, `EXTRA_PYTEST_ARGS`. For instance the following
invocation selects only the test matching 'test_source_suppress_export':

```
EXTRA_PYTEST_ARGS="-k test_source_suppress_export" TEST=tests/functional/suppress make test_analyzer_feature
```

### DIFF
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -49,6 +49,15 @@ PostgreSQL database backend.
 Further configuration options can be found here
 [Pytest configuration](https://docs.pytest.org/en/7.1.x/reference/customize.html).
 
+You can specify additional pytest arguments to the make targets by using the
+environmental variable `EXTRA_PYTEST_ARGS`:
+```sh
+EXTRA_PYTEST_ARGS='-k test_source_suppress_export'  TEST=tests/functional/suppress make test_analyzer_feature
+```
+The above example displays how you can select a specific testcase via the `-k`
+option passed through `EXTRA_PYTEST_ARGS`, within a specific testfile via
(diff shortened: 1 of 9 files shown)


docs/tests.md
```diff
@@ -49,6 +49,12 @@ PostgreSQL database backend.
 Further configuration options can be found here
 [Pytest configuration](https://docs.pytest.org/en/7.1.x/reference/customize.html).
 
+You can specify additional pytest arguments to the make targets by using the
+environmental variable `EXTRA_PYTEST_ARGS`:
+```sh
+EXTRA_PYTEST_ARGS='-k test_source_suppress_export'  TEST=tests/functional/suppress make test_analyzer_feature
```
bruntib, line 55: Could you please add one more sentence about the purpose of this `-k` flag? I think this is a very useful knowledge that this is the way of running a specific test function. Thanks!

bruntib (changes requested):

bruntib (approved):

PR #3943, opened 2023-06-22
[fix] Analysis shouldn't fail on non-existing directory

bruntib:
When the compilation command database contains the build information of a file in a non-existing directory then the pre-analysis phase of CTU shouldn't fail.

### DIFF
--- a/analyzer/codechecker_analyzer/pre_analysis_manager.py
+++ b/analyzer/codechecker_analyzer/pre_analysis_manager.py
@@ -127,9 +127,8 @@ def pre_analyze(params):
                                       ctu_temp_fnmap_folder)
 
     except Exception as ex:
-        LOG.debug_analyzer(str(ex))
+        LOG.error("Pre-analysis failed for %s: %s", action.source, str(ex))
         traceback.print_exc(file=sys.stdout)
-        raise
 
     try:
         if statistics_data:
--- a/analyzer/tests/functional/ctu/test_ctu.py
+++ b/analyzer/tests/functional/ctu/test_ctu.py
@@ -11,6 +11,7 @@
 
 import subprocess
 import glob
+import json
 import os
 import shutil
 import unittest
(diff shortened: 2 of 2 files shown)


analyzer/tests/functional/ctu/test_ctu.py
```diff
@@ -103,6 +104,31 @@ def teardown_method(self, method):
...
 
+    @skipUnlessCTUCapable
+    def test_ctu_non_existing_dir(self):
+        """
+        If the compilation database contains a build action with non-existing
+        directory then analysis shouldn't fail in the pre-analysis phase of
+        CTU.
+        """
+        self.teardown_method(self.test_ctu_non_existing_dir)
+        self.setUpWith('test_files_c', 'buildlog.json', 'reports')
```
Szelethus, line 115: Not exactly sure why you need these two. Isn't the default setup_method sufficient?

Szelethus (commented):
> When the compilation command database contains the build information of a file in a non-existing directory then the pre-analysis phase of CTU shouldn't fail.

Is that okay though? When can that naturally happen and not be a sign of a user error? I am not a fan on this patch.

analyzer/codechecker_analyzer/pre_analysis_manager.py
```diff
@@ -129,7 +129,6 @@ def pre_analyze(params):
     except Exception as ex:
         LOG.debug_analyzer(str(ex))
         traceback.print_exc(file=sys.stdout)
-        raise
```
Szelethus, line 132: Also, I strongly disagree with this blanket removal, even if we were to go through with what this patch proposes -- this gets rid of all exceptions, not just FileNotFoundError.

Szelethus (changes requested):

analyzer/tests/functional/ctu/test_ctu.py
bruntib: `teardown_method()` removes the previous report directory created by some other test function. I need to remove this otherwise `setUpWith()` fails to recreate this directory.

Szelethus (changes requested):
Please add a LOG warning or an error (and test it).

analyzer/tests/functional/ctu/test_ctu.py
Szelethus: That doesn't really answer what I was curious about -- I understand you need `teardown_method()` before manually invoking `setUpWith()`, but isn't the regular `setup_method` that runs automatically good enough? If not, why? Could you put the explanation in the code?

analyzer/codechecker_analyzer/pre_analysis_manager.py
```diff
@@ -127,9 +127,8 @@ def pre_analyze(params):
                                       ctu_temp_fnmap_folder)
 
     except Exception as ex:
```
Szelethus, line 129: Did you remove the special handling of FileNotFoundError? Why?

Szelethus (changes requested):
Its not your fault, but I'm not thrilled about the reasons behind this patch. If someone runs all their configure and setup scripts inside `CodeChecker log`, and logs build commands for temporary files, thats a user error.

I understand that we need to bend the knee on this one. But I'd prefer to ease up on the hard error / exception escape (before proper logging is implemented) as little as possible.

analyzer/tests/functional/ctu/test_ctu.py
bruntib: Now I understand your comment. I removed these two lines, thank you!

Szelethus (approved):
LGTM! I'm happy with where we landed with this all things considered.

PR #3944, opened 2023-06-27
Fix local local diff src code suppression

Szelethus:
Pretty much what it says on the tin.


Szelethus:
> ```
> CodeChecker cmd diff -b reports_a -n reports_b --new
> ```
> 
> This command shows 0 reports.

Actually, both according to what I see locally, as well as what can be see in the tests attached, that is not the case. The report does show up in the NEW set.

bruntib:
I think we should discuss this question again, because I still can't see new reports for my example above, even if the source files are in different folders.

bruntib (approved):

PR #3949, opened 2023-07-10
Eliminate "unknown" checker state

bruntib:
[[cmd] Eliminate default checker status](https://github.com/Ericsson/codechecker/commit/34143c0ed4d4d7cef5aed44a122ac08949854100)

Currently there are 3 kind of checkers regarding whether they are
running during analysis: enabled, disabled, unknown. Checkers can be
enabled if they are explicitly given to --enable flag, or they are the
member of "default" profile. Checkers can also be explicitly disabled.
The 3rd status is tricky: CodeChecker doesn't explicitly enable or
disable them, but gives the choice to the analyzer tool.

This behavior is bad, because a user cannot tell exactly which checkers
were executed during analysis. It is impossible to determine in case of
no error if a checker was disabled or just simply didn't report any
issue.

The goal in this commit is that every checker is either enabled or
disabled. It is the analyzer module's responsibility to assemble an
analysis command which executes at least the enabled checkers and turns
off or hides the disabled ones.

[[feat] ClangTidy enables only selected checkers](https://github.com/Ericsson/codechecker/commit/eef703567e1336770eb0e7d5a0eb7b1632cb6e2c)

Due to the previous commit every checker is either enabled or disabled.
This resulted very long analyzer commands in ClangTidy because every
checker was explicitly enabled or disabled in the clang-tidy invocation.
ClangTidy has a lot of checkers and also all compiler warnings listed by
diagtool are also considered as checkers.

The goal of this commit is to simplify the clang-tidy invocation command.
First we disable all checkers and enable only those that are necessary to
run:

clang-tidy -checks=-*,enabled1,enabled2...

This patch takes into account that clang-diagnostic-... checkers don't
turn on the corresponding warnings. They have to be enabled by -W...
compiler flag.

[[feat] Hide disabled ClangSA checkers in post-processing](https://github.com/Ericsson/codechecker/commit/bc2aa30326c1e85acb46e804e01e3ac88ed66ca6)

ClangSA checkers may rely on each other. For example unix.Malloc checker
may prevent a report from unix.MismatchedDeallocator. It is possible that
after disabling unix.Malloc the underlying analysis continues giving a
chance for unix.MismatchedDeallocator to emit a report.

In order to avoid this noisy behavior, ClangSA checkers are never
disabled during the analysis invocation. If CodeChecker wants to disable
a checker then its result should be hidden in a post-processing step.


analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -307,6 +340,7 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
                                 "instead.")
                     if state == CheckerState.enabled:
                         compiler_warnings.add('-W' + warning_name)
+                        enabled_checkers.add(checker_name)
```
Szelethus, line 343: Can you document why were doing this in code?

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -320,15 +354,21 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
                 elif warning_type == CheckerType.analyzer:
                     if state == CheckerState.enabled:
                         compiler_warnings.add('-W' + warning_name)
-                    elif state == CheckerState.disabled:
-                        checkers.append('-' + checker_name)
+                        enabled_checkers.add(checker_name)
```
Szelethus, line 357: I suppose the comments above are outdated.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -165,13 +165,51 @@ def get_warnings(env=None):
         raise
 
 
+def _add_star_for_group(
```
Szelethus, line 168: Wouldn't 'asterisk' be the correct term?

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -165,13 +165,51 @@ def get_warnings(env=None):
...
+    checkers: Iterable[str],
+    all_checkers: Set[str]
+) -> List[str]:
+    """
+    Since CodeChecker interprets checker name prefixes as checker groups, they
+    have to be added a '*' joker character when using them at clang-tidy
+    -checks flag. This function adds a '*' for each item in "checkers" if it's
+    a checker group, i.e. identified as a prefix for any checker name in
+    "all_checkers".
+    """
```
Szelethus, line 178: An example would help the explanation a lot.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -165,13 +165,51 @@ def get_warnings(env=None):
...
+    def is_checker_prefix(prefix: str, long: str) -> bool:
+        """
+        Returns True if a checker(-group) name is prefox of another
```
Szelethus, line 181: prefox->prefix

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -165,13 +165,51 @@ def get_warnings(env=None):
...
+    all_checkers: Set[str]
+) -> List[str]:
+    """
+    Since CodeChecker interprets checker name prefixes as checker groups, they
+    have to be added a '*' joker character when using them at clang-tidy
+    -checks flag. This function adds a '*' for each item in "checkers" if it's
+    a checker group, i.e. identified as a prefix for any checker name in
+    "all_checkers".
+    """
+    def is_checker_prefix(prefix: str, long: str) -> bool:
```
Szelethus, line 179: How about `is_group_prefix_of`?

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -165,13 +165,51 @@ def get_warnings(env=None):
...
+        Returns True if a checker(-group) name is prefox of another
+        checker name. For example bugprone-string is prefix of
+        bugprone-string-constructor but not of
+        bugprone-stringview-nullptr.
+        """
+        prefix_split = prefix.split('-')
+        long_split = long.split('-')
+        return prefix_split == long_split[:len(prefix_split)]
+
+    def need_star(checker: str) -> bool:
```
Szelethus, line 190: asterisk

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -165,13 +165,51 @@ def get_warnings(env=None):
         raise
 
 
+def _add_star_for_group(
+    checkers: Iterable[str],
```
Szelethus, line 169: `checkers` and `all_checkers`? Since this is a single purpose function, how about we call this `enabled_checkers`, or `subset_checkers`?

analyzer/tests/unit/test_checker_handling.py
```diff
@@ -278,14 +330,18 @@ def test_disable_clangsa_checkers(self):
 
         for arg in analyzer.construct_analyzer_cmd(result_handler):
             if arg.startswith('-checks='):
-                self.assertIn('-clang-analyzer-*', arg)
+                self.assertLess(*self._enable_disable_pos(
```
Szelethus, line 333: This is still a chore to digest. Can't you reduce it down to something easier to read? Like, `is_disabled`?

Szelethus (changes requested):

Szelethus:
Also, my usual complaint. You explain in the summary what this patch does, but not _why_ this is any good.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
bruntib: I added this documentation to the beginning. Would you think, it explains?
```.py
# In clang-tidy the warnings as compiler-diagnostics are enabled and
# disabled by -W<name> and -Wno-<name> respectively. The
# clang-diagnostic-<name> is needed, because by default all checkers
# are disabled by "checks=-*" and warnings have to be enabled as
# checkers with this name: clang-diagnostic-<name>.
```

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
bruntib: Which part do you think is outdated?

bruntib:
I fixed the things according to the comments, thank you!
As for the last one, I believe the commit messages answer the question of "what". If it doesn't then let's elaborate on it, I would appreciate if you could help to make comments as clear as possible. However, I think some comments have to describe "why" something happens, because many times it is hard to figure it out if something is done consciously or inherited from some outdated legacy.

Szelethus:
Here is what I mean under the _why_ missing from the summary.

Ideally, I would like to learn the following:
* How the program works now
* Why is this behaviour undesirable or incorrect, and what it should be instead
* What was the intent behind this behaviour back in the day (or at least an educated guess)
* If you considered multiple high level approaches to fix this, what were they, why did you pick the one found in the patch
* Briefly describe _what_ the actual patch is about, _how_ it fixed the problem.

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -288,6 +290,37 @@ def get_analyzer_config(cls) -> List[str]:
...
         return parse_clang_help_page(command, 'OPTIONS:')
 
+    def post_analyze(self, result_handler):
+        """
+        Disabled checkers are not actually disabled during analysis, because
+        they rely on each other under the hood. The disabled checkers' reports
+        are removed in this post-processing step.
+        """
+        if not os.path.isfile(result_handler.analyzer_result_file):
+            return
```
Szelethus, line 300: When does this happen? When the analysis crashed? Or no reports were emitted?

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -288,6 +290,37 @@ def get_analyzer_config(cls) -> List[str]:
...
+
+        try:
+            with open(result_handler.analyzer_result_file, 'rb') as f:
+                plist = plistlib.load(f)
+        except plistlib.InvalidFileException:
+            # It may happen that a compilation database contains a build action
+            # multiple times. For further details see analyzer_action_str() in
+            # ResultHandler. If this happens then the analysis of the same
```
Szelethus, line 308: That comment says nothing about _multiple_ build command occurances.

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -288,6 +290,37 @@ def get_analyzer_config(cls) -> List[str]:
...
+            # build action is analyzed in parallel several times, so the same
+            # output .plist file is changed by several threads. This may result
+            # an invalid .plist which fails plistlib.load(). This is not a big
+            # problem, because the second thread will also execute this
+            # post-processing and it happens rarely anyways.
+            # test_compile_uniqueing() fails undeterministically without this.
+            return
+
+        plist['diagnostics'] = list(filter(
```
Szelethus, line 317: Plist manipulation can be painfully slow. How hard does this trickery hit the performance?

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -266,30 +313,32 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
         clang-tidy and in some cases that would cause analysis error due to
         some ClangSA bug.
         """
-        checkers = []
         # Usage of a set will remove compiler warnings and clang-diagnostics
         # which are the same.
+        # In clang-tidy the warnings as compiler-diagnostics are enabled and
+        # disabled by -W<name> and -Wno-<name> respectively. The
+        # clang-diagnostic-<name> is needed, because by default all checkers
```
Szelethus, line 320: The part after "because" doesn't explain to me why we can't preserve `-W` and -`Wno`. It just runs in a circle.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -266,30 +313,32 @@ def get_checker_list(self, config) -> Tuple[List[str], List[str]]:
...
         for checker_name, value in config.checks().items():
             state, _ = value
 
             warning_name, warning_type = \
                 get_compiler_warning_name_and_type(checker_name)
+
+            # This warning must be given a parameter separated by either '=' or
+            # space. This warning is not supported as a checker name so its
+            # special usage is avoided.
+            if warning_name and warning_name.startswith('frame-larger-than'):
```
Szelethus, line 339: Do we need to this in this patch? Seems to be a new addition. If we must, we could at least emit a curtesy warning that this won't be enabled.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
Szelethus: > Would you think, it explains?

No.

analyzer/tests/functional/analyze_and_parse/test_files/compiler_warning_no_warn.output
```diff
@@ -18,11 +18,40 @@ CHECK#CodeChecker check --build "make compiler_warning_simple" --output $OUTPUT$
 [] - To store results use the "CodeChecker store" command.
 [] - See --help and the user guide for further options about parsing and storing the reports.
 [] - ----=================----
-Found no defects in compiler_warning.cpp
+[MEDIUM] compiler_warning.cpp:3:7: unused variable 'i' [clang-diagnostic-unused-variable]
```
Szelethus, line 21: Did you intend for this to appear here? Why? I'm sure I can figure out if I stare at it long enough, but this functional test has quite a lot of cryptic changes.

analyzer/tests/unit/test_checker_handling.py
```diff
@@ -108,9 +109,15 @@ def test_no_disabled_checks(self):
         """
         Test that ClangSA only uses enable lists.
         """
-        self.assertFalse(
-            any(arg.startswith('-analyzer-disable-checker')
-                for arg in self.__class__.cmd))
+        # TODO: This test is currently removed, because checkers that are not
+        # enabled are explicitly disabled. In a next commit ClangSA reports
+        # will be hidden instead of disabled. In that commit this test could be
+        # re-enabled.
```
Szelethus, line 115: Did you not do that already?

analyzer/tests/unit/test_checker_handling.py
```diff
@@ -272,14 +338,18 @@ def test_disable_clangsa_checkers(self):
 
         for arg in analyzer.construct_analyzer_cmd(result_handler):
             if arg.startswith('-checks='):
-                self.assertIn('-clang-analyzer-*', arg)
+                self.assertTrue(self._is_disabled(
+                    'clang-analyzer',
+                    arg[len('-checks='):].split(',')))
```
Szelethus, line 343: Yeah, this isn't any better. It should look like this:
```python
self.assertTrue(self._is_disabled('clang-analyzer', arg))
```
and other nasties should be inside `_is_disabled`. Do you see my point?

Szelethus (changes requested):
If I never had an issue with wanting to disable a checker but couldn't, would I notice any change in the new behaviour? Moving on, would I need to manually enable/disable clang-diagnostics to restore the pre-patch result set?

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
bruntib: I had this concern too, but this post-processing is applied only for ClangSA currently where we assume that this will be feasible (there are no verbose style checkers). For clang-tidy and cppcheck we don't even write these reports to the .plist. Nevertheless, we plan to apply this uniform post-processing method for those too, so it will be a relevant question later. Anyways, I'm not aware of a better solution with ClangSA, because it natively generates the .plst files.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
bruntib: I think it explains. We do preserve `-W` and `-Wno`, but it's not enough, because we disable every diagnostics by default.
```
clang-tidy -checks=-* main.cpp -- -W<some_warning>
```
This command doesn't emit any diagnostics, because `-checks=-*` disables it. That's why we need to add `clang-diagnostic-<some_warning>` too:
```
clang-tidy -checks=-*,clang-diagnostic-<some_warning> main.cpp -- -W<some_warning>
```
If the phrasing is not clear, then I'd agree to rewrite it.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
bruntib: This is a function which returns the available checkers. By this `if` statement this "checker" will not be displayed for the users, do they don't even know about the existence of it. They cant provide it in any analyzer command.

analyzer/tests/functional/analyze_and_parse/test_files/compiler_warning_no_warn.output
bruntib: Because we test here that `clang-diagnostic-unused-variable` being in the "default" profile is stronger than `-Wno-all` and `-Wno-extra` in the original build command.

analyzer/tests/unit/test_checker_handling.py
bruntib: Good catch, thank you!

analyzer/tests/unit/test_checker_handling.py
bruntib: If I do it this way then `test_default_checkers_are_not_disabled()` and `test_clangsa_analyzer_checks_are_disabled()` will be ugly. Which one should we save? :)

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
bruntib: I might have misunderstood, but in your original comment you suggested to provide a "courtesy warning" for the users which says that this checker/warning will not be enabled. That's why I answered that there is no need for such a warning, because CodeChecker users will not even know about the existence of this "checker" if we exclude it in this `get_checker_list()` function.

And we have to exclude it, because otherwise an invalid `clang-tidy` invocation will be assembled: `clang-tidy ... main.c -- -Wframe-larger-than=`. This is invalid, because this warning requires an integer parameter after `=`. So this is why I decided to hide this checker/warning at all. If I don't exclude this warning then one of our tests fail which enable all "clang-diagnostic-<blabla>" checkers:
`CodeChecker analyze -e clang-diagnostic ...`

analyzer/tests/unit/test_checker_handling.py
bruntib: Done, restructured.

analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
Szelethus: Aha, I understand now that this function asks clang-tidy for the list of recognized checks, and removed this particular one, so it will act as if the requested didn't exist, resulting in a warning.

But not the _right_ warning. It will say that the check doesn't exist, which is not true -- it does exist, we just don't don't allow running it under CodeChecker. Lets save the user some headache, his Clang binary is not broken -- please emit a warning about this.

warning: CodeChecker is incompatible with compiler warning 'frame-larger-than'. Please don't enable it.

web/tests/functional/diff_local/test_diff_local.py
```diff
@@ -77,15 +77,16 @@ def setup_class(self):
             'skip_list_file': skip_list_file,
             'check_env': test_env,
             'workspace': TEST_WORKSPACE,
-            'checkers': [],
+            'checkers': ['-d', 'clang-diagnostic'],
```
Szelethus, line 80: Sorry, it seems like you have to beat this into me. Don't we disable all compiler warnings out of the box? Why is this needed? Why did you need to change almost all invocations of CodeChecker in these tests? Its a mess.

web/tests/functional/report_viewer_api/__init__.py
```diff
@@ -91,7 +93,9 @@ def setup_class_common(workspace_name):
     # checkers to have some real difference.
     codechecker_cfg['checkers'] = ['-e', 'core.CallAndMessage',
                                    '-d', 'core.StackAddressEscape',
-                                   '-d', 'unix.Malloc'
+                                   '-d', 'unix.Malloc',
+                                   '-d', 'clang-diagnostic',
+                                   '-e', 'clang-diagnostic-division-by-zero'
```
Szelethus, line 98: Why is this needed? I am more interested in the _design_ reason, not the technical one. Suppose that you are new to this test file, and stumble upon this invocation -- what does it convey to you? We manually disable all warnings for some reason, but enable a specific one after that?

Was this an attempt to reduce the amount of changes needed in the tests, otherwise everything broke?

web/tests/functional/report_viewer_api/__init__.py
```diff
@@ -114,7 +123,11 @@ def setup_class_common(workspace_name):
     print("Third analysis of the test project was successful.")
 
     # Let's run the second analysis and updat the same run.
-    codechecker_cfg['checkers'] = ['-d', 'core.StackAddressEscape']
+    codechecker_cfg['checkers'] = ['-d', 'core.StackAddressEscape',
+                                   '-d', 'unix.Malloc',
```
Szelethus, line 127: Whats the reason behind this? Why did you disable MallocChecker?

web/tests/functional/report_viewer_api/test_report_filter.py
```diff
@@ -382,7 +382,7 @@ def test_detection_date_filters(self):
     def test_fix_date_filters(self):
         """ Filter by fix dates. """
         report_filter = ReportFilter(
-            detectionStatus=[DetectionStatus.RESOLVED])
+            detectionStatus=[DetectionStatus.OFF])
```
Szelethus, line 385: Can you provide more context for this change?

Szelethus (changes requested):
Ugh, sorry, after taking another look at the tests, I'm still a little confused.

web/tests/functional/diff_local/test_diff_local.py
bruntib: Some `clang-diagnostic-...` checkers have been added to the "default" profile (see the changes of `clang-tidy.json`), that's why some of them appeared in tests files. So either I rewrite all test files to eliminate the source of the report, or I disable `clang-diagnostic-...` checkers in order to remain conform with the previous behavior.

web/tests/functional/report_viewer_api/__init__.py
bruntib: Same answer as above. As for the design reason: in this test we have multiple analyses with different configurations. Here we want to test which checkers' reports have appeared and disappeared. We control this by explicitly enabling and disabling them in different analysis actions.

web/tests/functional/report_viewer_api/__init__.py
bruntib: Earlier the meaning of `--disable` flag was to actually disable the checker in the `clang` invocation. Now the meaning of `--disable` flag is to hide the results of the given checker. So after this patch `core.StackAddressEscape` is actually running, even if disabled, but its reports are hidden by post-processing. This caused `unix.Malloc` checker to report its own diagnostics.

web/tests/functional/report_viewer_api/test_report_filter.py
bruntib: Similar issue than the previous one. It comes from the "disable vs. hide" semantic change. The "OFF" status means that the `unix.Malloc` checker has been disabled and that's the reason for the report being resolved. Earlier there was a `unix.MismatchedDeallocator` checker report which simply became resolved without the checker being disabled. I agree if it seems a little fuzzy. If you need more precise elaboration about the exact reasons, then I could take some more effort in its inspection once... maybe sometime... in the future... :)

web/tests/functional/diff_local/test_diff_local.py
Szelethus: Aha, I see. In that case, can we not move the profile changed to its own PR?

Szelethus (commented):
I still strongly disagree with not having a curtesy warning for `larger-than-frame`. Its a piece of information we immediately possess and are choosing to withhold for no reason, or at least it sure seems to have no good reason in my view.

I will accept that this is not the time or place to divide this PR into smaller chunks, but am rather unhappy about it. Its possible that we would have landed parts of this pach much sooner if it was more managable in scope (this last part alone consumed ~2 weeks). Having separate commits is nice (and helped me greatly!), but some parts of this PR would have definitely deserved its own PR as well.

I come from a different reviewer environment, so please let me know if I'm the one that needs to change to workflow, or at least lets try better to find a common ground. Also, I understand that 90-95% of this patch did indeed need to land in a single piece.

Granted that @dkrupp makes the call on the `larger-than-frame` discussion and my last few nits are addressed, this PR LGTM.

Szelethus (approved):
LGTM! While I've made my fair share of criticisms on some of the circumstances around this patch, overall, I'm satisfied with how it turned out. Great work!

bruntib:
Thank you so much for the valuable review!

PR #3957, opened 2023-07-13
[test] Clang may need Z3 in logger tests

bruntib:
The build logger tests whether a build action using Clang compiler can be captured. In some environments Clang may be built with using Z3. The location of Z3 is provided in LD_LIBRARY_PATH in this case. So the value of this environment variable needs to be preserved.


vodorok (approved):
LGTM

PR #3963, opened 2023-07-20
[gui] Update run filter when setting the filter options

### LINKED ISSUE #3962: Web GUI date filter is faulty
**Describe the bug**
On the CodeChecker Web GUI, on the Runs tab, the history stored before and after filter doesn't seem to work, the information gets stored into the url, but it doesn't effect the output. Tried refreshing the browser, and refreshing via the `Reload the runs` button, also tried first giving the date info, then the text filter, and vice versa, the results are the same, the date filters don't effect the output.
See: https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/runs?run=spotbugs%20on%20spotbugs&stored-before=2023-07-19%2000%3A00%3A00&stored-after=2023-07-16%2000%3A00%3A00

**CodeChecker version**
The CodeChecker version is 6.22.2.

**To Reproduce**

1. Open the web GUI.
2. Click `Runs`.
3. Select a date in `History stored after...` and/or `History stored before...`.
4. See that the output doesn't change.

**Expected behaviour**
Only see runs stored between the given two dates.

Tested it on both Linux firefox and Windows Chrome.

vodorok:
Previously only the run history was affected, when the analyzed before/after options were set. This resulted in a confusing view when the displayed run entries last analyzed date would show outside of the specified date.
This patch fixes this issue as it will hide those runs where the before/after criteria is not met.

fixes #3962

### DIFF
--- a/web/server/vue-cli/src/components/Run/RunFilterToolbar.vue
+++ b/web/server/vue-cli/src/components/Run/RunFilterToolbar.vue
@@ -220,6 +220,7 @@ export default {
         ? this.dateTimeToStr(this.storedAfter) : undefined;
       this.updateUrl({ "stored-after": date });
 
+      this.$emit("on-run-filter-changed");
       this.$emit("on-run-history-filter-changed");
     }, 500));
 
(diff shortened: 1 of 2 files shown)


vodorok:
I plan to extend this pr with the proper GUI tests.

bruntib (approved):

PR #3964, opened 2023-07-21
[PyYAML] ~ bump to version 6.0.1 to fix cython 3 issue

ZiperRom1:
See [cython3 messup issue](https://github.com/yaml/pyyaml/issues/724)  in PyYAML

### DIFF
--- a/analyzer/requirements.txt
+++ b/analyzer/requirements.txt
@@ -1,5 +1,5 @@
 lxml==4.9.2
 portalocker==2.2.1
 psutil==5.8.0
-PyYAML==6.0
+PyYAML==6.0.1
 mypy_extensions==0.4.3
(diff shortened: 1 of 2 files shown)


bruntib (changes requested):
Thanks for the upgrade+
We have some further files where this dependency is applied. Could you check those too? Thank you!
```
./analyzer/requirements_py/dev/requirements.txt
./analyzer/requirements_py/osx/requirements.txt
./codechecker_common/requirements_py/dev/requirements.txt
```

bruntib:
Sorry, I've just noticed that there is another PR about the same issue: https://github.com/Ericsson/codechecker/pull/3961.
I would merge that one, because it was first in time, I hope, you don't mind.
Thank you again for the patch!

PR #3970, opened 2023-07-27
[bugfix] Don't trigger analyzer without enabled checkers

bruntib:
Sometimes users want to execute exactly one checker. This can be done by disabling all checkers and then enabling that one. Some analyzers fail if no checker is enabled. For example, if the user wants to enable one ClangSA checker and disable everything else explicitly, then ClangTidy fails.

After this patch an analyzer will not be triggered if all of its checkers are disabled.

 [cmd] Add --disable-all flag

Add --disable-all flag which disables all checkers of all analyzers. It
is equilent to using "--disable default".

Should be merged after #3949.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzer.py
@@ -28,7 +28,7 @@ from codechecker_statistics_collector.collectors.return_value import \
 from . import analyzer_context, analysis_manager, pre_analysis_manager, \
     checkers
 from .analyzers import analyzer_types
-from .analyzers.config_handler import CheckerState
+from .analyzers.config_handler import AnalyzerConfigHandler, CheckerState
 from .analyzers.clangsa.analyzer import ClangSA
 
 from .makefile import MakeFileCreator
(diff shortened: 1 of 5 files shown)


analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -160,9 +169,16 @@ def perform_analysis(args, skip_handlers, actions, metadata_tool,
                       "the Clang Static Analyzer.")
             return
 
-    actions = prepare_actions(actions, analyzers)
     config_map = analyzer_types.build_config_handlers(args, analyzers)
 
+    no_checker_analyzers = \
+        [a for a in analyzers if not __has_enabled_checker(config_map[a])]
+    for analyzer in no_checker_analyzers:
+        LOG.info("No checkers enabled for %s", analyzer)
```
Szelethus, line 177: Can we tell if the user explicitly enabled an analyzer? If so, we should hard error, shouldn't we?

bruntib: I can be convinced, but I would feel a hard error a little strong here.

Szelethus: I'll give you that one.

Szelethus (approved):
LGTM

PR #3971, opened 2023-07-27
[config] Add some checkers to "default" profile

bruntib:
These added checkers/warnings report violations of SEI-CERT rules.


Szelethus (approved):
LGTM

PR #3972, opened 2023-08-04
[bugfix] Allow the disabling of statisticsbased checkers

bruntib:
When --stats flag is given to "CodeChecker analyze" these two statistics-based checkers are enabled by default:

- statisticsbased.SpecialReturnValue
- statisticsbased.UncheckedReturnValue

These can't be disabled even with `-d/--disable` if they were enabled by `--stats`.

This patch fixes this bug: now users can disable either statistics-based checker, even if --stats is used.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
@@ -19,6 +19,10 @@ from typing import List
 from codechecker_common.logger import get_logger
 
 from codechecker_analyzer import analyzer_context, env
+from codechecker_statistics_collector.collectors.special_return_value import \
+    SpecialReturnValueCollector
+from codechecker_statistics_collector.collectors.return_value import \
+    ReturnValueCollector
 
 from .. import analyzer_base
 from ..config_handler import CheckerState
@@ -92,6 +96,19 @@ def parse_clang_help_page(
         res.append((flag, ' '.join(desc)))
 
     return res
+
+
+def _is_user_disabled_checker(checker, ordered_checkers):
+    """
+    This function returns True if the given checker is disabled by the user
+    explicitly by a --disable flag.
+    """
+    if not ordered_checkers:
+        return False
+
+    disabled_checkers = (c for c, enabled in ordered_checkers if not enabled)
+
+    return any(checker.startswith(c) for c in disabled_checkers)
 
 
 class ClangSA(analyzer_base.SourceAnalyzer):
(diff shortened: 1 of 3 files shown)


analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -94,6 +98,19 @@ def parse_clang_help_page(
     return res
 
 
+def _is_user_disabled_checker(checker, ordered_checkers):
+    """
+    This function returns True if the given statistics-based checker is
```
Szelethus, line 103: Seems like this is not a statisticsbased specific function. If it is important for those checkers specifically, document it why. I'd expect to see something like

"Since these checkers can be enabled through -e/--enable AND --stats, we need special handling for them to make sure -d/--disable overwrite --stats"

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -596,6 +613,39 @@ def construct_config_handler(cls, args):
             cmdline_checkers,
             'enable_all' in args and args.enable_all)
 
+        # If --stats flag is provided then enable statistics-based checkers
+        # unless explicitly disabled by the user.
+        if 'stats_enabled' in args:
+            # In some Clang versions these checkers may be in some checker
+            # group (e.g. alpha...), so let's find the actual checker name
+            # first.
```
Szelethus, line 621: I see what you wanted to say, but its a little sloppy. How about this:

"In some Clang versions, statisticsbased checkers are in the alpha package (alpha.statisticsbased), and maybe not in others. Lets figure out the actual checker name first"

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -596,6 +613,39 @@ def construct_config_handler(cls, args):
...
+                    cmdline_checkers):
+                handler.set_checker_enabled(special_return_value)
+
+            if unchecked_return_value and not _is_user_disabled_checker(
+                    unchecked_return_value,
+                    cmdline_checkers):
+                handler.set_checker_enabled(unchecked_return_value)
+
+        # Statistics collector checkers must be explicitly disabled
+        # as they trash the output.
```
Szelethus, line 642: You should also state that their job was done anyway in a preceding analysis phase. Otherwise I enjoy the sh!ttalk :)

Szelethus (commented):
Seems like you don't actually test whether you can disable those checkers if --stats is set. On another note, shouldn't we want if --stats is enabled but both statistics checkers are explicitly disabled?

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -596,6 +613,40 @@ def construct_config_handler(cls, args):
...
+                    cmdline_checkers):
+                handler.set_checker_enabled(special_return_value)
+
+            if unchecked_return_value and not _is_user_disabled_checker(
+                    unchecked_return_value,
+                    cmdline_checkers):
+                handler.set_checker_enabled(unchecked_return_value)
+
+        # Statistics collector checkers must be explicitly disabled as they
+        # trash the output. Thise checkers are executed in a preceding analysis
```
Szelethus, line 642:
```suggestion
        # trash the output. These checkers are executed in a preceding analysis
```

Szelethus: Also, almost there, how about you end the sentence with "so they are not needed here."

PR #3974, opened 2023-08-09
[GUI] Jump directly to documentation url without error modal

cservakt:
Currently on the report detail page there is a red error bubble where we can click on the checker documentation button to have more information about the error. The button click triggers a modal that contains already known information and the real checker documentation URL. 
The modal is unnecessary and it has two steps to navigate to the effective documentation. To reduce the step numbers, there are some modifications in the Report and the ReportStepMessage components. 
When the report page is loaded the first thing is to create an errorChecker, then getting its documentation and passing it to ReportStepMessage via props. The ReportStepMessage component present the checker documentation button if the URL exists. The button refers to the checker error documentation without displaying a modal. When the documentation is not available, the ReportStepMessage shows the "No documentation for checker." message in the bubble.

### DIFF
--- a/web/server/vue-cli/src/components/Report/Report.vue
+++ b/web/server/vue-cli/src/components/Report/Report.vue
@@ -1,10 +1,5 @@
 <template>
   <v-container fluid class="pa-0">
-    <checker-documentation-dialog
-      :value.sync="checkerDocumentationDialog"
-      :checker="selectedChecker"
-    />
-
     <analysis-info-dialog
       :value.sync="analysisInfoDialog"
       :report-id="reportId"
@@ -313,8 +308,7 @@ import ReportTreeKind from "@/components/Report/ReportTree/ReportTreeKind";
 import { SetCleanupPlanBtn } from "@/components/Report/CleanupPlan";
 
 import AnalysisInfoBtn from "./AnalysisInfoBtn";
-import CheckerDocumentationDialog from
-  "@/components/CheckerDocumentationDialog";
+
 import { ReportComments } from "./Comment";
 import GitBlameMixin from "./Git/GitBlame";
 import ToggleBlameViewBtn from "./Git/ToggleBlameViewBtn";
@@ -330,7 +324,6 @@ export default {
   components: {
     AnalysisInfoBtn,
     AnalysisInfoDialog,
-    CheckerDocumentationDialog,
     CopyBtn,
     ReportComments,
     ReportInfoButton,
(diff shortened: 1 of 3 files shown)


Szelethus (commented):
Nice, welcome to CodeChecker! :)

I recently left a comment in another PR that summarized what I usually like to see in the summary of the patch:
https://github.com/Ericsson/codechecker/pull/3949#issuecomment-1640004462
Yours is currently empty, so please fill it out.

web/server/vue-cli/src/components/Report/Report.vue
```diff
@@ -313,8 +308,7 @@ import ReportTreeKind from "@/components/Report/ReportTree/ReportTreeKind";
 import { SetCleanupPlanBtn } from "@/components/Report/CleanupPlan";
 
 import AnalysisInfoBtn from "./AnalysisInfoBtn";
-import CheckerDocumentationDialog from
-  "@/components/CheckerDocumentationDialog";
```
Szelethus, line 317: Does this mean that we can just outright delete that file? Seems like only `Report.vue` used it.

Szelethus (changes requested):
You seem to have forgotten to add a newline in between the title of your commit and the message of the commit, which results in the _entire_ summary to be in fact your title. This is the output of `git log --oneline`:
image

edit: I tried this PR out, and it works fine for me, but I lack the expertise on this part of the codebase to comment anything substantial :)

web/server/vue-cli/src/components/Report/Report.vue
cservakt: No, we cannot delete it. The Reports.vue also uses the CheckerDocumentationDialog.

web/server/vue-cli/src/components/Report/ReportStepMessage.vue
```diff
@@ -73,7 +79,8 @@ export default {
     index: { type: [ Number, String ], default: null },
     bus: { type: Object, default: null },
     prevStep: { type: Object, default: null },
-    nextStep: { type: Object, default: null }
+    nextStep: { type: Object, default: null },
+    docUrl: { type: String, default: "" }
```
bruntib, line 83: This is a minor comment, but "not having some kind of information" could be represented by `null` like in case of the other members. `docUrl` could be changed accordingly in this `ReportStepMessage.vue` and `Report.vue` file tool.

bruntib (changes requested):

bruntib (approved):

Szelethus (approved):
LGTM!

PR #3981, opened 2023-08-15
[GUI] Change the selected bubble's brightness

cservakt:
To highlight selected bubble on the report page currently we use only a 2px dashed border. It is hard to detect the active chip section when there are many bubbles in one report file. The active bubble is more highlighted when its opacity is 100 and 70% for the others. The font-weight is bold for active and lighter for non actives. The current modifications emphasize the selected bubble more then the dashed border.

### DIFF
--- a/web/server/vue-cli/src/components/Report/Report.vue
+++ b/web/server/vue-cli/src/components/Report/Report.vue
@@ -954,5 +954,12 @@ export default {
 
 ::v-deep .report-step-msg.current {
   border: 2px dashed var(--v-primary-base) !important;
+  opacity: 1;
+  font-weight: bold;
+}
+
+::v-deep .report-step-msg {
+  opacity: 0.7;
+  font-weight: lighter;
 }
 </style>


Szelethus:
For visual changes like this, please include some pictures!

Here is what the patch does when you select a specific event on the rigth hand side of the report page.
# Control flow events
image -> image
# Warnings
image -> image
# Macros
image -> image
# Notes
image -> image

This change is a big improvement from the existing implementation, but I noticed that the selected elements became somewhat darker, instead of brighter -- wouldn't that pop even more? I'm not a designer, so I actually don't know :)

cservakt:
> For visual changes like this, please include some pictures!
> 
> Here is what the patch does when you select a specific event on the rigth hand side of the report page.
> # Control flow events
> 
> image -> image
> # Warnings
> 
> image -> image
> # Macros
> 
> image -> image
> # Notes
> 
> image -> image
> 
> This change is a big improvement from the existing implementation, but I noticed that the selected elements became somewhat darker, instead of brighter -- wouldn't that pop even more? I'm not a designer, so I actually don't know :)

That's true. Maybe setting opacity is more usefull to highlight the bubbles. 
test2

bruntib (approved):

PR #3997, opened 2023-08-28
Configurability of analyzer versions

cservakt:
The modification can be used to validate the version of each analyzer. The user can specify all desired analyzers after the --analyzers flag, as was possible before. Also, the exact version number can be typed for each analyzer, separated by an '=' char. For example: 'CodeChecker analyze compile_commands.json -o reports --analyzers clangsa=14.0.0 cppcheck=2.7'. If the version number was not the same as in the current environment, the analyzer would be disabled. The user can enumerate the analyzers with or without versions, for example: 'CodeChecker analyze compile_commands.json -o reports --analyzers clangsa=14.0.0 cppcheck'.


analyzer/codechecker_analyzer/analyzers/analyzer_types.py
```diff
@@ -148,8 +149,11 @@ def check_supported_analyzers(analyzers):
 
     enabled_analyzers = set()
     failed_analyzers = set()
-
     for analyzer_name in analyzers:
+        analyzer_name, analyzer_version = analyzer_name.split('=', 1) \
```
Szelethus, line 153:
```suggestion
        analyzer_name, requested_version = analyzer_name.split('=', 1) \
```

Does this sound any better? Not sure, just an idea.

Szelethus (commented):
I'm a big fan of hard errors in favour of warnings. CodeChecker has a lot of output, and these warnings are hard to miss. Also, if I explicitly request an analyzer on a specific version, and the specified version is not found, I might prefer to fix that before sinking countless hours into an analysis.

This idea might clash with the philosophy we have with clang-tidy not having a single checker enabled (in which case we actually prefer a warning to a hard error).

Also, there migth be several versions of the binary available from config files or PATH, but we we don't try to look for them, is that correct?

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -347,7 +347,6 @@ def add_arguments_to_parser(parser):
                                dest='analyzers',
                                metavar='ANALYZER',
                                required=False,
-                               choices=analyzer_types.supported_analyzers,
                                default=argparse.SUPPRESS,
                                help="Run analysis only with the analyzers "
```
bruntib, line 351: The help message could be extended by some info about the version number. Don't forget to apply the changes in `docs/analyzer/user_guide.md` too.

analyzer/codechecker_analyzer/analyzers/analyzer_types.py
bruntib, line 153: We should consider using the same format as other tools are supporting when providing version info. For example, in requirements.txt you can write this: `my_dependency==1.2.3`. In the future `>` could be used too, but for now equality check is enough.

bruntib (changes requested):
Nice improvements, thanks! Also, some test code could check this new feature.

analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
```diff
@@ -152,9 +152,10 @@ def __add_plugin_load_flags(cls, analyzer_cmd: List[str]):
             analyzer_cmd.extend(["-load", plugin])
 
     @classmethod
-    def get_version(cls, env=None):
+    def get_version(cls, analyzer_binary=None, env=None):
```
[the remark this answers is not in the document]
cservakt: I think, it is better to have analyzer_binary parameter, because there is a version_compatible class method in the cppchecker analyzer that also calls the version_info method, and in this case the analyzer_binary parameter must be specified as configured_binary. The version_compatible methods of the analyzers just return true and they do not validate their version, but it is better to have the same syntax for all version_info methods.

Szelethus:
I don't intend to die on this hill, but I'm not yet convinced that a soft warning is the way to go. Breaking existing functionality is never a step to be taken lightly, but this is a new feature, we actually _can_ be quite strict.

I foresee users setting an explicit version of the analyzer trying to achieve a reproducable analysis across different machines. I don't really see the scenario where the desired behaviour would be to just (almost) silently disable it and go on.

Also, you haven't commented on the other idea, checking whether the analyzer with the requested version _is_ available from the path. I suspect we only get hold of a single binary and look no further. This could be especially cool for people that need to have multiple compiler versions on hand (which is rather common among C++ developers).

analyzer/codechecker_analyzer/analyzers/analyzer_types.py
cservakt: Thanks for the suggestion! I fixed the variable name.

cservakt:
> I don't intend to die on this hill, but I'm not yet convinced that a soft warning is the way to go. Breaking existing functionality is never a step to be taken lightly, but this is a new feature, we actually _can_ be quite strict.
> 
> I foresee users setting an explicit version of the analyzer trying to achieve a reproducable analysis across different machines. I don't really see the scenario where the desired behaviour would be to just (almost) silently disable it and go on.
> 
> Also, you haven't commented on the other idea, checking whether the analyzer with the requested version _is_ available from the path. I suspect we only get hold of a single binary and look no further. This could be especially cool for people that need to have multiple compiler versions on hand (which is rather common among C++ developers).


Now, we give error message if an analyzer version is incorrect. 

In the future, it would be great not only to validate the version number, but also to allow the user to select it.

analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -140,8 +139,7 @@ def perform_analysis(args, skip_handlers, actions, metadata_tool,
     analyzers = args.analyzers if 'analyzers' in args \
         else analyzer_types.supported_analyzers
     analyzers, errored = analyzer_types.check_supported_analyzers(analyzers)
-    analyzer_types.check_available_analyzers(analyzers, errored)
-
+    analyzer_types.check_available_analyzers(analyzers, errored, args)
```
Szelethus, line 142: I had a rather hard project a while ago, trying to factor `args` out of a function mostly responsible for `CodeChecker cmd diff` (#3863). The problem with `args` is that, in my mind, its supposed to be a very temporary object -- the 'main' file or function receives this object as a representation of the user's program arguments, parses it, and discards it.

How would you unit test a function that requires monstrositty of an object? You'd spend 30 minutes trying to assemble all the small little parts of args until the test finally runs. If you decide to thread `args` out of a function, and all the functions that it calls that also take `args`, you'll spend about a week doing it.

I am perfectly aware that I'm speaking of greater evils than what is seen here (pretty much a slippery slope argument), but these, imo, decent reasons against it :^)

analyzer/codechecker_analyzer/analyzers/analyzer_base.py
```diff
@@ -46,6 +46,24 @@ def config_handler(self):
...
+        to return the bin of analyzer.
+        """
+        pass
+
+    @classmethod
+    @abstractmethod
+    def analyzer_env(cls):
+        """
+        A subclass should have a analyzer_env method
+        to return the env for analyzer.
```
Szelethus, line 63: This is about about as descriptive, as saying this:
* return the context of the analyzer
* return the parameters of the analyzer
* return the stuff about the analyzer
* return the details of the analyzer
* return the analyzer data

You get my point :^) Do you mean to return the _environmental variables_? Why is this important? Do analyzers run with their own environmental variables?

analyzer/codechecker_analyzer/analyzers/analyzer_base.py
```diff
@@ -46,6 +46,24 @@ def config_handler(self):
     def construct_analyzer_cmd(self, result_handler):
         raise NotImplementedError("Subclasses should implement this!")
 
+    @classmethod
+    @abstractmethod
+    def analyzer_binary(cls):
+        """
+        A subclass should have a analyzer_binary method
+        to return the bin of analyzer.
```
Szelethus, line 54: Can this ever be `None`?

analyzer/codechecker_analyzer/analyzers/analyzer_base.py
```diff
@@ -63,6 +81,15 @@ def version_compatible(cls, configured_binary, environ):
         """
         raise NotImplementedError("Subclasses should implement this!")
 
+    @classmethod
+    @abstractmethod
+    def version_info(cls):
+        """
+        A subclass should have a version_info method
+        to return with the analyzer's version number.
```
Szelethus, line 89: What is the expected format? Like this: `0.3.5` or like this: `v1.1.`? Must it be valid as per [`StrictVersion`'s definition](https://epydoc.sourceforge.net/stdlib/distutils.version.StrictVersion-class.html)?

analyzer/codechecker_analyzer/analyzers/analyzer_types.py
```diff
@@ -184,7 +202,18 @@ def check_supported_analyzers(analyzers):
         # Check version compatibility of the analyzer binary.
         if analyzer_bin:
             analyzer = supported_analyzers[analyzer_name]
-            if not analyzer.version_compatible(analyzer_bin, check_env):
+            if requested_version:
+                bin_version = StrictVersion(str(analyzer.version_info()))
+                requested_version = StrictVersion(requested_version)
+                if requested_version != bin_version:
+                    LOG.error(
+                        f"Given version: {requested_version}, found version "
```
Szelethus, line 210:
```suggestion
                        f"Requested version: {requested_version}, found version "
```

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
```diff
@@ -331,29 +325,26 @@ def resolve_missing_binary(cls, configured_binary, env):
         return cppcheck
 
     @classmethod
-    def __get_analyzer_version(cls, analyzer_binary, env):
+    def version_info(cls):
         """
-        Return the analyzer version.
+        Run the Cppcheck version command
+        parse the output
+        and return the analyzer's version.
```
Szelethus, line 332:
```suggestion
        Run the Cppcheck version command parse the output
        and return the analyzer's version.
```

docs/analyzer/user_guide.md
```diff
@@ -254,7 +254,8 @@ analyzer arguments:
   --analyzers ANALYZER [ANALYZER ...]
                         Run analysis only with the analyzers specified.
                         Currently supported analyzers are: clangsa, clang-
-                        tidy.
+                        tidy, cppcheck. Version number can be also specified
```
Szelethus, line 257: Can we give an example? Also, did we ever change the actual `--help` output? I can't see it anywhere.

Szelethus (changes requested):

analyzer/codechecker_analyzer/analyzers/analyzer_types.py
```diff
@@ -184,7 +202,18 @@ def check_supported_analyzers(analyzers):
         # Check version compatibility of the analyzer binary.
         if analyzer_bin:
             analyzer = supported_analyzers[analyzer_name]
-            if not analyzer.version_compatible(analyzer_bin, check_env):
+            if requested_version:
+                bin_version = StrictVersion(str(analyzer.version_info()))
```
bruntib, line 206: In case of ClangSA the `analyzer.version_info()` returns a `ClangVersionInfo` object which is a custom object. This doesn't convert to a string that could be fed by `StrictVersion`. If `ClangSA.version_info()` would also return a `StrictVersion` then no conversion would be needed.

docs/analyzer/user_guide.md
```diff
@@ -254,7 +254,8 @@ analyzer arguments:
   --analyzers ANALYZER [ANALYZER ...]
                         Run analysis only with the analyzers specified.
                         Currently supported analyzers are: clangsa, clang-
-                        tidy.
+                        tidy, cppcheck. Version number can be also specified 
+                        for analyzers to verify them.
```
bruntib, line 258: Put this extra text to the argparse "help" attribute too. This documentation is supposed to be a copy-pase of the actual `CodeChecker analyze --help` page.

bruntib (changes requested):

analyzer/codechecker_analyzer/cmd/check.py
```diff
@@ -292,7 +292,6 @@ def add_arguments_to_parser(parser):
                                dest='analyzers',
                                metavar='ANALYZER',
                                required=False,
-                               choices=analyzer_types.supported_analyzers,
                                default=argparse.SUPPRESS,
                                help="Run analysis only with the analyzers "
```
dkrupp, line 296: I miss the version definition information from the help text.

Run analysis only with the analyzers specified. Currently supported analyzers are: clangsa, clang-tidy, cppcheck.
Optionally, you can specify the analyzer version as follows: clangsa==13.0.0, clang-tidy=12.1.0, cppcheck
The version string is prefix matched. If the version information is omitted, any analyzer version will be accepted, but all the lister analyzers must be available.

analyzer/codechecker_analyzer/analyzers/analyzer_types.py
```diff
@@ -111,22 +113,35 @@ def is_ignore_conflict_supported():
 def print_unsupported_analyzers(errored):
     """ Print error messages which occured during analyzer detection. """
     for analyzer_binary, reason in errored:
-        LOG.warning("Analyzer '%s' is enabled but CodeChecker is failed to "
-                    "execute analysis with it: '%s'. Please check your "
-                    "'PATH' environment variable and the "
-                    "'config/package_layout.json' file!",
-                    analyzer_binary, reason)
+        LOG.error("Analyzer '%s' is enabled but CodeChecker is failed to "
```
dkrupp, line 116: I don't like this error message because it does not tell me where it found the analyzer binary.

Consider this error message instead (we dont need the "Please check your PATH..." part) :
CodeChecker analyze . --analyzers clangsa==13.0.1 -o reports
[ERROR 2023-10-11 20:09] - clangsa analyzer with the required version "13.0.1" not found. clangsa analyzer with version "13.0" found instead at /app/vbuild/UBUNTU18-x86_64/clang/13.0.0/bin/clang
[ERROR 2023-10-11 20:09] - Failed to run command because one or more given analyzers **could not** be found on your machine

docs/analyzer/user_guide.md
```diff
@@ -1098,7 +1099,8 @@ analyzer arguments:
   --analyzers ANALYZER [ANALYZER ...]
                         Run analysis only with the analyzers specified.
                         Currently supported analyzers are: clangsa, clang-
-                        tidy.
+                        tidy, cppcheck. Version number can be also specified
```
dkrupp, line 1102: Please describe in this documentation the version matching sytnax.(prefix matching?) in this section of the document https://github.com/Ericsson/codechecker/blob/master/docs/analyzer/user_guide.md#configuring-clang-version

dkrupp (changes requested):
Please add testcases.


Also some use cases are not working:

1) specifying multiple required analyzers without version should work

CodeChecker analyze . --analyzers clang-tidy,clangsa -o reports

[ERROR 2023-10-11 20:15] - Analyzer 'clang-tidy,clangsa' is enabled but CodeChecker is failed to execute analysis with it: 'Analyzer unsupported by CodeChecker!'. Please check your 'PATH' environment variable, the 'config/package_layout.json' file and the --analyzers flag!
[ERROR 2023-10-11 20:15] - Failed to run command because one or more given analyzers cannot be found on your machine!

2) Specifying some analyzers with version should work, but not it throws an exception:

(CodeChecker venv-dev) ednikru@seliiuts02064[20:46][ednikru/test-projects/xerces-c]$ CodeChecker analyze . --analyzers clangsa==13.0.0, clang-tidy==13.0.0 -o reports
  File "/local/ednikru/codechecker/build/CodeChecker/lib/python3/codechecker_common/cli.py", line 209, in main
    sys.exit(args.func(args))
  File "/local/ednikru/codechecker/build/CodeChecker/lib/python3/codechecker_analyzer/cmd/analyze.py", line 1076, in main
    analyzer.perform_analysis(args, skip_handlers, actions, metadata_tool,
  File "/local/ednikru/codechecker/build/CodeChecker/lib/python3/codechecker_analyzer/analyzer.py", line 141, in perform_analysis
    analyzers, errored = analyzer_types.check_supported_analyzers(analyzers)
  File "/local/ednikru/codechecker/build/CodeChecker/lib/python3/codechecker_analyzer/analyzers/analyzer_types.py", line 207, in check_supported_analyzers
    requested_version = StrictVersion(requested_version)
  File "/app/vbuild/UBUNTU18-x86_64/python/3.9.17/lib/python3.9/distutils/version.py", line 40, in __init__
    self.parse(vstring)
  File "/app/vbuild/UBUNTU18-x86_64/python/3.9.17/lib/python3.9/distutils/version.py", line 137, in parse
    raise ValueError("invalid version number '%s'" % vstring)
ValueError: invalid version number '13.0.0,

See also the small suggestions inline.

PR #4000, opened 2023-09-05
Parse file speed up

cservakt:
The feature can be useful to speed up the parse process when you want to parse one or more specified files. The VSCode Plugin also runs the 'parse --file' command, so the correction can solve the slow file verification problem. It uses metadata.json for plist mapping if the result_source_files section is not empty and contains the pairing of source files and plist files. Before the new feature, the process read all existing plist files to find the specified source file's reports.

### DIFF
--- a/analyzer/codechecker_analyzer/cmd/parse.py
+++ b/analyzer/codechecker_analyzer/cmd/parse.py
@@ -15,6 +15,7 @@ import argparse
 import os
 import sys
 from typing import Dict, Optional, Set
+import fnmatch
 
 from codechecker_report_converter.util import dump_json_output
 from codechecker_report_converter.report import report_file, \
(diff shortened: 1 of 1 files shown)


analyzer/codechecker_analyzer/cmd/parse.py
```diff
@@ -392,7 +392,22 @@ def get_output_file_path(default_file_name: str) -> Optional[str]:
 
     for dir_path, file_paths in report_file.analyzer_result_files(args.input):
         metadata = get_metadata(dir_path)
-        for file_path in file_paths:
+
+        specified_file_paths = None
```
bruntib, line 396: I'm not sure if this variable is needed. `file_paths` could be assigned the new list if any.

analyzer/codechecker_analyzer/cmd/parse.py
```diff
@@ -392,7 +392,22 @@ def get_output_file_path(default_file_name: str) -> Optional[str]:
...
+        if metadata and 'files' in args:
+            # Mapping plists when files are specified to speed up parsing
+            specified_file_paths = [
+                key for key, val in
+                metadata['tools'][0]['result_source_files'].items()
+                if val in args.files
+            ] if 'tools' in metadata \
+                and len(metadata['tools']) > 0 \
+                and 'result_source_files' in metadata['tools'][0] \
+                else None
```
bruntib, line 406: If this `result_source_files` section is missing from `metadata.json` then we could fall back to the slow version by using `file_paths` again instead of giving no result due to this `None` value.

bruntib (changes requested):

bruntib (approved):

PR #4001, opened 2023-09-05
[bugfix] Make None assignments to be initializations, not overwrites

Szelethus:


bruntib (approved):
I don't really see the motivation behind this change, but looks good to me.

PR #4003, opened 2023-09-05
[enhancement] Add option to control clangsa header analysis and macro…

edwinzrodriguez:
… expansion

Added two options
    '--analyze-headers [on|off] default=on' to enable/disable clang SA analysis of headers
    '--expand-macros [on|off] default=on' to enable/disable macro expansion in generated plist

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
@@ -329,9 +329,11 @@ class ClangSA(analyzer_base.SourceAnalyzer):
                                      "-Xclang", plugin])
 
             analyzer_mode = 'plist-multi-file'
+            if config.analyze_headers:
+                analyzer_cmd.extend(['-Xclang',
+                                     '-analyzer-opt-analyze-headers'])
+
             analyzer_cmd.extend(['-Xclang',
-                                 '-analyzer-opt-analyze-headers',
-                                 '-Xclang',
                                  '-analyzer-output=' + analyzer_mode,
                                  '-o', analyzer_output_file])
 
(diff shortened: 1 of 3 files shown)


edwinzrodriguez:
I'm proposing this change to reduce analysis time for precommit approval.  The project I'd like to apply it to consists of approx 1600 source files and takes 5 1/2 hours to analyze with expand-macros=true and -analyzer-opt-analyze-headers enabled.  Not setting those options reduces analysis time to ~30 minutes, which is acceptable for a precommit check

Szelethus:
Hi!

Sorry for the late response, we definitely appreciate all the help we can get!

We already have an option to pass clang static analyzer flags through CodeChecker via the [`--saargs` option](https://github.com/Ericsson/codechecker/blob/master/docs/analyzer/checker_and_analyzer_configuration.md#clang-static-analyzer-special-configuration-options). Wouldn't that be a good solution for you?

The problem I foresee is that CodeChecker is (increasingly) less of wrapper around Clang, and more of a general static analysis driver tool, so creating analyzer specific flags may go contradictory to that. This doesn't mean we make exceptions (just look at all the CTU options that, as of now, specifically target ClangSA), I'm just not immediately sure we want to make one here.

Cheers,
Husi

bruntib:
Hi,
Thanks for the improvement. However, we wouldn't like to go in the direction of extending the interface of `CodeChecker analyze`. The analyzer configuration options shouldn't be exposed on this interface, because there will always be yet another analyzer option that somebody wants to introduce for `CodeChecker analyze`. I think these analyzer config options should be provided through `--analyzer-config` flag.

bruntib (changes requested):
Hi,
Thanks for the improvement. However, we wouldn't like to go in the direction of extending the interface of CodeChecker analyze. The analyzer configuration options shouldn't be exposed on this interface, because there will always be yet another analyzer option that somebody wants to introduce for CodeChecker analyze. I think these analyzer config options should be provided through `--analyzer-config` flag.

edwinzrodriguez:
How about just removing '-analyzer-opt-analyze-headers' and 'expand-macros=true' and require people who want it to add those options via --saargs

bruntib:
Hi,
Sorry for the extreme slow reply. So, I would say, that `-analyzer-opt-analyze-headers` should be turned on by default, because we wouldn't like to introduce CI-breaking changes for teams that integrated CodeChecker. I can be convinced about `expand-macros` if it causes massive performance loss, but we turned that on for the GUI functionality which displays these macros.
On the other hand I would still vote on using `--analyzer-config` flag when providing any analyzer config options.
Thank you!

PR #4006, opened 2023-09-07
[docs] Rewrite the docs for diff

Szelethus:
I intend to squash these commits, but this is what I'd consider to be pretty close to the final version of the new diff docs.

As I wrote more and more, it made sense to create a brand new file. The diff docs for specifically the GUI (https://github.com/Szelethus/codechecker/blob/diff_docs_rewrite/web/server/vue-cli/src/assets/userguide/userguide.md#compare-runs) seem to be a mess, but I'm not sure there is a need for rewriting, and if there is, I'm pretty sure it can wait a little :) 

Check out the final docs here: https://github.com/Szelethus/codechecker/blob/diff_docs_rewrite/docs/web/diff.md

Depends on #3995 and #3996.


docs/web/diff.md
```diff
@@ -0,0 +1,903 @@
...
+  ----=================----
+  ```
+
+### Using the web GUI
+
+Select "Test Run" to be in the "baseline" set by clicking on the left radio button on the right hand side. Select "Test Run Updated" to be in the "newline" set by clicking on the other one. Click on "DIFF" in the upper right corner.
+
+![image](https://github.com/Szelethus/codechecker/assets/23276031/75490051-86df-419d-998a-ec89942faca1)
+
+In the middle of the image, we can see the list of oustadning reports that appeared from the baseline to the newline. In the bottom left corner, by clicking on the cogwheel next to "Diff type", you can change diff type:
```
cservakt, line 507: The steps of the operation could be highlighted on the GUI images, e.g. with a square or an arrow.

docs/web/diff.md
```diff
@@ -0,0 +1,903 @@
...
+  ----=================----
+  ```
+
+## Review status rules and diffs
+
+Similarly to source code suppressions, you can set the review status of reports [on the GUI](../..//web/server/vue-cli/src/assets/userguide/userguide.md#review-status) as well, which will create a review status rule. Unlike source code suppressions, review status rules set the review status for _all reports_ matching that rule, _regardless_ whether the rule was created before or after the report was stored. Because these rules are stored on the server, **they play no part in local-local diffs as of yet**.
+
+:warning: Note: Review status rules act differently when runs are compared as opposed to when tags or dates are compared (see [below](#diff-on-tags-or-timestamps)).
+
+:warning: Note: If you are trying out with these examples, keep in mind that review status rules are _not_ deleted even if _all reports that it matches_ (by removing some or all runs in a given product) are. You will need to click on "Configuration" in the upper right corner, select "Review Status Rules", where these rules can be removed.
```
cservakt, line 617: Adding an image of this config might make it more understandable for the users.

docs/web/diff.md
```diff
@@ -0,0 +1,903 @@
...
+
+![image](https://github.com/Szelethus/codechecker/assets/23276031/c308038d-466f-453c-ad41-e69cb789ff4d)
+
+Click on the first report, emitted by `deadcode.DeadStores`.
+
+![image](https://github.com/Szelethus/codechecker/assets/23276031/1e6c47d6-d9f6-44c9-a770-08108ed53a04)
+
+Set its review status to false positive.
+
+![image](https://github.com/Szelethus/codechecker/assets/23276031/d642a5aa-17f3-4566-856f-a54bd37b20d8)
```
cservakt, line 658: I think when this picture was taken, the brightness adjustment patch remained in the cache, just like the other pictures where you can see the error bubble.

docs/web/diff.md
```diff
@@ -0,0 +1,903 @@
...
+----==== Checker Statistics ====----
+--------------------------------------------------
+Checker name        | Severity | Number of reports
+--------------------------------------------------
+core.DivideZero     | HIGH     |                 1
+deadcode.DeadStores | LOW      |                 1
+--------------------------------------------------
+----=================----
+```
+:warning: Note: For the sake of simplictiy, we only enabled the Clang Static Analyzer on these analyses using the `--analyzers=clangsa` flag. We chose omit this flag from our examples. Your actual results may also be different depending on the version of your clang binary.
```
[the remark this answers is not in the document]
Szelethus: The idea was that I'd like to keep these invocations as lean as possible, but still with a managable number of results. So I cheated, used a flag that I later omitted from the examples, which is something I'll likely get away with, but on the off chance somebody does actually try these out, I left this note.

I would strongly not to adjust all invocations to accomodate `--analyzers=clangsa` because it'd be a distractraction.

bruntib (approved):

web/server/vue-cli/src/assets/userguide/userguide.md
```diff
@@ -295,9 +295,14 @@ run history events.
 - filters will be saved in the URL query parameter list.
 
 ### Compare runs
+
 Calculates the difference between two analyses of the project, showing which
 reports have been fixed and which are newly introduced.
 
+:warning: Note: Here, you can read about _how_ to query the differences in
+between two results set. As to how this comparison works, refer to the docs
+on [`CodeChecker cmd diff`](../../../../../../docs/web/user_guide.md#show-differences-between-two-runs-diff).
```
bruntib, line 304: Maybe we should use absolute paths in documentations. That would be more resilient for potential moves of the current doc.

bruntib (approved):
A rebase could be useful to make sure that some previous commits really match in this patch.

web/server/vue-cli/src/assets/userguide/userguide.md
Szelethus: Do you know how to do that though? :)

docs/web/diff.md
```diff
@@ -0,0 +1,902 @@
...
+  ---------------------------------------------
+  Number of processed analyzer result files | 0
+  Number of analyzer reports                | 0
+  ---------------------------------------------
+  ----=================----
+  ```
+
+## Review status rules and diffs
+
+Similarly to source code suppressions, you can set the review status of reports [on the GUI](../..//web/server/vue-cli/src/assets/userguide/userguide.md#review-status) as well, which will create a review status rule. Unlike source code suppressions, review status rules set the review status for _all reports_ matching that rule, _regardless_ whether the rule was created before or after the report was stored. Because these rules are stored on the server, **they play no part in local-local diffs as of yet**.
```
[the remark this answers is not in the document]
Szelethus: I didn't want to dedicate yet another section to this, but I did add this link to the docs.

docs/web/diff.md
Szelethus: There are pictures in the already linked dics in the same section. You can argue whether the pictures are good enough there, but at least its something :^)

dkrupp (changes requested):
Please see my comments to the doc in https://github.com/Ericsson/codechecker/pull/4013

web/server/vue-cli/src/assets/userguide/userguide.md
bruntib:
```
[`CodeChecker cmd diff`](/docs/web/user_guide.md#show-differences-between-two-runs-diff)
```

bruntib: Now I realized, that this is a documentation which is displayed in the GUI. These paths may not be valid there anyway, so feel free to skip this comment of mine. I'm not even sure, that this "../../../" version works, because `user_guide.md` is not the part of the GUI. Maybe we could provide a link directly to this doc on GitHub.

Szelethus:
TODO: I linked my branch to the 6.23.0-rc1 release notes, that needs to point to the Ericsson/master branch after release.

docs/web/diff.md
```diff
@@ -1,6 +1,13 @@
 # Comparing analysis results (Diff)
 
-Compare the outstanding reports in a "before" (_baseline_) and an "after" (_new_ or newline) analysis sets, and display the differences in between them the in the same 
+## Introduction
+In CodeChecker you may compare analysis results either in the command line or in the web UI. With this tool, you can express _pull request analysis_ (did my pull request introduce/fix any bugs?), and _branch history analysis_ (did a change on my branch introduce/fix any bugs?). This document describes different use-cases for how locally stored analysis results, as well as remotely stored analysis results can be compared, and how other components of CodeChecker (source-code suppressions, review status rules, tags) affect the result of this set.
+
+From the outlook, diff sounds like a very simple feature (and it is!), but it also a powerful tool that allows CodeChecker, among other things, to be used as a gating tool. For this reason, you can express rather complicated querries with it using various filters -- this documents intends to help to understand the expected behaviour of diff in these cases as well.
```
bruntib, line 6:
```suggestion
From the outlook, diff sounds like a very simple feature (and it is!), but it also a powerful tool that allows CodeChecker, among other things, to be used as a gating tool. For this reason, you can express rather complicated queries with it using various filters -- this document intends to help to understand the expected behavior of diff in these cases as well.
```

docs/web/diff.md
```diff
@@ -1,6 +1,13 @@
...
+
+## How diffs work
+
+The term "diff" means the comparison of the outstanding reports in a "before" (_baseline_) and an "after" (_new_ or newline) analysis sets, and display the differences in between them the in the same
```
bruntib, line 10:
```suggestion
The term "diff" means the comparison of the outstanding reports in a "before" (_baseline_) and an "after" (_new_ or newline) analysis sets, and the display of differences between them in the same 
```

bruntib (changes requested):

dkrupp (approved):
LGTM

docs/web/diff.md
Szelethus: Behaviour is brittish english.

Szelethus:
I fixed all the remaining comments. Unless there is factually wrong information found here, I strongly insist on finally landing this patch. Its already miles better than the current docs, and we can always make it prettier later down the line.

bruntib (approved):

PR #4008, opened 2023-09-13
Added blame info feature for "cmd results" command

cservakt:
Blame information for reports was only available on the GUI. Now, if we wish to check the git commit info in the cli, we can do that with "CodeChecker cmd results --details" command We can only check blame info for runs that have a git repository. The server address and the run name should also be given, e.g.: "CodeChecker cmd results --details --url http://localhost:8001/Default test". The syntax is different in the GUI, as there you can check the blame info for each line of a reported file. In case of the cli, it would be huge output result to print long blame info for every report. A filtered one is added to the report data, which contains only the commit data of the given section.


web/api/report_server.thrift
```diff
@@ -304,27 +304,28 @@ struct ReviewStatusRule {
...
+  11: ReviewData      reviewData,         // Bug review status information.
   12: DetectionStatus detectionStatus, // State of the bug (see the enum constant values).
   13: string          detectedAt,      // Detection date of the report.
   14: string          fixedAt          // Date when the report was fixed.
-  15: i64             bugPathLength,   // Length of the bug path.
-  16: optional ReportDetails details,  // Details of the report.
-  17: optional string analyzerName,    // Analyzer name.
+  15: i64             bugPathLength,  // Length of the bug path.
+  16: optional ReportDetails details, // Details of the report.
+  17: optional BlameInfo blameInfo,   // Blmae info.
```
bruntib, line 323: New parameters should be added to the end of the list without changing previous numbers. As far as I know, these numbers play a role in API backward compatibility.

web/server/codechecker_server/api/report_server.py
```diff
@@ -1919,6 +1925,22 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
                         report.review_status_date,
                         report.review_status_is_in_source)
 
+                    blame_info = blame_infos[report.id] \
+                        if report.id in blame_infos else None
```
bruntib, line 1929:
```suggestion
                    blame_info = blame_infos.get(report.id)
```

web/server/codechecker_server/api/report_server.py
```diff
@@ -1901,9 +1901,15 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
...
                 # Get report details if it is required.
                 report_details = {}
-                if get_details:
-                    report_ids = [r[0].id for r in query_result]
+                blame_infos = {}
+                if get_details and len(query_result):
+                    report_ids, blames = zip(*[
+                        (
+                            r[0].id,
+                            (r[0].id, self.getBlameInfo(r[0].file_id))
```
bruntib, line 1909: We could consider caching fetched blame info in case there are many reports in a file.

bruntib (changes requested):

web/api/report_server.thrift
```diff
@@ -325,6 +325,7 @@ struct ReportData {
   // of custom labels that describe some properties of a report. For example the
   // timestamp in case of dynamic analyzers when the report was actually emitted.
   18: optional map<string, string> annotations,
+  19: optional BlameInfo blameInfo,    // Blmae info.
```
Szelethus, line 328: Its a typo, but more importantly, maybe this isn't the comment we need here :) How about
"Contains the git blame information of the report. May be NULL if the analysis was _not_ done in a git repository."

Szelethus (changes requested):
We definitely can't land this without tests.

Also, I saw the following blame info from a tinyxml2 analysis in json format:
```json
    "blameInfo": {
      "commits": {
        "a9cf3f9f3fe65df392caa5aecd3b77a260d7921f": {
          "author": {
            "name": "Lee Thomason",
            "email": "blabla123123123@sdkljfghsdjlkfgh.sdflkoghj"
          },
          "summary": "Switched to Artistic Style auto-formatting to allow integration of patches from other coding styles.",
          "message": "Switched to Artistic Style auto-formatting to allow integration of patches from other coding styles.\n",
          "committedDateTime": "2012-10-11 16:56:51-07:00"
        }
      },
      "blame": [
        {
          "startLine": 2646,
          "endLine": 2646,
          "commitHash": "a9cf3f9f3fe65df392caa5aecd3b77a260d7921f"
        }
      ]
    }
```
It may not be your doing, but are we sure we want the commit hash to be a key?

cservakt:
> We definitely can't land this without tests.
> 
> Also, I saw the following blame info from a tinyxml2 analysis in json format:
> 
> ```json
>     "blameInfo": {
>       "commits": {
>         "a9cf3f9f3fe65df392caa5aecd3b77a260d7921f": {
>           "author": {
>             "name": "Lee Thomason",
>             "email": "blabla123123123@sdkljfghsdjlkfgh.sdflkoghj"
>           },
>           "summary": "Switched to Artistic Style auto-formatting to allow integration of patches from other coding styles.",
>           "message": "Switched to Artistic Style auto-formatting to allow integration of patches from other coding styles.\n",
>           "committedDateTime": "2012-10-11 16:56:51-07:00"
>         }
>       },
>       "blame": [
>         {
>           "startLine": 2646,
>           "endLine": 2646,
>           "commitHash": "a9cf3f9f3fe65df392caa5aecd3b77a260d7921f"
>         }
>       ]
>     }
> ```
> 
> It may not be your doing, but are we sure we want the commit hash to be a key?

I think this is the best way to print the result in cli. Blame info belongs to a file, not to a report and each commit is identified by the commit hash. So we can only filter it to get a shorter form for the given report. If we want to get a different syntax when using the cli, we need to modify the blame info stuct in thirift, which is also used by the GUI.

web/api/report_server.thrift
cservakt: Thanks for your suggestion! I completely agree. It would be better if there was a more precise description.

web/api/report_server.thrift
```diff
@@ -325,6 +325,7 @@ struct ReportData {
   // of custom labels that describe some properties of a report. For example the
   // timestamp in case of dynamic analyzers when the report was actually emitted.
   18: optional map<string, string> annotations,
+  19: optional BlameInfo blameInfo,    // Contains the git blame information of the report if it exists.
```
Szelethus, line 328: This is longer than 79 columns.

Szelethus:
Shouldn't the value for key `commits` be a _list_ instead of a dict? How about this:

```json
"blameInfo": {
  "commits": [
    {
      "commit_hash": "a9cf3f9f3fe65df392caa5aecd3b77a260d7921f",
      "author": {
        "name": "Lee Thomason",
        "email": "blabla123123123@sdkljfghsdjlkfgh.sdflkoghj"
      },
      "summary": "Switched to Artistic Style auto-formatting to allow integration of patches from other coding styles.",
      "message": "Switched to Artistic Style auto-formatting to allow integration of patches from other coding styles.\n",
      "committedDateTime": "2012-10-11 16:56:51-07:00"
    },
    {
      "commit_hash": "a9cf3f9f3fe65df392caa5aecd3b77a260d7921f",
      "author": {
        "name": "Lee Thomason",
        "email": "blabla123123123@sdkljfghsdjlkfgh.sdflkoghj"
      },
      "summary": "Switched to Artistic Style auto-formatting to allow integration of patches from other coding styles.",
      "message": "Switched to Artistic Style auto-formatting to allow integration of patches from other coding styles.\n",
      "committedDateTime": "2012-10-11 16:56:51-07:00"
    }
  ],
  "blame": [
    {
      "startLine": 2646,
      "endLine": 2646,
      "commitHash": "a9cf3f9f3fe65df392caa5aecd3b77a260d7921f"
    }
  ]
}
```
Mind that the hash is no longer a key here either, but a value. This looks more idiomatic to me.

web/server/codechecker_server/api/report_server.py
```diff
@@ -1901,9 +1901,15 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
...
+                        ) for r in query_result])
                     report_details = get_report_details(session, report_ids)
+                    blame_infos = dict(blames)
```
Szelethus, line 1912: `getRunResults` is already stomach-churningly long. Can we put these lines a new function instead?

web/server/codechecker_server/api/report_server.py
```diff
@@ -1919,6 +1925,21 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
...
+                        commitHash = blame_data[0].commitHash \
+                            if len(blame_data) else None
+                        commitInfo = {cHash: commit for cHash, commit
+                                      in blame_info.commits.items()
+                                      if cHash == commitHash}
+                        blame_info = BlameInfo(
+                            commits=commitInfo,
+                            blame=blame_data
+                        )
+
```
Szelethus, line 1942: Same thing here.

web/server/codechecker_server/api/report_server.py
```diff
@@ -2004,6 +2032,22 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
...
```
Szelethus, line 2050: These code snippets are basically copy-paste from above, right? Can we do something about that?

web/tests/functional/report_viewer_api/test_get_run_results.py
```diff
@@ -447,3 +447,20 @@ def test_get_checker_labels(self):
...
+        """
+        runid = self._runid
+        simple_filter = ReportFilter()
+        run_results = self._cc_client.getRunResults([runid],
+                                                    100,
+                                                    0,
+                                                    None,
+                                                    simple_filter,
+                                                    None,
+                                                    True)
```
Szelethus, line 464: Can we name the parameters we're assign to here? [LLVM Coding Standards](https://llvm.org/docs/CodingStandards.html#comment-formatting) has a well written guidince on this. By no means are we strictly abiding all of these, but they make a lot of sense :)

web/tests/functional/report_viewer_api/test_get_run_results.py
```diff
@@ -447,3 +447,20 @@ def test_get_checker_labels(self):
...
+
+        self.assertTrue(any(res.blameInfo for res in run_results))
```
Szelethus, line 466: Can we test the json output as well? At least that is doesn't crash.

Szelethus (changes requested):

web/tests/functional/report_viewer_api/test_get_run_results.py
cservakt: We do not name the parameters in the above functions either. If we want to do it, then in all the other cases we must also assigne them.

web/tests/functional/report_viewer_api/test_get_run_results.py
cservakt: I think it is sufficient to check only the existence of the blame info. If someone modified the test files, the blame info whould be also changed, which whould break the test.

web/server/codechecker_server/api/report_server.py
cservakt: It is more readable form.

bruntib:
Done by https://github.com/Ericsson/codechecker/pull/4842

PR #4010, opened 2023-09-18
[QoL] Add config folder contents to the dev_package target

vodorok:
The `config` folder under `build/CodeChecker` wasn't symlinked before when using `dev_package` target of make, and resulted in  mystical errors occurring in case the developer forgot to manually update  the config contents.
This patch adds symlinking in short of one file which is created build time, (`commands.json`)


vodorok:
The `tree` command on the `package` target and `tree -l` (follow symlinks) on the modified `dev_package` target both gives the same output and in conclusion
`3 directories, 31 files`

bruntib (approved):

Issue #3968, opened 2023-07-25
`CodeChecker analyze` fails for 6.22.2 under Windows with "AttributeError: module 'os' has no attribute 'uname'"


petermbauer:
**Describe the bug**
Executing `CodeChecker analyze` leads to `AttributeError: module 'os' has no attribute 'uname'`.

```
 Traceback (most recent call last):
   File "D:\somedir\lib\site-packages\codechecker_common\cli.py", line 167, in main
     add_subcommand(subparsers, subcommand,
   File "D:\somedir\lib\site-packages\codechecker_common\cli.py", line 57, in add_subcommand
     command_module.add_arguments_to_parser(sc_parser)
   File "D:\somedir\Lib\site-packages\codechecker_analyzer/cmd\analyze.py", line 456, in add_arguments_to_parser
     clang_has_z3 = analyzer_types.is_z3_capable()
   File "D:\somedir\lib\site-packages\codechecker_analyzer\analyzers\analyzer_types.py", line 73, in is_z3_capable
     enabled_analyzers, _ = check_supported_analyzers([ClangSA.ANALYZER_NAME])
   File "D:\somedir\lib\site-packages\codechecker_analyzer\analyzers\analyzer_types.py", line 144, in check_supported_analyzers
     context = analyzer_context.get_context()
   File "D:\somedir\lib\site-packages\codechecker_analyzer\analyzer_context.py", line 314, in get_context
     return Context()
   File "D:\somedir\lib\site-packages\codechecker_common\singleton.py", line 20, in __call__
     super(Singleton, cls).__call__(*args, **kwargs)
   File "D:\somedir\lib\site-packages\codechecker_analyzer\analyzer_context.py", line 68, in __init__
     machine = os.uname().machine
 AttributeError: module 'os' has no attribute 'uname'
```

The problematic code has been introduced with 6.22.2: https://github.com/Ericsson/codechecker/compare/v6.22.1...v6.22.2#diff-b65c399d1bcab9aba1b50586c3ef10dfd319e80a985396c6739f83980a085a98R68

**CodeChecker version**
6.22.2

**To Reproduce**
see above

**Expected behaviour**
Run the analysis without an error

**Desktop (please complete the following information)**
 - OS: Windows
 - Version: 10
 - Python 3.10.0

**Additional context**
Add any other context about the problem.

vodorok:
[...]

PR #4012, opened 2023-09-20
replace os.uname() with platform.uname() for Windows support

petermbauer:
fix for #3968


whisperity (approved):
This patch works all the way back to Ubuntu 14.04 at least (I did not go further back :smirk:), so I believe this should be safe to merge.

bruntib (approved):

PR #4013, opened 2023-09-20
[cmd][diff] Ignore detection status for tags

Szelethus:
We already do this on the GUI, but on the cmdline, we default the
detection status to NEW, REOPENED and UNRESOLVED. This makes sense for
runs (where we check whether a report is outstanding at the time of the
query), but not for tags (as a report can have a RESOLVED detection
status but still be outstanding at the time of the tag).

### DIFF
--- a/web/client/codechecker_client/cmd_line_client.py
+++ b/web/client/codechecker_client/cmd_line_client.py
@@ -995,6 +995,27 @@ def get_diff_remote_runs(
         ttypes.SortType.FILENAME,
         ttypes.Order.ASC))]
 
+    # Generally speaking, we only compare outstanding reports in both the
+    # baseline and newline analyses, so it wouldn't make sense to show reports
+    # whose detection status indicates otherwise. However, for tags, we check
+    # whether a report is outstanding at the time of the tag, so in the context
+    # of that tag, a resolved report may actually be outstanding, so we clear
+    # the detection status filter.
+    # TODO: We don't support diffs with timestamps on the cmdline, but we
+    # should do this for them as well.
+    if base_run_tags or new_run_tags:
+        # If the value is not the default, lets warn the user that we overwrote
+        # it.
+        default_detection_status = [ttypes.DetectionStatus.NEW,
+                                    ttypes.DetectionStatus.REOPENED,
+                                    ttypes.DetectionStatus.UNRESOLVED]
+        if report_filter.detectionStatus != default_detection_status and \
+           report_filter.detectionStatus != []:
+            LOG.warning("--detection-status is ignored when comparing tags, "
+                        "showing reports regardless of detection status.")
+
+        report_filter.detectionStatus = []
+
     all_results = get_run_results(
         client, base_ids, constants.MAX_QUERY_SIZE, 0, sort_mode,
         report_filter, cmp_data, True)
--- a/web/tests/functional/diff_cmdline/test_diff_cmdline.py
+++ b/web/tests/functional/diff_cmdline/test_diff_cmdline.py
@@ -17,11 +17,11 @@ import shutil
 import unittest
 
 from codechecker_api.codeCheckerDBAccess_v6.ttypes import \
-        ReviewStatus, DiffType, ReportFilter
+        ReviewStatus, DiffType, ReportFilter, DetectionStatus
 
 from codechecker_client.cmd_line_client import \
     get_diff_local_dirs, get_diff_remote_run_local_dir, \
-    get_diff_local_dir_remote_run, get_diff_remote_runs
+    get_diff_local_dir_remote_run, get_diff_remote_runs, init_logger
 
 from libtest import codechecker, env
 from libtest.thrift_client_to_db import get_all_run_results
@@ -63,6 +63,8 @@ class TestDiffFromCmdLine(unittest.TestCase):
         test_config['codechecker_cfg'] = codechecker_cfg
 
         env.export_test_cfg(TEST_WORKSPACE, test_config)
+
+        init_logger(None, None)
 
     def teardown_class(self):
         """Clean up after the test."""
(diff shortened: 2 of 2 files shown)


bruntib (approved):
Please, rebase. Otherwise looks good.

docs/web/diff.md
```diff
@@ -0,0 +1,902 @@
+# Comparing analysis results (Diff)
+
+Compare the outstanding reports in a "before" (_baseline_) and an "after" (_new_ or newline) analysis sets, and display the differences in between them the in the same
```
dkrupp, line 3: Please start this section with a more descriptive introduction of the report states:

I would rewrite the starting stentence like this:

In CodeChecker a report can be either "Outstanding", or "Closed". Outstanding reports are the reports which should be fixed: are still detected, are unreviewed or confirmed to be true positive.
We call a reports "Closed", if it cannot be detected anymore (e.g. because it is fixed), or it is considered a false positive or an uninteresting (intentional) finding.

CodeChecker diff mechanism always compares two outstanding set of reports, a "before" (_baseline_) and an "after" (_new_ or newline) analysis sets, and display the differences in between them.

Formally:

A report is outsstanding if all ....

docs/web/diff.md
```diff
@@ -0,0 +1,902 @@
...
+  ```
+  ----======== Summary ========----
+  ---------------------------------------------
+  Number of processed analyzer result files | 0
+  Number of analyzer reports                | 0
+  ---------------------------------------------
+  ----=================----
+  ```
+
+## Review status rules and diffs
```
dkrupp, line 611: Refer to the use cases introduced in the introduction section.

E.g.

If you whish to store the suppressions on the server, you can use review status rules. With the review status rule you can set all instances of report type (reports with the same hash) e.g. false positive. This feature makes it easy to mark a report false positive in all runs at the same time.

docs/web/diff.md
```diff
@@ -0,0 +1,902 @@
...
+  ---------------------------------------------
+  Number of processed analyzer result files | 0
+  Number of analyzer reports                | 0
+  ---------------------------------------------
+  ----=================----
+  ```
+
+## Diff on tags or timestamps
+
+You can filter the analysis results or diff results based on which reports were outstanding _before_ a given timestamp. In the below image, on the web GUI, we diff the reports outstanding before the 1st of January, 2023 in "Test Run" against reports  outstanding before the 1st of January, 2023 in "Test Run Updated":
```
dkrupp, line 751: Refer to the introduction section "server based branch history analysis".

If a run is continously updated with new analysis results, it can store the past report states of all previous analysis executions. This can be useful to analyze how the reports were fixed, classified to false positives, or introduced over the development timeline.

docs/web/diff.md
```diff
@@ -0,0 +1,902 @@
...
+
+:warning: Note: Tags are created when the report is stored. As a result, you can diff a local analysis directory against a remote tag, but there are no "local tags".
+
+:no_entry: Note: You can compare tags through the command line and the web GUI as well. Currently, you can only compare timestamps on the web GUI.
+
+:no_entry: Note: Comparing tags older than the second-to-last is not supported, and may yield incorrect results in specific circumstances. Specifically, if from tag-to-tag a report is resolved, then becomes outstanding again, our results table irrecoverably drops the information that the report was one point resolved. This will result in CodeChecker calculating the set of outstanding reports incorrectly. You can read more about this in [issue](https://github.com/Ericsson/codechecker/issues/3884) in section "Tag diffs -> Local-Remote -> the footnote with the double asterisk (**)".
```
dkrupp, line 769: I would rephrase this as follows:

Comparing past tags, may yield incorrect results in specific circumstances.
CodeChecker does not store a full state change history of a report, it applies a simplification for reopened reports.

It only stores the date when a report was first detected in a run and the date when it was closed. If the report is closed and opened several times, it will only remember the first detection date, and the last close date. So CodeChecker assumes the worst: it regards a report which is on/and/off as it was always there until it is finally fixed.

In practice this means, that when you compare the run state at two past tags (after a report detection, but before final fix), a reopened report will be considered as always outstanding after first detection and not appear as new report in the diff.
You can read more about this in ...

docs/web/diff.md
```diff
@@ -0,0 +1,902 @@
+# Comparing analysis results (Diff)
+
+Compare the outstanding reports in a "before" (_baseline_) and an "after" (_new_ or newline) analysis sets, and display the differences in between them the in the same 
+format as `results`. A report is outstanding if all of the following is true:
+* its detection status is _new_, _reopened_ or _unresolved_,
```
dkrupp, line 5: A report is outstanding in a given time T if all the following is true:
- if it was detected before T
- if it was closed after T, or not closed at all
- a report can be closed if becomes undetected, considered false positive, intentional bug.

The server stores the past states of a reports with certain simplifications (see section **), so it can tell for any past time points whether the report was outstanding or closed.

The local report directory only stores a single present state.

In CodeChecker the most recent status of a report is also represented in the detection status:
-_new_: the report is new compared to the last analysis
-_reopened_: the report is new compared to the last analysis and it is reappearing
-_unresolved_: the report is still detected
-_resolved_: the report is not detected anymore

This is useful to quickly see if for example the analysis set of the "master" branch is updated, what are the new/unresolved reports we still have to work with.

dkrupp (changes requested):
Please refer to this document from the usage guide.
Also please connect the sections to real PR analysis, branch analysis use-cases which should be introduced in the intro section. The user should understand how to use these concepts in practice.

dkrupp (approved):
With the condition of fixin my comments in diff.md in https://github.com/Ericsson/codechecker/pull/4006#pullrequestreview-1661327221

Szelethus:
I cherry-picked the docs changes into #4006, so it no longer has any dependencies.

docs/web/diff.md
Szelethus: I added the "outstanding report" definition when we diff tags/timestamps.

I added links to the detection status and review status docs.

For the rest, this is just not the place to describe what detection statuses are, but rather their own page. On the fact that we ignore detection statuses for tags/timestamps there already is paragraph.

PR #4014, opened 2023-09-21
[GUI] Unnecessary reports columns

cservakt:
The "timestamp" and "testcase" columns are only needed for a special test case, therefore, they can be removed if their content is completely empty. When the "formattedReports" prop is changed, the values are verified again. If a report had timestamp or testcase, the columns would be displayed again.

### DIFF
--- a/web/server/vue-cli/src/views/Reports.vue
+++ b/web/server/vue-cli/src/views/Reports.vue
@@ -286,6 +286,7 @@ export default {
       reports: [],
       sameReports: {},
       hasTimeStamp: true,
+      hasTestCase : true,
       selected: [],
       namespace: namespace,
       pagination: {
@@ -332,6 +333,10 @@ export default {
           return this.hasTimeStamp;
         }
 
+        if (header.value === "testcase") {
+          return this.hasTestCase;
+        }
+
         return true;
       });
     },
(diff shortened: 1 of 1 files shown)


web/server/vue-cli/src/views/Reports.vue
```diff
@@ -375,6 +380,21 @@ export default {
...
+          report => report.timestamp
+        ).filter(ts => ts).length
+          ? true
+          : false;
+
+        this.hasTestCase = this.formattedReports.map(
+          report => report.testcase
+        ).filter(tc => tc).length
+          ? true
+          : false;
```
bruntib, line 396:
```suggestion
        this.hasTimeStamp =
          this.formattedReports.some(report => report.timestamp);
        this.hasTestCase =
          this.formattedReports.some(report => report.testcase);
```
May I recommend a shorter implementation?

bruntib (changes requested):
This is a nice patch, thanks! By the way, it is interesting to see `formattedReports` in the `watch` section. I've never seen a computed field in this section. But it works, anyway :)

web/server/vue-cli/src/views/Reports.vue
cservakt: Thank you for the suggestion. Using some() method is more stylish solution.

bruntib (approved):

PR #4020, opened 2023-09-25
Revert "Add config folder contents to the dev_package target"

bruntib:
This reverts commit 1ae30cd8c222f43bec1af774779527677e7bbca2.

The reason of this revert is that the original "make package" executed
this scrip too:

${PYTHON_BIN} ./scripts/build/extend_version_file.py -r $(ROOT) \
	$(CC_BUILD_DIR)/config/analyzer_version.json \
	$(CC_BUILD_DIR)/config/web_version.json

This command modifies the version json files, without this the CodeChecker
commands raise an error during usage.. However, if this modification is
done through the symlinks, then the original json files are modified too,
and shown as changed files in "git status".


whisperity:
Surprise LaTeX!

GitHub silently rolls out ${\LaTeX}$ mode, and this is how we get to know of it?! 🤣

PR #4030, opened 2023-10-05
Gcc analyzer native support

Szelethus:
WIP. Depends on #4011.


analyzer/codechecker_analyzer/analyzers/gcc/analyzer.py
```diff
@@ -0,0 +1,274 @@
...
+        return analyzer_context.get_context() \
+            .analyzer_binaries[cls.ANALYZER_NAME]
+
+    @classmethod
+    def get_version(cls, env=None):
+        """ Get analyzer version information. """
+        return cls.__get_analyzer_version(cls.analyzer_binary(), env)
+
+    def add_checker_config(self, checker_cfg):
+        LOG.error("Checker configuration for Gcc is not implemented yet")
```
bruntib, line 49: Maybe a warning instead, if CodeChecker doesn't fail on this?

analyzer/codechecker_analyzer/analyzers/gcc/analyzer.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def get_analyzer_mentioned_files(self, output):
```
bruntib, line 51: I think it's not necessary here. This is a ClangSA-specific thing for gathering other .c files that are needed for CTU.

config/labels/analyzers/gcc.json
```diff
@@ -0,0 +1,285 @@
+{
+  "analyzer": "gcc",
+  "labels": {
+    "-Wanalyzer-allocation-size": [
```
bruntib, line 4: These checker names should be exactly the same as what we provide for `--enable` flag, for example. These checker names should be listed by `CodeChecker checkers`. See Cppcheck, how we decided changing checker names compared to what Cppcheck provides itself.

bruntib (changes requested):
analyzer/.coverage might have been committed accidentally.

analyzer/codechecker_analyzer/analyzers/gcc/analyzer.py
```diff
@@ -0,0 +1,274 @@
...
+
+        # The analyzer version should be above 13.0.0 because the
+        # '-fdiagnostics-format=sarif-file' argument was introduced in this
+        # release.
+        if analyzer_version >= StrictVersion("13.0.0"):
+            return True
+
+        # FIXME: Maybe this isn't to place to emit an error, especially when
+        # we cycle over multiple binarier to find the correct one.
+        LOG.error("GCC binary found is too old at "
```
dkrupp, line 211: Not having gcc 13 on your machine and using codechecker is definitely not an error.


Now, the error message is like this
```
> CodeChecker analyzers
GCC binary found is too old at v7.5.0; minimum version is 13.0.0.
[ERROR 2023-10-11 13:49] - GCC binary found is too old at v7.5.0; minimum version is 13.0.0.
 clang-tidy
 clangsa
```

I would prefer something like this.
> CodeChecker analyzers
gcc /usr/bin/gcc GCC binary found is too old at v7.5.0; minimum version is 13.0.0. 
clang-tidy /local/workspace/llvm-project/build/bin/clang-tidy
clangsa /local/workspace/llvm-project/build/bin/clang-16

This way the user directly sees which analyzer binary is used.
Don't print the clang-tidy/clang binary location in this patch if you don't want, but it would look definitely better. :)

btw. where is cppcheck?

config/package_layout.json
```diff
@@ -3,7 +3,8 @@
     "analyzers": {
       "clangsa": "clang",
       "clang-tidy": "clang-tidy",
-      "cppcheck": "cppcheck"
+      "cppcheck": "cppcheck",
+      "gcc": "g++"
```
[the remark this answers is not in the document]
Szelethus: The same thing we do with clang. As I understand, these are functionally the same:
```
clang++ == clang -x cpp
clang == clang++ -x c
```

dkrupp (changes requested):
I am not ready with the review. I try to install gcc13 and then 
continue testing...

analyzer/codechecker_analyzer/analyzers/gcc/analyzer.py
```diff
@@ -0,0 +1,254 @@
...
+        return analyzer_context.get_context() \
+            .analyzer_binaries[cls.ANALYZER_NAME]
+
+    @classmethod
+    def get_version(cls, env=None):
+        """ Get analyzer version information. """
+        return cls.__get_analyzer_version(cls.analyzer_binary(), env)
+
+    def add_checker_config(self, checker_cfg):
+        LOG.warning("Checker configuration for Gcc is not implemented yet")
```
dkrupp, line 51: I think even a warning is not needed here. Simply don't provide any gcc analyzer checker options.

In the latest implementation I tested I even got an error here

[ERROR 2023-10-12 19:22] - Failed to get checker configuration options for 'gcc' analyzer(s)! Please try to upgrade your analyzer version to use this feature.

but actually this is not cause for a warning or an error.
Simply we dont provide any checker option yet.

config/labels/analyzers/gcc.json
```diff
@@ -0,0 +1,285 @@
...
+      "profile:security",
+      "profile:sensitive",
+      "severity:LOW"
+    ],
+    "gcc-null-argument": [
+      "doc_url:https://gcc.gnu.org/onlinedocs/gcc/Static-Analyzer-Options.html#index-Wanalyzer-null-argument",
+      "profile:extreme",
+      "profile:security",
+      "profile:sensitive",
+      "severity:LOW"
```
dkrupp, line 149: obviously these warnings are not low level for your info type of warnings, but high level warnings. we will need to categorize them.

docs/analyzer/user_guide.md
```diff
@@ -1168,11 +1168,13 @@ analyzer arguments:
...
-used to specify which analyzer tool should be used (by default, all supported
-are used). The tools are completely independent, so either can be omitted if
-not present as they are provided by different binaries.
+[_Clang-Tidy_](http://clang.llvm.org/extra/clang-tidy),
+[_Cppcheck_](http://cppcheck.sourceforge.net/) and
+[_GCC Static Analyzer_](https://gcc.gnu.org/wiki/StaticAnalyzer).
+`--analyzers` can be used to specify which analyzer tool should be used (by
+default, all supported are used). The tools are completely independent, so
+either can be omitted if not present as they are provided by different
+binaries.
```
dkrupp, line 1178: Please add a section for gcc describing any useful information about how it is invoked by default, what analyzer and checker options are supported and what are the limitations.

dkrupp (changes requested):
Nice job!
In general looks good.
- Please fix the binary exploration stuff. that's prohibitive even for the release-candidate
- some error and warning messages look too aggessive
- we will need to categorize the checkers into profiles and severities
- gcc static analyzer is said to be not working very well on c++ code at least in gcc 12. Should we recommend introduce a gcc analyzer option to only analyze c files? https://developers.redhat.com/articles/2022/04/12/state-static-analysis-gcc-12-compiler#
otherwise how the user gets rid off false warnings in C++ files.
- also add the gcc analyzer support to this doc https://github.com/Ericsson/codechecker/blob/master/docs/supported_code_analyzers.md

analyzer/codechecker_analyzer/analyzers/gcc/analyzer.py
```diff
@@ -0,0 +1,270 @@
...
+        analyzer_version = \
+            cls.__get_analyzer_version(configured_binary, environ)
+
+        # The analyzer version should be above 13.0.0 because the
+        # '-fdiagnostics-format=sarif-file' argument was introduced in this
+        # release.
+        if analyzer_version >= StrictVersion("13.0.0"):
+            return None
+
+        return f"GCC binary found is too old at v{analyzer_version.strip()}; "\
```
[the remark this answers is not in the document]
Szelethus: This does not print on my system, even if a newer gcc is unavailble from PATH.

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
```diff
@@ -361,9 +361,10 @@ def version_compatible(cls, configured_binary, environ):
         # The analyzer version should be above 1.80 because '--plist-output'
         # argument was introduced in this release.
         if analyzer_version >= StrictVersion("1.80"):
-            return True
+            return None
 
-        return False
+        return "CppCheck binary found is too old at " \
```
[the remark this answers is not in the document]
Szelethus: Does not print an error anymore. #4041 should solve not being able to give a recent enough gcc to CodeChecker. I wouldn't rush to print the binary location under the same breath -- not because thats not a good idea, I just wouldn't increase the size of this patch or the burden on us before release.

config/labels/analyzers/gcc.json
Szelethus: Thats an admirable goal, but I'd strongly argue for postponing it after rc1.

Szelethus:
I'm afraid you tested an outdated version :^) I'll see to the other comments ASAP.

analyzer/codechecker_analyzer/analyzers/gcc/analyzer.py
```diff
@@ -0,0 +1,255 @@
...
+            LOG.error(e.errno)
+        return []
+
+    @classmethod
+    def get_analyzer_config(cls):
+        """
+        Config options for gcc.
+        """
+        # TODO
+        return []
```
dkrupp, line 135: To return an empty list is valid here I think.

Now, I am getting this error:
```
workspace/test-projects/xerces-c]$ CodeChecker analyzers --analyzer-config gcc
[ERROR 2023-10-16 12:00] - Failed to get analyzer configuration options for 'gcc' analyzer! Please try to upgrade your analyzer version to use this feature.
```
However not supporting any option is not an error.

CodeChecker should write something like this instead:

`No analyzer options supported for gcc.`

dkrupp (changes requested):
- Please change the used gcc binary to gcc instead of g++
- please fix the error message for the gcc analyzer options

Then this is good to go for RC.

Szelethus:
> please fix the error message for the gcc analyzer options

I intend to fix that with the alternative solution for #4041 (where we will have at least one analyzer config  for each analyzer).

dkrupp (approved):
LGTM. Please follow-up todos in later patches.

bruntib (approved):

Issue #3672, opened 2022-05-20
Asynchronous `store` protocol


whisperity:
> This is a write-up of a technical specification for an idea that I am sure we have been having for several years now, but usually was discussed in private or only orally.

`store`, more specifically the API call to `massStoreRun` waits and blocks until the result of the store is returned to the client. As processing a store action takes a non-trivial amount of time on the serverside (and this operation is also executed only on one thread!), this means that returning from `massStoreRun` itself takes a non-trivial amount of time. The problem surfaces if the connection between the client and the server coughs, chokes, misbehaves, because it is only the networking stack in the kernel that is keeping the door open for the reply to arrive. While a disappearing client is no problem from the server's side and *data* won't be lost, CI jobs can hang indefinitely, or scripts that expect data to be available for `cmd` query after a return from `store` will break apart.

The proposal is to switch the blocking from relying on externalia like "the TCP stack" into a softer, but more local, blocking mechanism, while also turning the API itself asynchronous. This proposal is backwards compatible.

# Database changes

We already have information in `RunLock` as to what runs are undergoing a store. However, this is not enough, we need to store some semi-temporary information about store "attempts" or "sessions". This could go into its own table, per product, as this needs to be kept for a time even if the run lock is released. This table would contain the run name, a unique session token/identifier, and some status flag. The identifier might be auto-incremented, or a hash of the time when the lock was initialised, it is not a "secret" resource.

These identifiers should be garbage collected in the usual process.

# CLI changes

There are no changes needed on the CLI. Optionally, the `store` command might be extended with a `--no-block` argument which makes it immediately exit and return to shell once the server started processing the data, in case the user does not care about when the operation finished.

# API changes

A new endpoint, hereby referred to as `massStoreRun`**`Async`** shall be created. This function should return the aforementioned "store session token", or throw. The semantics of this function should be that once the server can confirm that processing of results can reasonably continue (cheap early checks like permission, the fact that the data is validly encoded before unpacking it, etc. should be performed) it returns.

To query whether the store operation has succeeded or not, a new function should be added, which returns status information (from the database) about the store. The information needed here is malleable, but at least a boolean: _"Is the operation still in progress?"_. (Consuming a successful result might want to remove the related information from the database, to ease garbage collection times at startup.)

# Implementation changes

[...]

PR #4039, opened 2023-10-11
feat(store): Explicitly time the client out if the connection hung

whisperity:
> Related to #3672, but does not implement it.

This is a partial solution that attempts to gracefully kill an ongoing `CodeChecker store` client if the reply to the `massStoreRun()` API call does not arrive in a timely manner. After reaching the timeout, the client throws an exception onto itself and presents an understandable error message about the situation. This implementation is temporary; thus, the timeout is a hard-coded limit.

The client still exits with a failure in these cases because there are no guarantees (and no deterministic ways to check) whether the timeout happened because the TCP connection failed and stuck (which is observed in a lot of cases, unfortunately) or whether the server indeed took more time than allotted to process the stored reports. A successful exit signal would result in CI scripts thinking the store was successful and then potentially breaking later, unreliably, on the fact that the stored run does not, in fact, exist or is in an outdated state, etc.

### DIFF
--- a/web/client/codechecker_client/cmd/store.py
+++ b/web/client/codechecker_client/cmd/store.py
@@ -17,6 +17,7 @@ import functools
 import hashlib
 import json
 import os
+import signal
 import sys
 import tempfile
 import zipfile
@@ -24,6 +25,9 @@ import zlib
 
 from collections import defaultdict, namedtuple
 from concurrent.futures import ProcessPoolExecutor
+from contextlib import contextmanager
+from datetime import timedelta
+from threading import Timer
 from typing import Dict, Iterable, List, Set, Tuple
 
 from codechecker_api.codeCheckerDBAccess_v6.ttypes import StoreLimitKind
@@ -713,6 +717,70 @@ def storing_analysis_statistics(client, inputs, run_name):
         os.remove(zip_file)
 
 
+class WatchdogError(Exception):
+    """
+    Custom exception type thrown by '_timeout_watchdog'. This is used instead
+    of the built-in 'TimeoutError' because TimeoutError <: OSError and other
+    handlers in the call chain during a timeout might catch the more generic
+    OSError instead, resulting in not running the appropriate handler.
+    """
+    def __init__(self, timeout: timedelta, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.timeout = timeout
+
+
+@contextmanager
+def _timeout_watchdog(timeout: timedelta, trap: int):
+    """
+    Sets up a context with a 'threading.Timer' which will signal 'trap' the
+    current process if 'timeout' passes. If and only if this 'trap' signal
+    arrives, a 'WatchdogError' is thrown for appropriate handling outside the
+    context.
+
+    The timer is started immediately as this function returns the context.
+
+    This is done in order to prevent the current process from hanging
+    indefinitely in case the stored data is too big and the server spends more
+    time processing it that what the network connection would stay alive for.
+
+    The fact that this works is **NOT GUARANTEED**! In some cases, based on
+    likely the kernel version, its implementation, and the exact network
+    topology, the client may enter an *UNINTERRUPTIBLE* sleep ('D') at which
+    point signals *MAY* not be accurately received. However, manual testing
+    revealbed that resurrection of hung clients is possible even if the process
+    is shown as "uninterruptible" by simply killing it from the outside.
+
+    This is a **TEMPORARY** measure to alleviate hangs until the proper
+    "pollable asynchronous store" feature is implemented, see:
+    http://github.com/Ericsson/codechecker/issues/3672
+    """
+    def _signal_handler(sig: int, frame):
+        if sig == trap:
+            signal.signal(trap, signal.SIG_DFL)
+            raise WatchdogError(timeout,
+                                "Timeout watchdog hit %d seconds (%s)"
+                                % (timeout.total_seconds(), str(timeout)))
+
+    pid = os.getpid()
+    timer = Timer(timeout.total_seconds(),
+                  lambda: os.kill(pid, trap))
+    LOG.debug("Set up timer for %d seconds (%s) for PID %d",
+              timeout.total_seconds(), str(timeout), pid)
+    try:
+        signal.signal(trap, _signal_handler)
+        timer.start()
+        yield timer
+    except Exception as e:
+        LOG.debug("External exception within timeout-sensing context: %s",
+                  str(e))
+        raise e
+    finally:
+        timer.cancel()
+        signal.signal(trap, signal.SIG_DFL)
+        LOG.debug("Timeout timer of %d seconds (%s) for PID %d stopped.",
+                  timeout.total_seconds(), str(timeout), pid)
+
+
 def main(args):
     """
     Store the defect results in the specified input list as bug reports in the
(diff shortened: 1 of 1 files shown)


whisperity:
## Example for the error message, as rendered to the terminal:

```
[INFO 2023-10-11 14:10] - Compressing report zip file...
[INFO 2023-10-11 14:10] - Compressing report zip file done (386EiB / 4TiB).
[INFO 2023-10-11 14:10] - Storing results to the server...
[WARNING 2023-10-11 14:10] - Timeout watchdog hit 3600 seconds (1:00:00)
  File "/home/whisperity/Ericsson/CodeChecker/build/CodeChecker/lib/python3/codechecker_client/cmd/store.py", line 872, in main
    client.massStoreRun(args.name,
  File "/home/whisperity/Ericsson/CodeChecker/build/CodeChecker/lib/python3/codechecker_client/thrift_call.py", line 45, in wrapper
    return func(*args, **kwargs)
  File "/home/whisperity/Ericsson/CodeChecker/venv_dev/lib/python3.10/site-packages/codechecker_api/codeCheckerDBAccess_v6/codeCheckerDBAccess.py", line 2325, in massStoreRun
    self.send_massStoreRun(runName, tag, version, zipfile, force, trimPathPrefixes, description)
  File "/home/whisperity/Ericsson/CodeChecker/venv_dev/lib/python3.10/site-packages/codechecker_api/codeCheckerDBAccess_v6/codeCheckerDBAccess.py", line 2340, in send_massStoreRun
    self._oprot.trans.flush()
  File "/home/whisperity/Ericsson/CodeChecker/venv_dev/lib/python3.10/site-packages/thrift/transport/THttpClient.py", line 184, in flush
    self.__http_response = self.__http.getresponse()
  File "/usr/lib/python3.10/http/client.py", line 1375, in getresponse
    response.begin()
  File "/usr/lib/python3.10/http/client.py", line 318, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python3.10/http/client.py", line 279, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/usr/lib/python3.10/socket.py", line 705, in readinto
    return self._sock.recv_into(b)
  File "/home/whisperity/Ericsson/CodeChecker/build/CodeChecker/lib/python3/codechecker_client/cmd/store.py", line 760, in _signal_handler
    raise WatchdogError(timeout,
store.WatchdogError: Timeout watchdog hit 3600 seconds (1:00:00)
[ERROR 2023-10-11 14:10] - Timeout!
        The server's reply did not arrive after 3600 seconds (1:00:00) elapsed since the server-side processing began.

        This does *NOT* mean that there was an issue with the run you were storing!
        The server might still be processing the results...
        However, it is more likely that the server had already finished, but the client did not receive a response.
        Usually, this is caused by the underlying TCP connection failing to signal a low-level disconnect.
        Clients potentially hanging indefinitely in these scenarios is an unfortunate and known issue.
                See http://github.com/Ericsson/codechecker/issues/3672 for details!

        This error here is a temporary measure to ensure an infinite hang is replaced with a well-explained timeout.
        A more proper solution will be implemented in a subsequent version of CodeChecker.
```

vodorok:
> ## Example for the error message, as rendered to the terminal:
> 
> ```
> [INFO 2023-10-11 14:10] - Compressing report zip file...
> [INFO 2023-10-11 14:10] - Compressing report zip file done (386EiB / 4TiB).
> [INFO 2023-10-11 14:10] - Storing results to the server...
> [WARNING 2023-10-11 14:10] - Timeout watchdog hit 3600 seconds (1:00:00)
> Traceback (most recent call last):
>   File "/home/whisperity/Ericsson/CodeChecker/build/CodeChecker/lib/python3/codechecker_client/cmd/store.py", line 872, in main
>     client.massStoreRun(args.name,
>   File "/home/whisperity/Ericsson/CodeChecker/build/CodeChecker/lib/python3/codechecker_client/thrift_call.py", line 45, in wrapper
>     return func(*args, **kwargs)
>   File "/home/whisperity/Ericsson/CodeChecker/venv_dev/lib/python3.10/site-packages/codechecker_api/codeCheckerDBAccess_v6/codeCheckerDBAccess.py", line 2325, in massStoreRun
>     self.send_massStoreRun(runName, tag, version, zipfile, force, trimPathPrefixes, description)
>   File "/home/whisperity/Ericsson/CodeChecker/venv_dev/lib/python3.10/site-packages/codechecker_api/codeCheckerDBAccess_v6/codeCheckerDBAccess.py", line 2340, in send_massStoreRun
>     self._oprot.trans.flush()
>   File "/home/whisperity/Ericsson/CodeChecker/venv_dev/lib/python3.10/site-packages/thrift/transport/THttpClient.py", line 184, in flush
>     self.__http_response = self.__http.getresponse()
>   File "/usr/lib/python3.10/http/client.py", line 1375, in getresponse
>     response.begin()
>   File "/usr/lib/python3.10/http/client.py", line 318, in begin
>     version, status, reason = self._read_status()
>   File "/usr/lib/python3.10/http/client.py", line 279, in _read_status
>     line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
>   File "/usr/lib/python3.10/socket.py", line 705, in readinto
>     return self._sock.recv_into(b)
>   File "/home/whisperity/Ericsson/CodeChecker/build/CodeChecker/lib/python3/codechecker_client/cmd/store.py", line 760, in _signal_handler
>     raise WatchdogError(timeout,
> store.WatchdogError: Timeout watchdog hit 3600 seconds (1:00:00)
> [ERROR 2023-10-11 14:10] - Timeout!
>         The server's reply did not arrive after 3600 seconds (1:00:00) elapsed since the server-side processing began.
> 
>         This does *NOT* mean that there was an issue with the run you were storing!
>         The server might still be processing the results...
>         However, it is more likely that the server had already finished, but the client did not receive a response.
>         Usually, this is caused by the underlying TCP connection failing to signal a low-level disconnect.
>         Clients potentially hanging indefinitely in these scenarios is an unfortunate and known issue.
>                 See http://github.com/Ericsson/codechecker/issues/3672 for details!
> 
>         This error here is a temporary measure to ensure an infinite hang is replaced with a well-explained timeout.
>         A more proper solution will be implemented in a subsequent version of CodeChecker.
> ```

I would prefer if we would not show a stacktrace to the user. It is a "planned" exceptional behaviour, the error message is enough.

whisperity:
> I would prefer if we would not show a stacktrace to the user. It is a "planned" exceptional behaviour, the error message is enough.

@vodorok I'm not sure about that. Without the stack trace, we do not know whether the timeout was triggered when **reading** from the socket (a.k.a., where we expect this) or during the **writing** of the socket (where such could, at least theoretically, also happen, but where we do NOT expect the program to hang for so long).

vodorok:
I checked the implementation, and you only guard the rpc call with the watchdog.
In the end does the stacktracematter when the watchdog triggers and shoots down the store command?

dkrupp (approved):
The stack trace is a but ugly IMHO, but I understand why we need it.

LGTM.

whisperity:
The only difference which *might* be meaningful is whether the send() fails versus the recv() failing. Nevertheless, I will put the stack trace behind the debug verbosity mode and update the patch.

vodorok:
@whisperity  Thanks that is a good compromise.

dkrupp:
I think it is much better like this. It is ok to be merged it.

PR #4040, opened 2023-10-12
[fix] Source code comment over review status rule

bruntib:
When a report is suppressed by a source code comment, then it should be stronger than a review status rule with the same bug hash. This is working fine when a new review status rule is set in the GUI for existing reports. Setting a review status rule has effect on the reports which don't have a review status provided in a source code comment.

However, if a new report is stored, then its review status is set based on the review status rule, even if it has a source code comment. This is fixed by this patch: the source code comment should be stronger even in a run store process.

### DIFF
--- a/web/tests/functional/review_status/test_review_status.py
+++ b/web/tests/functional/review_status/test_review_status.py
@@ -580,19 +580,19 @@ class TestReviewStatus(unittest.TestCase):
             lambda r: r.bugHash == UNCOMMENTED_BUG_HASH,
             reports1))
 
-        multi_confirmed_report = next(filter(
+        multi_confirmed_report1 = next(filter(
             lambda r: r.bugHash == MULTI_REPORT_HASH and
             r.reviewData.status == ReviewStatus.CONFIRMED,
             reports1))
-        multi_intentional_report = next(filter(
+        multi_intentional_report1 = next(filter(
             lambda r: r.bugHash == MULTI_REPORT_HASH and
             r.reviewData.status == ReviewStatus.INTENTIONAL,
             reports1))
-        multi_unreviewed_report = next(filter(
+        multi_unreviewed_report1 = next(filter(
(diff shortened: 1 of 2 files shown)


web/tests/functional/review_status/test_review_status.py
```diff
@@ -652,16 +652,55 @@ def test_review_status_changes(self):
...
+            lambda r: r.bugHash == MULTI_REPORT_HASH and
+            r.reviewData.status == ReviewStatus.FALSE_POSITIVE,
+            reports2))
+
+        self.assertIn(
+            (null_deref_report2.reviewData.status,
+             null_deref_report2.reviewData.comment),
+            map(lambda x:
+                (x['review_status'], x['review_status_comment']),
+                suppress_project_bugs[null_deref_report2.bugHash]))
```
Szelethus, line 684: Ugh... could we not make a new test function for these?

Szelethus (approved):
I suppose we still discourage mixing review status rules and source code comments, but otherwise, the patch makes a lot of sense. Localized rules should always overwrite general rules.

web/tests/functional/review_status/test_review_status.py
vodorok: I agree with @Szelethus  on this. If we ever have a regression it will much easier to see what went wrong.
Otherwise LGTM

vodorok (changes requested):

web/tests/functional/review_status/test_review_status.py
bruntib: Sorry, I don't understand this comment. This test is a continuation of a previous storage in the same test function. Here I test if the review statuses chance properly which were stored in a previous storage event. If I replaced these assertions to a different test function, then I couldn't ensure that it's executed after this current function.

vodorok (approved):
as discussed in person, refactoring this test case is out of the scope of this patch.

PR #4056, opened 2023-10-24
[analyzers] Deprecate --all and --details for analyzers

Szelethus:
* Deprecate both `--all` and `--details`, because they were more confusing than useful
* We list all supported (not available) analyzers by default, but print a warning message when its not available.
* We used to query the version of each analyzer binary in a uniform way, but clang, cppcheck, gcc prints their version using different flags, and in different formats. I changed this to call the analyzer plugins' get_binary_version method (which I also implemented in this patch).

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/analyzer_base.py
+++ b/analyzer/codechecker_analyzer/analyzers/analyzer_base.py
@@ -56,6 +56,17 @@ class SourceAnalyzer(metaclass=ABCMeta):
         """
         raise NotImplementedError("Subclasses should implement this!")
 
+    @abstractmethod
+    def get_binary_version(self, configured_binary, environ, details=False) \
+            -> str:
+        """
+        Return the version number of the binary that CodeChecker found, even
+        if its incompatible. If details is true, additional version information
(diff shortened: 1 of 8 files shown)


analyzer/codechecker_analyzer/analyzers/analyzer_base.py
```diff
@@ -56,6 +56,17 @@ def resolve_missing_binary(cls, configured_binary, environ):
         """
         raise NotImplementedError("Subclasses should implement this!")
 
+    @abstractmethod
+    def get_binary_version(self, configured_binary, environ, details=False) \
```
bruntib, line 60: `configured_binary` shouldn't be a parameter to this function. The concept is that analyzer plugins (i.e. sub-classes of this `SourceAnalyzer` class) return the analyzer's path through `analyzer_binary()` function. Currently that function uses the `analyzer_context` which actually stores this information (based on all kinds of config options, like environment variables, config files, etc.)

bruntib (changes requested):

bruntib (approved):

PR #4065, opened 2023-10-30
Remove a clang-tidy checker from default profile.

bruntib:
misc-include-cleaner checker results too many reports and causes an undeterministic failure at storage.


dkrupp (approved):
LGTM

PR #4066, opened 2023-10-30
[fix] Fix stacktrace when gcc analyzer version getter flag is not supported.

bruntib:
The command "CodeChecker analyzers" fails when a too old gcc doesn't support version getter flags.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/gcc/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/gcc/analyzer.py
@@ -200,6 +200,9 @@ class Gcc(analyzer_base.SourceAnalyzer):
         """
         analyzer_version = cls.get_binary_version(environ)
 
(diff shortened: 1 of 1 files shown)


analyzer/codechecker_analyzer/analyzers/gcc/analyzer.py
```diff
@@ -200,6 +200,9 @@ def is_binary_version_incompatible(cls, environ):
         """
         analyzer_version = cls.get_binary_version(environ)
 
+        if analyzer_version is None:
+            return False
```
Szelethus, line 204:
```suggestion
            return "GCC binary is too old to support -dumpfullversion."
```

Szelethus (changes requested):
This function returns an `Optional[str]`, not bool.

Szelethus (approved):
LGTM

PR #4067, opened 2023-10-30
[fix] Fix stacktrace when gcc analyzer version getter flag is not sup…

bruntib:
…ported.

The command "CodeChecker analyzers" fails when a too old gcc doesn't support version getter flags.


vodorok (approved):
LGTM, merge after passing tests

PR #4076, opened 2023-11-07
[cfg] Using multiprocess lib instead of multiprocessing

bruntib:
The standard `multiprocessing` lib has some issues mainly in Mac environment. Also, in Linux environment, too small system resources cause CodeChecker analyzer processes hang:
If the maximum number of processes (`ulimit -u`) is too small, then allocating a process pool with the standard `multiprocessing` lib raises a RuntimeError that can't be handled. With `multiprocess` lib another kind of exception is thrown, but the process itself doesn't hang.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ rules while writing your code:
 Order your `import` commands according to as follows:
 
   1. **System-wide** imports come first and foremost, e.g.
-    `import multiprocessing`.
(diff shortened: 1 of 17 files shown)


dkrupp (approved):
We tested the analysis and store on multiple open source projects.
The analysis/storage time did not increase and there is not additional error.

dkrupp:
fixes #4079

PR #4080, opened 2023-11-14
[fix] Bugfix at --enable-all and disable warning

bruntib:
**[fix] Bugfix at --enable-all and disable warning**

There was a bug in CodeChecker:

CodeChecker analyze --enable-all --disable clang-diagnostic-unused-variable

This command didn't disable that warning. The problem is that we represent --enable-all with -Weverything in the clang-tidy analyzer command, but -Wno-unused-variable has to be appended too. It was a false assumption that
"clang-tidy -checks='-clang-diagnostic-unused-variable' main.c -- -Weverything" does the job.

**[fix] Fix non-determinism of clang-diagnostic checker appearances**

Earlier we collected the enabled/disabled checkers in a set which
results non-determinism in the appearance of clang-tidy checkers.
Instead of sets we're using lists now.


dkrupp (approved):
LGTM

PR #4081, opened 2023-11-14
[report-converter] Check whether input paths exist

Szelethus:
We check (via `nargs`) whether any input paths were given, but we don't actually emit any errors if those paths point nowhere. This can be quite frustrating when the user erronously believes that the input file name was correct, is not given any errors, and left wondering why report-converter is not working.

This patch checks whether those paths actually exist.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/cli.py
+++ b/tools/report-converter/codechecker_report_converter/cli.py
@@ -151,12 +151,19 @@ class CollectFiles(argparse.Action):
 
         all_files = []
 
+        had_nonexistent_path = False
         for path in values:
-            if os.path.isfile(path):
+            if not os.path.exists(path):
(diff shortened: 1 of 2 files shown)


tools/report-converter/codechecker_report_converter/cli.py
```diff
@@ -151,12 +151,19 @@ def __call__(
 
         all_files = []
 
+        had_nonexistent_path = False
```
bruntib, line 154: Why do we need this variable? Why don't we `sys.exit(1)` right after `LOG.error()`?

bruntib (approved):

tools/report-converter/codechecker_report_converter/cli.py
bruntib: I see, thanks!

Issue #2497, opened 2019-12-17
Comments not updated with system comment about review status change when review status is changed


whisperity:
**Describe the bug**
If the review status is changed, a "system comment" is generated since #2032. However, unless the "report view" is closed and re-opened, or a manual comment is posted, the comment list is not updated and remains empty/how it was before.

**CodeChecker version**
6.11 - 6.22.2

**To Reproduce**
1. Open any report
2. Change review status
3. Open comment view
4. See nothing (or if comments were there previously, see nothing has changed)

**Expected behaviour**
Somehow make sure the comments are re-queried when the comment view is opened / or is open when the GUI knows the user did a change that invoked creation of a comment.

Automatically tell the sidebar to update its contents (incl. the "Comments (n)" number) when review status is changed. We should keep this consistent in future features if we know it'll involve a comment update.

The comments should be queried (or updated) when the user opens the comment view (as it is not open by default). I would personally prefer this: closing and opening the comment "sidebar" should do a query.

One, I think a bit more ugly approach, would be to offer a "refresh" button that refreshes just the comment sidebar.

**Desktop (please complete the following information)**
 - OS: Ubuntu Linux
 - Browser: Firefox
 - Version: 71.0

gyorb:
Yes, with the current UI this is a known behavior the page needs to be reloaded to show the new events.

We can think about how to improve it @csordasmarton ? Maybe connecting the update to some user action on the UI, like what you mentioned open/close the comments view?

Issue #3027, opened 2020-11-13
At review status change the comment list isn't reloaded - no indication that the status change indeed took place


whisperity:
**Describe the bug**
When the review status of a report is changed, a system comment #746 is generated. However, the comment pane does not reload/update with this fact until the report is reopened. This makes it hard to see at a glance that we indeed succeeded in saving the new review status (particularly problematic if the review message was elaborate and now we think it all went to nothing).

**CodeChecker version**
131861b7e94298f58cad883a533814573b89a0f3

~~~~
whisperity@476dc9a425ff:~$ CodeChecker-Release/bin/CodeChecker version
CodeChecker web version:
------------------------------------------------------------------------
Kind                          | Version
------------------------------------------------------------------------
Base package version          | 6.15.0
Package build date            | 2020-11-13T00:49
Git commit ID (hash)          | 131861b7e94298f58cad883a533814573b89a0f3
Git tag information           | 6.15
Server supported API (Thrift) | 6.35
Client API (Thrift)           | 6.35
------------------------------------------------------------------------
~~~~

**To Reproduce**
Steps to reproduce the behaviour:

1. Open the comment pane.
2. Change the review status of a bug.
3. Nothing changes in the comments.
4. Reopen the bug report and the comment list (F5, back to list and click on file again, etc...)
5. The system comment is visible.

**Expected behaviour**
The list of comments should be retrieved again from the server and the view updated.

**Desktop (please complete the following information)**
 - OS: Windows 10 19.09
 - Browser: Firefox
 - Version: 82.0.3

Issue #1368, opened 2018-02-14
Review status: Changing a bug in an already opened file shows wrong review status


whisperity:
Consider the case when two bugs with the same bug hash are found in a file. If the user opens the file and sets the review status then switches to the other "instance" of this issue, the old review status is shown.

Only a page reload or the closing and reopening of the file fixes this issue.

PR #4082, opened 2023-11-14
[GUI] Update reviewData when switching between reports

cservakt:
There was a bug when the report was loaded. The value of the review status dropdown menu did not change when switching to a report with a different status. 
Now, everytime when the report changes, the reviewData is updated as a computed property.


dkrupp:
Fixes #2497,  #1368,  #1367, #3026, #3027

dkrupp (approved):
I tested this and it looks like it fixes the updatin issues in the GUI

PR #4083, opened 2023-11-14
[fix] Extra arguments added twice to cppcheck

bruntib:
The removed line is duplicated: the extra arguments had been added a few lines above.


dkrupp (approved):
LGTM

PR #4084, opened 2023-11-15
[context][fix] Use the property instead of the field for analyzer env

Szelethus:
AnalyzerContext.__analyzer_env is lazy initialized via AnalyzerContext.analyzer_env. __parse_CC_ANALYZER_BIN became the first to use it, but erronously accessed the field directly -- which wasn't initialized at that time.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzer_context.py
+++ b/analyzer/codechecker_analyzer/analyzer_context.py
@@ -94,7 +94,7 @@ class Context(metaclass=Singleton):
 
     def __parse_CC_ANALYZER_BIN(self):
         env_var_bins = {}
-        if 'CC_ANALYZER_BIN' in self.__analyzer_env:
+        if 'CC_ANALYZER_BIN' in self.analyzer_env:
             had_error = False
             for value in self.__analyzer_env['CC_ANALYZER_BIN'].split(';'):
                 try:


dkrupp (changes requested):
Looks like it works fine based on my tests


So if CC_ANALYZERS_FROM_PATH is unset
the lookup order is 
CC_ANALYZER_BIN then
package_layout.json  then PATH

if CC_ANALYZERS_FROM_PATH is set
the lookup order is
CC_ANALYZER_BIN then PATH 

Could you please describe this in https://github.com/Ericsson/codechecker/blob/master/docs/analyzer/user_guide.md#configuring-clang-version ?

Could you please add a few Test cases?
given:

1)

unset CC_ANALYZERS_FROM_PATH
echo $CC_ANALYZER_BIN 
gcc:/usr/bin/gcc-13


```cat ./config/package_layout.json 
{
  "runtime": {
    "analyzers": {
      "clangsa": "/usr/bin/clang",
      "clang-tidy": "/usr/bin/clang-tidy",
      "cppcheck": "cppcheck",
      "gcc": "g++"
    },
    "clang-apply-replacements": "clang-apply-replacements"
  },
  "ld_lib_path_extra" : [],
  "path_env_extra" : []
}```

clangsa should be taken from /usr/bin/clang, the package config
gcc: gcc:/usr/bin/gcc-13 (from CC_ANALYZER_BIN)

2)

CC_ANALYZERS_FROM_PATH=1
echo $CC_ANALYZER_BIN 
gcc:/usr/bin/gcc-13
PATH=/opt/llvm/bin:$PATH

clangsa should be taken from the PATH: /opt/lllvm/bin
gcc:/usr/bin/gcc-13  (configured from CC_ANALYZER_BIN)

Szelethus:
I agree, but that is a separate issue worthy of a separate PR.

dkrupp (approved):
LGTM provided that you write a test case for this.

PR #4086, opened 2023-11-16
[gcc] Prune "...to here" messages

Szelethus:
They are unnecessary considering we have arrows for the same purpose.

WIP; still needs tests.


Szelethus:
Ugh. Turns out, notes cross reference each other by their ID... with this patch, this is broken (clearly, the first close was at note 12 before all the pruning)
image

PR #4093, opened 2023-11-22
[gcc] Add severities to checkers

### LINKED ISSUE #4091: severity levels are missing for the gcc checkers
checker severities are missing for all gcc checkers.

Szelethus:
I went ahead and added the missing severities. No checkers (as of now) are style checkers, so nothing got assigned to LOW. Here is some extra context:
* Pretty much every checker that warns about undefined behaviour was assigned HIGH. This seems consistent with our labels for other checkers.
* Leaks are not fatal errors, but leak checkers in other analyzers were categorized into HIGH, so I assigned HIGh to leak gcc checkers as well.
* The taint checker in ClangSA has HIGH severity, which I disagree with. Its an error to fix, but its not immediately breaking. Taint and other security checkers were assigned MEDIUM.

THIS PATCH ONLY ADDS SEVERITIES, NO PROFILES! I intend to fix those in a followup patch.

Here is the link to the GCC docs (they are rather sparse): https://gcc.gnu.org/onlinedocs/gcc/Static-Analyzer-Options.html

Fixes #4091.

### DIFF
--- a/config/labels/analyzers/gcc.json
+++ b/config/labels/analyzers/gcc.json
@@ -4,227 +4,227 @@
     "gcc-allocation-size": [
       "doc_url:https://gcc.gnu.org/onlinedocs/gcc/Static-Analyzer-Options.html#index-Wanalyzer-allocation-size",
       "profile:extreme",
-      "severity:UNSPECIFIED"
+      "severity:HIGH"
     ],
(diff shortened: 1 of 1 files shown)


bruntib (approved):

dkrupp (approved):
LGTM

PR #4094, opened 2023-11-22
Option validation of analyzer and checker configs

vodorok:
Draft for the time being...


whisperity:
Is #4204 superseding this?

PR #4095, opened 2023-11-22
[server] Use the `multiprocess` library on every platform

vodorok:
Multiprocessing is broken on osx, (probably because of the fork -> spawn change). This patch intends to correct it by using the newly added multiprocess library.

### DIFF
--- a/web/client/codechecker_client/blame_info.py
+++ b/web/client/codechecker_client/blame_info.py
@@ -1,12 +1,11 @@
 import json
 import os
-import sys
+from typing import Dict, Iterable, Optional
 import zipfile
 
-from concurrent.futures import ProcessPoolExecutor
 from git import Repo
 from git.exc import InvalidGitRepositoryError
-from typing import Dict, Iterable, Optional
+import multiprocess
 
 from codechecker_common.logger import get_logger
 
@@ -113,18 +112,9 @@ def assemble_blame_info(
 
     Returns the number of collected blame information.
     """
-    # Currently ProcessPoolExecutor fails completely in windows.
-    # Reason is most likely combination of venv and fork() not
-    # being present in windows, so stuff like setting up
-    # PYTHONPATH in parent CodeChecker before store is executed
-    # are lost.
-    if sys.platform == "win32":
-        file_blame_info = __collect_blame_info_for_files(file_paths)
-    else:
-        with ProcessPoolExecutor() as executor:
-            file_blame_info = __collect_blame_info_for_files(
-                file_paths, executor.map)
-
+    with multiprocess.Pool() as pool:
+        file_blame_info = __collect_blame_info_for_files(
+            file_paths, pool.map)
     # Add blame information to the zip for the files which will be sent
     # to the server if exist.
     for f, blame_info in file_blame_info.items():
--- a/web/client/codechecker_client/cmd/store.py
+++ b/web/client/codechecker_client/cmd/store.py
@@ -23,8 +23,9 @@ import tempfile
 import zipfile
 import zlib
 
+import multiprocess
+
 from collections import defaultdict, namedtuple
-from concurrent.futures import ProcessPoolExecutor
 from contextlib import contextmanager
 from datetime import timedelta
 from threading import Timer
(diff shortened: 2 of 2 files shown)


whisperity (changes requested):

web/client/codechecker_client/blame_info.py
```diff
@@ -120,6 +120,13 @@ def assemble_blame_info(
...
         file_blame_info = __collect_blame_info_for_files(file_paths)
+    elif sys.platform == "darwin":
+        # multiprocess might not be available on windows
+        # we import it here
+        import multiprocess
+        with multiprocess.Pool() as pool:
+            file_blame_info = __collect_blame_info_for_files(
+                file_paths, pool.map)
     else:
         with ProcessPoolExecutor() as executor:
```
whisperity, line 131: But why duplicate this logic?

At the top of the file, after all other imports:

```py
if sys.platform == "darwin":
  from multiprocessing import Pool as MultiprocessingPool
else:
  from concurrent.futures import ProcessPoolExecutor as MultiprocessingPool
```

If we want to be extra pedantic:

```py
if sys.platform == "win32":
  MultiprocessingPool = None
elif sys.platform == "darwin":
  from multiprocessing import Pool as MultiprocessingPool
else:
  from concurrent.futures import ProcessPoolExecutor as MultiprocessingPool
```
and then here simply this, because you no longer need to handle "darwin" and <anything that isn't "win32" and "darwin"> separately:

```suggestion
    else:
        with MultiprocessingPool() as pool:
            file_blame_info = __collect_blame_info_for_files(
                file_paths, pool.map)
```

whisperity: Actually, you could get even more pedantic (I see this is used in two files) by creating a compatibility module which imports the right thing based on platform, and for `win32`, defines itself as a context manager which does nothing when entering and exiting. (Not sure how to deal with the `.map` member...)

That would let us get rid of all the platform-specific distinctions in all the now affected locations. Compatibility shims are not the devil, and given that all "import" does is "set this module-global variable to point at a special dictionary object", you can make everything more readable by getting rid of conditions and indirections.

vodorok: I had the false impression, that the multiprocess lib is not supported on Windows, that's why the conditional import.
In the latest version, I completely eliminated the branch around the logic, and use multiprocess.Pool on all platforms.

bruntib (approved):

whisperity:
> Did some tests on windows, and I could use the multiprocess lib without issues. I will remove the branching completely.

@vodorok Let's just make sure these claims hold true for sufficiently old Python versions such as 3.9. Other than that, I think this is good to go.

web/client/codechecker_client/blame_info.py
```diff
@@ -1,9 +1,9 @@
...
 import os
-import sys
 import zipfile
 
-from concurrent.futures import ProcessPoolExecutor
+import multiprocess
+
 from git import Repo
 from git.exc import InvalidGitRepositoryError
 from typing import Dict, Iterable, Optional
```
whisperity, line 9:
```suggestion
import os
from typing import Dict, Iterable, Optional
import zipfile

from git import Repo
from git.exc import InvalidGitRepositoryError
import multiprocess
```

`typing` is part of the Python STL, while `multiprocess` is a 3PP library just like PyGit.

whisperity (commented):
(Barring the stylistic changes, LGTM! @vodorok)

whisperity (approved):
LGTM. But please re-check everything on real machines as well. Does the WINDOWS library version appropriately ignore the `pool.map` which was previously not passed? Or with this new library, everything works fine in a pool?

vodorok:
Do not merge yet

whisperity:
1. Create a new file `codechecker_common/multiprocessing.py` which deals with the platform-specific branching.

```py
if sys.platform in ["darwin", "win32"]:
  from multiprocessing import Pool as MultiProcessPool
elif sys.platform == ":
  from concurrent.futures import ProcessPoolExecutor as MultiProcessPool
```

2. In the usage places, import this new module and use the aliased name

```py
from codechecker_common.multiprocessing import MultiProcessPool

# ...

with MultiProcessPool(...) as pool:
  ...(..., pool.map)
```

vodorok:
Made a different implementation where the linux platform still uses the ProcessPoolExecutor class.
See #4118 
Will debug this after the release.

PR #4097, opened 2023-11-23
[gcc][docs] Elaborote on the configuration a bit more

Szelethus:


docs/analyzer/checker_and_analyzer_configuration.md
```diff
@@ -268,21 +268,35 @@ CodeChecker check -l ./compile_commands.json \
...
+they should only be enabled for experimentation.
 
 ## Example invocation
 
-TODO
+``` shell
+CodeChecker check -l ./compile_commands.json \
+  --analyzers gcc \ # Run GCC analyzer only
+  -e gcc \ # enable all checkers starting with "gcc"
+  -d gcc-double-free \ # disable gcc-double-free
```
bruntib, line 300: Is there a specific reason we want to disable this checker, or this is just for the sake of example? I believe, I know the answer, but maybe we shouldn't use a flag in a documentation which presents the usage of gcc analyzer.
Also, I would suggest using the full name of flags in documentations: `--enable`, `--disable`.

bruntib (approved):

docs/analyzer/checker_and_analyzer_configuration.md
Szelethus: With respect, I disagree with this comment -- and the enable/disable falgs seem consistent with other sections in this document.

PR #4099, opened 2023-11-23
[fix] fix venv_dev and multiprocess lib usage

bruntib:
Many functions and members of `multiprocess` lib are exposed dynamically. For this reason pylint is not able to tell what members this lib has (e.g. `multiprocess.Value()`. So these pylint reports have been suppressed.

Another purpose of this commit is to make "superset" relationship for venv_dev: if a CodeChecker developer installs venv_dev, then the content of venv should always be installed. venv_dev just extends them. `multiprocess` was a counterexample for this relationship: it was installed only by `venv` and not `venv_dev`. Such a mistake in the organization of requirements.txt files results missing libraries for developers.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/report/reports.py
+++ b/tools/report-converter/codechecker_report_converter/report/reports.py
@@ -8,7 +8,7 @@
 
 import logging
 
-from typing import Any, Callable, Iterable, List, Optional, Set
+from typing import Callable, Iterable, List, Optional, Set
 
 from codechecker_report_converter.report import Report, SkipListHandlers
 from codechecker_report_converter.report.hash import get_report_path_hash
@@ -17,8 +17,8 @@ LOG = logging.getLogger('report-converter')
 
 
 class GenericSuppressHandler:
-    get_suppressed: Callable[[Any, Report], bool]
-    store_suppress_bug_id: Callable[[Any, str, str, str, str], bool]
+    get_suppressed: Callable[[Report], bool]
+    store_suppress_bug_id: Callable[[str, str, str, str], bool]
 
 
 def get_mentioned_original_files(reports: List[Report]) -> Set[str]:
--- a/analyzer/codechecker_analyzer/analysis_manager.py
+++ b/analyzer/codechecker_analyzer/analysis_manager.py
@@ -724,6 +724,7 @@ def start_workers(actions_map, actions, analyzer_config_map,
     Start the workers in the process pool.
     For every build action there is worker which makes the analysis.
     """
+    # pylint: disable=no-member multiprocess module members.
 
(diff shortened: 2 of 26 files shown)


codechecker_common/skiplist_handler.py
```diff
@@ -103,12 +103,6 @@ def should_skip(self, source):
                 return sign == '-'
         return False
 
-    def __call__(self, source_file_path: str) -> bool:
```
vodorok, line 106: Please do this in a separate pr.

codechecker_common/skiplist_handler.py
```diff
@@ -117,11 +111,3 @@ def should_skip(self, file_path: str):
         handler.
         """
         return any(handler.should_skip(file_path) for handler in self)
-
```
vodorok, line 120: Please do this in a separate pr.

tools/report-converter/codechecker_report_converter/report/__init__.py
```diff
@@ -14,15 +14,16 @@
...
 
 from .. import util
 
 
 LOG = logging.getLogger('report-converter')
 
 
-SkipListHandlers = Callable[[str], bool]
+class SkipListHandlers(Protocol):
+    should_skip: Callable[[str], bool]
```
vodorok, line 26: Please do this in a separate pr.

tools/report-converter/codechecker_report_converter/report/reports.py
```diff
@@ -17,8 +17,8 @@
 
 
 class GenericSuppressHandler:
-    get_suppressed: Callable[[Any, Report], bool]
-    store_suppress_bug_id: Callable[[Any, str, str, str, str], bool]
+    get_suppressed: Callable[[Report], bool]
```
vodorok, line 20: Please do this in a separate pr.

tools/report-converter/codechecker_report_converter/report/__init__.py
```diff
@@ -473,7 +474,7 @@ def skip(self, skip_handlers: Optional[SkipListHandlers]) -> bool:
         if not skip_handlers:
             return False
 
-        return skip_handlers(self.file.original_path)
+        return skip_handlers.should_skip(self.file.original_path)
```
vodorok, line 477: Please do this in a separate pr.

vodorok (changes requested):
Please mention the pylint fixes in the commit title.
I think the skiplist refactoring should be done in a separate pr.

tools/report-converter/codechecker_report_converter/report/reports.py
bruntib: This change is related to mypy upgrade, so I would leave it here.

PR #4103, opened 2023-11-27
[analyze] Removing long enabled checker list at info log level

dkrupp:
At every analysis at the defult log level there was a wall of text checker list. It is enough to print this list at the verbose log level and redirect the user to the metadat.json file at default log level.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzer.py
@@ -256,8 +256,9 @@ def perform_analysis(args, skip_handlers, rs_handler: ReviewStatusHandler,
         metadata_info['analyzer_statistics']['version'] = version
 
         metadata_tool['analyzers'][analyzer] = metadata_info
-
-    LOG.info("Enabled checkers:\n%s", '\n'.join(
+    LOG.info("Enabled checker list can be found in " +
+             os.path.join(args.output_path, "metadata.json"))
+    LOG.debug("Enabled checkers:\n%s", '\n'.join(
         k + ': ' + ', '.join(v) for k, v in enabled_checkers.items()))
 
     if 'makefile' in args and args.makefile:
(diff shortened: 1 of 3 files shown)


analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -231,8 +231,9 @@ def perform_analysis(args, skip_handlers, actions, metadata_tool,
         metadata_info['analyzer_statistics']['version'] = version
 
         metadata_tool['analyzers'][analyzer] = metadata_info
-
-    LOG.info("Enabled checkers:\n%s", '\n'.join(
+    LOG.info("Enabled checker list can be found in " +
+             os.path.join(args.output_path, "metadata.json"))
```
whisperity, line 235:
```suggestion
    LOG.info("Enabled checker list can be found in %s",
             os.path.join(args.output_path, "metadata.json"))
```

Szelethus:
I totally support this patch -- the default output is still let a little loose on verbosity. With that said, metadata.json isn't very human readable without formatting, so we should consider prettifying it.

whisperity:
@Szelethus The `metadata.json` is not meant to be human-readable. It is a completely internal format and transitional medium, we should not suggest, direct, or expect users to do anything with it.

`CodeChecker parse` `--details`(?), however, could emit this list. Maybe as a proper list, not a huge concatenated string, which was also barely readable due to the difference in lengths. `clang-tidy --list-checks --checks='*'` emits a formatted list, one element per line.

whisperity (changes requested):
&nbsp;

analyzer/codechecker_analyzer/analyzer.py
whisperity, line 235:
```suggestion
```

analyzer/codechecker_analyzer/analyzer.py
```diff
@@ -231,8 +231,9 @@ def perform_analysis(args, skip_handlers, actions, metadata_tool,
...
+    LOG.debug("Enabled checkers:\n%s", '\n'.join(
         k + ': ' + ', '.join(v) for k, v in enabled_checkers.items()))
```
whisperity, line 237: Good, but maybe could be formatted into one line for each checker.

whisperity (changes requested):
(Reasoning above.)

whisperity (approved):

PR #4104, opened 2023-11-29
[analyzer][clang] Remove mention of Clang plugin from debug script

gamesh411:
Clangsa internal checkers used to be shipped as a dynamic library. This is not the case any more, so the support for those is removed.


bruntib (approved):
I approve this change, if you really won't need this option in the debug tools. So feel free to merge. However, I'd like to add that CodeChecker supports the usage of ClangSA analyzer plugins.

gamesh411:
included in #4112

PR #4105, opened 2023-11-29
[GUI] Update report sidebar structure and add review status reactivity

cservakt:
This PR gives a new report sidebar structure on the report-detail page. There are two main types in the report tree. One of them is the Outstanding category that contains outstanding reports classified into severity groups. The another category is the Closed that contains resolved, and not resolved but closed reports such as false positive, intentional.

Opening the a report-detail page, the report sidebar automatically opens a specific tree where the current report is located. If we changed the review status, the sidebar would react immediately.

### DIFF
--- a/web/server/vue-cli/e2e/pages/reportDetail.js
+++ b/web/server/vue-cli/e2e/pages/reportDetail.js
@@ -33,16 +33,20 @@ const bugTreeCommands = {
     return `${this.elements.node.selector}:nth-child(${index})`;
   },
 
-  getTreeNodeSelector(severityIndex, bugIndex, stepIndex) {
-    let selectors = [ this.selector, ">", this.node(severityIndex) ];
-
-    if (bugIndex !== undefined) {
+  getTreeNodeSelector(outstandingStateIndex, severityIndex, bugIndex, stepIndex) {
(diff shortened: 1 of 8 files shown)


Szelethus:
Please submit images when you are making GUI changes, that is kind of the point for these kinds of patches.

Szelethus (changes requested):
Please make it so that "outstanding" does not autoclose when the report I'm viewing gets placed in the "closed" category.

One another thing -- Are we sure that users know what "outstaning" mean? We have a tooltip icon next to the diff button (#3890), maybe we should put one around these categories too.

cservakt:
> Please make it so that "outstanding" does not autoclose when the report I'm viewing gets placed in the "closed" category.
> 
> One another thing -- Are we sure that users know what "outstaning" mean? We have a tooltip icon next to the diff button (#3890), maybe we should put one around these categories too.

Unfortunately, the sidebar is reactive, therefore, when the review status is changed, the side bar also changes. In the current PR the sidebar is not able to open submenus that does not part of the current reportdetail's menu tree.

I fixed the tooltip label for the outstanding icon. Thanks for the advice.

PR #4108, opened 2023-12-01
Clang-tidy arch parameter pass fix

dkrupp:
When -arch is specified in the compilation commmand '-arch ' was passed to clang-tidy which causes a
parsing error for clang tidy version 17


vodorok (approved):
LGTM

PR #4112, opened 2023-12-05
[analyzer][clang] Remove external plugin support

gamesh411:
Shipping plugins as external dynamic libraries for Clang Static Analyzer is generally frowned upon, and not supported. This patch removes the existing support for using external ClangSA analyzer plugins.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzer_context.py
+++ b/analyzer/codechecker_analyzer/analyzer_context.py
@@ -340,14 +340,6 @@ class Context(metaclass=Singleton):
         return self._data_files_dir_path
 
     @property
-    def checker_plugin(self):
-        # Do not load the plugin because it might be incompatible.
-        if env.is_analyzer_from_path():
-            return None
-
-        return os.path.join(self._data_files_dir_path, 'plugin')
-
-    @property
     def checker_labels(self):
         return self._checker_labels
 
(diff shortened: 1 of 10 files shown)


bruntib:
I agree that the plugins could be provided by `--saargs`, so I approve this patch. But what do you mean by "not supported"? Clang 17 still has a `-load` flag for loading plugins.

whisperity:
> But what do you mean by "not supported"? Clang 17 still has a `-load` flag for loading plugins.

There is also [`-fplugin=`](https://clang.llvm.org/docs/ClangPlugins.html), which provides a higher-level interface.

It is not supported in the sense that you will get no help ("support") in using that feature, including help about crappy results, crashes, complete and utter corruption of your filesystem, exploits, etc. The official stance of **CSA** (but not necessarily Clang as a whole!), is that if you use plugins and anything goes wrong, you're &mdash; for the lack of a better phrase &mdash; [**SOL**](https://www.urbandictionary.com/define.php?term=SOL). (It's like reporting a kernel panic of a [tainted kernel](https://docs.kernel.org/admin-guide/tainted-kernels.html) (a kernel that had proprietary `ko`s (drivers) loaded into it), the number one reply you'll get on the mailing list is that your crash report is disregarded.) It is your own and absolute responsibility to deal with every consequence that stems from having a plugin.

And because it is unsupported, there are no guarantees that whatever code you create and potentially distribute as a plugin will work with any arbitrary Clang version. They don't care about API or ABI compatibility.

whisperity:
@bruntib will have comments.

whisperity (changes requested):
@bruntib will have comments.

whisperity (changes requested):
@gamesh411 I'll summarise what I discussed with @bruntib prior because it looks like they hadn't gotten around to publishing their comments so far.

PR #4113, opened 2023-12-05
[cfg] Add a config file to change review statuses

bruntib:
[feat] Ignore reports in post-processing

Reports can be ignored on filepath or checker name basis. This means
that these reports will not only be considered as "closed" but totally
ignored from the .plist files. These reports are thus invisible for all
other CodeChecker commands like "diff", "parse", "store", etc.

 [feat] Change review status config YAML format

The format of the review status config is changed. A version number is
also included to identify the format in case further modifications are
needed in the future.

### DIFF
--- a/analyzer/codechecker_analyzer/analysis_manager.py
+++ b/analyzer/codechecker_analyzer/analysis_manager.py
@@ -25,6 +25,7 @@ import multiprocess
 import psutil
 
 from codechecker_common.logger import get_logger
+from codechecker_common.review_status_handler import ReviewStatusHandler
 
 from codechecker_statistics_collector.collectors.special_return_value import \
     SpecialReturnValueCollector
@@ -217,8 +218,11 @@ def prepare_check(action, analyzer_config, output_dir,
     return source_analyzer, rh
 
 
-def handle_success(rh, result_file, result_base, skip_handlers,
-                   capture_analysis_output, success_dir):
+def handle_success(
+    rh, result_file, result_base, skip_handlers,
+    rs_handler: ReviewStatusHandler,
+    capture_analysis_output, success_dir
+):
     """
     Result postprocessing is required if the analysis was
     successful (mainly clang tidy output conversion is done).
@@ -229,7 +233,7 @@ def handle_success(rh, result_file, result_base, skip_handlers,
         save_output(os.path.join(success_dir, result_base),
                     rh.analyzer_stdout, rh.analyzer_stderr)
 
-    rh.postprocess_result(skip_handlers)
+    rh.postprocess_result(skip_handlers, rs_handler)
 
     # Generated reports will be handled separately at store.
 
@@ -316,7 +320,8 @@ def handle_reproducer(source_analyzer, rh, zip_file, actions_map):
 
 
 def handle_failure(
-    source_analyzer, rh, zip_file, result_base, actions_map, skip_handlers
+    source_analyzer, rh, zip_file, result_base, actions_map, skip_handlers,
+    rs_handler: ReviewStatusHandler
 ):
     """
     If the analysis fails a debug zip is packed together which contains
@@ -331,7 +336,7 @@ def handle_failure(
     checks = source_analyzer.config_handler.checks()
     state = checks.get('clang-diagnostic-error', (CheckerState.enabled, ''))[0]
     if state == CheckerState.enabled:
-        rh.postprocess_result(skip_handlers)
+        rh.postprocess_result(skip_handlers, rs_handler)
 
     # Remove files that successfully analyzed earlier on.
     plist_file = result_base + ".plist"
@@ -485,7 +490,7 @@ def check(check_data):
     skiplist handler is None if no skip file was configured.
     """
     actions_map, action, analyzer_config, \
-        output_dir, skip_handlers, quiet_output_on_stdout, \
+        output_dir, skip_handlers, rs_handler, quiet_output_on_stdout, \
         capture_analysis_output, generate_reproducer, analysis_timeout, \
         ctu_reanalyze_on_failure, \
         output_dirs, statistics_data = check_data
@@ -603,12 +608,13 @@ def check(check_data):
 
             if success:
                 handle_success(rh, result_file, result_base,
-                               skip_handlers, capture_analysis_output,
-                               success_dir)
+                               skip_handlers, rs_handler,
+                               capture_analysis_output, success_dir)
             elif not generate_reproducer:
                 handle_failure(source_analyzer, rh,
                                os.path.join(failed_dir, zip_file),
-                               result_base, actions_map, skip_handlers)
+                               result_base, actions_map, skip_handlers,
+                               rs_handler)
 
         if rh.analyzer_returncode == 0:
             handle_analysis_result(success=True)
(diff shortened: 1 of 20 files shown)


codechecker_common/review_status_handler.py
```diff
@@ -49,7 +49,8 @@ class ReviewStatusHandler:
         'checker_filter',
         'report_hash_filter',
         'review_status',
-        'message']
+        'message',
+        'ignore']
```
Szelethus, line 55: I know I'm nitpicking, but did we not end up deciding that `ignore` is not a review status? So why is it implemented in this file?

codechecker_common/review_status_handler.py
```diff
@@ -170,6 +190,27 @@ def set_review_status_config(self, config_file):
...
+        written to the .plist files and CodeChecker should consider them as not
+        existing reports in any context.
+
+        TODO: Should "ignore" state be a review status? Currently it's a
+            standalone field.
+        TODO: If it's a standalone field, then maybe we should find another
+            name for this class, because it handles not only review statuses.
+        """
+        if self.__data is None:
+            return False
```
Szelethus, line 231: Whats the reason for this?

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -985,8 +985,13 @@ def main(args):
...
     rs_handler = review_status_handler.ReviewStatusHandler(args.output_path)
-    if 'review_status_config' in args:
-        rs_handler.set_review_status_config(args.review_status_config)
+
+    try:
+        if 'review_status_config' in args:
+            rs_handler.set_review_status_config(args.review_status_config)
+    except ValueError as ex:
+        LOG.error(ex)
+        sys.exit(1)
```
Szelethus, line 994: This is odd -- did we just let ValueErrors bleed through previously?

codechecker_common/review_status_handler.py
```diff
@@ -84,61 +86,85 @@ def __parse_codechecker_review_comment(
...
+    def __validate_review_status_yaml_data(self):
+        """
+        This function validates the data read from review_status.yaml file and
+        raises ValueError with the description of the error if the format is
+        invalid.
+        """
+        if not isinstance(self.__data, dict) or '$version' not in self.__data:
+            raise ValueError(
+                f"{self.__review_status_yaml} should be a dictionary with "
+                "the key '$version'.")
```
Szelethus, line 147: Can we do something better?
How about we
1. Emit a separate warning for the file not being a valid yaml file (why call it a "dict" in the first place?), and when the "version" key is not found
2. Give a hint: "hint: Currently the latest version of this format is '1.0'. Consider adding 'version: 1.0' to the top of `<file>.yaml`."

docs/analyzer/user_guide.md
```diff
@@ -2610,38 +2610,64 @@ Review status can be configured by a config file in YAML format. This config
...
+
+  - filters:
+      filepath: /path/to/3pp/lib/*
+    actions:
+      ignore: true
+      reason: The reports of this lib are ignored completely.
+```
+
+The review settings rules have two componetes: `filters` and `actions`. A rule
+is applied for a report if the filter fields match. The following filter
```
Szelethus, line 2648: for *every* report

Szelethus (changes requested):
I have raised some minor things, I won't die on a hill for them. Here are the things I want addressed:
1. Why did you implement ignoring for each analyzer individually?
2. Please write out the parameter types wherever possible.
3. Avoid saying "plists" -- refer to it only when you are actually writing to or reading from plist files. Most of the time when you mention plists, you mean report objects (reports, for short).

dkrupp (changes requested):
1) I think it is a problem that this feature does not have a Test Cases as it is quite complex.
2) yaml schema/content validation should be done on the client side 
    - before uploading to the server
    - before executing the anaysis
    - at the parse phase
3) It should be supported to modify the review status config file after analysis. (the review status config file should be a symlink to from the report directory)
4) We should drop an error if the symlink is invalid

analyzer/codechecker_analyzer/analysis_manager.py
```diff
@@ -783,6 +785,7 @@ def signal_handler(signum, frame):
                          analyzer_config_map.get(build_action.analyzer_type),
                          output_path,
                          skip_handlers,
+                         rs_handler,
```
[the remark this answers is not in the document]
bruntib: Yes, I considered similar solutions. For example AnalyzerContext is a singleton in CodeChecker which is available everywhere in the analyzer without wiring it through such monster functions with >10 parameters. The simplification of these functions is among the plans.
I also considered handling "skipfile" and "review status" config by the same class, because the "ignore" option of the review status config file has a similar "skip" effect.
But in the end I chose using the same pattern what skip handler uses. I can tell you later why I chose this instead of the other options.

analyzer/codechecker_analyzer/analyzers/clangsa/result_handler.py
```diff
@@ -43,7 +48,13 @@ def postprocess_result(self, skip_handlers: Optional[SkipListHandlers]):
             reports = report_file.get_reports(
                 self.analyzer_result_file, self.checker_labels,
                 source_dir_path=self.source_dir_path)
-            reports = [r for r in reports if not r.skip(skip_handlers)]
+            reports = (r for r in reports if not r.skip(skip_handlers))
+
+            if rs_handler:
+                reports = (r for r in reports
+                           if not rs_handler.should_ignore(r))
```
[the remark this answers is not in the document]
bruntib: Actually skipping and ignoring are really similar. But the difference is that skipped files are not analyzed at all. Ignored reports are analyzed, but their reports are removed in post-processing.
The "list" is not iterated twice. That's why I changed square brackets to parentheses: this is just creating a generator object. The list is generated only once.

analyzer/codechecker_analyzer/analyzers/result_handler_base.py
```diff
@@ -181,7 +182,13 @@ def clean_results(self):
...
                 LOG.debug(oserr)
 
-    def postprocess_result(self, skip_handlers: Optional[SkipListHandlers]):
+    # TODO: If the parameters are not optional then we can get rid of some
+    #   extra checks.
+    def postprocess_result(
+        self,
+        skip_handlers: Optional[SkipListHandlers],
+        rs_handler: Optional[ReviewStatusHandler]
+    ):
```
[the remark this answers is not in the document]
bruntib: I agree that it shouldn't be the analyzers' responsibility to ignore some reports. But it shouldn't be the analyzers' responsibility to use skip handlers. I didn't want to apply a different way of using review status handler and skip handlers, and I also didn't want to refactor the whole skiplist handling in this regard. So my conclusion was to apply review status handler conform with skip list handlers and this refactoring will be done later.

analyzer/codechecker_analyzer/analyzers/clangsa/result_handler.py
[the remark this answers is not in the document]
Szelethus: Sorry, meant to say "almost" the same thing :)

codechecker_common/review_status_handler.py
```diff
@@ -83,45 +87,88 @@ def __parse_codechecker_review_comment(
...
         raises ValueError with the description of the error if the format is
         invalid.
         """
-        if not isinstance(self.__data, list):
+        if not isinstance(self.__data, dict) or '$version' not in self.__data:
             raise ValueError(
-                f"{self.__review_status_yaml} should be a list of review "
-                "status descriptor objects.")
+                f"{self.__review_status_yaml} should be a dictionary with "
+                "the key '$version'.")
```
Szelethus, line 148: I still want this separated. I can't properly unittest this.

Szelethus (changes requested):

Szelethus (changes requested):
Some comments regarding the warning messages.

Szelethus (approved):
Approving to not block merging in the case that you are done working on this, but I'm already offline.

PR #4127, opened 2023-12-11
[fix] Fixing OFF and UNAVAILABLE detectionStatus in the report sidebar

cservakt:
It is a bug fix to locate correctly both OFF and UNAVAILABLE reports in the sidebar. These kind of reports are not outstanding, so they should be classified as closed.


dkrupp (approved):
I tested and it works as intended.

PR #4131, opened 2023-12-13
Fix remote server path resolution

vodorok:


codechecker_common/review_status_handler.py
```diff
@@ -320,7 +320,7 @@ def get_review_status_from_source(
         # the source files are extracted as the original file path.
         # TODO: Should original_path be strip('/') at store?
         source_file_name = os.path.realpath(os.path.join(
-            self.__source_root, report.file.original_path))
+            self.__source_root, report.file.original_path.strip('/')))
```
whisperity, line 323: Question: Should this use `lstrip` instead of plain `strip`?

whisperity:
Another question: Shouldn't this patch first go into the `release-v6.23.0` branch (which does not exist yet) and then be cherry-picked onto master? Or is going into master first and then getting backported to the fix-branch the better way?

vodorok:
@We are planning to release a patch release today (or tomorrow), as it affects all servers that is not running in a docker container.

codechecker_common/review_status_handler.py
vodorok: Valid question, the goal is to make `report.file.original_path` not absolute. lstrip would only remove the forward slash from the front of the path string.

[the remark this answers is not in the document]
vodorok: Good question, I need to better understand if path manipulation can be achieved at this stage.
Eg.: can  Mallory construct an `original_path ` that can navigate out of the temp folder, and cause harm there.

codechecker_common/review_status_handler.py
```diff
@@ -316,15 +317,22 @@ def get_review_status_from_source(
...
-        # TODO: Should original_path be strip('/') at store?
-        source_file_name = os.path.realpath(os.path.join(
-            self.__source_root, report.file.original_path))
+        if self.__source_root:
+            # sanitize paths like mass store run
+            source_file_name = path_for_fake_root(
+                report.file.original_path, self.__source_root)
+        else:
+            # in this branch we are operating on the original filesystem,
+            # and we need the path to he original file
```
Szelethus, line 326:
```suggestion
            # and we need the path to the original file
```

Szelethus: "On this branch the server runs on the same file system where the source file is"

Szelethus: Its very interesting and not at all trivial why we need to branch on this -- can you create an issue to remind us to document this properly?

Szelethus (approved):

codechecker_common/review_status_handler.py
vodorok: Thanks, for the review, will make an issue tomorrow

PR #4140, opened 2023-12-20
Fix typo 'vesion' in sarif conversion template

SirMutantRat:
The Sarif schema has a 'version' field but the template is mis-typed as 'vesion'. Adding the missing 'r' allows the code to be read as Sarif data (e.g. by the VSCode Sarif viewer plugin.


bruntib (approved):
Thanks for the fix!

PR #4141, opened 2023-12-22
[analyzer] Ignore `-fno-tree-dominator-opts`

bo-dani:
The `-fno-tree-dominator-opts` GCC option is not recognized by clang.


bruntib (changes requested):
Thanks for the patch!
Could you, please, change the commit message so it includes `-fno-tree-dominator-opts` instead of `-fno-move-loop-invariants`?
Thank you!

bo-dani:
> Thanks for the patch! Could you, please, change the commit message so it includes `-fno-tree-dominator-opts` instead of `-fno-move-loop-invariants`? Thank you!

Oops, brainfart from my side :)

PR #4143, opened 2024-01-02
[report-converter][fix] Don't crash when we can't find the checker name for ubsan

Szelethus:
We forgot to inizialize `checker_name` to its default value, which lead to a read-before-assignment crash.

### DIFF
--- a/tools/report-converter/tests/unit/analyzers/ubsan_output_test_files/ubsan1_nonmatching_msg.plist
+++ b/tools/report-converter/tests/unit/analyzers/ubsan_output_test_files/ubsan1_nonmatching_msg.plist
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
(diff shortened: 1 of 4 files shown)


tools/report-converter/tests/unit/analyzers/ubsan_output_test_files/ubsan1_nonmatching_msg.plist
```diff
@@ -0,0 +1,69 @@
...
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>diagnostics</key>
+	<array>
+		<dict>
+			<key>category</key>
+			<string>unknown</string>
+			<key>check_name</key>
+			<string>UndefinedBehaviorSanitizer</string>
```
whisperity, line 11: How can this be that, if it was supposedly initialised to `None`?

Szelethus: The idea is that when `checker_name` is `None`, we fall back to a default checker name later down the road, which is the name of the sanitizer.

bruntib (approved):

PR #4152, opened 2024-01-15
[store] Unique reports _before_ storing

Szelethus:
We used to zip every result file in the result directory and send it to the server. This is tremendously wasteful, as many report files contained the very same reports (for instance, if the report originated from a header file).

In this patch, I unique all reports, dump them in a tmpfile, and store that to the server. For a clang-tidy --enable-all analysis on xerces, this reduced the size of the zipfile drastically:

BEFORE:
```
Compressing report zip file done (1.6GiB / 62.2MiB).
[...]
real	8m32,381s
user	15m57,436s
sys	0m14,708s
```
AFTER: 
```
Compressing report zip file done (413.0MiB / 9.0MiB).
[...]
real	5m45,621s
user	15m54,858s
sys	0m14,458s
```

While this doesn't speed up CodeChecker parse or CodeChecker diff, it still speeds up the server quite a bit, both for storage and query times.

I also did a little cleanup on the debug logs -- the one in report converter spit out a line for literally _every report_, that feels a little unnecessary. I also added some logs to show that the parsing of result files is in progress, though that could be a little more verbose.

edit:
In addition, I needed to fix plist parsing at one point, and also, I had a looooot of trouble with `AnalysisInfo` not being set for each report. The problem was in this nasty ass loop:

```python
    LOG.info("Building report zip file...")
    with zipfile.ZipFile(zip_file, 'a', allowZip64=True) as zipf:
        # Add the files to the zip which will be sent to the server.
        for file_path in files_to_compress:
            _, file_name = os.path.split(file_path)

            # Create a unique report directory name.
            report_dir_name = hashlib.md5(os.path.dirname(
                file_path).encode('utf-8')).hexdigest()

            zip_target = os.path.join('reports', report_dir_name, file_name)
            zipf.write(file_path, zip_target)
```

Now, since we support multiple report dir storages at once, in the zipped file, each of those report dirs get their own report dir. However, do you also not see the clear implementation of that in the form of "for each input dir, create an output dir"? That nasty `os.path.dirname` part does all that, and my initial implementation didn't set the dirname right. Since this manifested in errors seemingly unrelated parts of the code, I was lead on a month-long chase to find out that this was the issue.

Also, since we handle all input report dirs all at once in a rather cryptic manner, I also chose to just brute-force my implementation into the code, instead of refactoring it into an easy-to-digest "for each dir in input dir, zip dir". That might be worth a followup refactoring patch.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/report/parser/plist.py
+++ b/tools/report-converter/codechecker_report_converter/report/parser/plist.py
@@ -519,6 +519,11 @@ class Parser(BaseParser):
                         self._create_macro_expansion(
                             macro_expansion, file_index_map))
 
+            if report.annotations:
+                diagnostic["report-annotation"] = dict()
+                for key, value in report.annotations.items():
+                    diagnostic["report-annotation"][key] = value
+
             data['diagnostics'].append(diagnostic)
 
         return data
--- a/tools/report-converter/codechecker_report_converter/report/hash.py
+++ b/tools/report-converter/codechecker_report_converter/report/hash.py
@@ -205,5 +205,6 @@ def get_report_path_hash(report: Report) -> str:
     if not report_path_hash:
         LOG.error('Failed to generate report path hash: %s', report)
 
+    # This might be a little too verbose even for the verbose output.
     # LOG.debug(report_path_hash)
     return __str_to_hash(report_path_hash)
--- a/tools/report-converter/codechecker_report_converter/report/hash.py
+++ b/tools/report-converter/codechecker_report_converter/report/hash.py
@@ -205,6 +205,4 @@ def get_report_path_hash(report: Report) -> str:
     if not report_path_hash:
         LOG.error('Failed to generate report path hash: %s', report)
 
-    # This might be a little too verbose even for the verbose output.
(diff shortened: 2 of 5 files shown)


bruntib (changes requested):

tools/report-converter/codechecker_report_converter/report/parser/plist.py
```diff
@@ -519,6 +519,11 @@ def convert(
                         self._create_macro_expansion(
                             macro_expansion, file_index_map))
 
+            if report.annotations:
+                diagnostic["report-annotation"] = dict()
+                for key, value in report.annotations.items():
+                    diagnostic["report-annotation"][key] = value
```
bruntib, line 525: Why do we store the report annotations in the diagnostic here? I can't see it's usage.

Szelethus: Because we parse a lot of plist files, merge its contents, and spit it back out. This means that we parse report-annotation info as well, and need it output it as well -- this last part was simply missing prior to this patch.

bruntib (approved):

PR #4156, opened 2024-01-16
feat(gui): Checker status auditing

whisperity:
Closes #4096. Blocked by #4089.

> _Full description WIP._


whisperity:
@cservakt I'm working on refactoring this to have some of the logic reusable as we discussed! It's just a non-trivial process to get everything right.

whisperity:
@cservakt I did the refactoring and now the `getAnalysisInfo()` in the Dialog's code is much simpler, and deals with only calling transformation codes. So for your feature, you should be able to reuse the mixin from other locations, which **will** do the simple checks (such as heuristically reasoning why the checker list is not available) and load the **raw** values (analysis command and **raw, unsorted and ungrouped** checker list &mdash; so essentially only what is received from the API), but doing the "fancy" transformations is a separate **opt-in** call, which is driven by the main Dialog's code.

web/server/vue-cli/src/mixins/analysis-info-handling.mixin.js
```diff
@@ -0,0 +1,268 @@
...
+
+  methods: {
+    loadAnalysisInfo() {
+      const analysisInfoFilter = new AnalysisInfoFilter({
+        // Query a run's analysis info only if a run history ID is not explicit.
+        runId: (!this.runHistoryId ? this.runId : null),
+
+        runHistoryId: this.runHistoryId,
+        reportId: this.reportId
+      });
```
cservakt, line 181: As we discussed, it would be better to have parameter list for loadAnalysisInfo method that contains runId, runHistoryId, reportId instead of using the props. You have already called in this way the loadAnalysisInfo() in the AnalysisInfoDialog.vue component.

web/server/vue-cli/src/mixins/analysis-info-handling.mixin.js
```diff
@@ -0,0 +1,268 @@
...
+              }
+
+              analysisInfo.populateAnalyzers();
+
+              resolve(analysisInfo);
+            }));
+      });
+    },
+
+    decideNegativeCheckerStatusAvailability(analysisInfo) {
```
cservakt, line 210: Same problem here. It would be better to get runId, runHistorId and reportId as a parameter of the function.

web/server/vue-cli/src/mixins/analysis-info-handling.mixin.js
whisperity: Blimey... I think the original attempt at the refactoring perhaps was a bit *too* eager then...

cservakt (approved):
Thank you for the rebase! Now, it looks good to me.

whisperity:
I've created **tests** for this feature, but I will upload them as a separate patch after this is merged s.t. this review is not derailed by concerns specific only to the tests.

PR #4161, opened 2024-01-29
[debug][scripts] Add experimental script for clustering similar crashes

gamesh411:
This tool is specific to the aggregate failzip format, that is specifically emitted by our CI, and uses data analysis techniques, which I am only fleetingly familiar with, so take this as it is: a prototype that can be improved.
What are your thoughts?


dkrupp (approved):
Looks like a very interesting idea. Thanks for the tool!

PR #4164, opened 2024-02-06
Adding OpenSSF Scorecard badge to README

gkunz:
This adds the OpenSSF Scorecard badge to the README. We spent effort on adopting Scorecards best-practices, so we should be proudly showing this to the users of CodeChecker.


vodorok (approved):
I agree with you, thanks for the effort! Looks good to me!

PR #4167, opened 2024-02-14
refactor(database): Make `zlib`-compressed values more explicit

whisperity:


whisperity:
Putting this into `6.23.2` so far, because this should not be merged together with #4089 to ensure that the migrations that need to be performed when the server is upgraded to not take too much time.

The time measurements for this patch will be ongoing soon...

whisperity:
Putting this back into draft because there are some weird performance characteristics on large real-world databases. I'll investigate...

whisperity:
> [!CAUTION]
> Deferred until #4171 is implemented.

Alright, it looks like this patch, as it stands right now, while correct in isolation, is intractable to execute in a timely manner for large enough (think multiple millions of rows) databases. While each individual action often takes less than **1 second**, the number of records affected (because this is a per-record update that can't be grouped or optimised further), it becomes somewhere around 30 minutes per database. Which would be fine if you had one database or maybe two, but not if you have like 50 that each take half an hour, and then these migrations are executed in series. See #4171.

PR #4172, opened 2024-02-25
[feat] Display dynamic columns in static HTML 

bruntib:
"CodeChecker parse -e html" command produces a static HTML that displays
the reports. However, the dynamic analysis related columns were missing
from this table: testcase and timestamp are now displayed too, if any.

Note: this PR consists of 2 commits. The first one is deployed in another PR (#4168). After that's merged, this one can be rebased.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/report/output/html/static/js/buglist.js
+++ b/tools/report-converter/codechecker_report_converter/report/output/html/static/js/buglist.js
@@ -16,7 +16,12 @@ var BugList = {
   },
 
   _cmp3 : function (a, b) {
-    return a < b ? -1 : a > b ? 1 : 0;
+    if (a === null)
+      return -1;
+    else if (b === null)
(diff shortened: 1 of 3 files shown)


tools/report-converter/codechecker_report_converter/report/output/html/static/js/buglist.js
```diff
@@ -138,6 +151,18 @@ var BugList = {
     review_col.appendChild(document.createTextNode(data['review-status']));
     row.appendChild(review_col);
 
+    var testcase_col = document.createElement('td');
```
vodorok, line 154: I would prefer to use variables declared with `let` as they are less bugprone for the avg. developer. (or use `const` if they are not modified.)

vodorok (approved):
Can be merged if my nitpicks are not worthy of fixing.

PR #4180, opened 2024-03-04
Allow specifiying relative paths in the `--file` option without `*`

vodorok:
With this modification, the `--file` option would allow passing paths with folder prefixes.

Consider a compilation json that contains a `.` in the directory fields, with the following example directory layout:
```
.                            
├── compile_commnands.json
├── include                                
│   └── lib.h                    
├── lib                                                                               
│   └── lib.cpp                                                                       
├── Makefile                                                                          
└── src                                                                               
    ├── a.cpp            
    └── b.cpp  
```
Before this patch, you could only specify path to source files with the `--file` option, that were not absolute is to add a `*` before the directory name.  
`CodeChecker analyze ./compile_commnands.json -o ./reports --file "*src/b.cpp"`

After this patch the following two methods will also work:
`CodeChecker analyze ./compile_commnands.json -o ./reports --file "./src/b.cpp"`
`CodeChecker analyze ./compile_commnands.json -o ./reports --file "src/b.cpp"`

TBD.:
* Documentation changes

### DIFF
--- a/analyzer/codechecker_analyzer/analysis_manager.py
+++ b/analyzer/codechecker_analyzer/analysis_manager.py
@@ -20,6 +20,7 @@ import traceback
 import zipfile
 
 from threading import Timer
+from pathlib import Path
 
 import multiprocess
 import psutil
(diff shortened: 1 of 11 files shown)


dkrupp:
Implemented by https://github.com/Ericsson/codechecker/pull/4250

PR #4191, opened 2024-03-12
fix(migration): Migrate reports with appropriate default checker ID

whisperity:
> Follows up #4089.

The first row in a table with an `AUTOINCREMENT` ID has the value 1, not 0. This was synchronised in the migration script to have `server_default="1"` such that all default constructed `report` rows have a **valid** `FOREIGN KEY` to the `checkers` table (the `__FAKE__` checker). But the hand-written upgrade query filtered for `0` to get a batch of "unhandled rows", essentially preventing appropriate migration.

### DIFF
--- a/web/server/codechecker_server/migrations/report/versions/c3dad71f8e6b_store_information_about_enabled_and_disabled_checkers_for_a_run.py
+++ b/web/server/codechecker_server/migrations/report/versions/c3dad71f8e6b_store_information_about_enabled_and_disabled_checkers_for_a_run.py
@@ -196,8 +196,13 @@ def upgrade():
         LOG.info("Done filling 'checkers', %d unique entries.", checker_count)
 
     def upgrade_reports():
+        Base = automap_base()
+        Base.prepare(conn, reflect=True)
+        Report = Base.classes.reports  # 'reports' is the table name!
+        Checker = Base.classes.checkers
+
         db = Session(bind=conn)
-        report_count = db.execute("SELECT COUNT(id) FROM reports;").scalar()
+        report_count = db.query(Report.id).count()
         if not report_count:
             return
 
@@ -214,6 +219,10 @@ def upgrade():
                      if batch < num_batches else report_count,
                      float(batch) / num_batches * 100)
 
+        fake_chk_id = db.query(Checker.id) \
+            .filter(Checker.analyzer_name == FakeChecker[0],
+                    Checker.checker_name == FakeChecker[1]) \
+            .scalar()
         for i in range(0, num_batches):
             # FIXME: "UPDATE ... SET ... FROM ..." is only supported starting
             # with SQLite version 3.33.0 (2020-08-14). Until this version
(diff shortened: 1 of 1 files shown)


cservakt:
I noticed that instead of utilizing SQLAlchemy's built in methods, raw SQL queries were executed directly. While this approach can work, it may introduce potential risks (for example when table or column name is changed).

Would it be possible to refactor the queries?

web/server/codechecker_server/migrations/report/versions/c3dad71f8e6b_store_information_about_enabled_and_disabled_checkers_for_a_run.py
```diff
@@ -214,6 +214,10 @@ def _print_progress(batch: int):
                      if batch < num_batches else report_count,
                      float(batch) / num_batches * 100)
 
+        fake_chk_id = db.execute(
+            "SELECT id FROM checkers WHERE "
+            f"analyzer_name = '{FakeChecker[0]}' AND "
+            f"checker_name = '{FakeChecker[1]}';").scalar()
```
cservakt, line 220:
```suggestion
        fake_chk_id = db.query(Checker.id).filter(
            Checkers.analyzer_name == FakeChecker[0],
            Checkers.checker_name == FakeChecker[1]
        ).scalar()
```

whisperity: @cservakt You can not use this construct here, because we are in a migration context.

whisperity:
> I noticed that instead of utilizing SQLAlchemy's built in methods, raw SQL queries were executed directly. 
> 
> Would it be possible to refactor the queries?

That is very common during migration: [[1]](https://github.com/Ericsson/codechecker/blob/v6.23.0/web/server/codechecker_server/migrations/report/versions/75ae226b5d88_review_status_for_each_report.py#L71), [[2]](https://github.com/Ericsson/codechecker/blob/v6.23.0/web/server/codechecker_server/migrations/report/versions/dabc6998b8f0_analysis_info_table.py#L59), [[3]](https://github.com/Ericsson/codechecker/blob/v6.23.0/web/server/codechecker_server/migrations/report/versions/e89887e7d3f0_add_bug_path_length.py#L25). When this script is executing, the schema in the database is not available in an ORM-y way, and there is a very high chance that the schema in the database "right now" does not match what the db model definitions in the Python package contain.

`automap(reflect` allows for some very limited functionality to figure out some things and go back to relying on SQLAlchemy to generate queries.

> While this approach can work, it may introduce potential risks (for example when table or column name is changed).

Not even `automap` solves this. You have to rely on the **exact** table and column name, hence the

```py
Report = Base.classes.reports  # 'reports' is the table name!
Checker = Base.classes.checkers
```

in the earlier code.

I can try changing this specific query to use `automap` here, as the `Checkers` table (or its contents) is not mutated in this function.

web/server/codechecker_server/migrations/report/versions/c3dad71f8e6b_store_information_about_enabled_and_disabled_checkers_for_a_run.py
whisperity: Nevermind, this one query (and its corresponding counterpart in `downgrade()`) I was able to refactor.

The bigger queries that deal with inner queries, pagination, UPDATE...SET...FROM are unlikely to be refactorable to something SQLAlchemy can stomach.

cservakt (approved):
LGTM

vodorok (approved):
LGTM

PR #4195, opened 2024-03-20
[fix] Wait for run_lock when freeing it

bruntib:
A store event inserts a line to the run_lock table in order to prevent concurrent storage to the same run. The run_lock record is removed at the end of storage. The insertion and deletion of the run_lock record are locked operations on the database level. In CodeChecker we used "nowait" lock which means that these insert and delete operations fail and throw an exception if they can't happen immediately.

This is too strict in case of removing the run_lock object, because the thrown exception is forwarded to the user who can't handle it anyway. For this reason the run_lock removal is waiting for the database lock to be free. In the worst case the exception will still be thrown after the configured statement_timeout, but ideally that should be an unlikely event.

### DIFF
--- a/web/server/codechecker_server/api/mass_store_run.py
+++ b/web/server/codechecker_server/api/mass_store_run.py
@@ -421,9 +421,15 @@ class MassStoreRun:
         """ Remove the lock from the database for the given run name. """
         # Using with_for_update() here so the database (in case it supports
         # this operation) locks the lock record's row from any other access.
+        # The "nowait" option is turned off, because it results a 'could not
+        # obtain lock on row in relation "run_locks"' error message in case
+        # this query happens simultaneously with a concurrent locking of
+        # another parallel store event. With "nowait" option the query will be
+        # blocked until the lock is undone. In the worst case when
+        # statement_timeout is reached, the exception will be thrown anyway.
         run_lock = session.query(RunLock) \
             .filter(RunLock.name == self.__name) \
-            .with_for_update(nowait=True).one()
+            .with_for_update(nowait=False).one()
         session.delete(run_lock)
         session.commit()
 


whisperity:
> With "nowait" option the query will be blocked until the lock is undone. In the worst case when statement_timeout is reached, the exception will be thrown anyway.

Unfortunately, this adds the requirement that the server *should* be configured with a [`STATEMENT TIMEOUT`](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-STATEMENT-TIMEOUT), which is **not** the default case. If the server operator does not configure this value, the store will hang forever, am I reading this right?

Can you please check if using [`SELECT ... FOR UPDATE `**`SKIP LOCKED`**](https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE) would be more beneficial for us? If I am reading the docs right (and I am not claiming I am!) then this would mean that in case the locking can not be done, we get back an empty result (so we use `.one_or_none()` to fetch the maybe-row) and we can thus deduce that another transaction happened to beat us to the punch, so there is no lock for us (the currently committed `store`) to remove. There is a [`skip_locked` parameter in `with_for_update()`](https://docs.sqlalchemy.org/en/13/core/selectable.html#sqlalchemy.sql.expression.GenerativeSelect.with_for_update).

> To prevent the operation from waiting for other transactions to commit, use either the `NOWAIT` or `SKIP LOCKED` option. With `NOWAIT`, the statement reports an error, rather than waiting, if a selected row cannot be locked immediately. **With `SKIP LOCKED`, any selected rows that cannot be immediately locked are skipped. Skipping locked rows provides an inconsistent view of the data, so this is not suitable for general purpose work, but can be used to avoid lock contention with multiple consumers accessing a queue-like table.**

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -421,9 +421,15 @@ def __free_run_lock(self, session: DBSession):
...
+        # The "nowait" option is turned off, because it results a 'could not
+        # obtain lock on row in relation "run_locks"' error message in case
+        # this query happens simultaneously with a concurrent locking of
+        # another parallel store event. With "nowait" option the query will be
+        # blocked until the lock is undone. In the worst case when
+        # statement_timeout is reached, the exception will be thrown anyway.
         run_lock = session.query(RunLock) \
             .filter(RunLock.name == self.__name) \
-            .with_for_update(nowait=True).one()
+            .with_for_update(nowait=False).one()
```
vodorok, line 432: Should only this one usage site be modified?

vodorok (commented):
Maybe it is worth to try `skip locked`.  (Even though most real-world database applications need to be guarded with statement timeouts.)

whisperity:
> (Even though most real-world database applications need to be guarded with statement timeouts.)

Still, guidelines are not __requirements__. If we have this as a __requirement__ without which the server will **misbehave**, then the server startup process should query the db, check if there is a statement timeout, and show a warning to the operator.

bruntib:
A different solution will be provided later.

PR #4198, opened 2024-03-25
fix(migration): Do not emit log output for `checker_id` normalisation

whisperity:
> Follows up #4089, #4191.

In case a new product is added to an empty server, what happens in the background is a full "migration" from the "empty" state to the newest schema, which results in the log output that is removed by this patch on the standard output of the server.

These lines are not directly needed, but there is no good reason to hide them behind an `if len(...):` check either. In case actual rows are migrated (which never happens when a new product is added to the server), the appropriate output showing the row count of the `UPDATE` statement will be printed to the output normally.


vodorok (approved):
LGTM

PR #3866, opened 2023-04-05
[analyzer] Add a new flag to demote missing checker errors to a warning

Szelethus:
Altough discouraged, if someone really insists on having a soft warning when a checker is supplied to --enable or --disable, they can use the new flag '--no-missing-checker-error', which is disabled by default.

PR #4204, opened 2024-03-28
Add functionality to validate analyzer and checker options

noraz31:
When specifying invalid checker and analyzer config, the CodeChecker analyze command used to give a warning but continued running.
This functionality throws an error and prevents the analysis from running in this case.
CODECHKR-5784

### DIFF
--- a/analyzer/codechecker_analyzer/cmd/analyze.py
+++ b/analyzer/codechecker_analyzer/cmd/analyze.py
@@ -865,8 +865,7 @@ def is_checker_config_valid(checker_conf: List[CheckerConfig]) -> bool:
                 f"--checker-config' command to see available checkers."
             )
             return False
-        else:
-            return True
+        return True
 
 
 def get_affected_file_paths(
--- a/analyzer/codechecker_analyzer/cmd/analyze.py
+++ b/analyzer/codechecker_analyzer/cmd/analyze.py
@@ -26,7 +26,8 @@ from codechecker_analyzer import analyzer, analyzer_context, \
 from codechecker_analyzer.analyzers import analyzer_types, clangsa
 from codechecker_analyzer.arg import \
     OrderedCheckersAction, OrderedConfigAction, existing_abspath, \
-    analyzer_config, checker_config
+    analyzer_config, checker_config, AnalyzerConfig, CheckerConfig
+
 from codechecker_analyzer.buildlog import log_parser
 
 from codechecker_common import arg, logger, cmd_config, review_status_handler
@@ -788,7 +789,9 @@ LLVM/Clang community, and thus discouraged.
                                default=argparse.SUPPRESS,
                                help="Emit a warning instead of an error when "
                                     "an unknown checker name is given to "
-                                    "either --enable or --disable.")
+                                    "either --enable, --disable,"
+                                    "--analyzer-config and/or "
+                                    "--checker-config.")
 
     logger.add_verbose_arguments(parser)
     parser.set_defaults(
@@ -800,37 +803,43 @@ def is_analyzer_config_valid(analyzer_conf: List[AnalyzerConfig]) -> bool:
     Ensure that the analyzer_config parameter is set to a valid value
     by verifying if it belongs to the set of allowed values.
     """
-    analyzer = analyzer_conf.__getitem__(0).analyzer
-
-    if analyzer in analyzer_types.supported_analyzers:
-        analyzer_class = analyzer_types.supported_analyzers[analyzer]
-    else:
-        LOG.error(f"Invalid --analyzer-config parameter: {analyzer} is not "
-                  f"a supported analyzer. The supported analyzers are "
-                  f"{', '.join(a for a in analyzer_types.supported_analyzers)}"
-                  f".")
-        return False
-
-    if not isinstance(analyzer_conf, list):
-        return False
-
-    analyzers = [analyzer[0] for analyzer in
-                 analyzer_class.get_analyzer_config()]
-
-    # Make sure "[clang-tidy] Allow to override checker list #3203" works
-    if analyzer_class.ANALYZER_NAME == 'clang-tidy':
-        analyzers.append('take-config-from-directory')
+    wrong_configs = []
+    supported_analyzers = analyzer_types.supported_analyzers
 
     for cfg in analyzer_conf:
-        if cfg.analyzer == analyzer_class.ANALYZER_NAME and \
-                cfg.option not in analyzers:
-            LOG.error(f"Invalid --analyzer-config parameter: "
-                      f"{analyzer_class.ANALYZER_NAME} has no config named "
-                      f"{cfg.option}. Use the 'CodeChecker analyzers "
-                      f"--analyzer-config {analyzer_class.ANALYZER_NAME}' "
-                      f"command to check available configs.")
-            return False
-        return True
+        if cfg.analyzer not in supported_analyzers:
+            wrong_configs.append((cfg.analyzer, None))
+            continue
+
+        analyzer_class = supported_analyzers[cfg.analyzer]
+        analyzer_name = analyzer_class.ANALYZER_NAME
+        analyzers = [analyzer[0] for analyzer in
+                     analyzer_class.get_analyzer_config()]
+
+        # Make sure "[clang-tidy] Allow to override checker list #3203" works
+        if analyzer_name == 'clang-tidy':
+            analyzers.append('take-config-from-directory')
+
+        if cfg.analyzer == analyzer_name and cfg.option not in analyzers:
+            wrong_configs.append((analyzer_name, cfg.option))
+
+    if wrong_configs:
+        for analyzer_name, option in wrong_configs:
+            if option is None:
+                LOG.error(
+                    f"Invalid argument to --analyzer-config: {analyzer_name} "
+                    f"is not a supported analyzer. Supported analyzers are: "
+                    f"{', '.join(a for a in supported_analyzers)}.")
+            else:
+                LOG.error(
+                    f"Invalid argument to --analyzer-config: {analyzer_name} "
+                    f"has no config named {option}. Use the 'CodeChecker "
+                    f"analyzers --analyzer-config {analyzer_name}' command to "
+                    f"see available configs.")
+
+        return False
+
+    return True
 
 
 def is_checker_config_valid(checker_conf: List[CheckerConfig]) -> bool:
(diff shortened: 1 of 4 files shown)


Szelethus (changes requested):
Please check the analyzer name as well, and clear up the wording in the error msg.

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -797,6 +795,81 @@ def add_arguments_to_parser(parser):
...
+        if cfg.analyzer == analyzer_class.ANALYZER_NAME and \
+                cfg.option not in analyzers:
+            LOG.error(f"Invalid --analyzer-config parameter: "
+                      f"{analyzer_class.ANALYZER_NAME} has no config named "
+                      f"{cfg.option}. Use the 'CodeChecker analyzers "
+                      f"--analyzer-config {analyzer_class.ANALYZER_NAME}' "
+                      f"command to check available configs.")
+            return False
+        else:
+            return True
```
Szelethus, line 834: [Don’t use else after a return](https://llvm.org/docs/CodingStandards.html#id41).
```suggestion
        return True
```

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -797,6 +795,81 @@ def add_arguments_to_parser(parser):
...
+                cfg.checker not in checkers:
+            LOG.error(
+                f"Invalid --checker-config parameter: "
+                f"{analyzer_class.ANALYZER_NAME} has no checker named "
+                f"{cfg.checker}. Use the 'CodeChecker checkers "
+                f"--checker-config' command to see available checkers."
+            )
+            return False
+        else:
+            return True
```
Szelethus, line 870: [Don’t use else after a return](https://llvm.org/docs/CodingStandards.html#id41).
```suggestion
        return True
```

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -797,6 +795,81 @@ def add_arguments_to_parser(parser):
         func=main, func_process_config_file=cmd_config.process_config_file)
 
 
+def validate_analyzer_parameter(analyzer_conf):
```
Szelethus, line 798: I usually prefer boolean function names to start with `is_`, `was_`, `should_` etc. to imply that its a boolean function. Also, we're specifically checking analyzer configs, not any random parameter to the analyzer.
```suggestion
def is_analyzer_config_valid(analyzer_conf: List[AnalyzerConfig]) -> bool:
```

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -797,6 +795,81 @@ def add_arguments_to_parser(parser):
...
+                      f"{analyzer_class.ANALYZER_NAME} has no config named "
+                      f"{cfg.option}. Use the 'CodeChecker analyzers "
+                      f"--analyzer-config {analyzer_class.ANALYZER_NAME}' "
+                      f"command to check available configs.")
+            return False
+        else:
+            return True
+
+
+def validate_checker_parameter(checker_conf):
```
Szelethus, line 837:
```suggestion
def is_checker_config_valid(checker_conf: List[CheckerConfig]) -> bool:
```

analyzer/tests/unit/test_checker_handling.py
```diff
@@ -396,6 +399,20 @@ def test_disable_clangsa_checkers(self):
...
+        --analyzer-config or --checker-config parameter is specified.
+        """
+
+        analyzer_cfg = [AnalyzerConfig('asd', '123', 'value')]
+        checker_cfg = [CheckerConfig('clangsa', 'asd', '123', 'value')]
+
+        analyzer_func = validate_analyzer_parameter(analyzer_cfg)
+        checker_func = validate_checker_parameter(checker_cfg)
+
+        [self.assertFalse(func) for func in [analyzer_func, checker_func]]
```
Szelethus, line 414: This code looks smarter than it should be :^)
```suggestion
        self.assertFalse(validate_analyzer_parameter(analyzer_cfg))
        self.assertFalse(validate_checker_parameter(checker_cfg)
```

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -797,6 +795,81 @@ def add_arguments_to_parser(parser):
         func=main, func_process_config_file=cmd_config.process_config_file)
 
 
+def validate_analyzer_parameter(analyzer_conf):
+    """
+    Ensure that the analyzer_config parameter is set to a valid value
+    by verifying if it belongs to the set of allowed values.
+    """
+    analyzer = analyzer_conf.__getitem__(0).analyzer
```
Szelethus, line 803: Why didn't you use the bracket (`[]`) operator? Also, you check all analyzer configs here, but only check the very first analyzer config for a valid analyzer field?

For this example, the second analyzer config this has an invalid analyzer field, but you miss it:
```
CodeChecker check -b "g++ test.cpp" --analyzer-config clangsa:prune-paths=false --analyzer-config:OOGABOOGA:no-false-positives=true
```

Szelethus (changes requested):
Analyzer value checking is still a little off. My other comments are just nitpicking.

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -797,6 +796,79 @@ def add_arguments_to_parser(parser):
...
+        if analyzer_class.ANALYZER_NAME == 'clang-tidy':
+            analyzers.append('take-config-from-directory')
+
+        if cfg.analyzer == analyzer_class.ANALYZER_NAME and \
+                cfg.option not in analyzers:
+            LOG.error(f"Invalid --analyzer-config parameter: "
+                      f"{analyzer_class.ANALYZER_NAME} has no config named "
+                      f"{cfg.option}. Use the 'CodeChecker analyzers "
+                      f"--analyzer-config {analyzer_class.ANALYZER_NAME}' "
+                      f"command to check available configs.")
```
Szelethus, line 831: Same here.

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -797,6 +796,79 @@ def add_arguments_to_parser(parser):
         func=main, func_process_config_file=cmd_config.process_config_file)
 
 
+def is_analyzer_config_valid(analyzer_conf: List[AnalyzerConfig]) -> bool:
```
Szelethus, line 799: You exit this function upon finding the very first invalid config. You could instead emit an error for all invalid configs, and exit in the end (this could save precious failed reruns of an ill-configured CI job).

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -797,6 +796,79 @@ def add_arguments_to_parser(parser):
...
+                      f"{analyzer_class.ANALYZER_NAME} has no config named "
+                      f"{cfg.option}. Use the 'CodeChecker analyzers "
+                      f"--analyzer-config {analyzer_class.ANALYZER_NAME}' "
+                      f"command to check available configs.")
+            return False
+
+    return True
+
+
+def is_checker_config_valid(checker_conf: List[CheckerConfig]) -> bool:
```
Szelethus, line 837: Much of the same here with the too early exit and parameter/argument things.

analyzer/tests/unit/test_checker_handling.py
```diff
@@ -396,6 +399,18 @@ def test_disable_clangsa_checkers(self):
...
+        """
+        This test checks whether the analyze command detects if a wrong
+        --analyzer-config or --checker-config parameter is specified.
+        """
+
+        analyzer_cfg = [AnalyzerConfig('asd', '123', 'value')]
+        checker_cfg = [CheckerConfig('clangsa', 'asd', '123', 'value')]
+
+        self.assertFalse(is_analyzer_config_valid(analyzer_cfg))
+        self.assertFalse(is_checker_config_valid(checker_cfg))
```
Szelethus, line 412: Would liketo see some more tests for
* Correct analyzer/checker configs,
* Valid analyzer for --analyzer-config, but invalid config name,
* Valid analyzer for --checker-config and
  * invalid checker name + valid checker option name,
  * valid checker name + invalid checker option name,
* Invalid analyzer for --checker-config.

Szelethus (changes requested):
Almost there!

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -790,13 +789,106 @@ def add_arguments_to_parser(parser):
...
         func=main, func_process_config_file=cmd_config.process_config_file)
 
 
+def is_analyzer_config_valid(analyzer_conf: List[AnalyzerConfig]) -> bool:
+    """
+    Ensure that the analyzer_config parameter is set to a valid value
+    by verifying if it belongs to the set of allowed values.
+    """
+    if not isinstance(analyzer_conf, list):
+        return False
```
Szelethus, line 807: Shouldn't we emit an error message here if we terminate CodeChecker?

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -790,13 +789,106 @@ def add_arguments_to_parser(parser):
...
+    return True
+
+
+def is_checker_config_valid(checker_conf: List[CheckerConfig]) -> bool:
+    """
+    Ensure that the checker_config parameter is set to a valid value
+    by verifying if it belongs to the set of allowed values.
+    """
+    if not isinstance(checker_conf, list):
+        return False
```
Szelethus, line 854: same here

analyzer/codechecker_analyzer/cmd/analyze.py
```diff
@@ -934,6 +1026,20 @@ def main(args):
...
+        'checker_config': is_checker_config_valid
+    }
+
+    config_validator_res = [validate_func(getattr(args, conf))
+                            for conf, validate_func in config_validator.items()
+                            if conf in args]
+
+    if False in config_validator_res \
+            and 'no_missing_checker_error' not in args:
+        sys.exit(1)
```
Szelethus, line 1041: If the validation failed and this suppression flag isn't present, try to advertise it, somewhat like what I've done in #3866:
```suggestion
        LOG.info("Although it is not reccomended, if you want to "
                 "suppress errors relating to unknown "
                 "analyzer/checker configs, consider using the option "
                 "'--no-missing-checker-error'")
        sys.exit(1)
```

Szelethus (changes requested):

Szelethus:
Also, please create tests for the interaction of the suppression flag and this validation checking.

analyzer/codechecker_analyzer/cmd/analyze.py
noraz31: To be removed. Will be verified elsewhere.

PR #4207, opened 2024-04-02
feat(script): Verify the existence of checker config `doc_url` pages and find appropriate older releases for gone (removed, dealpha, etc.) checkers

whisperity:
The checker label configuration files most often contain a documentation page link that we suggest to the user when viewing the details of a report. These JSON files are always hard-baked into a released package, and the server serves information based on what is available in the deployed image. As all of these links point to **external** resources, these links are very susceptible to [**link rot**](http://enwp.org/Link_rot).

For example, suppose that an analysis was stored with the `alpha.Foo` checker, with the URL pointing to `.../alpha/Foo.html`. Once the underlying analyser's documentation changes (usually for two reasons: improving the checker and removing it from alpha, or the checker becoming completely removed from upstream!), this link is now dead. Newer reports stored with `core.Foo` (`.../core/Foo.html`) will point to a proper documentation, but re-routing `alpha.Foo`'s documentation page to `core.Foo`'s would be an invalid action, as the behaviour of the checker might have changed meanwhile, rendering the contents of the new document inapplicable to the old report! In addition, nothing prevents the user from running an older analyser with/through a newer CodeChecker package, and uploading new results from the `alpha.` version even when after `core.` analyser's release.

This patch introduces an opt-in tool which reads the configuration files and **verifies** whether the URL is available to a hypothetical user. If not, it attempts to employ a heuristic pipeline to attempt a URL that corresponds to the checker with the currently dead link, first by fixing the typos in the URL, and if that is still unsuccessful, trying the documentation sites of older releases. For now, this fixing logic is only implemented for the **LLVM**-based analysers, *Clang SA* and *Clang-Tidy*, as implementing it requires an accurate understanding of the documentation structure of the specific analyser.

### DIFF
--- a/scripts/labels/label_tool/checker_labels.py
+++ b/scripts/labels/label_tool/checker_labels.py
@@ -1,0 +1,140 @@
+# -------------------------------------------------------------------------
+#
+#  Part of the CodeChecker project, under the Apache License v2.0 with
+#  LLVM Exceptions. See LICENSE for license information.
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# -------------------------------------------------------------------------
+"""Provides I/O with the configuration files that describe checker labels."""
+from collections import deque
+import json
+import pathlib
+from typing import Dict, List, Optional, cast
+
+from codechecker_common.checker_labels import split_label_kv
+
+from .output import Settings as OutputSettings, error, trace
+
+
+_ConfigFileLabels = Dict[str, List[str]]
+
+SingleLabels = Dict[str, Optional[str]]
+Labels = Dict[str, Dict[str, str]]
+
+
+def _load_json(path: pathlib.Path) -> Dict:
+    try:
+        with path.open("r") as file:
+            return json.load(file)
+    except OSError:
+        import traceback
+        traceback.print_exc()
+
+        error("Failed to open label config file '%s'", path)
+        raise
+    except json.JSONDecodeError:
+        import traceback
+        traceback.print_exc()
+
+        error("Failed to parse label config file '%s'", path)
+        raise
+
+
+def _save_json(path: pathlib.Path, data: Dict):
+    try:
+        with path.open("w") as file:
+            json.dump(data, file, indent=2)
+            file.write('\n')
+    except OSError:
+        import traceback
+        traceback.print_exc()
+
+        error("Failed to write label config file '%s'", path)
+        raise
+    except (TypeError, ValueError):
+        import traceback
+        traceback.print_exc()
+
+        error("Failed to encode label config file '%s'", path)
+        raise
+
+
+class MultipleLabelsError(Exception):
+    """
+    Raised by `get_checker_labels` if multiple labels exist for the same key.
+    """
+
+    def __init__(self, key):
+        super().__init__("Multiple labels with key: %s", key)
+        self.key = key
+
+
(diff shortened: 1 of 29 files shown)


scripts/labels/label_tool/checker_labels.py
```diff
@@ -0,0 +1,137 @@
...
+    for the `key` label. Raises `MultipleLabelsError` if there is at least
+    two labels with the same `key`.
+    """
+    try:
+        label_cfg = cast(_ConfigFileLabels, _load_json(path)["labels"])
+    except KeyError:
+        error("'%s' is not a label config file", path)
+        raise
+
+    prefix = f"{key}:"
```
bruntib, line 83: The format of the label files allows whitespaces around the colon (`:`) character and at the two ends of the label. Currently we don't have label files containing whitespaces in labels, but at least a TODO may be worth here. Another option could be prohibiting whitespaces in label files.

bruntib (approved):

scripts/labels/label_tool/checker_labels.py
```diff
@@ -0,0 +1,137 @@
...
+    filtered_labels = {
+        checker: [label.replace(prefix, '', 1)
+                  for label in labels if label.startswith(prefix)]
+        for checker, labels in label_cfg.items()}
+    if OutputSettings.trace():
+        deque((trace("No '%s:' label found for '%s/%s'",
+                     key, analyser, checker)
+               for checker, labels in filtered_labels.items()
+               if not labels), maxlen=0)
```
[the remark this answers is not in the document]
whisperity: It is much faster, considering that the generator is actually generating `None`s, so there is no use assigning the result to a variable. It's faster to just discard it, and CPython even has a dedicated hot path implementation for this use case!

http://stackoverflow.com/questions/50937966/fastest-most-pythonic-way-to-consume-an-iterator/50938015#50938015 I was surprised to learn about this fact as well.

scripts/labels/label_tool/checker_labels.py
whisperity: So, just to clarify, `" foo : bar"` as a label **is** the most extreme but still valid format? A.k.a., the "label" is the full `key:value` string in your interpretation.

I think it will be easy to have this tool support that.

bruntib: I wasn't sure about its necessity when I added this many `strip()`s, but currently this is the implementation:
https://github.com/Ericsson/codechecker/blob/11554c62c8642af6ad3f35514aff41598d080b25/codechecker_common/checker_labels.py#L82-L95

scripts/labels/label_tool/checker_labels.py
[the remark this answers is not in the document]
bruntib: Wow, good to know, thanks :)

scripts/labels/label_tool/checker_labels.py
whisperity: @bruntib Thanks for the code link! If I come up with a simpler solution I'll just include changes to this file as part of this patch as well!

But due to the massively non-ergonomic and unusual complexity of creating a working Python package out of CodeChecker's source, I'd like not to rely on loading stuff directly from the main source tree.

Alternatively, I'll just have to say you can't use these scripts prior to having run `make package` and that's it...

PR #4210, opened 2024-04-05
[GUI] New "Checker Coverage" statistics tab to show all enabled checkers and …

cservakt:
It is a new feature based on #4089 PR. The new Checker Coverage statitics tab on the Statistics page can show all enabled checkers for runs that are selected (or for all runs if no run selected) in the report filter. The table lists all checkers that were enabled in at least one of runs according to the latest analysis. It also shows checker severity, status, number of closed and outstanding reports.

Status can inform the user that the specific checker was "Enabled in all" runs or "Enabled in all runs except these" where "runs" and "these" words are links to list appropriate runs.

Closed and outstanding report counts depend on review and detection status. These statistics represent the number of closed and outstanding reports that belong to runs that were created with new DB schema.


bruntib (approved):

web/api/report_server.thrift
```diff
@@ -396,6 +396,17 @@ struct CheckerCount {
 }
 typedef list<CheckerCount> CheckerCounts
 
+struct CheckerInfo {
+  1: string       checkerName,  // Analyzer name of the checker.
```
whisperity, line 400: Wrong comment.

web/api/report_server.thrift
```diff
@@ -396,6 +396,17 @@ struct CheckerCount {
...
+  2: string       analyzerName, // Analyzer name of the checker.
+  3: Severity     severity,     // Severity level of the checker.
+  4: list<i64> enabled,      // Runs' names in which the checker enabled.
+  5: list<i64> disabled,     // Runs' names in which the checker disabled.
```
whisperity, line 404: Formatting is wrong, also, "Run names" is enough, no need to add a genitive.

web/api/report_server.thrift
```diff
@@ -396,6 +396,17 @@ struct CheckerCount {
...
+  6: i64          closed        // Number of closed reports.
+  7: i64          outstanding   // Number of outstanding reports.
+}
+typedef map<i64, CheckerInfo> CheckerInfos
```
whisperity, line 408: What is the key int here?

web/api/report_server.thrift
```diff
@@ -852,6 +863,14 @@ service codeCheckerDBAccess {
                                  5: i64          offset)
                                  throws (1: codechecker_api_shared.RequestFailed requestError),
 
+  // It gives statistics for specific runs which checkers were
+  // enabled and how many reports are opened or closed.
+  // If the run id list is empty the statistics
+  // will be counted for all of the runs.
+  CheckerInfos getCheckerInfo(1: list<i64>    runIds,
```
whisperity, line 870: Documentation for the required permission level is missing.

whisperity, line 870: Also, this function (and the associated return struct) could've really used a much better name... `CheckerInfo` is as vague as it can get, especially since this API function seems to conflate multiple things into the same return object (both details about a checker, which could've been its own function; and the number of bugs specific to the checkers, which also could've been its own function).

Maybe `getCheckerStatusVerificationDetails()`... At least that will immediately make it apparent that this is a **very** purpose-tailored function.

web/server/codechecker_server/api/report_server.py
```diff
@@ -42,7 +42,8 @@
     ReviewStatusRuleFilter, ReviewStatusRuleSortMode, \
     ReviewStatusRuleSortType, RunData, RunFilter, RunHistoryData, \
     RunReportCount, RunSortType, RunTagCount, \
-    SourceComponentData, SourceFileData, SortMode, SortType
+    SourceComponentData, SourceFileData, SortMode, SortType, \
+    ReviewStatus as API_ReviewStatus, DetectionStatus, CheckerInfo
```
whisperity, line 46: 20 lines later (@✏️) the alphabetic order is kept in the change, but here it was violated.

web/server/codechecker_server/api/report_server.py
```diff
@@ -58,6 +59,7 @@
 from ..database.database import conv, DBSession, escape_like
 from ..database.run_db_model import \
     AnalysisInfo, AnalysisInfoChecker as DB_AnalysisInfoChecker, \
+    AnalysisInfoChecker, \
```
whisperity, line 62: **Marker: ✏️**

web/server/codechecker_server/api/report_server.py
```diff
@@ -2861,6 +2863,187 @@ def getCheckerCounts(self, run_ids, report_filter, cmp_data, limit,
...
+                    Checker.id.label("checker_id"),
+                    Checker.checker_name,
+                    Checker.analyzer_name,
+                    Checker.severity,
+                    Report.bug_id,
+                    Report.detection_status,
+                    Report.review_status,
+                )
+                .join(RunHistory)
+                .join(AnalysisInfo, RunHistory.analysis_info)
```
whisperity, line 2897: Is having a connected row here always guaranteed, or is it not a problem if there isn't one?

web/server/codechecker_server/api/report_server.py
```diff
@@ -2861,6 +2863,187 @@ def getCheckerCounts(self, run_ids, report_filter, cmp_data, limit,
...
+                session.query(
+                    (func.string_agg(
+                        cast(Run.id, sqlalchemy.String).distinct(),
+                        ','
+                        ).label("run_id")
+                     if session.bind.dialect.name == "postgresql"
+                     else func.group_concat(Run.id.distinct()).label("run_id"))
+                    if report_filter.isUnique
+                    else Run.id.label("run_id"),
+                    Checker.id.label("checker_id"),
```
whisperity, line 2888: (Style nit: Perhaps it would improve readability if these were factored out into their own variables, and then reused. Lots of indentations here, and the `X if B else Y` syntax isn't helping the readability.)

web/server/codechecker_server/api/report_server.py
```diff
@@ -2861,6 +2863,187 @@ def getCheckerCounts(self, run_ids, report_filter, cmp_data, limit,
...
+                                subquery.c.detection_status.in_(list(map(
+                                    detection_status_str,
+                                    (DetectionStatus.OFF,
+                                     DetectionStatus.UNAVAILABLE)
+                                ))),
+                                False
+                            )
+                        ],
+                        else_=True
+                    ).label("isEnabled"),
```
whisperity, line 2940: (Style ditto.)

**Marker: 🦄**

web/server/codechecker_server/api/report_server.py
```diff
@@ -2861,6 +2863,187 @@ def getCheckerCounts(self, run_ids, report_filter, cmp_data, limit,
...
+                                    subquery.c.review_status.in_(list(map(
+                                        review_status_str,
+                                        (API_ReviewStatus.UNREVIEWED,
+                                         API_ReviewStatus.CONFIRMED))))
+                                ),
+                                True
+                            )
+                        ],
+                        else_=False
+                    ).label("isOpened"),
```
whisperity, line 2960: (Style ditto.)

web/server/codechecker_server/api/report_server.py
```diff
@@ -2861,6 +2863,187 @@ def getCheckerCounts(self, run_ids, report_filter, cmp_data, limit,
...
+                                subquery.c.detection_status.in_(list(map(
+                                    detection_status_str,
+                                    (DetectionStatus.OFF,
+                                     DetectionStatus.UNAVAILABLE)
+                                ))),
+                                False
+                            )
+                        ],
+                        else_=True
+                    ),
```
whisperity, line 2981: Style ditto, **especially** considering this is the same as line 2928 (@🦄).

web/server/codechecker_server/api/report_server.py
```diff
@@ -2861,6 +2863,187 @@ def getCheckerCounts(self, run_ids, report_filter, cmp_data, limit,
...
+                    if checker_id in checker_stats \
+                    else CheckerInfo(
+                        checkerName=checker_name,
+                        analyzerName=analyzer_name,
+                        enabled=[],
+                        disabled=all_run_id,
+                        severity=severity,
+                        closed=0,
+                        outstanding=0
+                    )
```
whisperity, line 3026: What is the expected value for the saturation of the dict's keys, and whether this code takes the true or the false branch? It is said that in Python it's not nice to do `X = D[K] if K in D else V` because it will do **two** lookups in the key set. Depending on whether we expect the key to be already there or not, it's better to use either

**exceptions**

```py
try:
  X = d[K]
except KeyError:
  X = V
```

or the null-coalescing **get()**: `X = D.get(K, V)`

web/server/codechecker_server/api/report_server.py
```diff
@@ -2861,6 +2863,187 @@ def getCheckerCounts(self, run_ids, report_filter, cmp_data, limit,
...
+
+                if is_enabled:
+                    for r in (run_ids.split(",")
+                              if type(run_ids) is str
```
whisperity, line 3030: `isinstance(run_ids, str)`

web/server/codechecker_server/api/report_server.py
```diff
@@ -2861,6 +2863,187 @@ def getCheckerCounts(self, run_ids, report_filter, cmp_data, limit,
...
+                              else [run_ids]):
+                        run_id = int(r)
+                        if run_id not in checker_stat.enabled:
+                            checker_stat.enabled.append(run_id)
+                        if run_id in checker_stat.disabled:
+                            checker_stat.disabled.remove(run_id)
```
whisperity, line 3036: Perhaps it would be useful to have `set`s here for quicker lookup, and only convert them to a list at the end before the value of `checker_stat` is completely calculated. (Also, wouldn't that help with returning a sorted list to the API request as well?)

web/server/vue-cli/src/components/AnalysisInfoDialog.vue
```diff
@@ -272,6 +273,8 @@ export default {
 
       var analysisInfo = await this.loadAnalysisInfo(
         this.runId, this.runHistoryId, this.reportId);
+      decideNegativeCheckerStatusAvailability(
+        analysisInfo, this.runId, this.runHistoryId, this.reportId);
```
whisperity, line 277: This should be put below the `highlightedCmds` assignment such that it is visually together with the other API call that is related to the checker status.

web/server/vue-cli/src/components/Statistics/StatisticsHelper.js
```diff
@@ -1,5 +1,9 @@
 import { ccService, handleThriftError } from "@cc-api";
-import { ReportFilter, ReviewStatus } from "@cc/report-server-types";
+import {
+  ReportFilter,
+  ReviewStatus,
+  // AnalysisInfoFilter
```
whisperity, line 5: Stale comment. It does not look like this import is indeed used.

whisperity (commented):
It seems like this was merged so fast I had no time to check it in time. Hopefully, the comments can be incorporated in a subsequent patch after rc but before the actual release.

web/api/report_server.thrift
whisperity: But wait, you're saying you're returning run names here, but the returned type is list of ints. So what is returned here? Maybe it got turned into Run IDs?

web/server/vue-cli/src/components/Statistics/CheckerCoverage/CheckerCoverageStatistics.vue
```diff
@@ -0,0 +1,334 @@
...
+            stat.enabledInAllRuns
+              ? "Enabled in all selected runs"
+              : "Not enabled in all selected runs",
+            stat.closed,
+            stat.outstanding
+          ];
+        })
+      ];
+
+      this.toCSV(data, "codechecker_analysis_statistics.csv");
```
whisperity, line 220: (The downloaded filename are not in line with the name of the component it is getting obtained from.)

web/api/report_server.thrift
cservakt: You're right. I didn't update the comments properly. Enabled and disabled list are run id lists not names.

web/api/report_server.thrift
cservakt: The key for CheckerInfo is a checker id.

web/server/codechecker_server/api/report_server.py
whisperity: Storno, because the data structure is updated often but it's directly in the Thrift type so it would need back-and-forth conversions.

PR #4212, opened 2024-04-11
Do not allow parse if the <report dir> does not exist 

### LINKED ISSUE #4186: `parse` should fail when path does not exist
When `parse` is given a path that doesn't exist, it doesn't output any error. Instead it outputs
```
---------------------------------------------
Number of processed analyzer result files | 0
Number of analyzer reports                | 0
---------------------------------------------
----=================----
```

I was using `./report` with analyze and `./reports` with parse because of copy-paste and couldn't understand why I'm getting no results for an hour. :(

`parse` should print an error and terminate with a non-zero code when the given path does not exist.

**Additional context**

noraz31:
This change prevents the CodeChecker parse command
from running if the entered report dir or filepath 
parameter does not exist.
Up until now, the parse command ran with any parameters,
reporting zero findings if a wrong parameter was entered.
Fixes #4186


analyzer/codechecker_analyzer/cmd/parse.py
```diff
@@ -283,6 +283,11 @@ def main(args):
         LOG.error(fnerr)
         sys.exit(1)
 
+    for input_path in args.input:
+        if not os.path.exists(input_path):
+            LOG.error(f"Input path {input_path} does not exist!")
+            sys.exit(1)
```
Szelethus, line 289: You could consider emitting a warning for all non-existent paths before exiting.

analyzer/tests/functional/cmdline/test_cmdline.py
```diff
@@ -250,6 +250,16 @@ def test_analyzer_config(self):
...
 
+    def test_parse_incorrect_file_path(self):
+        """
+        This test checks whether the parse command stops running if a
+        non-existent path is specified.
+        """
+
+        parse_cmd = [env.codechecker_cmd(), 'parse', '/asd/123/qwe']
+
+        self.assertEqual(1, run_cmd(parse_cmd)[0])
```
Szelethus, line 261: Maybe try to match the string in the error message?

Szelethus (approved):
LGTM!

PR #4215, opened 2024-04-14
[fix] Prevent overlapping report groups

bruntib:
When getRunResults() API function queries the reports, the query uses LIMIT and OFFSET for returning the given page of the results due to their huge number. However, it is possible that different pages have overlapping reports. This way the resulting report list may contain duplicate reports.
In this commit we apply an ordering on report id in order to prevent this.

### DIFF
--- a/web/server/codechecker_server/api/report_server.py
+++ b/web/server/codechecker_server/api/report_server.py
@@ -2089,6 +2089,13 @@ class ThriftRequestHandler:
                                        sort_type_map,
                                        order_type_map)
 
+                # Most queries are using paging of reports due their great
+                # number. This is implemented by LIMIT and OFFSET in the SQL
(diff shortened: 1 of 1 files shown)


whisperity:
Does this change still work if there is something in the "sort types" in the API? Which sorting takes priority in that case?

cservakt (approved):

cservakt:
> Does this change still work if there is something in the "sort types" in the API? Which sorting takes priority in that case?

In SQL Alchemy `.order_by(col1, col2)` is equal to `.order_by(col1).order_by(col2).` The generated raw SQL query would be "SELECT col1, col2, ... FROM <table_name> ORDER BY col1, col2;" In both cases.

whisperity (approved):

PR #4225, opened 2024-04-19
feat(script): Automatically generate `doc_url` and `severity` from ToCs, documentations, and analyser outputs

whisperity:
> [!IMPORTANT]
> :no_entry: Blocked by #4207.

There were some existing scripts in `/scripts/labels` which generated labels (not just `doc_url`, but also `severity`, in some cases) from analyser-specific knowledge.
With the introduction of the `label_tool` in the previous patch #4207, it is logical to subsume these scripts to be part of that infrastructure, which is done in this patch.

This patch is **partially** a *refactor* and **partially** a *fix*, as the aforementioned scripts were mostly 3 years old and some ceased to work since, e.g., the generation of labels for Clang-related stuff failed to do anything because the XPath expressions inside were targeting things that are not as such in the current upstream.
The URL generation for Clang SA and Clang-Tidy previously could not do anything with checkers that did not have any labels prior, which is not the case anymore.


dkrupp (approved):
LGTM

PR #4224, opened 2024-04-18
chore(config): Add automatically generated `doc_url`s and `severity`s

whisperity:
Add the URLs and severities for checkers that were missing this data, and auto-generate the severities and URLs from analyser-curated table-of-contents wherever possible. Changes created by applying #4225.

PR #4235, opened 2024-05-03
 chore(config): [clang-tidy] Update SEI-CERT Guideline mappings

whisperity:
> [!IMPORTANT]
>
> :no_entry: **Blocked** by #4224.

Ensure that checkers that target SEI CERT C & C++ guideline rules and recommendations are appropriately labelled as such: `guideline:sei-cert`, `profile:security`, and the guideline label.

In case the checker is implemented through an alias (or even multiple aliases), apply the labels to **only** the **main** checker, and strip the profile and guidelines associations from the aliases. (Due to the lack of proper alias handling in both Clang-Tidy and CodeChecker, firing all aliases would result in multiple detections of the same match.)

Ensure, in addition, that `profile:default` and `profile:sensitive` checkers are also always in the more broad superset profiles `profile:sensitive` and `profile:extreme`.

### DIFF
--- a/analyzer/tests/functional/cmdline/test_cmdline.py
+++ b/analyzer/tests/functional/cmdline/test_cmdline.py
@@ -152,7 +152,7 @@ class TestCmdline(unittest.TestCase):
                         '--guideline', 'sei-cert']
         _, out, _ = run_cmd(checkers_cmd)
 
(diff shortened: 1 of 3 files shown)


dkrupp (approved):
LGTM

PR #4236, opened 2024-05-03
[fix] Links in static HTML files should be relative

bruntib:
The links in the static HTML files generated by "CodeChecker parse -e html" should be relative, so the output folder of the above command is portable.

### DIFF
--- a/tools/report-converter/tests/unit/output/html/plist_to_html_test.py
+++ b/tools/report-converter/tests/unit/output/html/plist_to_html_test.py
@@ -186,4 +186,10 @@ class PlistToHtmlTest(unittest.TestCase):
         index_html = os.path.join(output_dir, "index.html")
 
         with open(index_html, 'r', encoding="utf-8", errors="ignore") as f:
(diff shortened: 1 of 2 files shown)


tools/report-converter/tests/unit/output/html/plist_to_html_test.py
```diff
@@ -186,4 +186,10 @@ def test_html_for_inclusion(self):
         index_html = os.path.join(output_dir, "index.html")
 
         with open(index_html, 'r', encoding="utf-8", errors="ignore") as f:
-            self.assertEqual(len(re.findall('"link": "', f.read())), 3)
+            content = f.read()
+
+            # There are 3 reports in the test file.
+            self.assertEqual(len(re.findall('"link": "', content)), 3)
+            # The links should be relative so the static HTML folder is
+            # prtable.
```
whisperity, line 194:
```suggestion
            # portable.
```

vodorok (approved):
LGTM!

PR #4242, opened 2024-05-16
Make sure that the Number of runs is displayed correctly

noraz31:
In some rare cases, when an error occurs while deleting 
runs, the number of runs to be deleted is deducted from 
the "Number of runs" value displayed in the
"Products" view, even though some runs do not get
deleted. For this reason, the number of runs could go 
negative upon the next run deletion.
This change is to fix this issue.

### DIFF
--- a/web/server/codechecker_server/api/report_server.py
+++ b/web/server/codechecker_server/api/report_server.py
@@ -3424,10 +3424,27 @@ class ThriftRequestHandler:
             # cascades and such. Deleting runs in separate transactions don't
             # exceed a potential statement timeout threshold in a DBMS.
             runs = []
+            deleted_run_cnt = 0
+
             for run in q.all():
-                runs.append(run.name)
-                session.delete(run)
-                session.commit()
+                try:
+                    runs.append(run.name)
+                    session.delete(run)
+                    session.commit()
+                    deleted_run_cnt += 1
+                except Exception as e:
+                    # TODO: Display alert on the GUI if there's an exception
+                    # TODO: Catch SQLAlchemyError instead of generic
+                    #  exception once it is confirmed that the exception is
+                    #  due to a large run deletion timeout based on server
+                    #  log warnings
+                    # This exception is handled silently because it is
+                    # expected to never occur, but there have been some rare
+                    # cases where it occurred due to underlying reasons.
+                    # Handling it silently ensures that the Number of runs
+                    # counter is not affected by the exception.
+                    LOG.warning(f"Suppressed an exception while "
+                                f"deleting run {run.name}. Error: {e}")
 
             session.close()
 
@@ -3436,7 +3453,7 @@ class ThriftRequestHandler:
 
         # Decrement the number of runs but do not update the latest storage
         # date.
-        self._set_run_data_for_curr_product(-1 * len(runs))
+        self._set_run_data_for_curr_product(-1 * deleted_run_cnt)
 
         # Remove unused comments and unused analysis info from the database.
         # Originally db_cleanup.remove_unused_data() was used here which


web/server/codechecker_server/api/report_server.py
```diff
@@ -3424,10 +3424,22 @@ def removeRun(self, run_id, run_filter):
...
+                    session.commit()
+                    run_no += 1
+                except Exception:
+                    # TODO: Display alert on the GUI if there's an exception
+                    # This exception is handled silently because it is
+                    # expected to never occur, but there have been some rare
+                    # cases where it occurred due to underlying reasons.
+                    # Handling it silently ensures that the Number of runs
+                    # counter is not affected by the exception.
+                    pass
```
Szelethus, line 3442: I appreciate the detailed explanation; silently suppressing an exception can lead to a lot of headache, so whenever we do it, we better make sure we need to.

It just so happens that in my view, we could just emit a simple LOG message _and then_ be done with it. The error occuring rarely is an argument for, not against suppressing this silently. That way, we would have at least SOME idea that something went wrong in here.
```suggestion
                except Exception as e:
                    # TODO: Display alert on the GUI if there's an exception
                    # This exception is handled silently because it is
                    # expected to never occur, but there have been some rare
                    # cases where it occurred due to underlying reasons.
                    # Handling it silently ensures that the Number of runs
                    # counter is not affected by the exception.
                    LOG.warning(f"Suppressed an exception here: {e}")
                    pass
```

Szelethus (changes requested):
I think silently suppressing the generic exception object is never a good idea. Maybe there are exceptions (hehe), but this just isn't it in my view.

Szelethus (approved):
I'd be thrilled to learn what kind of exceptions can come out of here, but also submit to the fact we can't just blindly release them first. LGTM.

web/server/codechecker_server/api/report_server.py
```diff
@@ -3424,10 +3424,24 @@ def removeRun(self, run_id, run_filter):
             # cascades and such. Deleting runs in separate transactions don't
             # exceed a potential statement timeout threshold in a DBMS.
             runs = []
+            run_no = 0
```
cservakt, line 3427: I'm not sure `run_no` is a clear variable name. Maybe deleted_run_cnt or something else would be better.

web/server/codechecker_server/api/report_server.py
```diff
@@ -3424,10 +3424,24 @@ def removeRun(self, run_id, run_filter):
...
+                except Exception as e:
+                    # TODO: Display alert on the GUI if there's an exception
+                    # This exception is handled silently because it is
+                    # expected to never occur, but there have been some rare
+                    # cases where it occurred due to underlying reasons.
+                    # Handling it silently ensures that the Number of runs
+                    # counter is not affected by the exception.
+                    LOG.warning(f"Suppressed an exception while "
+                                f"deleting run {run.name}. Error: {e}")
+                    pass
```
bruntib, line 3444: Unnecessary `pass` statement.

web/server/codechecker_server/api/report_server.py
```diff
@@ -3424,10 +3424,24 @@ def removeRun(self, run_id, run_filter):
...
+                    session.delete(run)
+                    session.commit()
+                    run_no += 1
+                except Exception as e:
+                    # TODO: Display alert on the GUI if there's an exception
+                    # This exception is handled silently because it is
+                    # expected to never occur, but there have been some rare
+                    # cases where it occurred due to underlying reasons.
+                    # Handling it silently ensures that the Number of runs
+                    # counter is not affected by the exception.
```
bruntib, line 3441: One reason of the exception can be an SQL statement timeout while a too large run is being deleted. We should include another TODO here which suggests the catching of `SQLAlchemyError` instead of the generic `Exception` when we can make this sure based on the warnings messages below, in the server logs.

bruntib (changes requested):

bruntib (approved):

PR #4244, opened 2024-05-16
Showing closed reports

cservakt:
[feat] Showing closed reports on Checker Coverage tab

It is an additional PR for `Checker Coverage` tab to list closed reports not just outstandings. The number of closed report can be clickable for the user if it is not zero.

Achieving this, it is necessary to add a `report status` filter to the report filter section that has two options: `OUTSTANDING` and `CLOSED`. A report is outstanding when its review status is unreviewed/confirmed and its detaction status is new/unresolved/reopened. When the new filter is set, the review and the detection status filters are not taken into account.


bruntib:
The review status and detection status filter options could be disabled in the GUI when report status is selected. Otherwise this patch looks good.

dkrupp:
1) The "Report Status" filter should behave like the other filters

> When the new filter is set, the review and the detection status filters are not taken into account.
I think that the "Report Status" filter should behave the same way as the other filters. 
A differnet beheviour for this specific filter would be very disturbing.

So the "Review Status", "Detection Status" should be an additional restricting filter. For example one might want to see the "closed" report whose review status is "false positive".

2) The link from the coverage statistics should select teh "closed reports" only
When the user clicks on the "closed reports" in the Checker coverage statistics tab, the filter should be set to show the "closed reports" only and the "Review Status" and "Detection Status" filters should be left empty.

3) the new default filter should be the "oustanding reports"
The new default filter should also be changed to show the "Outstanding" reports and the review status, detection status filters should be left empty.

We should orient the users towards using the Report Status filter instead of the "Review Status" and "Detection Status" filters. They should use these latter filters only if they want a detailed selection by the review/detection status.

cservakt:
> 1. **The "Report Status" filter should behave like the other filters**
> 
> 
> > When the new filter is set, the review and the detection status filters are not taken into account.
> > I think that the "Report Status" filter should behave the same way as the other filters.
> > A differnet beheviour for this specific filter would be very disturbing.
> 
> So the "Review Status", "Detection Status" should be an additional restricting filter. For example one might want to see the "closed" report whose review status is "false positive".
> 
>     2. **The link from the coverage statistics should select teh "closed reports" only**
>        When the user clicks on the "closed reports" in the Checker coverage statistics tab, the filter should be set to show the "closed reports" only and the "Review Status" and "Detection Status" filters should be left empty.
> 
>     3. **the new default filter should be the "oustanding reports"**
>        The new default filter should also be changed to show the "Outstanding" reports and the review status, detection status filters should be left empty.
> 
> 
> We should orient the users towards using the Report Status filter instead of the "Review Status" and "Detection Status" filters. They should use these latter filters only if they want a detailed selection by the review/detection status.

Thanks for the suggestions! Now the `report filter` works as expected. The `report status` filter does not exclude the `review` and `detection status` filters.

Clicking to the number of the `closed` (or `outstanding`) reports, the url and the report filter are set correctly.

Changing the default filter for the `reports` and the `statistics` menu should be a different patch where we reconstruct the entire `report filter`.

dkrupp (changes requested):
When testing I found the following minor faults:

1) Please fix the help message of the report status field. Remove the text that the "review status and detection status is not prevalent"

2) The Outstanding report number in the filter menu is not correctly counted. It always shows 0.

Otherwise looks good!

cservakt:
> When testing I found the following minor faults:
> 
>     1. Please fix the help message of the report status field. Remove the text that the "review status and detection status is not prevalent"
> 
>     2. The Outstanding report number in the filter menu is not correctly counted. It always shows 0.
> 
> 
> Otherwise looks good!

I have fixed these minor bugs.

bruntib (approved):

dkrupp (approved):

PR #4245, opened 2024-05-22
[script] Script for querying all reports

bruntib:
This script can query all reports for all products from a CodeChecker server.


scripts/result_listing/query_all_reports.py
```diff
@@ -0,0 +1,124 @@
...
+                'CodeChecker', 'cmd', 'runs',
+                '--url', f'{args.url}/{product}',
+                '-o', 'json'],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE
+        ) as proc:
+            runs = list(__get_keys_from_list(proc.stdout.read()))
+
+            with Pool(args.jobs) as p:
+                p.map(result_getter(args), map(f, runs))
```
cservakt, line 118: I am just wondering wouldn't be better to have this section in a helper function.

scripts/result_listing/compare_results.py
```diff
@@ -0,0 +1,57 @@
...
+            found_mismatch = True
+
+    for report in result2:
+        if report not in result1:
+            print("Report not found in result1:")
+            print(json.dumps(report, indent=4))
+            found_mismatch = True
+
+    args.result1.close()
+    args.result2.close()
```
cservakt, line 51: I think there is a code duplication here. Could you create a function to avoid it?

cservakt (approved):
LGTM

PR #4247, opened 2024-05-27
[feature] Adding guideline label to Checker Coverage tab

cservakt:
It's an additional PR to integrate the guideline column into checker coverage statistics table. Showing the guideline rule for those checks that cover specific rules e.g. sei-cert or misra-c.

### DIFF
--- a/web/server/vue-cli/src/components/Statistics/BaseStatisticsTable.vue
+++ b/web/server/vue-cli/src/components/Statistics/BaseStatisticsTable.vue
@@ -326,6 +326,26 @@
       </span>
     </template>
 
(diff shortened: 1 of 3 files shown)


dkrupp (changes requested):
This patch looks good.
I would recommend the following changes:
- Please show the guideline associations in the following format: sei-cert:str51-cpp
- if there are more htan one guideline rule per checker, list them one in a row
- The column name is misspelled: Guidline -> Guideline

dkrupp (approved):
The sei cert rule listing works works well in teh GUI even if there are multiple associated rules per one checker.

The CSV generation also works fine.

The PR looks goot to me.

PR #4275, opened 2024-06-24
feat(script): Verify checker labelling invariants (e.g., `guideline:X` ⇔ `X:rule`, `default` ⊆ `sensitive` ⊆ `extreme`) with the label-tool

whisperity:
There were multiple occasions in which the labelling of the checkers was lacking some serious invariants, such that enabling `sensitive` should also enable `default`, or that a checker with
`guideline:sei-cert` has to be in `profile:security`, or that `guideline:sei-cert` should not appear without a `sei-cert:` label and vice versa.

Added a tooling to the label tool framework that verifies (and wherever possible, such as with the **⊆** relations, auto-fixes) this.


whisperity:
Added a rule to tear out `alpha.` checkers from the normal profiles. We shall be able to catch things like #4284 automatically in the future.

PR #4281, opened 2024-06-29
[analyzer] Adds -fno-freestanding to ignored GCC compiler flags

ArchieAtkinson:
`-ffreestanding` and `-fno-freestanding` are common in embedded systems but clang does not support it. This PR adds them both to the ignored GCC compiler flags list.

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -94,7 +94,7 @@ IGNORED_OPTIONS_GCC = [
     '-fno-delete-null-pointer-checks',
     '-fno-defer-pop',
     '-fno-extended-identifiers',
-    '-fno-freestanding'
+    '-fno-freestanding',
     '-fno-jump-table',
     '-fno-keep-inline-dllexport'
     '-fno-keep-static-consts',
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -94,6 +94,7 @@ IGNORED_OPTIONS_GCC = [
     '-fno-delete-null-pointer-checks',
     '-fno-defer-pop',
     '-fno-extended-identifiers',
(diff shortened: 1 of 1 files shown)


bruntib:
Is it possible that only an old Clang version doesn't support this flag? I can see this in the current Clang docs: https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-ffreestanding

ArchieAtkinson:
> Is it possible that only an old Clang version doesn't support this flag? I can see this in the current Clang docs: https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-ffreestanding

Odd, I'm running 17.0.6 and I get this:
```
[INFO 2024-07-04 18:39] - Starting static analysis ...
[ERROR 2024-07-04 18:39] - Analyzing main.cpp with clang-tidy  failed!
[ERROR 2024-07-04 18:39] - 
error: unknown argument: '-fno-freestanding' [clang-diagnostic-error]

[ERROR 2024-07-04 18:39] - 
```

ArchieAtkinson:
> Is it possible that only an old Clang version doesn't support this flag? I can see this in the current Clang docs: https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-ffreestanding

Updated to only ignore -fno-freestanding, which isn't supported by clang ([Godbolt](https://godbolt.org/z/dMGGrKdzs)) while -freestanding is ([Godbolt](https://godbolt.org/z/W5Wv47eGM))

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -94,6 +94,7 @@
     '-fno-delete-null-pointer-checks',
     '-fno-defer-pop',
     '-fno-extended-identifiers',
+    '-fno-freestanding'
```
bruntib, line 97:
```suggestion
    '-fno-freestanding',
```
Thanks for the investigation of this flag. Make sure to add a comma to the end of line, and we can merge it. Thank you!

bruntib (approved):

PR #4291, opened 2024-07-09
chore(config): Apply invariant fixes from `label-tool`

whisperity:
Apply the automatic fixes of invariants such as _"`profile:sensitive` and `profile:extreme` should be present if `profile:default` is present"_ and _"`guideline:sei-cert` should be present if a `sei-cert:RULE-01`-like label is present"_ to the config files, as emitted by #4275.

### DIFF
--- a/config/labels/analyzers/clangsa.json
+++ b/config/labels/analyzers/clangsa.json
@@ -875,7 +875,6 @@
       "profile:default",
       "profile:extreme",
       "profile:security",
-      "profile:security",
       "profile:sensitive",
       "sei-cert:env31-c",
       "sei-cert:env34-c",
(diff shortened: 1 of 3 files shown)


config/labels/analyzers/clangsa.json
```diff
@@ -1077,6 +1076,7 @@
     ],
     "unix.cstring.NullArg": [
       "doc_url:https://clang.llvm.org/docs/analyzer/checkers.html#unix-cstring-nullarg-c",
+      "guideline:sei-cert",
```
whisperity, line 1079: This is the only behavioural change, but if someone uses `analyse --disable default --enable guideline:sei-cert` (or however it should be specified...) then they will start seeing this checker, which changes the potentially reported set of bugs.

whisperity (commented):
@bruntib @dkrupp @vodorok I'm not sure if this should go into 6.25 only, or 6.24.1 as well.

bruntib (approved):

PR #4294, opened 2024-07-12
[Fix] Report sorting in unique mode

cservakt:
Fixing report sorting on unique mode. After modifying the DB schema in #4089 PR, the unique mode query of getRunResults endpoint has been changed, therefore, the report sorting is not working properly.

Now, the unique mode query is redesigned and it use row_number() function to filter unique reports correctly. Where clause is also modified. It is getting rid of report annotation filter. Filtering annotation remains in having clause.

### DIFF
--- a/web/server/codechecker_server/api/report_server.py
+++ b/web/server/codechecker_server/api/report_server.py
@@ -20,7 +20,7 @@ import time
 import zlib
 
 from copy import deepcopy
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict, defaultdict, namedtuple
 from datetime import datetime, timedelta
 from typing import Any, Dict, List, Optional, Set, Tuple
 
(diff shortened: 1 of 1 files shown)


web/server/codechecker_server/api/report_server.py
```diff
@@ -1962,45 +1962,29 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
                     ReportAnnotations.value)])).label(f"annotation_{col}")
 
             if report_filter.isUnique:
+                # A report annotation filter cannot set in WHERE clause if we
```
bruntib, line 1965:
```suggestion
                # A report annotation filter cannot be set in WHERE clause if we
```

web/server/codechecker_server/api/report_server.py
```diff
@@ -1962,45 +1962,29 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
...
+                # use annotation parameters in aggregate functions to create
+                # pivot table. Instead filtering report annotations in WHERE
```
bruntib, line 1967:
```suggestion
                # a pivot table. Instead of filtering report annotations in WHERE
```

web/server/codechecker_server/api/report_server.py
```diff
@@ -1962,45 +1962,29 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
...
+                # clause, we should use HAVING clause only for filtering
+                # aggregate functions.
+                # TODO: Fixing report annotation filter in every report server
+                # enpoint function.
```
bruntib, line 1971:
```suggestion
                # endpoint function.
```

web/server/codechecker_server/api/report_server.py
```diff
@@ -1962,45 +1962,29 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
...
+                report_filter_annotations = report_filter.annotations
```
bruntib, line 1972:
```suggestion
                annotations_backup = report_filter.annotations
```
Maybe such a name could cause less misunderstanding compared to that one-character difference.

bruntib (approved):

PR #4299, opened 2024-07-22
[feat] Print statistics after analysis through Makefile

bruntib:
"CodeChecker analyze --makefile ..." is generating a Makefile that contains analyzer commands. With "make" command one can execute analysis using this generated Makefile.

This commit is about printing some statistics about the analysis based on the Makefile. The statistics contain the number of successful and failed analysis.


bruntib:
This will be solved with another implementation.

PR #4300, opened 2024-07-23
[fix] Don't capture cc1 by the logger.

bruntib:
After make-4.3 the compilers are executed in a way that cc1 binary is also captured by the logger. This accomplishes only a partial phase of the compilation process and this shouldn't be captured. As a solution we use the fact that having a slash (/) in a compiler's name in CC_LOGGER_GCC_LIKE makes a suffix search.


dkrupp (approved):
LGTM

PR #4307, opened 2024-08-01
[fix] Don't enable checkers by suffix

bruntib:
Currently, checkers can be enabled by checker name prefix. For example, "--enable unix" enables all checkers starting with "unix". However, checkers shouldn't be enabled by suffix match: "--enable Malloc" shouldn't enable "unix.Malloc" checker.

The source of the confusion was that having "unix.Stream" checker in the "default" profile also enabled "alpha.unix.Stream", which is not an intended behavior.


dkrupp (approved):
I did not find any mention in the documentation of postfix matching checkers in the enable list, so the docs dont need to be updated.  LGTM.

PR #4311, opened 2024-08-14
[cmd] Checker name prefixes are meant along separator characters

bruntib:
Checkers can be enabled by groups, e.g. unix.cstring enables all checkers starting with this prefix. However, unix.cs is not a checker group so this shouldn't be allowed. In this case an error message is printed and analysis doesn't start.

### DIFF
--- a/analyzer/tests/unit/test_checker_handling.py
+++ b/analyzer/tests/unit/test_checker_handling.py
@@ -264,6 +264,27 @@ class CheckerHandlingClangSATest(unittest.TestCase):
         self.assertTrue(all_with_status(CheckerState.ENABLED)
                         (cfg_handler.checks(), low_severity))
 
+        # Enable checkers with a checker group prefix.
+        cfg_handler = ClangSA.construct_config_handler(args)
+        cfg_handler.initialize_checkers(checkers,
+                                        [('default', False),
+                                         ('cplusplus.NewDelete', True)])
+        self.assertTrue(
+            all_with_status(CheckerState.ENABLED)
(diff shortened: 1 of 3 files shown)


analyzer/tests/unit/test_checker_handling.py
```diff
@@ -264,6 +264,18 @@ def f(checks, checkers):
         self.assertTrue(all_with_status(CheckerState.ENABLED)
                         (cfg_handler.checks(), low_severity))
 
+        # Enable checkers with a checker group prefix.
```
dkrupp, line 267: this test is enabling one concrete checker alpha.security.ArrayBound and tests that alpha.security.ArrayBoundv2 should not be enabled. It is not testing "checker group prefix"

Could you please add another test case which enables "alpha.security" group. Then alpha.security.ArrayBound and also alpha.security.ArrayBoundv2 should be enabled.

dkrupp (changes requested):
Please add another testcase for testing enabling checker groups.

NagyDonat:
Instead of `alpha.security.ArrayBound` and `alpha.security.ArrayBoundV2`, consider using the checkers `cplusplus.NewDelete` and `cplusplus.NewDeleteLeaks` because their names are more stable. We will hopefully rename/rearrange the bounds checkers within a year, while these are already stable checkers and their names won't change.

bruntib:
@dkrupp, @NagyDonat Thanks for the comments! I fixed the tests accordingly.

dkrupp (approved):
LGTM

PR #4315, opened 2024-08-16
[feat] Implicit include paths added with -idirafter

bruntib:
GCC has implicit include paths that are forwarded to Clang. Until now these paths were added with -isystem flag, but sometimes the priority of this is too high: https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html The implicit include paths should be searched the last, so -isystem has been changed to -idirafter.
In case of potential backward compatibility break
--add-gcc-include-dirs-with-isystem has been introduced for "CodeChecker analyze" command.


dkrupp (approved):
I tested this on both tmux and postgres and the idirafter/isystem parameters were appended correctly to the analysis commands. Both analysis were successful. So it will make a difference only in fewer cases.

LGTM

PR #4317, opened 2024-08-17
feat(server): Asynchronous server-side background task execution

whisperity:
> [!IMPORTANT]
>
> **⛔ Blocked** by #4170.
>
> This is patch **2** of the _Asynchronous Store Protocol_ (#3672).

This patch implements the whole support ecosystem for **server-side background tasks**, in order to help lessen the load, occupancy, and potentially detrimental blocking behaviour on API handler processes, occupied by long-running operations in the web-server, such as _`massStoreRun()`_.

## API workers vs. background/task workers

In this patch, the processing model of the `CodeChecker server` is extended with the concept of **background workers**. These special processes deal with consuming _tasks_ off of the server's queue, and executing them. The previous clone processes of a _server_ are now and henceforth termed _API workers_, and they do the same as prior: respond to Thrift RPC calls.

The server will start **`background_worker_processes`** number of background processes, which is, by default, the same number as _`worker_processes`_, the number of API handlers. (Which defaults to `$(nproc)`.)

### Self-healing against dead children

A **fix** is also implemented in this patch which is the result of a side effect that stemmed from reviewing and reworking both the threading and the signalling model of a `CodeChecker server` process tree. Previously, exceptions escaping from an _API worker_ process (or it getting OOM-killed, etc.) would result in an initially large number of workers to die off one by one, leaving the server with a monotonously descending number of workers to dispatch RPC requests to. In this patch, the server is scripted to handle `SIGCHLD` and respawn dead child processes (both _API_ and _Task_) when given the opportunity.

### Graceful server shutdown

The server's life cycle and signal handling logic refactoring also includes changes which makes the server process(es) more graceful in their termination, which is needed to accurately clean-up pending tasks (see later).

In case a brutal and unclean server shutdown is needed, following this patch, the `SIGINT` or `SIGTERM` signal must be sent **twice** to the server.
(Alternatively, just use `SIGKILL`! 😉)

## Task management

A **Task** is represented by two things in strict co-existence: a lightweight, `pickle`-able implementation of the task's execution logic (subclassed from `AbstractTask` and overriding `_implementation()`); and a corresponding `BackgroundTask` entity in the database. These records reside in the _CONFIGURATION_ database, and the task information is server-, or service-wide, shared across all products.

The database-synchronised record contains several human-readable metadata about the task, such as a one-line `summary` field and the larger, log-like `comments` column, with several timestamps recorded for the crucial moments during the task's execution.

The most important flag is a task's _status_, which can be: _`ALLOCATED`_, _`ENQUEUED`_, _`RUNNING`_, _`COMPLETED`_, _`FAILED`_, _`CANCELLED`_, and _`DROPPED`_.

### Normal path

The life cycle of a task is as follows:

1. A **`Task`** is created as the result of an API request (executed in the _API worker_ process).
    a. (Currently, for the sake of both testing and demonstration, only the _`createDummyTask()`_ API, and only in a test context, can create a task.)
2. A unique identifier, termed a **task token** is _`ALLOCATED`_, together with the `BackgroundTask` database entity.
3. The _API worker_ **pushes** the `Task` object into a shared, synchronised queue within the server's memory space.
   At this point, the task is considered _`ENQUEUED`_.
    a. _`AbstractTask`_ subclasses **MUST** be `pickle`-able and reasonably small.
    b. The library offers means to store additional large data on the file system, in a temporary directory specific to the task.
4. The **task token** is returned to the user via the RPC API call, and the _API worker_ is free too respond to other requests.
5. In a loop with some frequency, the user exercises the _`getTaskInfo()`_ API (executed in the context of any _API worker_ process, synchronised over the database) to query whether the task was completed, **if** the user wishes to receive this information.
6. A _background worker_ process (which is a separate context, but executed in the same process tree as the server's API requests — there is **NO** publishing of tasks _between_ machines!) wakes up to the enqueuement, and pops the `Task` object from the queue.
   After some bookkeeping, the task will be _`RUNNING`_.
7. **`MyTaskClass::_implementation()`** is called, which executes the task's primary business logic.
8. If **`_implementation()`** returns without a failure, the task will be considered successfully _`COMPLETED`_.
   Any exception escaping from the method will set the task to _`FAILED`_, and exception information is logged into the `BackgroundTask.comments` column of the database.
   Together, these two are the _"Normal termination states."_.
9. Eventually, the user will receive (if waiting for this) the fact that the task had completed.
10. The _background worker_ process goes back to sleep until a new `Task` is available from the queue.

### Abnormal path 1: admin cancellation

At any point following _`ALLOCATED`_ status, but most likely in the _`ENQUEUED`_ and _`RUNNING`_ statuses, a **`SUPERUSER`** may issue a _`cancelTask()`_ order.
This will set `BackgroundTask.cancel_flag`, and the task is expected (although not required!) to poll its own _`should_cancel()`_ status internally in checkpoints, and terminate gracefully to this request. This is done by **`_implementation()`** exiting by raising a **`TaskCancelHonoured`** exception.
(If the task does not raise one, it will be allowed to conclude normally, or fail in some other manner.
Tasks cancelled gracefully will have the _`CANCELLED`_ status.

For example, a background task that performs an action over a set of input files generally should be implemented like this:

```py3
def _implementation(tm: TaskManager):
    for file in INPUTS:
      if tm.should_cancel(self):
          ROLLBACK()
          raise TaskCancelHonoured(self)

      DO_LOGIC(file)
```

### Abnormal path 2: server shutdown

Alternatively, at any point in this life cycle, the server might receive the command to terminate itself (kill signals `SIGINT`, `SIGTERM`; alternatively caused by `CodeChecker server --stop`). Following the termination of _API workers_, the _background workers_ will also shut down one by one.
At this point, the default behaviour is to cause a special _cancel event_ which tasks currently _`RUNNING`_ may still gracefully honour, as-if it was a `SUPERUSER`'s single-task cancel request. All other tasks that have not started executing yet and are in the _`ALLOCATED`_ or _`ENQUEUED`_ status will never start.

All tasks not in a _normal termination state_ will be set to the _`DROPPED`_ status, with the `comments` field containing a log about the specifics of in which state the task was dropped, and why. (Together, _`CANCELLED`_ and _`DROPPED`_ are the _"abnormal termination states"_, indicating that the task terminated due to some external influence.)

## Task querying

The **`getTaskInfo()`** API, querying the status of **one** task, is available to the user who caused the task to spawn, the _`PRODUCT_ADMIN`s_ of the product associated with the task (if any), and `SUPERUSER`s.

The **`getTasks()`** API, which queries multiple tasks based on a _filter_ set, is available only to _`PRODUCT_ADMIN`s_ (results restricted to the products they are admin thereof) and `SUPERUSER`s (unrestricted).
Just about anything that is available as information in the database about a task can be queried.

## `--machine-id`

Unfortunately, servers don't always terminate gracefully (cue the aforementioned `SIGKILL`, but also the container, VM, or the host machine could simply die during execution, in ways the server is not able to handle). Because tasks are not shared across server processes, and there are crucial bits of information in the now dead process's memory which would have been needed to execute the task, a server later restarting in place of a previously dead one should be able to identify which tasks its "predecessor" left behind without clean-up.

This is achieved by storing the running computer's identifier, configurable via `CodeChecker server --machine-id`, as an additional piece of information for each task. By default, the machine ID is constructed from `gethostname():portnumber`, e.g., `cc-server:80`.

In containerised environments, relying on `gethostname()` may not be entirely stable!
For example, Docker exposes the first 12 digits of the container's unique hash as the _"hostname"_ of the insides of the container. If the container is started with `--restart always` or `--restart unless-stopped`, then this is fine, however, more advanced systems, such as _Docker swarm_ will **create a new container** in case the old one died (!), resulting in a new value of `gethostname()`.

In such environments, service administrators must pay additional caution and configure their instances by setting `--machine-id` for subsequent executions of the "same" server accordingly. If a server with machine ID **`M`** starts up (usually after a container or "system" restart), it will set every task not in any "termination states" and associated with machine ID **`M`** to the _`DROPPED`_ status (with an appropriately formatted comment accompanying), signifying that the _previous instance_ "dropped" these tasks, but had no chance of recording this fact.


web/server/codechecker_server/task_executors/task_manager.py
```diff
@@ -0,0 +1,229 @@
...
+                            f"{MAX_TOKEN_RANDOM_RETRIES} retries!") from ie
+
+    def create_task_data(self, token: str) -> Path:
+        """
+        Creates a temporary directory which is **NOT** cleaned up
+        automatically, and suitable for putting arbitrary files underneath
+        to communicate large inputs (that should not be put in the `Queue`)
+        to the `execute` method of an `AbstractTask`.
+        """
+        task_tmp_root = Path(tempfile.gettempdir()) / "codechecker_tasks" \
```
bruntib, line 102: It's not a real question, I'm just thinking loudly: we put things under `/tmp`, because it's server side, right? Only the client is using the workspace directory instead of `/tmp` for temporary files, right?

web/server/codechecker_server/task_executors/abstract_task.py
```diff
@@ -0,0 +1,183 @@
...
+        if not self._data_path:
+            return
+
+        try:
+            shutil.rmtree(self._data_path)
+        except Exception as ex:
+            LOG.warning("Failed to remove background task's data_dir at "
+                        "'%s':\n%s", self.data_path, str(ex))
+
+    def _implementation(self, _task_manager: "TaskManager") -> None:
```
bruntib, line 89: Shouldn't it get an `@abstractmethod` annotation?

web/api/tasks.thrift
```diff
@@ -0,0 +1,162 @@
...
+struct TaskFilter {
+   1: list<string>                   tokens,
+   2: list<string>                   machineIDs,
+   3: list<string>                   kinds,
+   4: list<TaskStatus>               statuses,
+  // If empty, it means "all", including those of no username.
+   5: list<string>                   usernames,
+  // If True, it means filter for **only** "no username".
+  // Can not be set together with a non-empty "usernames".
+   6: bool                           filterForNoUsername,
```
bruntib, line 71: How about this?
```.thrift
5: optional list<string> usernames
```
In that case `None` and `[]` could be distinguished and `filterForNoUsername` wouldn't be needed. This technique is used for some `ReportFilter` fields, too. I feel it's easier to make a mistake when the client's creator doesn't know that these two fields belong together.
This comment applies for `filterForNoProductID`, too.

web/server/codechecker_server/api/tasks.py
```diff
@@ -0,0 +1,391 @@
...
+                raise RequestFailed(ErrorCode.GENERAL,
+                                    f"Task '{token}' does not exist!")
+
+            has_right_to_query_status: bool = False
+            should_set_consumed_flag: bool = False
+
+            if db_task.username == self._get_username():
+                has_right_to_query_status = True
+                should_set_consumed_flag = db_task.is_in_terminated_state
+            elif db_task.product_id is not None:
```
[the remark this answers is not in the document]
whisperity: Because different workflows are executed. You can only query the status of the task if:

1. You are the user who was responsible for creating the task. (In this case, but only for terminated tasks, the `consumed` flag is set to true.)
2. The task is associated with a product (this might not always be the case!) and you are an admin of said product. (In this case, the `consumed` flag is left intact.)
3. You are a superuser. (In this case, the `consumed` flag is left intact.)

If this was not an `elif`, a product admin or superuser querying their own task would leave the `consumed` flag intact.

web/server/codechecker_server/task_executors/task_manager.py
```diff
@@ -0,0 +1,229 @@
...
+    def _get_task_record(self, task_obj: "AbstractTask") -> DBTask:
+        """
+        Retrieves the `DBTask` for the task identified by `task_obj`.
+
+        This class should not be mutated, only the fields queried.
+        """
+        with DBSession(self._database_factory) as session:
+            try:
+                db_task = session.query(DBTask).get(task_obj.token)
+                session.expunge(db_task)
```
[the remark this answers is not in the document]
whisperity: Removes the `DBTask` object from the `Session`'s internal cache. Results in clearing the `__sa_state` member, tearing off the connection between the Python object and the ORM-handled "view" of it. This ensures that callers of `get_task_record()`, even if they capture a reference to what is being returned here, are NOT allowed to mutate the state of the object in a way that the database state is also mutated.

(In order to safely mutate a `DBTask` entity, `_mutate_task_record` exists.)

web/server/codechecker_server/task_executors/task_manager.py
whisperity: Yes. This is all server-side. Just a temporary directory that's more identifiable than a truly "random" temporary directory for the task's tmp data (like the unzip of a mass-store-run). Because servers can fail in a way that the cleaning of temp directories are not executed, as we are no longer using the tempfile... context managers here.

web/server/codechecker_server/task_executors/abstract_task.py
whisperity: Unfortunately, `@abstractmethod` isn't a built-in, but coming from the `abc` library, which makes the object not pickleable.

Also, methods that raise `NotImplementedError()` are type-checked by PyRight and there is a linter warning in the editor that your class is not proper.

web/api/tasks.thrift
whisperity: Alright, I will try reworking the implementation for this...

whisperity: Okay, I implemented the requested changes!

bruntib:
This patch has been merged in this PR after resolving merge conflicts: https://github.com/Ericsson/codechecker/pull/4603
@whisperity Thank you for this great work!

PR #4319, opened 2024-08-21
[fix] fix compare_results.py sciprt

bruntib:
The list.sort() function doesn't return a sorted list but sorts inplace.


noraz31 (approved):
LGTM

PR #4330, opened 2024-08-28
[fix] Missing analyzer error

cservakt:
`CodeChecker analyzers` gives python error for the user if a supported analyzer is missing. In addition, CodeChecker `analyze` and `check` should have provided an error and sys.exit when the specified (with `--analyzers`) analyzer was missing.

With this PR, `CodeChecker analyzers` can give a warning message with `NOT FOUND` status if a supported analyzer can't be found on the machine.

In case of explicitly defined missing analyzer, the `analyze` and `check` command will be failed and will give warrning and error message for the user.


bruntib:
How about having `None` as default value to `args.analyzers`? That could simplify the implementation, because that way we can distinguish these cases:
- `None`: All analyzers are enabled implicitly. No error if some of them is not available.
- `[<all_analyzers_listed_here...>]`: All analyzers are enabled explicitly. Error if some of them is not available.

cservakt:
> distinguish

Thank you for the suggestion. I also thought about it to use None as default for args.

PR #4333, opened 2024-08-30
Bump CI to ubuntu-24.04

pdgendt:
CI fails to install g++-13 on older ubuntu versions.


pdgendt:
Clearly it's not as simple as this. How does the `pypi` release get done currently?

bruntib:
Hi,
Thanks for the patch, however, we plan to support 20.04 at least until April next year which is the maintenance deadline of this Ubuntu version.
We deployed 6.24.0 to Pypi successfully, and we'll deploy 6.24.1 hopefully soon, too.

PR #4337, opened 2024-09-05
Environment initialization for binaries

dkrupp:
CodeChecker calls various binaries during analysis: clang, clang-tidy,
clang-extdef-mapping etc.
If the binary is delivered within the CodeChecker package,
LD_LIBRARY_PATH needs to be extended with it.

Otherwise the environment of the caller shell should be used
for excuting binaries.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzer.py
@@ -355,7 +355,8 @@ def perform_analysis(args, skip_handlers, rs_handler: ReviewStatusHandler,
     end_time = time.time()
     LOG.info("Analysis length: %s sec.", end_time - start_time)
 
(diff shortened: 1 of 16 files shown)


bruntib:
Great improvement, I like the idea of not trusting the callers to provide the environment to these functions.
Some further locations should apply this approach:
```
ctu_autodetection.py:78
  invoke_binary_checked
  
clangtidy/analyzer.py:145
  get_warnings
  
cppcheck/analyzer.py:243
  get_analyzer_checkers
  
gcc/analyzer.py:89
  get_analyzer_checkers

codechecker_analyzer/cmd/analyzers.py:123,126
  main
```

dkrupp:
Thanks for the review. I added the environment initialization to the additional places.

bruntib (approved):

PR #4351, opened 2024-09-18
Github Actions: stop previous jobs when a new one was pushed

stt08:
The following modification will cancel previous tasks/runs when a new one was pushed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,10 @@ on: [push, pull_request]
 
 permissions: read-all
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -5,6 +5,10 @@ on: [push, pull_request]
 
 permissions: read-all
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
oleg-derevenetz, line 9: `github.ref` for pull requests [has the form](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) of `refs/pull/<pr_number>/merge`, i.e. it already includes the pull request number.

stt08: changes applied

feyruzb (approved):
I tried the changes and they work as expected

PR #4353, opened 2024-09-23
adding codechecker analysis workflow

dkrupp:


bruntib:
I converted this PR to a draft, because this is not a PR to merge directly. This will be the basis of a workflow which uses CodeChecker for analyzing PRs.

PR #4357, opened 2024-10-03
Resolve paths when blaming files

tomhughes:
Both `git check-ignore` and `git blame` will error if run on a symbolic link, so resolve the path before running them.


cservakt (approved):
LGTM

bruntib (approved):

dkrupp (approved):
LGTM

vodorok (approved):
LGTM

bruntib:
@tomhughes Thanks for the patch. Could you, please, force-push a dummy change, so the test workflow is triggered? For some reason they don't start. Thank you!

tomhughes:
I've forced a run now - there is one failure but it doesn't look related.

PR #4359, opened 2024-10-10
[fix] Cppcheck unknown C/C++ standard

bruntib:
Cppcheck tool is failing from version 2.15 if "--std=" flag is given an unknown value. For example, "--std=c++0x" is not known, but "--std=c++11" is. Before cppcheck-2.15 the latest known standard version was silently used.

In this patch the unknown standard versions are mapped to the known ones.


dkrupp (approved):
LGTM

PR #4363, opened 2024-10-14
[fix] Better SQL SELECT instead of a timeout query

bruntib:
This query runs when filtering by files or components with "anywhere on bugpath" option. In this case the following query was generated:

`SELECT <columns> FROM reports WHERE reports.id IN (id1, id2, ...);`

The ID list at "IN" block can be so huges that it eats up all the memory and times out the query. This ID list is now replaced with a nested select:

`SELECT <columns> FROM reports WHERE reports.id IN
  (SELECT report_id FROM <some tables>);`

### DIFF
--- a/web/server/codechecker_server/api/report_server.py
+++ b/web/server/codechecker_server/api/report_server.py
@@ -534,39 +534,31 @@ def get_source_component_file_query(
 
 def get_reports_by_bugpath_filter(session, file_filter_q) -> Set[int]:
     """
-    This function returns a set of report IDs that are related to any file
+    This function returns a query for report IDs that are related to any file
     described by the query in the second parameter, either because their bug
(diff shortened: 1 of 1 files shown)


cservakt:
I am not sure using `in` operator is the best solution. What if we join all the tables to the reports like this:

`SELECT 
	<columns> 
FROM reports 
INNER JOIN files
	ON reports.file_id = files.id
LEFT JOIN bug_path_events
	ON bug_path_events.file_id = files.id
LEFT JOIN bug_report_point
	ON bug_report_point.file_id = files.i
LEFT JOIN extended_report_data
	ON extended_report_data.file_id = files.id
WHERE
	<file_filters>
	AND (bug_path_events.id IS NOT NULL OR bug_report_point.id IS NOT NULL OR extended_report_data.id IS NOT NULL);`

PR #4372, opened 2024-10-21
[analyzer] Disable clang flags that cause major undeterminism issues

Szelethus:
Turn off Z3 timeouts added to the Clang Static Analyzer in https://github.com/llvm/llvm-project/pull/95129. We are now aware that these options can often cause nondeterministic behaviour.


config/config_files/codechecker.json
```diff
@@ -4,6 +4,10 @@
     "--enable=core.CallAndMessage",
     "--analyzer-config",
     "clangsa:unroll-loops=true",
+    "--analyzer-config",
```
dkrupp, line 7: isn't this config file just an example? Will it get in effect by default?

dkrupp (commented):
isn't this config file just an example? Will it get in effect by default?

config/config_files/codechecker.json
Szelethus: I initially intended to put the change here, I'm just not sure whether hardcoding such a change is appropriate:
https://github.com/Ericsson/codechecker/blob/master/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py#L392
@bruntib?

dkrupp:
I guess this patch can be abandoned... or?

Szelethus:
Tahts right, sorry!

PR #4374, opened 2024-10-23
fixing malparsed windows paths

### LINKED ISSUE #4277: CodeChecker not parsing correctly windows escaped paths in compile_commands.json
**Describe the bug**
Rightly or wrongly, the compile_commands.json file is being created with both unix forward slash, and escaped windows backward slash.
if you tell me that the bug is further upchain so to speak, because the database should be in consistent then I would need to unwrap Zephyr RTOS and CMake. To start my investigation, this seems like the logical place.

This is an extract from the database as created by Zephyr RTOS/CMake:
```
{
  "directory": "A:/projects/AMPD/AE/board/firmware/build",
  "command": "A:\\tools\\zephyr\\zephyr-sdk-0.16.8\\arm-zephyr-eabi\\bin\\arm-zephyr-eabi-gcc.exe
```

as you can see the directory is in unix format, but the command is in windows format, with the assumption of one layer of escaping.
It is only the directory that is in unix format, the rest of the contents is formatted like the command shown above. (i'm missing the rest because in poking around I have no broken my build env, but what to shoot this report off ASAP as it's a fundamental blocker for me right now). 

I'm developing on windows with a mix of windows native tools and mingw tools. (Could that be a factor?)

My build logs (for the codechecker stage, the main build is successful)
```
Failed to get and parse version of: A:toolszephyrzephyr-sdk-0.16.8arm-zephyr-eabibinarm-zephyr-eabi-gcc.exe
[WinError 2] The system cannot find the file specified
```

**CodeChecker version**
```
[INFO 2024-06-25 08:53] - CodeChecker analyzer version:
Base package version | 6.23.1
Package build date   | 2023-12-14T14:38
Git commit ID (hash) | 2a8fa6e711a4ff591280a79fe8798dee2507d984
Git tag information  | 6.23.1

[INFO 2024-06-25 08:53] - CodeChecker web version:
Base package version                | 6.23.1
Package build date                  | 2023-12-14T14:38
Git commit ID (hash)                | 2a8fa6e711a4ff591280a79fe8798dee2507d984
Git tag information                 | 6.23.1
Server supported Thrift API version | 6.54
Client Thrift API version           | 6.54
```

**To Reproduce**

(On Windows)
1. Follow the Zephyr getting started page to install tools: https://docs.zephyrproject.org/latest/develop/getting_started/index.html
2. Follow the Zephyr CodeChecker page to install and demo codechecker: https://docs.zephyrproject.org/latest/develop/sca/codechecker.html
3. observe the file path to gcc that does not contain any folders path separators. 

**Expected behaviour**
I expected it to parse the path to GCC correctly.

 - OS: Windows
 - Browser: N/A
 - Version: 23H2 22631.3593

**Additional context**
Mix of tooling. CMake was installed via Winget, make is coming from msys2, as is ninja.
This isn't as per the STR, which uses chocolatey, i'm in the process of verifying that aspect of the setup.

tuix:
Fixes #4277

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -1163,7 +1163,7 @@ def extend_compilation_database_entries(compilation_database):
             source_files = []
             source_dir = entry['directory']
 
-            options = shlex.split(entry['command'])
+            options = shlex.split(entry['command'], posix=(os.name=='posix'))
             for opt in options:
                 if opt.startswith('@'):
                     response_file = os.path.join(source_dir, opt[1:])


analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -1163,7 +1163,7 @@ def extend_compilation_database_entries(compilation_database):
             source_files = []
             source_dir = entry['directory']
 
-            options = shlex.split(entry['command'])
+            options = shlex.split(entry['command'], posix=(os.name=='posix'))
```
marc-h38, line 1166: The `posix=` parameter is designed to manage nuances between different Unix shells, so it's very mysterious how this helps with Windows and why?!

I realize this is still just a draft, but some... draft comments and a commit message draft would not hurt.

Generally speaking, `shlex` is NOT meant to help prepare Windows commands ("outputs"). According to its documentation, it's meant as a convenience to help parsing _inputs_ to the program, like a simple configuration file or some parameters.

Longer `shlex` example and story in https://github.com/zephyrproject-rtos/west/commit/94f8a04ac83467b008f1a2aa5c4b437981367cf9

ferkulat: @marc-h38 
"The posix= parameter is designed to manage nuances between different Unix shells, so it's very mysterious how this helps with Windows and why?!"
 When the `posix=false` is not passed on windows, `shlex` is eating up back slashes:
 
 
image

And here with `posix=False`:
image

here is some related issue:
#https://github.com/mesonbuild/meson/issues/5726

marc-h38: > > The posix= parameter is designed to manage nuances between different Unix shells, 

Correction:  `posix=True` means `backwards-compatibility=False`. This could/should be better documented. I found the evidence here:
https://github.com/python/cpython/issues/38314#issuecomment-5287735090

> > so it's very mysterious how this helps with Windows and why?!

> When the posix=false is not passed on windows, shlex is eating up back slashes:

Right, escaping rules are different between posix and non-posix:
https://docs.python.org/3/library/shlex.html#parsing-rules

However this is just one difference among a long list of other differences. shlex is effectively two pretty different parsers in one. Those other differences should  be understood and evaluated in this particular context and the commit message should tell that story. When not, the "Pragrammatic Programmer" book names this "programming by coincidence".

Or maybe I should say "called" in the past tense now since AI is writing all the code ;-)

marc-h38:
Just found by chance that #5017 is another tentative fix.

analyzer/codechecker_analyzer/buildlog/log_parser.py
marc-h38: > here is some related issue: #https://github.com/mesonbuild/meson/issues/5726

Apparently, the fix there was to stop (ab)using `shlex` for cases it was not designed to handle, and replace it with something else entirely.

ferkulat:
> Just found by chance that #5017 is another tentative fix.

no, #5017 is a fix for back slashes in the "file" field of compile_database.json. but not for failing to extract the compiler path, like this fix

PR #4375, opened 2024-10-24
[fix] Get product configuration with view permission

bruntib:
The getProductConfiguration() function on Product endpoint requires a current product in the URL for checking view permission. The requirement of having view permission has been added in
8953b30f6d17597635ec59bb943683aacb216619. However there is no "current product" in the Product endpoint URL queries, but the product id is provided through a function parameter.


dkrupp (changes requested):
please check the permission handling of the removeProduct() function. It should need the same treatment.

dkrupp (approved):

PR #4377, opened 2024-10-29
Handle ambigous profile/group parameters

noraz31:
Introduce the -e prefix: option, to improve handling of ambigous parameters.
such as "security" which is a checker prefix group and a profile at the same time

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/config_handler.py
+++ b/analyzer/codechecker_analyzer/analyzers/config_handler.py
@@ -12,8 +12,10 @@ Static analyzer configuration handler.
 
 from abc import ABCMeta
 from enum import Enum
+from string import Template
 import collections
 import platform
+import sys
 import re
 
 from codechecker_analyzer import analyzer_context
@@ -149,10 +151,12 @@ class AnalyzerConfigHandler(metaclass=ABCMeta):
           and "debug" checker groups. "osx" checker group is also not included
           unless the target platform is Darwin.
         - Command line "--enable/--disable" flags.
-          - Their arguments may start with "profile:" or "guideline:" prefix
-            which makes the choice explicit.
-          - Without prefix it means a profile name, a guideline name or a
-            checker group/name in this priority order.
+          - Their arguments may start with "prefix:", "profile:" or
+            "guideline:" label which makes the choice explicit.
+          - Without a label, it means either a profile name, a guideline name
+            or a checker prefix group/name, unless there is a clash
+            between these names, in which case an error is emitted.
+            If there is a name clash, using a label is mandatory.
 
         checkers -- [(checker name, description), ...] Checkers to add with
                     their description.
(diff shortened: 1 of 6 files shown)


analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -210,23 +212,40 @@ def initialize_checkers(self,
         profiles = checker_labels.get_description('profile')
         guidelines = checker_labels.occurring_values('guideline')
 
+        templ = Template("The ${entity} name '${identifier}' conflicts with a "
+                         "checker(-group) name. Please use -e ${entity}: to "
```
vodorok, line 216: We should be very specific about what conflicts with what. So if you can provide an exact error message instead of `...with a checker(-group)...` then please be specific.

analyzer/codechecker_analyzer/checkers.py
```diff
@@ -23,7 +23,8 @@ def available(ordered_checkers, available_checkers):
         if checker_name.startswith('profile:') or \
                 checker_name.startswith('guideline:') or \
                 checker_name.startswith('severity:') or \
-                checker_name.startswith('sei-cert:'):
+                checker_name.startswith('sei-cert:') or \
+                checker_name.startswith('group:'):
```
vodorok, line 27: If you plan to introduce a new checker enable prefix, please add this to the documentation and help of CodeChecker analyze and CodeChecker check.

analyzer/tests/unit/test_checker_handling.py
```diff
@@ -216,10 +225,15 @@ def f(checks, checkers):
 
         # Enable "security" profile checkers without "profile:" prefix.
         cfg_handler = ClangSA.construct_config_handler(args)
-        cfg_handler.initialize_checkers(checkers,
-                                        [('security', True)])
```
vodorok, line 220: Why are you replacing the security prefix? enabling with sensitive?

vodorok (changes requested):
I feel like this change  is a backward incompatible change, and would requires  another reviewers opinion.
@dkrupp  what is your opinion?

analyzer/codechecker_analyzer/checkers.py
noraz31: Will do. Thanks!

analyzer/codechecker_analyzer/analyzers/config_handler.py
noraz31: That's right, AFAIK it is always a checker group name that is conflicting, so I'll correct this.

analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -210,23 +212,40 @@ def initialize_checkers(self,
...
             elif identifier in profiles:
                 if identifier in reserved_names:
-                    LOG.warning("Profile name '%s' conflicts with a "
-                                "checker(-group) name.", identifier)
-                for checker in checker_labels.checkers_by_labels(
-                        [f'profile:{identifier}']):
-                    self.set_checker_enabled(checker, enabled)
+                    LOG.error(templ.substitute(entity="profile",
+                                               identifier=identifier))
+                    sys.exit(1)
```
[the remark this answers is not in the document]
dkrupp: I think yes. These are the exit code listed by CodeChecker analyze --help
Exit status
0 - Successful analysis
1 - CodeChecker error
3 - Analysis of at least one translation unit failed
128+signum - Terminating on a fatal signal whose number is signum

analyzer/tests/unit/test_checker_handling.py
```diff
@@ -199,7 +208,7 @@ def f(checks, checkers):
 
         # Enable alpha checkers explicitly.
         cfg_handler = ClangSA.construct_config_handler(args)
-        cfg_handler.initialize_checkers(checkers, [('alpha', True)])
+        cfg_handler.initialize_checkers(checkers, [('group:alpha', True)])
```
dkrupp, line 211: why do we need this change? "alpha" prefix group still should work as there is no profile named "alpha"

Maybe have one testcase with 
cfg_handler.initialize_checkers(checkers, [('group:alpha', True)])
and one with 
cfg_handler.initialize_checkers(checkers, [('alpha', True)])

both should enable the same set of checkers.

analyzer/tests/unit/test_checker_handling.py
```diff
@@ -244,13 +257,6 @@ def f(checks, checkers):
         self.assertTrue(all_with_status(CheckerState.DISABLED)
                         (cfg_handler.checks(), cert_guideline))
 
-        # Disable "sei-cert" guideline checkers.
```
dkrupp, line 247: why do we have to remove these tests?

analyzer/tests/unit/test_checker_handling.py
noraz31: Will do, and also add some more tests in the next patch to cover some more scenarios.

analyzer/tests/unit/test_checker_handling.py
noraz31: Will undo this

analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -210,23 +212,40 @@ def initialize_checkers(self,
...
 
+        templ = Template("The ${entity} name '${identifier}' conflicts with a "
+                         "checker group name. Please use -e ${entity}: to "
+                         "select checkers of a ${entity} or use -e group: "
+                         "to select checkers which have a name "
+                         "starting with '${identifier}'.")
+
         for identifier, enabled in cmdline_enable:
-            if ':' in identifier:
+            if "group:" in identifier:
```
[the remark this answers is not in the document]
noraz31: Got rid of the "group" label. Now the "prefix" label is used everywhere.
Thanks for the review!

dkrupp (approved):
LGTM

vodorok (approved):
LGTM

PR #4379, opened 2024-11-05
Bring code borrowed from `http.server` in sync with upstream

Discookie:
Remove translate_path() to rely on http.server's implementation, and fix handling of leading slashes.

### DIFF
--- a/web/server/codechecker_server/server.py
+++ b/web/server/codechecker_server/server.py
@@ -79,6 +79,8 @@ class RequestHandler(SimpleHTTPRequestHandler):
     def __init__(self, request, client_address, server):
         self.path = None
         super().__init__(request, client_address, server)
+        # GET requests are served from www_root.
+        self.directory = server.www_root
 
     def log_message(self, *args):
         """ Silencing http server. """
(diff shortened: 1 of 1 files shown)


web/server/codechecker_server/server.py
```diff
@@ -237,11 +239,14 @@ def do_GET(self):
...
             self.path = self.path.replace(f"{product_endpoint}/", "", 1)
+            # Remove extra leading slashes, see cpython#93789.
+            self.path = '/' + self.path.lstrip('/')
 
         if self.path == '/':
             self.path = "index.html"
 
         # Check that the given path is a file.
+        # The base directory is set to www_root.
         if not os.path.exists(self.translate_path(self.path)):
```
cservakt, line 248: You already deleted this translate_path fuction so I think you would have modified this if state.

cservakt (commented):
Can you get rid off the unused imports like posixpath and urllib to run pycodestyle test without any problem?

web/server/codechecker_server/server.py
Discookie: I deleted the re-implementation of the translate_path function, this calls SimpleHTTPRequestHandler.translate_path() - translating the path with the root set to www_root.

Discookie:
Linter passes as of latest commit.

cservakt (approved):

dkrupp (approved):
LGTM

PR #4380, opened 2024-11-08
[feat] Adding report annotation for json export

cservakt:
It is an additional feature for parseing to have report annotation fields when the user wants to export the results to json.


bruntib (approved):
The JSON output of `CodeChecker parse` has a `version` attribute which is currently 1. In what situations should we increase this number? I think, the extension of the current format, like adding `annotations` is not harmful, but maybe we should discuss this question sometime.

PR #4384, opened 2024-11-21
[feat] Unique reports by different annotations

bruntib:
Two report should also be considered different if they differ only in their annotations.


dkrupp (approved):
LGTM

PR #4392, opened 2024-11-29
[fix] Resolve checker enable/disable ambiguity

cservakt:
This PR fixes the enabling and disabling checker options, focusing on resolving ambiguities in option processing.

 Key updates:
  **Ambiguity handling:**
> When ambiguous checker options are provided with `-e` or `-d` flags, an error is now raised. The user receives suggestions for specifying the option by using namespaces (checker, prefix, guideline, etc.).
> If a checker name matches multiple checkers as a prefix, it triggers an ambiguity error. Suggestions are provided to help users choose between the checker name or the prefix.

 **Additional namespaces:**
>A new `checker` namespace is created. All label types can be a namespace as well, specified with a `:` separator before the checker option.

 **Default profile settings improvements:**
> Only those checkers enabled that has profile:default label in the label files. Prefix-based configuration is no longer applied for the default profile, ensuring precise and predictable behavior.

 **Getting rid of `W` prefix:**
> It is no longer available to enable or disable a warrning with W.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/config_handler.py
+++ b/analyzer/codechecker_analyzer/analyzers/config_handler.py
@@ -12,7 +12,6 @@ Static analyzer configuration handler.
 
 from abc import ABCMeta
 from enum import Enum
-from string import Template
 import collections
 import platform
 import sys
@@ -86,7 +85,7 @@ class AnalyzerConfigHandler(metaclass=ABCMeta):
         """
         self.__available_checkers[checker_name] = (state, description)
 
-    def set_checker_enabled(self, checker_name, enabled=True):
+    def set_checker_enabled(self, checker_name, enabled=True, is_strict=False):
         """
         Explicitly handle checker state, keep description if already set.
         """
@@ -94,7 +93,8 @@ class AnalyzerConfigHandler(metaclass=ABCMeta):
         regex = "^" + re.escape(str(checker_name)) + "\\b.*$"
 
         for ch_name, values in self.__available_checkers.items():
-            if re.match(regex, ch_name):
+            if (is_strict and ch_name == checker_name) \
+               or (not is_strict and re.match(regex, ch_name)):
                 _, description = values
                 state = CheckerState.ENABLED if enabled \
                     else CheckerState.DISABLED
@@ -118,7 +118,7 @@ class AnalyzerConfigHandler(metaclass=ABCMeta):
         """
         return self.__available_checkers
 
-    def __gen_name_variations(self):
+    def __gen_name_variations(self, only_prefix=False):
         """
         Generate all applicable name variations from the given checker list.
         """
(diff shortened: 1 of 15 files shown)


analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -207,48 +207,68 @@ def initialize_checkers(self,
...
+                identifier = identifier.split(":", 1)[1]
+                self.initialize_checkers_by_namespace(
+                    identifier_namespace, identifier, enabled)
 
             else:
-                self.set_checker_enabled(identifier, enabled)
+                all_options = dict(zip(labels, map(
+                    lambda label: checker_labels.get_description(label).keys()
+                    if checker_labels.get_description(label)
+                    else checker_labels.occurring_values(label), labels)))
```
bruntib, line 233: Can't we always use `occurring_values()`? Why do we need this `if`?

analyzer/codechecker_analyzer/analyzers/config_handler.py
```diff
@@ -118,7 +118,7 @@ def checks(self):
         """
         return self.__available_checkers
 
-    def __gen_name_variations(self):
+    def __gen_name_variations(self, only_prefix=False, unique=False):
```
bruntib, line 121: When is it useful getting non-unique results?

bruntib (dismissed):

analyzer/codechecker_analyzer/analyzers/config_handler.py
cservakt: Previously, it was necessary to collect all name variations. I just didn't want to create a new backward incompatibility function. But it is ok for me to have just unique mode.

dkrupp (changes requested):
1. Please update the help of CodeChecker analyze --help and mention the :checker prefix option
2. Update the analyzer user guide https://github.com/Ericsson/codechecker/blob/master/docs/analyzer/user_guide.md#toggling-checkers
and remove any -Wwarning syntax suggestion. Describe the new "checker:" and "prefix:" syntax. Describe the name clash resolution logic 

These are the direct remarks. More comments will come soon.

dkrupp (changes requested):
1)
Please also fix this section
https://github.com/Ericsson/codechecker/blob/master/docs/analyzer/user_guide.md#toggling-checkers
"
Command line --enable/--disable flags.
Their arguments may start with profile: of guideline: prefix which makes the choice explicit.
Without prefix it means a profile name, a guideline name or a checker group/name in this priority order."

Describe the checker:, prefix: and describe that priority order does not matter, an ambiguity causes error.

2) remove all old --enable W*  warning enable syntax from
https://github.com/Ericsson/codechecker/blob/master/docs/analyzer/user_guide.md#toggling-compiler-warnings

section

dkrupp (changes requested):
Looks good.

I found 1 problem though:
CodeChecker analyze -e checker:nonextisting-checker

should give an error, but it accepts a non existing checker name.

dkrupp (approved):

Issue #4360, opened 2024-10-11
LD_PRELOAD cannot be preloaded, possible regression with #4312


gamesh411:
**Describe the bug**
On Ubuntu-24.04, when analyzing bitcoin v0.20.1, the logging phase fails.

**CodeChecker version**
master a6663b3349c54a5d00687bd3d3695ead19b8e060

**To Reproduce**
```
# Fresh build of CodeChecker on master
git clone https://github.com/Ericsson/codechecker.git
cd codechecker
make venv package
source venv
export PATH="$PWD/build/CodeChecker/bin:$PATH"
cd ..
# Clone and log bitcoin project
git clone https://github.com/bitcoin/bitcoin.git -b v0.20.1
cd bitcoin
./autogen.sh && ./configure --disable-wallet --disable-static --disable-tests --without-gui
CodeChecker log -b 'make -j' -o compile_commands.json
```
Results in this error for all TUs:
```
CXX      lib/libunivalue_la-univalue_get.lo
ERROR: ld.so: object 'ldlogger.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
```

**Expected behaviour**
I expect the log to produce a compilation database, but it is empty.

**Desktop (please complete the following information)**
 - OS: WSL Ubuntu-24.04
 - Browser: not relevant
 - Version: not relevant

**Additional context**
I see that the issue came up in #3858, and I suspect that the regression to this error happened #4312.
I think the change should not just be reverted, but instead, the machine architecture of the TU analyzed should be added to LD_LIBRARY_PATH env variable of the compiler invocation to be logged.

gamesh411:
I have tested with the latest master version (already CMake-based) Bitcoin project, and there is no mention of it overriding the LD_LIBRARY_PATH. The issue still holds.
IMO we need this https://github.com/Ericsson/codechecker/pull/3919, but with the modification that ensures we also support a different-architecture binary than the host OS.

Right now, I see 2 ways of doing this:
1) we detect the architecture of the compiler we use for logging
2) blindly pass all the shared objects to the binary by adding all of them (maybe multiple absolute path entries for LD_PRELOAD, or just a relative path, and adding all the machine-specific directories to LD_PRELOAD_PATH)

dkrupp:
[...]

PR #4394, opened 2024-12-02
[analyzer] Include arch subdirectories in LD_LIBRARY_PATH for logging

gamesh411:
The issue with logging (see https://github.com/Ericsson/codechecker/issues/4360 for details) came up on my Ubuntu24 machine, where only specifying the <data_files>/ldlogger/lib in the LD_LIBRARY_PATH was not enough for the dynamic loader to load the ldlogger.so inside <data_files>/ldlogger/lib/x86_64, so I observed there is no implicit architecture resolution in the logic of the loader. There were other tests on other distributions where this implicit architecture resolution seems to exist (rhel7, rhel8, ubuntu22, among others).

The solution implemented by this patch is adding all <data_files>/ldlogger/lib/* folders in addition to the current  <data_files>/ldlogger/lib folder to the LD_LIBRARY_PATH.

I have tested it on Ubuntu24, and the multiple .so files that are possibly reachable due to all of them being added are not an issue. The logger works as expected, and if I remove only the relevant ldlogger.so (x86_64 for my machine) but keep the others, the CodeChecker log command correctly detects that I meet neither the ld-logger nor the intercept-build requirement.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzer_context.py
+++ b/analyzer/codechecker_analyzer/analyzer_context.py
@@ -71,12 +71,19 @@ class Context(metaclass=Singleton):
         # Original caller environment of CodeChecker for external binaries
         self.__original_env = None
 
-        self.logger_lib_dir_path = os.path.join(
-            self._data_files_dir_path, 'ld_logger', 'lib')
-
-        if not os.path.exists(self.logger_lib_dir_path):
-            self.logger_lib_dir_path = os.path.join(
+        # Find the path which has the architectures for the built ld_logger
+        # shared objects.
+        ld_logger_path = Path(self._data_files_dir_path, 'ld_logger', 'lib')
+        if not ld_logger_path.is_dir():
+            ld_logger_path = Path(
                 self._lib_dir_path, 'codechecker_analyzer', 'ld_logger', 'lib')
+
+        # Add the ld_logger_path and all children (architecture) paths to be
+        # later used in the LD_LIBRARY_PATH environment variable during
+        # logging of compiler invocations.
+        self.logger_lib_dir_path = ':'.join(
+            [str(ld_logger_path)] +
+            [str(arch) for arch in ld_logger_path.iterdir() if arch.is_dir()])
 
         self.logger_bin = None
         self.logger_file = None
--- a/analyzer/codechecker_analyzer/analyzer_context.py
+++ b/analyzer/codechecker_analyzer/analyzer_context.py
@@ -73,17 +73,18 @@ class Context(metaclass=Singleton):
 
         # Find the path which has the architectures for the built ld_logger
         # shared objects.
-        ld_logger_path = Path(self._data_files_dir_path, 'ld_logger', 'lib')
+        ld_logger_path = Path(self._data_files_dir_path, "ld_logger", "lib")
         if not ld_logger_path.is_dir():
             ld_logger_path = Path(
-                self._lib_dir_path, 'codechecker_analyzer', 'ld_logger', 'lib')
+                self._lib_dir_path, "codechecker_analyzer", "ld_logger", "lib"
+            )
 
-        # Add the ld_logger_path and all children (architecture) paths to be
-        # later used in the LD_LIBRARY_PATH environment variable during
-        # logging of compiler invocations.
-        self.logger_lib_dir_path = ':'.join(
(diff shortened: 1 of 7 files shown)


gamesh411:
The issue with logging (see #4360 for details) came up on my Ubuntu24 machine, where only specifying the `<data_files>/ldlogger/lib` in the LD_LIBRARY_PATH was not enough for the dynamic loader to load the `ldlogger.so` inside  `<data_files>/ldlogger/lib/x86_64`, so I observed there is no implicit architecture resolution in the logic of the loader. There were other tests on other distributions where this implicit architecture resolution seems to exist (rhel7, rhel8, ubuntu22, among others).

The solution implemented by this patch is adding all `<data_files>/ldlogger/lib/*` folders in addition to the current ` <data_files>/ldlogger/lib` folder.

I have tested it on Ubuntu24, and the multiple .so files that are possibly reachable due to all of them being added are not an issue. The logger works as expected, and if I remove only the relevant `ldlogger.so` (x86_64 for my machine) but keep the others, the CodeChecker log command correctly detects that I meet neither the ld-logger nor the intercept-build requirement.

dkrupp:
sounds good. did you test this on older distros? (ubuntu 20, 22,rhel8,rhel9)

dkrupp:
You may want to try out this implmentation on the following test

test.c:
```
int main(void){
  return 0;
}
```

builder.c:
```
#include<stdlib.h>
int main(void){
  const char* command="gcc ./test.c";
  system(command);
  return 0;
}
```
`gcc -m32 ./builder.c -o builder32`



`CodeChecker log -b ./builder32 -o compile_command32.json`

gamesh411:
Hi, I have tested with the suggested method, and both logs have the correct content.
I tested the above approach but modified the builder to make it even more convoluted like this:

```
#include<stdlib.h>
int main(void) {
  system("/path/to/32bitcompiler/bin/clang -o test_build_with_clang32 test.c");
  system("/path/to/64bitcompiler/bin/clang -o test_build_with_clang64 test.c");
  return 0;
}
```
Then, build this to `builder32` and `builder64`, and log them like this:
```
strace -fte trace=%process,%file CodeChecker log -b './builder32' -o ccdb32.json 2>&1 | tee strace_log32
strace -fte trace=%process,%file CodeChecker log -b './builder64' -o ccdb64.json 2>&1 | tee strace_log64
```
Then, the contents of *BOTH* `ccdb32.json` and `ccdb64.json`: 
```
[
	{
		"directory": "...",
		"command": "/path/to/32bitcompiler/bin/clang -o test_build_with_clang32 test.c",
		"file": "test.c"
	}
	,
	{
		"directory": "...",
		"command": "/path/to/64bitcompiler/bin/clang -o test_build_with_clang64 test.c",
		"file": "test.c"
	}
]
```
All in all, I see this is the expected behaviour.

analyzer/codechecker_analyzer/analyzer_context.py
```diff
@@ -71,13 +71,20 @@ def __init__(self):
...
+        ld_logger_path = Path(self._data_files_dir_path, 'ld_logger', 'lib')
+        if not ld_logger_path.is_dir():
+            ld_logger_path = Path(
                 self._lib_dir_path, 'codechecker_analyzer', 'ld_logger', 'lib')
 
+        # Add the ld_logger_path and all children (architecture) paths to be
+        # later used in the LD_LIBRARY_PATH environment variable during
+        # logging of compiler invocations.
+        self.logger_lib_dir_path = ':'.join(
+            [str(ld_logger_path)] +
```
dkrupp, line 85: I think we dont need to add CodeChecker/ld_logger/lib/
and then we can also remove the  'i486' 'i586' 'i686' subdirectories from analyzer/tools/build-logger/Makefile line 79
`  for x86dir in 'i386' 'i486' 'i586' 'i686'; do \`

dkrupp (changes requested):
I like the proposed solution and I tested and it works on ubuntu 22 too
I think we can remove the i486, i586, i686 subdirs of ld_logger/lib
and just keep i386 dir.

Also could you please add the testcase which I proposed in my comment as functional test?

gamesh411:
@dkrupp Thanks for the suggestions. I implemented and verified all of them on Ubuntu24 and this Ubuntu20 test environment.

PR #4401, opened 2024-12-09
[feat][report-converter][fix][server] Add support for bug path arrows into sarif parser.

jstevens176:
The current status quo supports SARIF but doesn't allow you to use the "show arrows" feature within the report detail view. This PR extracts the relevant data from the SARIF files and provides the database the means to draw the arrows, making "show arrows" usable.

This exposed an issue with the current server ingest process, in that it assumes a certain format for the BugPath positions and event objects. This PR also adds a number of checks to correctly extract the data as required.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/report/parser/sarif.py
+++ b/tools/report-converter/codechecker_report_converter/report/parser/sarif.py
@@ -143,25 +143,29 @@ class Parser(BaseParser):
 
         thread_flow_info = ThreadFlowInfo()
 
-        # TODO: Currently, we only collect bug path events.
-
         for code_flow in result.get("codeFlows", []):
             for thread_flow in code_flow.get("threadFlows", []):
                 for location_data in thread_flow["locations"]:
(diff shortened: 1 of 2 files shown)


dkrupp:
@jstevens176  thanks for the PR.

Could you please specify a sample sarif file which is not stored properly into CodeChecker?

I tried the following and it worked.
``

mkdir reports_sarif

 clang -c --analyze --analyzer-output sarif ./os_command_injection/CWE78_OS_Command_Injection__char_connect_socket_system_02.c -o output.sarif  -Xclang -analyzer-config -Xclang expand-macros=true -Xclang -analyzer-checker=optin.taint -o ./reports_sarif/output.sarif
``

then store it into codechecker
``
CodeChecker store ./reports_sarif/ -n sarif --url http://localhost:8002/Default
``

The store is successful. Arrows are there.

image

Could you please attach a sample sarif file which is not stored/visualized properly?
Also please add a testcase to the PR which demonstrates the fixed issue.

PR #4412, opened 2025-01-06
[test] Add tests for different report hashes

bruntib:
There are some tests already which check if two reports with different annotations are considered different reports.
Now we also check if two reports with different report hash are considered different.


noraz31 (approved):
LGTM

dkrupp (approved):
LGTM

PR #4413, opened 2025-01-06
Add a mode where CTU files are retained

tomrittervg:
For debugging purposes I wanted to retain the CTU files.  In case this sounds useful, I'll throw the patch here.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzer.py
@@ -148,6 +148,7 @@ def perform_analysis(args, skip_handlers, filter_handlers,
 
     ctu_collect = False
     ctu_analyze = False
+    ctu_delete = False
     ctu_dir = ''
     if 'ctu_phases' in args:
         ctu_dir = os.path.join(args.output_path, 'ctu-dir')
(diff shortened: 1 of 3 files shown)


bruntib:
Hi,

Thanks for the patch! However, as a rule of thumb, we would like to go in the direction of not introducing further flags unless it's unavoidable otherwise.

`CodeChecker analyze` command has `--ctu-collect` and `--ctu-analyze` flags. The first one is producing the CTU files and the second one is using them for analysis. The `--ctu` flag is almost equivalent with providing both flags. When we need to debug the generated CTU files, then we run `--ctu-collect` only. Would you think, this is a good methodology for testing CTU?

bruntib:
I understand the motivation and I agree with having a solution for debugging reasons. I'm just suggesting an alternative solution to execute the analysis in two phases. I mean if you execute `CodeChecker analyze` with `--ctu-collect` and `--ctu-analyze` separately, then the analysis is the same and `ctu-dir` is not removed.

tomrittervg:
> I understand the motivation and I agree with having a solution for debugging reasons. I'm just suggesting an alternative solution to execute the analysis in two phases. I mean if you execute `CodeChecker analyze` with `--ctu-collect` and `--ctu-analyze` separately, then the analysis is the same and `ctu-dir` is not removed.

Oh, I misunderstood or mis-remembered the code then, I thought it was still removed after ctu-analyze even if it was two phases, but [that doesn't seem to be the case](https://github.com/Ericsson/codechecker/blob/66efe2592f1da5dc813ceca0dfaf76e4424f6ac1/analyzer/codechecker_analyzer/analyzer.py#L397).  

Thanks, I will migrate to that pattern at some point.

PR #4415, opened 2025-01-08
[fix] Proper encoding of standard output

bruntib:
"CodeChecker cmd" commands were crashing if they would print a character that can't be decoded by the current system encoding.


cservakt (approved):
LGTM

PR #4416, opened 2025-01-08
[gui] Display full filename

bruntib:
No need for truncating long filenames in GUI.


cservakt (commented):
Adding `:title` prop to tags of ReportTreeLabel can be a great external feature for checking filename and others easily.

cservakt (approved):
Now it looks good to me.

PR #4424, opened 2025-01-10
[report-converter] Improve documentation maintainability

gamesh411:
Reduce documentation duplication by:
- Create a concise package-level README focused on quick start (and silence a warning, as MANIFEST.in expects a README.md)
- Update setup.py to use a focused description, and as the scope of the tool expanded over time, keywords and general description now acknowledges that
- Point to official readthedocs.io documentation as single source of truth
- Remove redundant information from package documentation

### DIFF
--- a/tools/report-converter/README.md
+++ b/tools/report-converter/README.md
@@ -1,0 +1,46 @@
+# CodeChecker Report Converter
+
+A Python tool to parse and convert various code analyzer outputs into CodeChecker format. It can create HTML reports from analyzer outputs and store them in a CodeChecker server.
+
+## Quick Start
+
+```bash
+# Create a Python virtualenv and set it as your environment
+make venv
+source $PWD/venv/bin/activate
+
+# Build and install report-converter package
+make package
+```
+
+## Basic Usage
+
+```bash
+# Convert analyzer output to CodeChecker format
+report-converter --type clang-tidy --output /path/to/reports clang-tidy-output.txt
+
+# Generate HTML report
+plist-to-html --input /path/to/reports/*.plist --output /path/to/html
+```
+
+## Documentation
+
+For comprehensive documentation, including:
+- Complete list of supported analyzers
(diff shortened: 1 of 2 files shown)


tools/report-converter/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+## Basic Usage
+
+```bash
+# Convert analyzer output to CodeChecker format
+report-converter --type clang-tidy --output /path/to/reports clang-tidy-output.txt
+
+# Generate HTML report
+plist-to-html --input /path/to/reports/*.plist --output /path/to/html
```
bruntib, line 23: I can't find this `plist-to-html` script.

bruntib (changes requested):

tools/report-converter/README.md
gamesh411: There is a entry point defined for the report-converter package, which is named `plist-to-html`:
/tools/report-converter/report_converter.egg-info/entry_points.txt
This is a way for Python packages to provide a CLI script once installed.
I just went ahead and included this one in the docs as well.

bruntib: Maybe, that entry point should be removed from setup.py, because it really doesn't exist. Or, am I doing something wrong?

gamesh411: I have found these references to the plist-to-html converter tool (besides the markdown docs):
```
tools/report-converter/setup.py|37 col 14| 'plist-to-html = codechecker_report_converter.report.output.html.cli:main'
tools/report-converter/codechecker_report_converter/report/output/html/cli.py|52 col 15| prog="plist-to-html",
```

So I think this is still a thing in this repository (at least it still appears as the prog argument for this cli.py's ArgParse).
Maybe we should remove it from there as well if we do not intend to have a separate plist-to-html runnable Python file built.
@bruntib What do you think?

PR #4425, opened 2025-01-18
Upgrade Thirft API compiler version to 0.21.0

gamesh411:
The current version of Thrift is quite old, and the generated Python stubs are triggering deprecation warnings in modern Python versions (I tested it with Python 3.12). Additionally, the official Docker image used for compiling Thrift definition files is outdated. To address this, I have introduced a customization point in the build system, making it easier to configure when building the CodeChecker API packages.


web/api/Makefile
```diff
@@ -1,5 +1,13 @@
 THRIFT_OPTS = -r -o /data
-THRIFT_VERSION = "0.12.0"
+
+# The default for THRIFT_DOCKER_IMAGE is the official thrift image
+THRIFT_DOCKER_IMAGE ?= thrift
+# The current default for THRIFT_VERSION is latest version of the currently
+# unmaintained official thrift image on dockerhub
+# This is the version of the Docker image, and currently we are operating on
```
gamesh411, line 7: This statement about version matching is just too speculative and refer to a bigger context, so I remove them.

gamesh411:
There is a fix for the issues we have been having with Thrift on Python 3.12+. Let's wait for Thrift release 0.22 and upgrade to that one. I'm closing this.

PR #4440, opened 2025-01-28
[web] Timestamp annotation is not checked anymore

bruntib:
Earlier it was checked whether "timestamp" annotation had a proper "timedate" format. It turns out that sometimes it's more convenient to order reports based on a different time measure. For this reason we allow any string as "timestamp" annotation.


bruntib:
I abandon this change because we chose another implementation.

PR #4443, opened 2025-01-28
Fix multiprocess compatibility on non-linux systems

gamesh411:
Enhance multiprocessing compatibility across platforms by introducing a `create_pool` function in the compatibility module. This function abstracts the differences between `multiprocess.Pool` and `concurrent.futures.ProcessPoolExecutor`, ensuring consistent pool initialization. The changes were applied to both 'analysis_manager.py' and 'pre_analysis_manager.py', addressing API differences.


gamesh411:
We will have to investigate more as part of a platform compatibility effort.

PR #4447, opened 2025-02-04
[analyzer] Add --inferargs flag

bruntib:
Analyzers can be given further flags that are forwarded verbatim to the analyzer invocation. These flags are provided in a text file which is given to --<analyzer>args flag. This flag was missing at Facebook Infer plugin.


cservakt (commented):
The patch looks good to me, but please fix the pylint detected issues. Right now, env is ambiguous.

cservakt (approved):

Issue #4437, opened 2025-01-27
Codechecker Store fails with Python 3.12


amai2012:
**Describe the bug**
Codechecker fails to store results using python 3.12 (e.g. from Ubuntu 24.04 LTS)

**CodeChecker version**
6.24.4

**To Reproduce**
Steps to reproduce the behaviour:

1. Executing the command '$ CodeChecker store ./reports --url https://codechecker:8001/Default --name PROJECT`'

triggers
```
$ CodeChecker store ./reports --url https://codechecker:8001/Default --name PROJECT
[INFO 2025-01-27 09:27] - Storing analysis results for run 'PROJECT'
[INFO 2025-01-27 09:27] - Checking local passwords or tokens in /root/.codechecker.passwords.json
[INFO 2025-01-27 09:27] - Password file not found.
[INFO 2025-01-27 09:27] - Checking for local valid sessions.
Traceback (most recent call last):
  File "/builds/projectvenv/lib/python3.12/site-packages/codechecker_common/cli.py", line 208, in main
    sys.exit(args.func(args))
             ^^^^^^^^^^^^^^^
  File "/builds/projectvenv/lib/python3.12/site-packages/codechecker_client/cmd/store.py", line 895, in main
    client = libclient.setup_client(args.product_url)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/projectvenv/lib/python3.12/site-packages/codechecker_client/client.py", line 256, in setup_client
    session_token = get_new_token(protocol, host, port, cred_manager)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/projectvenv/lib/python3.12/site-packages/codechecker_client/client.py", line 235, in get_new_token
    return perform_auth_for_handler(auth_client, host, port, cred_manager)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/projectvenv/lib/python3.12/site-packages/codechecker_client/client.py", line 144, in perform_auth_for_handler
    auth_response = auth_client.getAuthParameters()
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/projectvenv/lib/python3.12/site-packages/codechecker_client/thrift_call.py", line 41, in wrapper
[...]

PR #4448, opened 2025-02-05
[web] Patch Thrift deprecation for Python >=3.12

gamesh411:
All of the currently available versions of Thrift use Python code that is deprected and gives an error when run on Python >=3.12. This change manually fixes the installed Thrift package. NOTE: This is a HACK and should be removed as soon as Thrift creates an updated version and CodeChecker is ported to use that version.


gamesh411:
This patch aims to help the issue #4437 .
This only patches the virtual environment that CodeChecker runs in.
I think packages built from this just presuppose Thrift being installed in the site-packages, so this workaround only works when CodeChecker is run from venv or venv_dev.
Another route would be forking thrift itself and providing alternaltive pypi packages.
None of the above is satisfactory or feasible to maintain in the long run, but this version could at least work as a workaround for locally packaged CodeChecker.

PR #4450, opened 2025-02-06
[feat] Introduce chronological order for dynamic reports

bruntib:
Reports' can be assigned some labels. It is useful especially for dynamic analyzer reports for storing their timestamps or test cases during which they appeared.
Now it is possible to store a "chronological-order" annotation which is a simple integer counter.


cservakt (commented):
Currently, the report filters only include testcase. Users might also need the new chronological-order annotation as a filter. Should we consider adding it?

dkrupp (changes requested):
Where is the expected PLIST format is documented? Shouldn't this new chornological-order field be documented? could you please add it?

dkrupp (approved):
LGTM

PR #4456, opened 2025-02-12
[feat] Introduce cc-verbatim-args-file

bruntib:
The `--saargs`, `--tidyargs` and `--cppcheckargs` files are now
deprecated. These config files can now be provided through
`--analyzer-config` flag with a common config named
`cc-verbatim-args-file`.


dkrupp (approved):
LGTM

PR #4463, opened 2025-02-19
Configure UTF-8 output encoding globally

gamesh411:
Force UTF-8 encoding for stdout/stderr streams to handle locale-related issues across platforms. This replaces the manual encoding workaround and ensures consistent behavior regardless of the system's locale settings.

- Fixes crashes when printing Unicode characters on systems with non-UTF-8 locales (e.g. Chinese GB18030 or Windows CP1252)
- Centralizes encoding configuration in init.py
- Removes redundant per-file encoding logic

This approach works cross-platform by directly wrapping the standard streams rather than relying on environment variables, which can be unreliable on Windows.


gamesh411:
@bruntib What do you think about this approach for handling the issue that was the root cause for this commit:
#4415 ?

bruntib (approved):

PR #4464, opened 2025-02-20
[fix] Rename cmd modules to avoid conflict with built-in cmd

gamesh411:
Rename the module defined in 'codechecker_common/cmd' to 'codechecker_common/cli_commands' (as there is already a cli.py file in the same folder) in order to avoid ambiguity with Python's built-in cmd module when using `pdb.set_trace()`. The new name better reflects its purpose and eliminates potential naming conflicts.

Rename all 'cmd' directories to 'cli' for consistency with 'codechecker_common/cli_commands'.


gamesh411:
BTW I'm up for discussing what a good module name would be:
I started with 'commandline', then considered 'command_line' (because of grammar :D ), and arrived at 'cli'. But cli.py was already in the codechecker_common directory, so I opted for 'codechecker_cli' for just that folder.

bruntib (approved):
This is a nice improvement for sake of debugging, thanks!
Could you, please, fix the merge conflicts? Otherwise looks good.

Issue #4355, opened 2024-09-24
CodeChecker cmd runs - should provide analysis information (list of enabled checkers)


dkrupp:
'CodeChecker cmd runs <args> --enabled-checkers' should list all checkers which were enabled in at least one of the selected runs.

for example

'CodeChecker cmd runs -n master --enabled-checkers' shoudl list the enabeld checkers for the run named "master".

PR #4468, opened 2025-02-24
CodeChecker cmd runs

xb058t:
This PR implements the feature requested in issue #4355

### DIFF
--- a/web/client/codechecker_client/cmd_line_client.py
+++ b/web/client/codechecker_client/cmd_line_client.py
@@ -19,6 +19,7 @@ import re
 import sys
 import shutil
 import time
+import json
 from typing import Dict, Iterable, List, Optional, Set, Tuple
 
 from codechecker_api.codeCheckerDBAccess_v6 import constants, ttypes
@@ -638,6 +639,55 @@ def handle_list_runs(args):
     sort_mode = ttypes.RunSortMode(sort_type, sort_order)
 
     runs = get_run_data(client, run_filter, sort_mode)
+
+    if hasattr(args, 'enabled_checkers') and args.enabled_checkers:
+        results = get_run_results(client,
+                                  [run.runId for run in runs],
+                                  constants.MAX_QUERY_SIZE,
+                                  0,
+                                  None,
+                                  ttypes.ReportFilter(),
+                                  None,
+                                  None)
+        enabled_checkers = {}
+        for result in results:
+            if result.analyzerName not in enabled_checkers:
+                enabled_checkers[result.analyzerName] = set()
+            enabled_checkers[result.analyzerName].add(result.checkerId)
+        
+        if args.output_format == 'plaintext':
+            print("Enabled checkers:")
+            for analyzer in enabled_checkers:
+                print(analyzer + ":")
+                for checker in enabled_checkers[analyzer]:
+                    print("  " + checker)
+
+        elif args.output_format == 'csv':
+            separator = ','
+            print("Analyzer" + separator + "Checker")
+            for analyzer in enabled_checkers:
+                for checker in enabled_checkers[analyzer]:
+                    print(analyzer + separator + checker)
+
+        elif args.output_format == 'json':
+            enabled_checkers = {
+                analyzer: list(checkers)
+                          for analyzer, checkers in enabled_checkers.items()}
+            print(json.dumps(enabled_checkers, indent=2))
+
+        elif args.output_format == 'table':
+            header = ['Analyzer', 'Checker']
+            rows = [
+                (analyzer, checker)
+                for analyzer in enabled_checkers
+                for checker in enabled_checkers[analyzer]]
+            print(twodim.to_str(args.output_format, header, rows))
+
+        else:
+            LOG.error("Unsupported output format: %s", args.output_format)
+            return 1
+
+        return 0
 
     if args.output_format == 'json':
         # This json is different from the json format printed by the
--- a/web/client/codechecker_client/cmd/cmd.py
+++ b/web/client/codechecker_client/cmd/cmd.py
@@ -1298,6 +1298,11 @@ def add_arguments_to_parser(parser):
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
         description="List the analysis runs available on the server.",
         help="List the available analysis runs.")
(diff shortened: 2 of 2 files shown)


web/client/codechecker_client/cmd_line_client.py
```diff
@@ -639,6 +640,47 @@ def handle_list_runs(args):
...
+        elif args.output_format == 'table':
+            header = ['Analyzer', 'Checker']
+            rows = [
+                (analyzer, checker)
+                for analyzer, checkers in enabled_checkers.items()
+                for checker in checkers]
+            print(twodim.to_str(args.output_format, header, rows))
+        else:
+            LOG.error("Unsupported output format: %s", args.output_format)
+        return
```
dkrupp: @xb058t  Please do as @cservakt  suggested and merge the enabled-checkers processing starting at line 684.

Also please add the enabled checkers list when `CodeChecker cmd runs --details -o json` is invoked by the user.


[INFO 2026-02-13 15:50] - Checking for local valid sessions.
[
   {
      "tinyxmls" : {
         "analyzerStatistics" : {
            "clang-tidy" : {
               **"enabledCheckers":["checker1","checker2"]**
               "failed" : 0,
               "failedFilePaths" : [],
               "successful" : 2,
               "version" : "18.1.3"
            },
`

web/client/tests/unit/test_cmd_line_client.py
```diff
@@ -0,0 +1,191 @@
+import io
```
dkrupp, line 1: This unit test does not test the database query.

So please add a function test too in https://github.com/Ericsson/codechecker/blob/master/web/tests/functional/cmdline/test_cmdline.py

where the other function tests of CodeChecker cmd runs is tested.

web/client/codechecker_client/cmd_line_client.py
```diff
@@ -639,6 +640,47 @@ def handle_list_runs(args):
...
+            for analyzer, checkers in enabled_checkers.items():
+                print(analyzer + ":")
+                for checker in checkers:
+                    print("  " + checker)
+        elif args.output_format == 'csv':
+            separator = ';'
+            print("Analyzer" + separator + "Checker")
+            for analyzer, checkers in enabled_checkers.items():
+                for checker in checkers:
+                    print(analyzer + separator + checker)
```
bruntib: Please, use the standard `csv` module instead.

web/client/codechecker_client/cli/cmd.py
```diff
@@ -1298,6 +1298,11 @@ def add_arguments_to_parser(parser):
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
         description="List the analysis runs available on the server.",
         help="List the available analysis runs.")
+
+    runs.add_argument(
+        '--enabled-checkers',
```
bruntib: I'm not sure if the introduction of this new flag is a good idea. If no run name is specified, then the current implementation lists the union of all checkers enabled in any of the runs:
`CodeChecker cmd runs --enabled-checkers`
What is the use of such a query? At least the `--name` flag should be mandatory

I would say that the list of enabled checkers belongs to the given runs, so they should be listed per run. The default output format of `CodeChecker cmd runs` is an ASCII-art table that is not convenient to display this long information. My suggestion is to put this checker list in the JSON format only as @dkrupp mentioned.

The following functional test could be extended to test this feature in entirety: https://github.com/Ericsson/codechecker/blob/463081e3fbdc3d419ffc0d6fe41f9f7153e3f3ee/web/tests/functional/cmdline/test_cmdline.py#L224

Ps.:
In a later and independent discussion we could talk about whether `--details` flag and other output formats like plaintext, table (these are the same, by the way), csv, etc. are necessary.

dkrupp: After discussing with @bruntib we have the following proposal:

Remove the `--details` flag and the `--enabled-checkers` flags from the `CodeChecker cmd runs` command.

The enabled checker list should always be printed in the `-o json` mode, but not in the `plaintext`, `csv`, and `table` modes. 

So the user would use the command like
`CodeChecker cmd runs -n <run_name>`  or without the run name selector, like 
`CodeChecker cmd runs`  when all runs are printed.

Make sure that the json output is not null for any fields like `runCmd`.

The json format is okay like this
```
[INFO 2026-02-13 15:50] - Checking for local valid sessions.
[
{
"tinyxmls" : {
   "analyzerStatistics" : {
   "clang-tidy" : {
    "enabledCheckers":["checker1","checker2"]
   "failed" : 0,
"failedFilePaths" : [],
"successful" : 2,
"version" : "18.1.3"
},
```

Make sure that the json output is indented properly.

web/client/codechecker_client/cmd_line_client.py
```diff
@@ -651,9 +645,23 @@ def handle_list_runs(args):
...
+                        if checker_info.enabled:
+                            enabled_checkers[analyzer].add(checker_name)
+            run.analyzerStatistics = client.getAnalysisStatistics(
+                run.runId, None) or {}
+            for analyzer in run.analyzerStatistics:
+                stat = run.analyzerStatistics[analyzer]
+                stat.enabledCheckers = sorted(
+                    list(enabled_checkers.get(analyzer, set())))
             results.append({run.name: run})
         print(CmdLineOutputEncoder().encode(results))
```
bruntib:
```suggestion
        print(json.dumps(results, cls=CmdLineOutputEncoder, indent=4))
```
This would print the output indented.

web/tests/functional/cmdline/test_cmdline.py
```diff
@@ -230,23 +230,15 @@ def test_runs_analysis_statistics(self):
...
-
         self.assertEqual(0, ret)
         for run in json.loads(res):
             for data in run.values():
                 self.assertEqual(
                     data['analyzerStatistics']['clangsa']['failedFilePaths'],
                     [])
+                self.assertNotEqual(
+                    data['analyzerStatistics']['clangsa']['enabledCheckers'],
+                    [])
```
bruntib, line 241: We should check if the list contains a specific checker that is known to be enabled. The problem with this check is that this assertion is true if the value is `None` for some reason.

PR #4471, opened 2025-02-26
[fix] CodeChecker checkers --label option:value doesn't list checkers

bruntib:
--labels flag is working properly in these cases:

CodeChecker checkers --label
CodeChecker checkers --label severity

But doesn't work in this case:

CodeChecker checkers --label severity:HIGH

This bug is now fixed.


cservakt (approved):
LGTM

PR #4490, opened 2025-03-17
Add Vue redirect on invalid product name

Discookie:
Previously when a product endpoint was opened in the browser that was invalid on the server side, such as `localhost:8001/$$`, there were confusing errors, as the API endpoints threw inconsistent errors separately. This change adds an early bail-out when encountering such an endpoint.


bruntib:
I don't really understand this fix. I can't see any change in behavior. I get the following image when providing an invalid product name with the current CodeChecker version:
image

Discookie:
I'll fix the underlying issue in a different way.

PR #4491, opened 2025-03-17
[bugfix] Emit an error instead of crash for missing CC_LIB_DIR

Szelethus:
If you are developing CodeChecker, you should likely build the 'dev_package' target instead of 'package', as it symlinks instead of copied implementation files. However, CodeChecker will immediately crash if the CC_LIB_DIR environmental variable is not set. Instead, lets just emit an error.

### DIFF
--- a/analyzer/codechecker_analyzer/cmd/analyzers.py
+++ b/analyzer/codechecker_analyzer/cmd/analyzers.py
@@ -14,6 +14,8 @@ analyzers available in CodeChecker.
 import argparse
 import subprocess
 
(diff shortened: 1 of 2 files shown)


bruntib (changes requested):
Thanks, this is a long-awaited fix. However, I think the solution should be removing `realpath()` here, which resolves the symlink that is used in `dev_package`:
https://github.com/Ericsson/codechecker/blob/90df5ab5a6748ad55b0ad5aebf7da30d765089f6/codechecker_common/cli.py#L98

PR #4497, opened 2025-03-20
fix(report-converter): Support null column in eslint reports

SweetVishnya:
ESLint reports may contain bugs with "column": null. Thus, report-converter tool raises an exception:

TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

We set column number to 0 in such cases.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/analyzers/eslint/analyzer_result.py
+++ b/tools/report-converter/codechecker_report_converter/analyzers/eslint/analyzer_result.py
@@ -59,7 +59,7 @@ class AnalyzerResult(AnalyzerResultBase):
                     get_or_create_file(
                         os.path.abspath(file_path), file_cache),
                     int(bug['line']),
-                    int(bug['column']),
+                    int(bug['column']) if bug['column'] else 0,
                     bug['message'],
                     bug['ruleId']))
 


SweetVishnya:
@bruntib, can you take a look, please?

bruntib (approved):
Thanks for the fix! I will approve this change, but could you, please show an ESLint report where the column number is not available?
This question has two motivations:
- I would just like to make sure that this fix won't hide a potential parser bug of the report converter. 
- I would like to test that it causes any rendering issue in CodeChecker GUI to place a message bubble to a non-existing column number.

Thank you!

SweetVishnya:
@bruntib, here are some examples:

```json
      {
        "ruleId": "jsdoc/newline-after-description",
        "severity": 2,
        "message": "There must be no newline after the description of the JSDoc block.",
        "line": 38,
        "column": null,
        "nodeType": "Block",
        "endLine": 38,
        "endColumn": null
      },
      {
        "ruleId": "jsdoc/require-description-complete-sentence",
        "severity": 2,
        "message": "Sentence must end with a period.",
        "line": 116,
        "column": null,
        "nodeType": "Block",
        "endLine": 116,
        "endColumn": null
      },
      {
        "ruleId": "jsdoc/check-types",
        "severity": 2,
        "message": "Invalid JSDoc @param \"opt_propertyBag\" type \"Object\"; prefer: \"object\".",
        "line": 233,
        "column": null,
        "nodeType": "Block",
        "endLine": 233,
        "endColumn": null
      },
      {
        "ruleId": "jsdoc/check-tag-names",
        "severity": 2,
        "message": "Invalid JSDoc tag (preference). Replace \"constructor\" JSDoc tag with \"class\".",
        "line": 234,
        "column": null,
        "nodeType": "Block",
        "endLine": 234,
        "endColumn": null
      },
```

PR #4499, opened 2025-03-20
Update psutils to version with wheels

elupus:
The 5.8.0 version has very limited set of wheels available.  Update to 5.9.8 which has wheels for many more python versions and os:es

https://pypi.org/project/psutil/5.8.0/#files
Vs

https://pypi.org/project/psutil/5.9.8/#files


bruntib:
Another PR is being created: https://github.com/Ericsson/codechecker/pull/4501
Would you think, that will solve this issue?

elupus:
Yes that would work as well. Its more future proof. At the risk of some broken combinations, but that is better handled by packagers.

PR #4512, opened 2025-03-28
[fix] Display chronological order in GUI

bruntib:
The report annotation "chronological_order" uses underscore instead of dash. Due to this mistake chronological order doesn't appear in the GUI.


dkrupp (commented):
LGTM

dkrupp (approved):
lgtm

PR #4514, opened 2025-03-31
[WIP] Add tests for Bazel

Szelethus:
I'm still ironing out a few things, so WIP for now.


dkrupp (changes requested):
It's really great that you made this test working! Congrats for that!

I see that these tests are coming from the https://github.com/Ericsson/codechecker_bazel/
repo.

These tests are testing the bazel rules laid out in the bzl files in that repo:  codechecker_config, codechecker, codechecker_test, code_checker_test, clang_tidy_test etc.

I think a much more natural place for this github action would be in the codechecker_bazel repo, beacause they are essentially testing the implementation of those rules.

It would look odd, to need to modify these tests in the codechecker repo, when an action is renamed or rewritten in the other bazel repo.

The codechekcer bazel repo has CodeChecker as its dependency and not the other way around.

I understand the motivation for having a regression test in codechecker as codechecker changes more frequently, but still I would not duplicate test code this way.

Since codechecker command line is fairly stable (or backward compatible), I would think these tests would not break frequently. 

Vscode plugin (which also depends on CodeChecker command line IF) tests are also not placed in the CodeChecker repo, rather in the vscode plugin repo. 

So I think we should follow a similar login there.

What do you think?

Szelethus:
Moved tests here.
https://github.com/Ericsson/codechecker_bazel/pull/5

PR #4517, opened 2025-04-02
integrated signum fetching and using it as optional username

feyruzb:
integrated fetching of signum from `id_token` so it can be used as username for Microsoft.
Can be configured in configuration file what will be username:
`"username": "mail"` or `"username": "signum"`.

### DIFF
--- a/docs/web/authentication.md
+++ b/docs/web/authentication.md
@@ -378,7 +378,11 @@ CodeChecker also supports OAuth-based authentication. The `authentication.method
 
       * `user_groups_url`
 
-          `Microsoft` specific field to request security groups that the user is member of.
+          `Microsoft`-specific field used to request security groups that the user is member of.
+
+      * `jwks_url`
+
+          `Microsoft`-specific field used to request public signing keys for decoding JWT tokens.
 
       * `scope`
 
--- a/docs/web/authentication.md
+++ b/docs/web/authentication.md
@@ -391,12 +391,26 @@ CodeChecker also supports OAuth-based authentication. The `authentication.method
           * `username`
 
               Field for the username.
-          * `email`
 
-              Field for the email.
-          * `fullname`
+        The `username` field defines what value will be used as the user's unique identifier (i.e. their "username") depending on the OAuth provider.
 
-              Field for the fullname.
+        Supported Providers and Options:
+
+        `github`
+          * Key: `login`
+          * Important: To ensure consistency with other providers, we currently fetch the user's primary email address instead.
+          * If the user's email is hidden (private), this will result in an error.
+
+        `google`
+          * Key: `email` – uses the user's email address.
+
+        `microsoft`
+
+          * You can choose between:
+
+            Key: `mail` – uses the user's email address.
+
+            Key: `signum` – uses the user's Signum (a company-specific unique identifier).
 ~~~{.json}
 "method_oauth": {
       "enabled": false,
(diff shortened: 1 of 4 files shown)


dkrupp (changes requested):
Please add the documentation of this new field to this document
https://github.com/Ericsson/codechecker/blob/master/docs/web/authentication.md#oauth-authentication

Is the handling of groups intentionally added to this PR?
Can you please factor it out into another PR? (anyway the graph API implementation is already in the baseline code, so I am not sure we need it at all)

web/server/codechecker_server/api/authentication.py
```diff
@@ -368,21 +369,45 @@ def performLogin(self, auth_method, auth_string):
...
+                    jwks_response.raise_for_status()
+                    jwks_fetched = jwks_response.json()
+
+                    jwt_decoder = JsonWebToken(['RS256'])
+                    claims = jwt_decoder.decode(id_token, key=jwks_fetched)
+                    claims.validate()
+
                     user_groups_url = oauth_config["user_groups_url"]
-                    response = oauth2_session.get(user_groups_url).json()
+                    group_response = oauth2_session.get(user_groups_url)
```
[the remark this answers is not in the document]
feyruzb: I don't use id token for fetching groups, that is still using `Microsoft graph API` but groups handling as it was before wasn't secure enough and could fail without notifying us.
That is why this more secure way was introduced.
As per request, I will create a separate pr for upgrading group fetching.

dkrupp (changes requested):
please simplify the user_info_mapping configuration

docs/web/authentication.md
```diff
@@ -391,32 +391,46 @@ CodeChecker also supports OAuth-based authentication. The `authentication.method
...
+      "example_provider": {
+        "enabled": false,
+        "client_id": "client id",
+        "client_secret": "client secret",
+        "authorization_url": "https://accounts.google.com/o/oauth2/auth",
+        "callback_url": "http://localhost:8080/login/OAuthLogin/provider",
+        "token_url": "https://accounts.google.com/o/oauth2/token",
+        "user_info_url": "https://www.googleapis.com/oauth2/v1/userinfo",
+        "user_emails_url": "https://api.github.com/user/emails",
+        "scope": "openid email profile",
```
dkrupp, line 427: I am getting this error at authentication 
User info fetch failed: "jwks_url'

I guess it is missing from the config file

dkrupp (changes requested):

web/server/codechecker_server/api/authentication.py
```diff
@@ -373,41 +374,67 @@ def performLogin(self, auth_method, auth_string):
...
                     response = oauth2_session.get(user_groups_url).json()
+
                     for group in response["value"]:
                         if group.get("onPremisesSyncEnabled") and \
                                 group.get("securityEnabled"):
                             groups.append(group["displayName"])
-                username = user_info[
-                    oauth_config["user_info_mapping"]["username"]]
-                LOG.info("User info fetched, username: %s", username)
+
```
dkrupp, line 396: please print the username and the groups of the user in debug log level

dkrupp (changes requested):

docs/web/authentication.md
dkrupp: But why is it still missing from this authentication md .json example?

dkrupp (changes requested):

dkrupp (approved):
LGTM

PR #4520, opened 2025-04-04
report_converter: add support for ruff json input

melven:


melven:
Ok, this still fails but I don't really understand why. I tried to make the ruff test very similar to the pylint test.

melven:
Sorry @bruntib 

This was a problem with an absolute path in the report file from ruff. I replaced it with a relative path for testing.

melven:
Hi @bruntib 

Now the unit test succeeds. There is one further problem with the GUI but I don't understand the error message or why this is related to my changes...

tools/report-converter/tests/unit/analyzers/ruff_output_test_files/Makefile
```diff
@@ -0,0 +1,3 @@
+# copied from pylint, could
```
gulyasgergely902, line 1: This comment in unnecessary here. Please remove.

gulyasgergely902 (approved):
Great work overall with a minor "issue". Even with that I approve the change.

PR #4521, opened 2025-04-07
Restrict the SQL database creation to the config directory

Discookie:
Previously, SQL files could be created anywhere on the server. This change restricts them always to inside the config directory.

The old behavior can be brought back by creating symlinks manually in the config dir.


bruntib (changes requested):
Please, fix the web tests, otherwise looks good.

PR #4527, opened 2025-04-11
Enhance --trim-path-prefix to also trim paths in report messages

gamesh411:
The clang-diagnostic-implicit-int-conversion clang warning mentions absolute file system paths to refer to unnamed enums with source locations.
The feature --trim-path-prefix also seems aligned with the feature of removing this path prefix from the diagnostic messages.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/report/__init__.py
+++ b/tools/report-converter/codechecker_report_converter/report/__init__.py
@@ -13,7 +13,9 @@ import itertools
 import json
 import logging
 import os
+import re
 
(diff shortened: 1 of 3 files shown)


gamesh411:
The clang-diagnostic-implicit-int-conversion clang warning mentions absolute file system paths to refer to unnamed enums with source locations.
The feature --trim-path-prefix also seems aligned with the feature of removing this path prefix from the diagnostic messages.

dkrupp:
@bruntib  could you please review this PR?

PR #4530, opened 2025-04-13
[feat] Add JSCPD report converter

noraz31:
Introduce code similarity detection to CodeChecker by adding a JSCPD report converter


tools/report-converter/codechecker_report_converter/analyzers/jscpd/parser.py
```diff
@@ -0,0 +1,93 @@
...
+                reports.append(report)
+        return reports
+
+    def get_report(self, file_path: str, duplication: dict) -> Report:
+        """ Parse a report from the given output file. """
+
+        checker_name = 'jscpd'
+        duplication_file_1 = duplication['firstFile']
+        file_path_1 = os.path.join(
+            os.path.dirname(file_path),
```
bruntib, line 50: The generated paths by default are relative, but in most cases not relative to this input json file's directory. I would suggest removing this `os.path.join()` and use the file path as is. If the file isn't found then we should suggest the usage of `jscpd -a` flag that generates absolute paths.

tools/report-converter/tests/unit/analyzers/jscpd_output_test_files/jscpd-report.json
```diff
@@ -0,0 +1,133 @@
+{
```
bruntib, line 1: In the test file we should check the generated plist and check its content. Currently the test doesn't use the report converter.

bruntib (changes requested):

tools/report-converter/codechecker_report_converter/analyzers/jscpd/parser.py
noraz31: I've updated the code accordingly. Thanks!

bruntib (approved):

PR #4533, opened 2025-04-14
Extended tests for OAuth

feyruzb:


web/api/authentication.thrift
```diff
@@ -69,6 +73,18 @@ service codeCheckerAuthentication {
                       2: string authString)
                       throws (1: codechecker_api_shared.RequestFailed requestError),
 
+  // Handles reguesting state stored in database for OAuth
+  bool validateOAuthSession(1: string state),
+
+  // Handles reguesting access_token stored in database for OAuth verification
+  bool validateOAuthTokenSession(1: string access_code)
```
dkrupp, line 80: Remove this from the public API! This is a huge security vulnerability! This way, an unauthenticated user can scan for valid oauth access tokens!

Move this function to be server internal and use it from the test.

web/server/codechecker_server/api/authentication.py
```diff
@@ -245,11 +245,62 @@ def createLink(self, provider):
                 code_verifier=pkce_verifier
             )
 
-        self.__insertOAuthSession(state=state,
-                                  code_verifier=pkce_verifier,
-                                  provider=provider)
+        self.insertOAuthSession(state=state,
```
dkrupp, line 248: leave this as an internal API function

web/server/codechecker_server/api/authentication.py
```diff
@@ -245,11 +245,62 @@ def createLink(self, provider):
...
+                                code_verifier=pkce_verifier,
+                                provider=provider)
         return url
 
+    @timeit
+    def validateOAuthTokenSession(self, access_token):
```
dkrupp, line 254: don't add it as a public function. Make it a server internal function. You can also move this db query to the test code.

web/server/codechecker_server/api/authentication.py
```diff
@@ -245,11 +245,62 @@ def createLink(self, provider):
...
+        with DBSession(self.__config_db) as session:
+            access_token_db, *_ = \
+                session.query(OAuthToken.access_token) \
+                .filter(OAuthToken.access_token == access_token) \
+                .first()
+        return access_token_db is not None \
+            and access_token_db == access_token
+
+    @timeit
+    def validateOAuthSession(self, state):
```
dkrupp, line 270: don't add it as a public function. Make it a server internal function. You can also move this db query to the test code.

web/server/codechecker_server/api/authentication.py
```diff
@@ -245,11 +245,62 @@ def createLink(self, provider):
...
+        state_db = None
+        with DBSession(self.__config_db) as session:
+            state_db, *_ = \
+                session.query(OAuthSession.state) \
+                .filter(OAuthSession.state == state) \
+                .first()
+        return state_db is not None and state_db == state
+
+    @timeit
+    def getOAuthRow(self, state):
```
dkrupp, line 285: This is a test helper. move this to the test code.

web/tests/functional/authentication/test_authentication.py
```diff
@@ -206,11 +206,24 @@ def try_login(self, provider, username, password):
             f"&code_challenge_method={code_challenge_method}" \
             f"&code_challenge={code_challenge}"
 
+        validate_result = auth_client.validateOAuthSession(state)
```
dkrupp, line 209: here you are testing if the oauth session was created properly,right?

web/tests/functional/authentication/test_authentication.py
```diff
@@ -206,11 +206,24 @@ def try_login(self, provider, username, password):
...
         self.session_token = auth_client.performLogin(
             "oauth", provider + "@" + auth_string)
 
         return self.session_token
 
+    def test_oauth_token_session(self):
+        auth_client = env.setup_auth_client(
+            self._test_workspace, session_token='_PROHIBIT')
+
+        self.try_login("github", "admin_github", "admin")
```
dkrupp, line 222: you could also create an assertion here for the return value of the try_login() function as it should be a valid codechecker session token

dkrupp (changes requested):
1) Remove the added functions from the thrift API!
the thrift api must not change! Opening up the thrift api for testing creates a huge security risk.
2) Please add negative test cases too for the performLogin function. If the user calls back with an invalid authstring then the authentication must be denied by the server.
2) Please add a testcase oauthsession expiration. When it is expired, the oauthsession record must be removed from the db and the performLogin()  authentication must be denied.
3) Please add a testcase for oauthtoken expiration. A oauth token expiration, the CodeCheker session should also expire and be removed. Then all API requests to CodeChecker (which need authentication) should fail afterwards.

web/client/codechecker_client/helpers/authentication.py
```diff
@@ -48,6 +48,21 @@ def performLogin(self, auth_method, auth_string):
     def createLink(self, provider):
         pass
 
+    @thrift_client_call
+    def validateOAuthSession(self, state):
+        pass
+
+    @thrift_client_call
+    def validateOAuthTokenSession(self, access_code):
```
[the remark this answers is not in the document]
feyruzb: Fixed.

web/tests/functional/authentication/test_authentication.py
feyruzb: Yes.

Discookie (changes requested):
In general these are still very much surface-level tests. I'd be much more interested in seeing the invariants around username, expiration time, timeout of login attempts, setting permission levels properly etc. enforced by asserts, instead of just the fact that a login happened.
I'd also like to eventually see more tests for malformed OAuth login attempts, provider="", PKCE missing etc., but those can always be added on in separate PRs.

Seconded the strong caution against exposing the internal functions via a public interface.

web/tests/functional/authentication/oauth_server.py
```diff
@@ -84,19 +118,39 @@ def login_tester(self):
...
+                    self.code_challenges[code] = {
+                        "code_challenge": code_challenge,
+                        "code_challenge_method": params[
+                            'code_challenge_method']}
+
                     return self.show_json({"code": code,
-                                           "state": state,
-                                           "code_challenge": code_challenge,
-                                           "code_challenge_method": ccm})
+                                           "state": state})
```
Discookie, line 134: Is this correct? Showing an arbitrary state and a valid code seems weird here.

Discookie (commented):
The tests themselves look good now, much better coverage than before.
I'm still missing an assert for creating a link for an invalid provider, and verifying the username/email somewhere, but other than that the scope looks fine now for this PR.

One thing to note is that some of these tests will break the moment #4559 is merged, because these also rely on provider names.
Some testcases will need to be changed to use templates instead.

web/tests/functional/authentication/test_authentication.py
```diff
@@ -207,38 +215,247 @@ def try_login(self, provider, username, password):
...
+                                 provider=provider_r,
+                                 expires_at=expires_at_r)
+
+        result = env.validate_oauth_session(session_factory, state_r)
+        # validate that the session was inserted
+        self.assertTrue(result, "Session that will be removed "
+                        "was not inserted")
+
+        # user that should login successfully and remove the old session
+        session_token = self.try_login("github", "admin_github", "admin")
```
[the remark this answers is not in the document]
feyruzb: Because every login attempt is triggering a table purge of old OAuth sessions.

web/tests/functional/authentication/test_authentication.py
```diff
@@ -207,38 +215,247 @@ def try_login(self, provider, username, password):
...
+        self.assertTrue(result, "No entry found in database, "
+                        "unexpected behavior")
+
+    def test_oauth_insert_session_failure(self):
+        """
+        Testing insertion of wrong session data."""
+        session_factory = env.create_sqlalchemy_session(self._test_workspace)
+
+        with self.assertRaises(TypeError):
+            env.insert_oauth_session(session_factory,
```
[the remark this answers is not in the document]
feyruzb: Test rewritten. now tests if mock server rejects a non existent user.

web/tests/functional/authentication/test_authentication.py
```diff
@@ -207,38 +215,247 @@ def try_login(self, provider, username, password):
...
+
+        with self.assertRaises(RequestFailed):
+            self.try_login("github", "user_csrf", "user")
+
+        with self.assertRaises(RequestFailed):
+            self.try_login("google", "user_csrf", "user")
+
+    def test_oauth_pkce_attack_protection(self):
+        """
+        Tests if after logic authorization server returns a code_verifier
```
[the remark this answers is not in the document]
feyruzb: Fixed, rephrased.

web/tests/functional/authentication/test_authentication.py
```diff
@@ -207,38 +215,247 @@ def try_login(self, provider, username, password):
...
+
+        with self.assertRaises(RequestFailed):
+            self.try_login("github", "user_pkce", "user")
+
+        with self.assertRaises(RequestFailed):
+            self.try_login("google", "user_pkce", "user")
+
+    def test_oauth_incomplete_token_data(self):
+        """
+        Tests if the token data was received incomplete results
```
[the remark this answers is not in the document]
feyruzb: All information besides access_token is missing in returned data. 
In real authorization code, presence of all needed fields is checked separately.

dkrupp (changes requested):
Looks much better now. Could you please check my further comments?

web/tests/functional/authentication/test_authentication.py
```diff
@@ -215,37 +223,247 @@ def try_login(self, provider, username, password):
...
+        """
+        auth_client = env.setup_auth_client(
+            self._test_workspace, session_token='_PROHIBIT')
+
+        link_github = auth_client.createLink("github")
+        link_google = auth_client.createLink("google")
+
+        self.assertIsNotNone(link_github,
+                             "Authorization link for Github created empty")
+
```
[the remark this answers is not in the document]
feyruzb: Fixed, added asserts for states of created links to validate their existence in database.

web/tests/functional/authentication/oauth_server.py
```diff
@@ -84,19 +118,39 @@ def login_tester(self):
             if "username" in query_params:
                 query = f"{params['username']}:{params['password']}"
                 query_result = self.users_by_data.get(query, None)
-                if query_result:
-                    state = params['state']
+                # csrf attack case
+                if params['username'] == "user_csrf":
```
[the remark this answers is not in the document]
feyruzb: Great suggestion — I removed the special case from the mock server and instead added the corrupted state directly at the same point in the code where the PKCE attack case is handled.

web/tests/functional/authentication/test_authentication.py
```diff
@@ -215,37 +223,247 @@ def try_login(self, provider, username, password):
...
-        auth_string = f"{link}?code={code}&state={state}" \
-            f"&code_challenge_method={code_challenge_method}" \
-            f"&code_challenge={code_challenge}"
+
+        code, state = data['code'], data['state']
+        auth_string = f"{link}?code={code}&state={state}"
+
+        # PKCE attack case
+        if username == "user_pkce":
+            env.change_oauth_session_data(
```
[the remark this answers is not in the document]
dkrupp: I fixed it as I suggested. I hope it is okay with you.

PR #4539, opened 2025-04-17
Simplify oauth interface

gulyasgergely902:
What have been done:
- If only one oauth provider exists the SSO button reads "Login with <providername>" and without opening the modal redirects straight to the provider login page on click. If more than one exists, the standard SSO Login opens on click with the modal to choose the desired provider.


dkrupp (approved):
LGTM

PR #4540, opened 2025-04-17
Personal access token

bruntib:


bruntib:
I'm still planning to redesign the old API functions for token handling, because they return the token, and can't handle the unique name of a token.
I'll also write a proper commit message.

web/server/codechecker_server/database/config_db_model.py
```diff
@@ -144,6 +144,47 @@ def __init__(self, token, user_name, groups, description=None,
...
+        description,
+        auth_session_id
+    ):
+        self.user_name = user_name
+        self.token_name = token_name
+        self.token = token
+        self.description = description
+        self.auth_session_id = auth_session_id
+        self.last_access = datetime.now()
+        self.expiration = self.last_access + timedelta(days=365)
```
Discookie, line 185: This default expiration should be configurable, and ideally, user-selectable (with a configurable upper limit).

web/server/vue-cli/src/components/Layout/PersonalAccessTokenBtn.vue
```diff
@@ -0,0 +1,182 @@
...
+        </p>
+        <p>
+          Description: <strong>{{ token.description }}</strong>
+        </p>
+        <p>
+          Make sure to save this token, because it won't be available anymore
+          after closing this window.
+        </p>
+      </template>
+    </confirm-dialog>
```
Discookie, line 49: It would be nice if instead of a Cancel button here, there was a Copy to clipboard button, or something similar.

docs/web/authentication.md
```diff
@@ -631,17 +631,25 @@ Personal tokens can be written instead of the user's password in the
...
                                  [--url SERVER_URL]
                                  [--verbose {info,debug,debug_analyzer}]
 
 Creating a new personal access token.
 
-optional arguments:
+options:
   -h, --help            show this help message and exit
+  --name NAME           A unique name that identifies the access token.
+                        (default: None)
```
Discookie, line 643: Wouldn't it make sense to make this a positional argument, same as when deleting a token?

web/server/codechecker_server/api/authentication.py
```diff
@@ -721,3 +723,102 @@ def getTokens(self):
                     str(t.last_access)))
 
             return result
+
+    @timeit
+    def newPersonalAccessToken(self, name, description):
```
Discookie, line 728: Please trim whitespace from the name of the PAT - unexpected spaces already caught me off-guard, and they don't show in the web UI.

gulyasgergely902 (approved):

Discookie (commented):
Couple usability snags I found skimming over it:
- The no-permission error when creating tokens is always `The server must be start by using privilaged access to execute this action.`, even when the user is not logged in.
- You can log in with a PAT from the web interface via the username/password login flow. This causes the browser to stay logged in until the PAT expires.

web/client/codechecker_client/token_client.py
```diff
@@ -38,7 +38,7 @@ def handle_add_token(args):
     client = init_auth_client(protocol, host, port)
 
     description = args.description if 'description' in args else None
-    session = client.newToken(description)
+    session = client.newPersonalAccessToken(args.name, description)
```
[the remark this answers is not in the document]
bruntib: Thanks, I fixed it!

dkrupp (changes requested):
When LDAP and Oauth authentication is allowed, the personal access token based login does not work from the CLI.  The reason is likely that authentication is tried through LDAP first and then when it fails because the personal access token does not match the LDAP password. So personal access based authentication should always be tried first and successful, the user should be let in, otherwise the other password based authentication methods can be tried.

web/server/vue-cli/src/components/Layout/PersonalAccessTokenBtn.vue
```diff
@@ -0,0 +1,182 @@
...
+                  <th>Last access</th>
+                  <th>Expiration</th>
+                  <th>Description</th>
+                  <th>Delete</th>
+                </tr>
+              </thead>
+              <tbody>
+                <tr v-for="tok in accessTokens" :key="tok.token">
+                  <td>{{ tok.name }}</td>
+                  <td>{{ tok.lastAccess }}</td>
```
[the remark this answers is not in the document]
bruntib: done

web/server/codechecker_server/database/config_db_model.py
bruntib: Will be done later.

docs/web/authentication.md
bruntib: Done

web/server/codechecker_server/api/authentication.py
bruntib: Done

dkrupp (changes requested):
The PA login with LDAP users is now fine, thanks for the fix.

I noticed that with older clients (6.25.1), the CodeChecker cmd token list returns the session tokens instead of the PA tokens.

dkrupp (approved):
LGTM

PR #4546, opened 2025-04-30
[feat] Return custom message in cli upon failed authentication

noraz31:
A customizable message from the server config file should be shown in cli upon failed authentication via CodeChecker cmd login.
As part of this PR, such message was added to the server config file for feature demonstration. This can be removed before merging.

### DIFF
--- a/web/server/config/server_config.json
+++ b/web/server/config/server_config.json
@@ -17,6 +17,7 @@
     "enabled" : false,
     "realm_name" : "CodeChecker Privileged server",
     "realm_error" : "Access requires valid credentials.",
+    "failed_auth_message": "",
     "session_lifetime" : 300,
     "refresh_time" : 60,
     "logins_until_cleanup" : 30,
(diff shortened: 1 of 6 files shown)


web/server/config/server_config.json
```diff
@@ -17,6 +17,7 @@
     "enabled" : false,
     "realm_name" : "CodeChecker Privileged server",
     "realm_error" : "Access requires valid credentials.",
+    "failed_auth_message": "Personal access token based authenication is allowed only.",
```
dkrupp, line 20: I think you should just leave it blank. Then the user will know that it is a possibility.

"failed_auth_message": "",

dkrupp (changes requested):
Please add a description of the new configuration parameter in 
https://github.com/Ericsson/codechecker/blob/master/docs/web/authentication.md#server-side-configuration

Otherwise looks good and works as expected.

web/server/config/server_config.json
noraz31: Removed the message

dkrupp (approved):
LGTM. Thanks!

PR #4550, opened 2025-05-05
[version] Bump python version to 3.9

### LINKED ISSUE #4549: Bump minimum Python from 3.8 to 3.9
Python 3.8 is [EOL](https://devguide.python.org/versions/#unsupported-versions) since 2024-10-07

pdgendt:
Python 3.8 is EOL, bump to 3.9.

Fixes #4549

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -17,7 +17,7 @@ apps:
     command: CodeChecker/bin/CodeChecker
     environment:
       PATH: ${SNAP}/usr/bin:${PATH}
-      PYTHONPATH: ${SNAP}/usr/lib/python3.8
+      PYTHONPATH: ${SNAP}/usr/lib/python3.9
   report-converter:
     command: CodeChecker/bin/report-converter
   post-process-stats:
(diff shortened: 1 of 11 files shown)


pdgendt:
I also do wonder why CI runs **only** with 3.8 currently. It would make sense to update the PR to run CI with `3.9` - `3.13`.

snap/snapcraft.yaml
```diff
@@ -17,7 +17,7 @@ apps:
     command: CodeChecker/bin/CodeChecker
     environment:
       PATH: ${SNAP}/usr/bin:${PATH}
-      PYTHONPATH: ${SNAP}/usr/lib/python3.8
+      PYTHONPATH: ${SNAP}/usr/lib/python3.9
```
pdgendt, line 20: I honestly don't know if this is correct.

pdgendt:
Any feedback here?

snap/snapcraft.yaml
Discookie: This is all that can be done with snaps to specify a Python version.

Discookie (approved):
No feedback other than LGTM :D 

Running CI with all possible Python versions would be a bit overkill IMO - I think it's enough to run with the lowest supported version, which is now 3.9.
We updated the Docker image's Python version a while back, but missed Snap and others, so thanks!

PR #4551, opened 2025-05-05
Utilize personal access token expiration date

gulyasgergely902:


web/server/codechecker_server/migrations/config/versions/7ed50f8b3fb8_new_table_for_personal_access_tokens.py
```diff
@@ -50,25 +43,6 @@ def upgrade():
...
-
-    op.execute(
-        f"""
-        INSERT INTO personal_access_tokens (user_name, token_name, token,
-            description, last_access, expiration, auth_session_id)
-        SELECT user_name, {random_string}, token, description, last_access,
-            '{one_year_later}', id
-        FROM auth_sessions
-        WHERE can_expire = false
-        """)
```
bruntib, line 71: The personal access tokens should still be migrated from the `auth_sessions` table. We wouldn't like to lose the already generated personal access tokens.

web/server/codechecker_server/session_manager.py
```diff
@@ -495,6 +480,10 @@ def __try_personal_access_token(self, auth_string):
         if not personal_access_token:
             return False
 
+        if (personal_access_token.expiration - datetime.now()) \
```
bruntib, line 483: It is a question of coding style, but can these date fields be simply compared with the `<` operator? It might be more intuitive than checking the result of a subtraction, but you may chose according to your preference.

bruntib (changes requested):

bruntib (approved):

PR #4559, opened 2025-05-13
Add OAuth templates, simplify OAuth configuration flow

Discookie:
Simplified the configuration of OAuth by introducing variables and templates.
In the general case, the user shouldn't ever need to specify any OAuth endpoint URLs.

Instead of relying on the name of the OAuth provider, the configuration now uses so-called "templates" to specify a behavior when logging in. 
Even if the user wants to override a default URL, `{variables}` such as `{host}` help streamline defining URLs.

Example OAuth configuration using templates:
```jsonc
"method_oauth": {
  "enabled": true,
  "shared_variables" : {
    "host": "https://example.com:8123" // must include protocol
  },
  "providers" : {
    "microsoft": {
      "enabled": true,
      "client_id": "<ExampleClientID>",
      "client_secret": "<ExampleClientSecret>",

      "template": "ms_entra/v2.0",
      "variables": {
        "tenant_id": "common" // replace with tenant ID, if not using multi-tenant mode
      },

      "user_info_mapping": { // optional
        "username": "email"
      }
    }
  }
}
```

while an URL could be specified as: ``"authorization_url": "https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/authorize"``

Currently a template must be specified, but in the future there could be handling for a "default" OAuth endpoint type.
Supported providers are the same ones as before: `github => github/v1`, `google => google/v1`, `microsoft => ms_entra/v2.0`, named after the OAuth API versions of each service.

[See the docs for more details.](https://github.com/Discookie/codechecker/blob/oauth-config/docs/web/authentication.md#oauth-authentication)

[See the oauth_templates.py file for the currently available templates.](https://github.com/Discookie/codechecker/tree/oauth-config/web/codechecker_web/server/oauth_templates.py)

### DIFF
--- a/docs/web/authentication.md
+++ b/docs/web/authentication.md
@@ -337,8 +337,12 @@ Specific behavior related to each provider is configured by a provider `template
 
   * `shared_variables`
 
-      A key-value table that is used to set variables across all providers, for convenience.
-      If using the `host` variable, it should be in the format `http://example.com`, including the protocol.
+    A key-value table that is used to set variables across all providers, for convenience.
+
+    Any variable can be specified. If using the `host` variable, it should be in the format `https://example.com`, including the protocol.
+    
+    The `callback_url`'s default value uses the `host` and `provider` variables.
+    Template `ms_entra/v2.0` uses the `tenant_id` variable.
 
  * `providers`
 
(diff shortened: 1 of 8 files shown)


docs/web/authentication.md
```diff
@@ -402,73 +442,73 @@ CodeChecker also supports OAuth-based authentication. The `authentication.method
 
         Currently there are only 2 options for this:
          * `username`
-         * `email`
+         * `email` (*default*)
 
         Expected output for each provider:
 
-        `github`
+        `github/v1`
```
dkrupp, line 449: the provider name is "github" not github/v1

docs/web/authentication.md
```diff
@@ -402,73 +442,73 @@ CodeChecker also supports OAuth-based authentication. The `authentication.method
...
           * `username` - user's GitHub `login` will be user's identifier
           * `email` - a request will be sent to fetch the primary email of account to be used as the user's identifier.
             If it is hidden, an error will be thrown.
 
-        `google`
+        `google/v1`
           * Only supports `email`, and user's Gmail email will be considered his username.
 
-        `microsoft`
+        `ms_entra/v2.0`
```
dkrupp, line 457: provider name is microsoft?

docs/web/authentication.md
```diff
@@ -402,73 +442,73 @@ CodeChecker also supports OAuth-based authentication. The `authentication.method
...
-      "authorization_url": "https://login.microsoftonline.com/<tenant-id>/oauth2/v2.0/authorize",
-      "callback_url": "https://<server_host>/login/OAuthLogin/microsoft",
-      "token_url": "https://login.microsoftonline.com/<tenant-id>/oauth2/v2.0/token",
-      "user_groups_url": "https://graph.microsoft.com/v1.0/me/memberOf",
-      "jwks_url": "https://login.microsoftonline.com/<tenant-id>/discovery/v2.0/keys",
-      "user_info_url": "https://graph.microsoft.com/v1.0/me",
-      "scope": "User.Read email profile openid offline_access",
+      "template": "google/v1", // still needed for the response handling logic
+      "authorization_url": "https://accounts.google.com/o/oauth2/auth",
+      "callback_url": "https://<server_host>/login/OAuthLogin/google",
```
dkrupp, line 501: can the user use variables also here or only in the templates?

docs/web/authentication.md
```diff
@@ -402,73 +442,73 @@ CodeChecker also supports OAuth-based authentication. The `authentication.method
...
       "client_id": "<ExampleClientID>",
       "client_secret": "<ExampleClientSecret>",
-      "authorization_url": "https://login.microsoftonline.com/<tenant-id>/oauth2/v2.0/authorize",
-      "callback_url": "https://<server_host>/login/OAuthLogin/microsoft",
-      "token_url": "https://login.microsoftonline.com/<tenant-id>/oauth2/v2.0/token",
-      "user_groups_url": "https://graph.microsoft.com/v1.0/me/memberOf",
-      "jwks_url": "https://login.microsoftonline.com/<tenant-id>/discovery/v2.0/keys",
-      "user_info_url": "https://graph.microsoft.com/v1.0/me",
-      "scope": "User.Read email profile openid offline_access",
+      "template": "google/v1", // still needed for the response handling logic
```
dkrupp, line 499: are all config_key/value pair is overridden by these definitions?

docs/web/authentication.md
```diff
@@ -328,12 +328,18 @@ groups. For more information [see](permissions.md#managing-permissions).
...
 #### OAuth Configuration options <a name="oauth-configuration-options"></a>
   * `enabled`
 
     Indicates if OAuth authentication is enabled (required for any methods below)
 
+  * `shared_variables`
+
+      A key-value table that is used to set variables across all providers, for convenience.
+      If using the `host` variable, it should be in the format `http://example.com`, including the protocol.
+
```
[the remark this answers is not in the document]
Discookie: Any variable can be specified and used. Added details about the variables used by the templates' default values.

web/server/config/server_config.json
```diff
@@ -56,47 +56,45 @@
...
-          "user_groups_url" : "https://graph.microsoft.com/v1.0/me/memberOf",
-          "jwks_url": "https://login.microsoftonline.com/<tenant-id>/discovery/v2.0/keys",
-          "user_info_url": "https://graph.microsoft.com/v1.0/me",
-          "scope": "User.Read email profile openid offline_access",
+          "template": "ms_entra/v2.0",
+          "variables": {
+            "tenant_id": "common"
+          }
+        },
+        "custom": {
```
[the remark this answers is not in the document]
Discookie: Left in by accident - a relic from when the `default` template was still considered accepted. Removed.

docs/web/authentication.md
Discookie: Each available template, instead.

docs/web/authentication.md
Discookie: They can, correct. Added example.

docs/web/authentication.md
Discookie: The template sets default values, which can be overridden by the user. The template's default will never have higher priority.

dkrupp (approved):
Thanks for the fixes. Looks good now.

PR #4562, opened 2025-05-15
Ensure the compiler has no L18Ned output

cmorty:
Set LC_ALL to C, as the parser will not detect include paths, if the compiler output is localized.


cmorty:
@bruntib : I updated the code to just set `LC_ALL` and added a comment.

To reproduce, install a locale supported by `gcc` (e.g. `de_DE.UTF-8`). Set the locale (`export LANG=de_DE.UTF-8`) and run CodeChecker on a `compile_commands.json` using `gcc`. `compiler_includes` in `compiler_info.json` will be empty.

bruntib (approved):
Thanks for the hint. I couldn't reproduce the issue tough, but the change seems safe.

PR #4565, opened 2025-05-15
Add plist documentation

noraz31:


docs/tools/plist.md
```diff
@@ -0,0 +1,301 @@
+# Static Analysis PLIST Format Specification
+
+Plist files are used in CodeChecker to store and parse diagnostic 
+information from static analysis tools.
+This documentation explains the structure of plist files generated by
```
dkrupp, line 5: parsed by CodeChecker

docs/tools/plist.md
```diff
@@ -0,0 +1,301 @@
...
+  * [1. `bug_path_event` — Bug Path Events](#1-bug_path_event--bug-path-events)
+  * [2. `notes` — Notes](#2-notes--notes)
+  * [3. `macro_expansions` — Macro Expansions](#3-macro_expansions--macro-expansions)
+  * [4. `report-annotation` — Report Annotations](#4-report-annotation--report-annotations)
+* [License](#license)
+
+
+## Plist files in CodeChecker
+
+Plist files are produced by the report-converter in CodeChecker from the
```
dkrupp, line 22: When mentioning the report-converter, refer to this doc https://github.com/Ericsson/codechecker/blob/master/docs/tools/report-converter.md

docs/tools/plist.md
```diff
@@ -0,0 +1,301 @@
...
+* [Diagnostic Entry Examples](#diagnostic-entry-examples)
+  * [1. `bug_path_event` — Bug Path Events](#1-bug_path_event--bug-path-events)
+  * [2. `notes` — Notes](#2-notes--notes)
+  * [3. `macro_expansions` — Macro Expansions](#3-macro_expansions--macro-expansions)
+  * [4. `report-annotation` — Report Annotations](#4-report-annotation--report-annotations)
+* [License](#license)
+
+
+## Plist files in CodeChecker
+
```
dkrupp, line 21: Maybe you can start with this: Plist is the default static and dynamic analysis inter-change file format of CodeChecker. They are created as a results of the analysis: `CodeChecker analyze -o report_directory" invocation in the report directory. The proprietary report format of each invoked analyzer tool is converted into `plist` format, by the report-converter tools. Then these plist reports are consumed by the `CodeCheckecker "store/diff/parse` subcommands.

docs/tools/plist.md
```diff
@@ -0,0 +1,301 @@
...
+			<key>name</key>
+			<string>report-converter</string>
+			<key>version</key>
+			<string>x.y.z</string>
+		</dict>
+	</dict>
+</dict>
+</plist>
+```
+
```
dkrupp, line 84: Here please explain the meaning of each key.
* `category:string` The category of the checker emitted the report. Currently unused by the CodeChecker tooling
* `check_name:string` The name of the checker emitted the report. This checker name is visualized in the GUI and used in the report suppressions.
* `description:string` The short description of the analysis report.
...

docs/tools/plist.md
```diff
@@ -0,0 +1,301 @@
...
+      <integer>1</integer>
+      <key>line</key>
+      <integer>3</integer>
+    </dict>
+    <key>message</key>
+    <string>Second bug path event</string>
+  </dict>
+</array>
+```
+
```
dkrupp, line 198: Please describe the meaning of each keys.

what are  the kind types?
what is the meaning of depth? (how it is visualized?)

docs/tools/plist.md
```diff
@@ -0,0 +1,301 @@
...
+### 4. `report-annotation` — Report Annotations
+
+Report Annotations are a special kind of metadata associated with reports in 
+CodeChecker. They were originally introduced to incorporate dynamic analysis 
+tools better, where additional contextual data is useful. However, 
+Report Annotations can be associated with any Report, including the ones
+generated by static analysis tools.
+
+A report-annotation may include:
```
dkrupp, line 273: Supported report-annotations for storing dynamic analysis results:

docs/tools/plist.md
```diff
@@ -0,0 +1,301 @@
...
+- **`timestamp`** (`date-time`, ISO 8601): Timestamp when the issue was generated, recorded, or annotated
+- **`testcase`** (`string`): Customizable name of the Testcase
+- **`chronological_order`** (`int`): Sequence number of the issue
```
dkrupp, line 276: In dynamic analysis the order of issues are important. With the chronological order field, the order of dynamic anlaysis events can be recorded.

docs/tools/plist.md
```diff
@@ -0,0 +1,301 @@
...
+
+Report Annotations are a special kind of metadata associated with reports in 
+CodeChecker. They were originally introduced to incorporate dynamic analysis 
+tools better, where additional contextual data is useful. However, 
+Report Annotations can be associated with any Report, including the ones
+generated by static analysis tools.
+
+A report-annotation may include:
+- **`timestamp`** (`date-time`, ISO 8601): Timestamp when the issue was generated, recorded, or annotated
+- **`testcase`** (`string`): Customizable name of the Testcase
```
dkrupp, line 275: When recording dynamic analysis executions, the reports can be associated to test cases during which the report occurred.

dkrupp (commented):
Please link the document into the Developer Documentations list
https://github.com/Ericsson/codechecker?tab=readme-ov-file#developer-documentations

With the title: "Static/Dynamic Analsyis Plist Format Specification"

docs/tools/plist.md
```diff
@@ -0,0 +1,301 @@
...
+		<dict>
+			<key>category</key>
+			<string>unknown</string>
+			<key>check_name</key>
+			<string>division-by-zero</string>
+			<key>description</key>
+			<string>The message of the report object is shown here</string>
+			<key>issue_hash_content_of_line_in_context</key>
+			<string>...</string>
+			<key>location</key>
```
Discookie, line 45: This location can also be a span, which is an array of 2 dicts of col/file/line. Other locations can also optionally take the form of a span.
The Clang Static Analyzer emits location spans, among others.

Discookie (dismissed):
In general, it would be nice to have the full plist source files in a folder somewhere, the ones that are used to make the images. Something like `macro_expansion.plist` next to `macro_expansion.png`.

Please link to the Apple plist specification in some form as well.

docs/tools/plist.md
noraz31: Thanks for the review! As far as I can see, CodeChecker only generates locations to plist files as represented here. If that's not correct, could you please give me an example scenario where the location is a span?

docs/tools/plist.md
```diff
@@ -0,0 +1,394 @@
...
+</plist>
+```
+
+---
+
+## Diagnostic Entry Examples
+
+The `.plist` file may contain the following kinds of entries:
+
+### 1. `bug_path_event` — Bug Path Events
```
bruntib:
```suggestion
### 1. `path` — Bug Path Events
```

docs/tools/plist.md
```diff
@@ -0,0 +1,394 @@
...
+    </dict>
+    <key>message</key>
+    <string>This is a note</string>
+  </dict>
+</array>
+```
+
+#### UI Representation:
+
+![Web interface showing a note](images/note.png)
```
bruntib: Could the images be included directly instead of a link?

docs/tools/plist.md
```diff
@@ -0,0 +1,394 @@
...
+information from static analysis tools.
+This documentation explains the structure of plist files parsed by 
+CodeChecker, illustrating how they are represented in the code and what 
+they look like on the CodeChecker UI.
+
+## Table of Contents
+* [Plist files in CodeChecker](#plist-files-in-codechecker)
+* [Structure of a .plist File](#structure-of-a-plist-file)
+* [Diagnostic Entry Examples](#diagnostic-entry-examples)
+  * [1. `bug_path_event` — Bug Path Events](#1-bug_path_event--bug-path-events)
```
bruntib:
```suggestion
  * [1. `path` — Bug Path Events](#1-bug_path_event--bug-path-events)
```

docs/tools/plist.md
```diff
@@ -0,0 +1,394 @@
...
+ - `location:dict`
+    - `col:integer` Position of the line of the file the entry applies to.
+    - `file:integer` Enumerated value of the file the entry applies to. 
+    The corresponding files are listed under the `files` key. The indexing of 
+    these files start from 0.
+    - `line:integer` Line of the file the entry applies to.
+ - `files:string[]` File paths referenced in the location (e.g. source 
+ files, headers). Can contain relative and absolute paths. Absolute paths 
+ must start with `/`.
+- `kind:string` Specifies the type of the entry. Values used in CodeChecker
```
bruntib: I'm a little lost here. Is this `kind` the property of the whole report, or is this related to the different "Diagnostic entry"s?
Maybe a shortened example for this "kind" and also a "diagnostic entry" could go to the example .plist above. Not the full structure, but an indication of which level they belong in the structure.

docs/tools/plist.md
noraz31: The images could possibly be embedded using base64, but as far as I know this is the more common and more maintainable solution, and some other .md files in the project include images the same way. It's a fair point that the images should be with all the others used in the documentations, so I've moved these - now, it should be consistent with the other .md files

docs/tools/plist.md
noraz31: Thanks, this is a good catch. The kind property is only used within entries in the path array (which represents the bug path), so I think it's best to remove this from here and have this explanation in the 1. `path` — Bug Path Events paragraph. There are also examples in the `path` plist for both values (control, event).

PR #4566, opened 2025-05-16
[refactor] Make analyzer and checker options typed

bruntib:
The analyzer_base.get_analyzer_config() and
analyzer_base.get_checker_config() functions returned a list of tuples. The tuple members were accessed by indexing. It was hard to understand what the tuple members represent and what are their possible values.

This commit introduces AnalyzerConfig and CheckerConfig types. The functions mentioned above return the list of these objects respectively.

cc-verbatim-args-file is a common analyzer option introduced by not the analyzers but CodeChecker. The parameter of this option must be an existing file path. The typed handling of analyzer options helps to emit an error when the provided path doesn't exist.


analyzer/codechecker_analyzer/cli/analyze.py
```diff
@@ -861,85 +861,99 @@ def add_arguments_to_parser(parser):
...
     """
-    wrong_configs = []
+    wrong_config_messages = []
     supported_analyzers = analyzer_types.supported_analyzers
 
-    for cfg in analyzer_conf:
-        if cfg.analyzer not in supported_analyzers:
-            wrong_configs.append((cfg.analyzer, None))
+    analyzer_configs = {
+        analyzer_name: analyzer_class.get_analyzer_config()
```
elupus, line 875: This breaks when clangsa binaries are not existing: get_analyzer_config does not protect for missing binary.

elupus: Fix here: https://github.com/Ericsson/codechecker/pull/4660

PR #4567, opened 2025-05-19
[feat] Implement configurable Personal Acces Token expiry

gulyasgergely902:
A new optional parameter was added to cmd token new command: "--expiration". This parameter sets the expiration date by days with a minimum of 1 and the maximum of 365. If other number was set, the expiration will default to 365.

A new field was added to the web GUI to set the expiration date of the Personal Access Token.


web/client/codechecker_client/token_client.py
```diff
@@ -38,8 +38,9 @@ def handle_add_token(args):
     client = init_auth_client(protocol, host, port)
 
     description = args.description if 'description' in args else None
+    expiration = args.expiration if 'expiration' in args else None
     personal_access_token = \
-        client.newPersonalAccessToken(args.name, description)
+        client.newPersonalAccessToken(args.name, description, expiration)
 
     print("The following access token has been generated for your account: " +
           personal_access_token.token)
```
Discookie, line 46: Print either the expiration date, or the token lifetime here, if possible.

Discookie (commented):
I don't know if the server configuration for max token lifetime should be added in this PR or not.

If that's going to be a separate PR, then LGTM after my 1 comment is addressed.

web/client/codechecker_client/token_client.py
gulyasgergely902: The scope of this ticket is only for setting the expiration of the personal access tokens. Setting a max limit on server side will be handled on a separate ticket.

Discookie (approved):

PR #4568, opened 2025-05-19
[cfg] Update clang-tidy, clangsa and cppcheck configurations

gamesh411:
- Add new checks to clang-tidy config.
- Update the ClangSA configuration by adding new webkit-related checks and de-alpha-ed checkers.
- Add new checks to CppCheck config.
- Reorganise a few misplaced (according to alphabetical ordering) entries.

### DIFF
--- a/config/labels/analyzers/clangsa.json
+++ b/config/labels/analyzers/clangsa.json
@@ -248,13 +248,6 @@
       "profile:sensitive",
       "severity:HIGH"
     ],
-    "unix.BlockInCriticalSection": [
-      "doc_url:https://clang.llvm.org/docs/analyzer/checkers.html#alpha-unix-blockincriticalsection-c",
-      "profile:default",
-      "profile:extreme",
-      "profile:sensitive",
-      "severity:LOW"
-    ],
     "alpha.unix.Chroot": [
       "doc_url:https://clang.llvm.org/docs/analyzer/checkers.html#alpha-unix-chroot-c",
       "profile:extreme",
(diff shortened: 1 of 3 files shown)


gamesh411:
CppCheck rules have no documentation (nothing that I could find), so I went by their names and the similarities to other existing checks.

On the Open Source test suite, we got this report count for the not-yet categorised checks/checkers across the board:
Checker	ReportCount
core.FixedAddressDereference 5
cppcheck-checkLevelNormal 103
cppcheck-ctunullpointer 51
cppcheck-legacyUninitvar 215
cppcheck-redundantContinue 31

For the other checkers in the first commit, there were no reports.

gamesh411:
I have added these checks/checkers/warnings to the default profile:

ClangSA
core.FixedAddressDereference
    (not many false positives, but may "hide" the findings of other nullpointer finding checkers, as in the open source results, 
    there were no fixed address pointers other than the NULL macro, i.e., (void*)0 )
unix.Chroot

Clang Tidy
bugprone-tagged-union-member-count
bugprone-unintended-char-ostream-output

I am not sure whether we should add compiler warnings in general to the default profile, but these are the ones that I consider ok to be on by default:

clang-diagnostic-dangling-capture
clang-diagnostic-default-const-init
clang-diagnostic-default-const-init-field
clang-diagnostic-default-const-init-field-unsafe
clang-diagnostic-default-const-init-unsafe
clang-diagnostic-default-const-init-var
clang-diagnostic-default-const-init-var-unsafe
clang-diagnostic-uninitialized-explicit-init
clang-diagnostic-unnecessary-virtual-specifier

config/labels/analyzers/clangsa.json
```diff
@@ -1225,6 +1272,14 @@
       "profile:extreme",
       "profile:sensitive"
     ],
+    "unix.cstring.NotNullTerminated": [
```
dkrupp, line 1275: you can add it to the default profile as it only finds very basic cases. no false positives expected

bruntib (approved):

dkrupp (commented):
please see one tiny comment

PR #4581, opened 2025-05-28
Thrift upgrade

bruntib:


Discookie (changes requested):
`package-lock.json` was not updated after regenerating the API files. Did the API upgrade go through properly?
No usability issues.

Discookie (approved):

PR #4584, opened 2025-05-29
[fix] Emit error message when SQLite DB is not under workspace dir

bruntib:
When the SQLite database file's directory is given as an absolute path that is not under the workspace directory, then a meaningful error message should be printed to the user.


Discookie (approved):
One small comment, otherwise lgtm

PR #4597, opened 2025-06-10
[fix] Fix migration logging

bruntib:
Python "logging" module handles a hierarchy of loggers. Loggers can be given handlers to handle the actual log message. If a logger doesn't have a handler then the parent's handler will be used. "logger.handlers" is an array containing the direct handlers, but "logger.hasHandlers()" returns true even if the parent has a handler. So, the array can be empty and "hasHandlers()" still returns True.


noraz31 (commented):
LGTM

PR #4479, opened 2025-03-06
Cache __contains_no_intrinsic_headers and thus speedup parse_options ~2x

irishrover:
When analyzing Chromium or any other large project, __contains_no_intrinsic_headers is called thousands of times for the same dir names. It is an expensive OS call.

PR #4607, opened 2025-06-22
[fix] Parallelize parse_unique_log to speed-up ~nproc times

irishrover:
See https://github.com/Ericsson/codechecker/pull/4479#issuecomment-2704631041 parse_options is very slow.
In this PR I parallelize calls to it.

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -26,6 +26,7 @@ from typing import Dict, List, Optional
 
 from codechecker_analyzer.analyzers import clangsa
 
+from codechecker_common.compatibility import multiprocessing
 from codechecker_common.logger import get_logger
 from codechecker_common.util import load_json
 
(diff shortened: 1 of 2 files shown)


irishrover:
Gentle ping)

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -1215,12 +1217,72 @@ class CompileActionUniqueingType(Enum):
...
+            results.append((action, skip))
+
+        except Exception as ex:
+            LOG.debug("Error parsing options: %s", str(ex))
+            results.append((None, True))
+
+    return results
+
+
+def _create_task_bundles(tasks, chunk_size):
```
bruntib: `codechecker_common.util.chunks()` is a similar function. Could that be used instead?

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -1292,19 +1360,44 @@ def parse_unique_log(compilation_database,
...
-
-            # Skip parsing the compilaton commands if it should be skipped
+            args = (
+                entry,
+                compiler_info_file,
+                keep_gcc_include_fixed,
+                keep_gcc_intrin,
+                analyzer_clang_version
+            )
+            parallel_args.append(args)
```
bruntib: This seems to be like a huge memory consumption in case of big input. We already have some "out of memory" issues during analysis. Couldn't the jobs be thrown to a pool directly somehow?

irishrover: Thank you for you feedback. 
I fixed that - now the code uses generators instead of explicit lists. I also switched to `imap_unordered` with a built-in support of chunks and not requiring lists, just generators.

analyzer/codechecker_analyzer/buildlog/log_parser.py
irishrover: In the new version no manual chunking is required

irishrover:
Hm, cannot reproduce the tests failures locally. 
@bruntib, can you please check- do these tests fail with my PR?

analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -15,6 +15,7 @@
 
 import glob
 import json
+import multiprocessing
```
bruntib: Please, use `codechecker_common.compatibility.multiprocessing`

bruntib (changes requested):
One last minor comment, otherwise looks good. Thank you!

analyzer/codechecker_analyzer/buildlog/log_parser.py
irishrover: > Please, use `codechecker_common.compatibility.multiprocessing`

By the way, I don't fully understand why there's a difference for Linux vs MacOSX & Windows, why can't we use multiprocessing instead of futures in Linux?

bruntib: It would require further investigation, but we experienced malfunctions on different platforms with the different libraries (hanging zombie processes and such). It might have been fixed since then, but I would be a little worried changing this without extensive testing on different platforms.

PR #4618, opened 2025-07-06
Fix SeverityIcon color error

Marsman1996:
The original code only updates the color once, so the severity icon in the checker retains the color of the first checker the user opens.

For example, if a `low` severity checker is opened first, the icon will appear green in all checker tabs, regardless of their actual severity:
image


bruntib (approved):
Thanks for the fix, nice catch!

PR #4621, opened 2025-07-08
[fix] Fix updating fixed_at time

gulyasgergely902:
Fix updating fixed_at time on every store command. Now it should only update if there is no fixed_at timestamp present, yet.

### DIFF
--- a/web/tests/functional/review_status/test_review_status.py
+++ b/web/tests/functional/review_status/test_review_status.py
@@ -885,14 +885,13 @@ int main() {
         self.assertEqual(report.fixedAt, fixed_at)
 
         # A reopened and false positive report gets its "fixed at" date from
-        # ReviewStatus table. This is the date when the false positive review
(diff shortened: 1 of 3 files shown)


web/tests/functional/review_status/test_review_status.py
```diff
@@ -892,7 +892,7 @@ def get_report():
         report = get_report()
         self.assertEqual(report.detectionStatus, DetectionStatus.REOPENED)
         self.assertEqual(report.reviewData.status, ReviewStatus.FALSE_POSITIVE)
-        self.assertNotEqual(report.fixedAt, fixed_at)
+        self.assertEqual(report.fixedAt, fixed_at)
```
bruntib, line 894: Please, modify the explanatory comment above this section accordingly.

bruntib (changes requested):

bruntib (approved):

PR #4622, opened 2025-07-08
[fix] Blank page on invalid session token

bruntib:
When the browser has an auth session token that is not found on server side, then a blank page was loaded. As a workaround the user had to manually clear the token from the browser and reload the page. Now we redirect the user back to the login page when this situation occurs.

The problem was that the "authParams" was cached in a state that its "sessionStillActive" field is false. Now these auth parameters are not cached but loaded from the server.


Discookie (approved):
Some way to test this case would be nice, but otherwise LGTM.

PR #4624, opened 2025-07-10
Addition of Gitlab authentication.

feyruzb:
Added the following:
- Gitlab OAuth template.
- Gitlab configuration file section.
- Documentation regarding username selection options of Gitlab.

### DIFF
--- a/web/server/codechecker_server/api/authentication.py
+++ b/web/server/codechecker_server/api/authentication.py
@@ -484,9 +484,17 @@ class ThriftAuthHandler:
                         username = claims.get("Signum")
                     else:
                         username = user_info.get("mail")
+                elif template == "gitlab/v1":
+                    print(user_info)
+                    if username_key == "username":
+                        username = user_info.get("preferred_username")
+                    elif username_key == "email":
+                        print(user_info)
+                        username = user_info.get("email")
+                    elif username_key == "name":
+                        username = user_info.get("name")
 
                 LOG.debug(f"groups fetched for {username}, are: {groups}")
-
                 LOG.info("Username fetched, for username: %s", username)
             except Exception as ex:
                 LOG.error("Username fetch failed: %s", str(ex))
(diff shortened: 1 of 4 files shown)


LorandMarton:
Hi! Thanks for adding support for GitLab authentication - much appreciated!

Quick question: Will this also work with self-hosted GitLab instances? If not, would it be possible to make the template configurable (e.g., allowing a custom domain and endpoints) to support self-hosted GitLab as well?

feyruzb:
> Hi! Thanks for adding support for GitLab authentication - much appreciated!
> 
> Quick question: Will this also work with self-hosted GitLab instances? If not, would it be possible to make the template configurable (e.g., allowing a custom domain) to support self-hosted GitLab as well?

Hello, current configuration supports variables overriding, so if you specify variables mentioned below according to your custom GitLab instance in server_config file, it should work. 

- client_id
- client_secret
- authorization_url
- token_url
- user_info_url

It should look, like this:

image

web/server/codechecker_server/api/authentication.py
```diff
@@ -484,9 +484,17 @@ def performLogin(self, auth_method, auth_string):
                         username = claims.get("Signum")
                     else:
                         username = user_info.get("mail")
+                elif template == "gitlab/v1":
+                    print(user_info)
```
bruntib: Please, remove this print.

web/server/codechecker_server/api/authentication.py
```diff
@@ -484,9 +484,17 @@ def performLogin(self, auth_method, auth_string):
...
+                    if username_key == "username":
+                        username = user_info.get("preferred_username")
+                    elif username_key == "email":
+                        print(user_info)
```
bruntib: Please, remove this print.

web/server/codechecker_server/api/authentication.py
feyruzb: Fixed.

web/server/codechecker_server/api/authentication.py
feyruzb: Fixed.

bruntib (approved):

PR #4627, opened 2025-07-10
Url format checker doesn't recognize - as part of url 

feyruzb:
Fixed URL format bug according to suggestion from user.


dkrupp (approved):
LGTM

PR #4631, opened 2025-07-14
[fix] Fix statistics page components

gulyasgergely902:
Fix statistics page minor issues and inconsistencies:
- Panels "Number of outstanding reports" / "Number of resolved reports" and "Number of failed files" / "Number of checkers reporting faults" are separated since they are two different metrics.
- "Component statistics" table removed since it is a redundant duplicate.
- Pie chart moved to "Severity statistics" page.


bruntib (changes requested):
- The Component statistics table appears under "Severity statistics" tab.
- Also, at "Severity statistics" tab there is some label gitch: 

Image

gulyasgergely902:
@bruntib Fixed the label glitch also renamed the table so it represents what is should better.

PR #4633, opened 2025-07-15
Manage secrets outside of server_config.json

barnabasdomozi:
In this PR, I added an option to store secrets outside of `server_config.json`.

To use this feature, we can replace sensitive data with `$SECRET:NAME_OF_SECRET$`, then `NAME_OF_SECRET` will be read from a separate `secrets.json` file when we initially load `server_config.json`.
This change is also backward compatible (e.g., secrets can still be stored in `server_config.json`).

See the updated documentation in `docs/web/server_config.md` for more info.


web/server/codechecker_server/session_manager.py
```diff
@@ -409,6 +410,45 @@ def __get_config_dict(self):
...
+                elif isinstance(v, str):
+                    secret_matched = secret_re.search(v)
+                    env_matched = env_re.search(v)
+
+                    if secret_matched or env_matched:
+                        var_name = v.split(':')[1][:-1]
+                        if secret_matched and var_name in secrets_dict:
+                            d[k] = secrets_dict[var_name]
+                        elif env_matched and var_name in os.environ:
+                            d[k] = os.environ.get(var_name)
```
bruntib:
```suggestion
                            d[k] = os.environ[var_name]
```
Just to be conform with the previous branch, but feel free to ignore this minor comment.

bruntib (approved):
Please, fix the lint test, otherwise looks good to me.

web/server/codechecker_server/session_manager.py
```diff
@@ -409,6 +410,45 @@ def __get_config_dict(self):
             # have been parsed from it.
             raise ValueError("Server configuration file was invalid, or "
                              "empty.")
+
+        LOG.debug(self.__secrets_file)
+        if os.path.exists(self.__secrets_file):
+            secrets_dict = load_json(self.__secrets_file, {})
+            check_file_owner_rw(self.__secrets_file)
+        else:
+            secrets_dict = {}
```
Discookie, line 419: This fallibility is already handled by `load_json()`.
```suggestion
        secrets_dict = load_json(self.__secrets_file, {})
        if secrets_dict != {}:
            check_file_owner_rw(self.__secrets_file)
```

Discookie:
(Similarly, there should be at least a log warning, if not a ValueError, on a nonexistent environment variable.)

web/server/codechecker_server/session_manager.py
barnabasdomozi: The server should function correctly if `server_secrets.json` doesn't exist (eg., because secrets are not used at all). 
This is intentional, as `load_json()` will log an error message that it could not open the file.

Discookie: That's fine then.

bruntib (approved):

PR #4634, opened 2025-07-16
6512 documentation on GitHub gating

gulyasgergely902:
Add documentation on GitHub PR gating with CodeChecker into CONTRIBUTING.md


CONTRIBUTING.md
```diff
@@ -95,65 +97,84 @@ from product_db_access_handler import ThriftProductHandler
 # ... your code here
 ```
 
+### Code quality
+
+Any change in the Pull Request against the codechecker repository will be checked
+with CodeChecker itself for potential issues. If any issue found, the Pull Request
```
dkrupp: If any **new** issue is found

CONTRIBUTING.md
```diff
@@ -95,65 +97,84 @@ from product_db_access_handler import ThriftProductHandler
 # ... your code here
 ```
 
+### Code quality
```
dkrupp, line 100: I think it would be good to write down in this section which script should be run locally and from where to check the gating criteria ...

gulyasgergely902: I'd make a different user for this to avoid "demo" users find the production checks. Make codechecker product invisible to "demo" user.

CONTRIBUTING.md
```diff
@@ -95,65 +97,103 @@ from product_db_access_handler import ThriftProductHandler
...
+ "client_autologin" : true,
+ "credentials": {
+  "https://codechecker-demo.eastus.cloudapp.azure.com": "codechecker:codechecker"
+ }
+}
+```
+
+* Run `ci/github_analysis/codechecker_gate_pr.sh` to analyze the code.
+
+__Important:__ Make sure you are running CodeChecker 6.25.1 when running the script!
```
Discookie: If this version is hardcoded elsewhere, it needs to be referenced in the developer docs, as it needs changing on a new release.

gulyasgergely902: The version `6.25.1` is required since the server which the gating is using, runs CodeChecker 6.25.1 and the version matching is required. If the server will be updated to a new version, then can this version be updated as well.

dkrupp: I guess the requirement is that the client needs to be newer than the 6.25.1 (the server version), isn't it?

So I think we should write here that the client used must be always on a lower version than the server at https://codechecker-demo.eastus.cloudapp.azure.com

gulyasgergely902: The requirement is the client must be on the same or lower version than the server so in this case it should be 6.25.1 or lower. I'll modify it so it reflects this.

PR #4642, opened 2025-07-25
Show error on reaching report limit

gulyasgergely902:
This change fixes the error message which previously appeared in the server logs instead of the client logs. Now it shows to the user when the report limit was reached.

### DIFF
--- a/web/client/codechecker_client/cli/store.py
+++ b/web/client/codechecker_client/cli/store.py
@@ -619,7 +619,7 @@ def assemble_zip(inputs,
 
     # Fail store early if too many reports.
     p = prod_client.getCurrentProduct()
-    if len(unique_reports) > p.reportLimit:
+    if len(unique_report_hashes) > p.reportLimit:
         LOG.error(f"""Report Limit Exceeded
 
 This report folder cannot be stored because the number of reports in the
--- a/web/client/codechecker_client/cli/store.py
+++ b/web/client/codechecker_client/cli/store.py
@@ -424,6 +424,15 @@ def parse_analyzer_result_files(
     return analyzer_result_file_reports
 
 
+class ReportLimitExceedError(Exception):
+    """
+    Custom exception type thrown by 'assemble_zip'. This is used when the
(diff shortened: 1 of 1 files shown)


web/client/codechecker_client/cli/store.py
```diff
@@ -922,6 +931,9 @@ def main(args):
                          client,
                          prod_client,
                          context.checker_labels)
+        except ReportLimitExceedError as ex:
+            print(ex)
```
bruntib: The error message was printed properly above to the "error" log level. We shouldn't use pure `print()` function for logging but we could replace that error log here.

web/client/codechecker_client/cli/store.py
```diff
@@ -619,7 +628,7 @@ def assemble_zip(inputs,
 
     # Fail store early if too many reports.
     p = prod_client.getCurrentProduct()
-    if len(unique_reports) > p.reportLimit:
+    if len(unique_report_hashes) > p.reportLimit:
```
dkrupp, line 631: For me it is not clear why this change is needed. Could you pls explain?

Could you please add a testcase which shows what error this change fixes?

gulyasgergely902: The variable `unique_reports` contain the reports in the reports directory. Using `unique_reports` by itself is not working right since its length in only 1 (there is only 1 reports directory), to get the right value I would need to dig deeper in this structure. There were 2 viable options which I choose from. Either use `unique_reports` but not with length but a different way or use the already existing and incrementing by report `unique_report_hashes` with a simple length.

gulyasgergely902: As for the test case: the test is already implemented in `test_store.py:test_store_limit` which tests if the limit is lower than the to be stored count.

dkrupp: But how can it happen that the test is passing with and without this patch (while the test was not changed)?
Then the test is not fully testing this feature  or?

PR #4645, opened 2025-07-28
Rework config_directory

barnabasdomozi:
In this PR, I reworked `--config-directory` and `--workspace` settings. The current behavior is that these parameters are essentially the same.

PR changes:
Server configuration files and SSL certs are now stored in `--config-directory`, while SQLite databases, logs, temporary files in `--workspace`.

If `--config-directory` is not set, it will be the same as `--workspace`.

For the review: `args.dbdatadir` was also removed since I haven't found any usage of that in the entire codebase.

### DIFF
--- a/web/server/codechecker_server/cli/server.py
+++ b/web/server/codechecker_server/cli/server.py
@@ -85,14 +85,11 @@ def add_arguments_to_parser(parser):
                         default=default_workspace,
                         required=False,
                         help="Directory where CodeChecker can store analysis "
-                             "result related data, such as the database. "
-                             "(Cannot be specified at the same time with "
-                             "'--sqlite' or '--config-directory'.)")
+                             "result related data, such as the database.")
 
     parser.add_argument('-f', '--config-directory',
                         type=str,
                         dest="config_directory",
-                        default=default_workspace,
                         required=False,
                         help="Directory where CodeChecker server should read "
                              "server-specific configuration (such as "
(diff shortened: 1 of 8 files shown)


web/server/codechecker_server/cli/server.py
```diff
@@ -373,54 +367,22 @@ def arg_match(options):
...
-        # the latter one is expected to replace the earlier.
-        options = ['--config-directory', '--workspace']
-        options_short = ['--config-directory', '-w']
-        if set(arg_match(options)) == set(options) or \
-                set(arg_match(options_short)) == set(options_short):
-            parser.error("argument --config-directory: not allowed with "
-                         "argument --workspace")
-
         # If workspace is specified, sqlite is workspace/config.sqlite
         # and config_directory is the workspace directory.
```
bruntib: Adapt this comment to the changes, because these statements are not true.

web/server/codechecker_server/server.py
```diff
@@ -814,6 +815,7 @@ def __init__(self,
         LOG.debug("Initializing HTTP server...")
 
         self.config_directory = config_directory
+        self.workspace_directory = workspace_directory
```
bruntib, line 818: Why does this class need the workspace directory?

web/server/codechecker_server/cli/server.py
barnabasdomozi: Done

web/server/codechecker_server/server.py
barnabasdomozi: It is used by the `ThriftProductHandler` class, for example on line 297 in file `product_server.py`.

bruntib (approved):

PR #4654, opened 2025-08-20
Add memory-safety guideline

salticecream:
This PR adds a new guideline, `memory-safety`, which can be useful for developers who are interested in catching specifically memory-safety bugs. 

Furthermore, my suggestion (included in the PR) is to raise `unix.MismatchedDeallocator` and `unix.Malloc` to `HIGH` severity. The former _always_ leads to UB on true positives, and the latter almost always.

### DIFF
--- a/config/guidelines/memory-safety.yaml
+++ b/config/guidelines/memory-safety.yaml
@@ -1,5 +1,5 @@
 guideline: memory-safety
-guideline_title: Memory-safety related CWEs based on Juliet test suite
+guideline_title: Memory-safety related CWEs
 rules:
 - rule_id: cwe-121
   title: Stack-Based Buffer Overflow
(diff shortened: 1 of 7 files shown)


bruntib:
Please, extend this list accordingly: https://github.com/Ericsson/codechecker/blob/834807db3178a926b1b8fdf0e75de684986db334/web/server/vue-cli/src/components/Statistics/Guideline/GuidelineStatistics.vue#L177
Also, add the new list item to the selected items' list: https://github.com/Ericsson/codechecker/blob/834807db3178a926b1b8fdf0e75de684986db334/web/server/vue-cli/src/components/Statistics/Guideline/GuidelineStatistics.vue#L211
Yes, you're right, this should be added dynamically in the future :)

Also, the severity of some checkers are changed (from medium to high, for example). This breaks some tests: https://github.com/Ericsson/codechecker/actions/runs/17100480035/job/48495383987?pr=4654
I would say that severity change of such important checkers should be done in an independent pull request if we decide to change them.

Thank you!

PR #4656, opened 2025-09-02
Fix CodeChecker analyzer config exception

gulyasgergely902:
Fix a CodeChecker exception which occurs when the analyze command has an analyzer config set which' analyzer is not enabled.

This fix collects all the enabled analyzers and skips the config checks for the analyzers which are not enabled.

### DIFF
--- a/analyzer/codechecker_analyzer/cli/analyze.py
+++ b/analyzer/codechecker_analyzer/cli/analyze.py
@@ -902,14 +902,18 @@ def is_analyzer_config_valid(
             continue
 
     for cfg_arg in analyzer_config_args:
+        if enabled_analyzers and cfg_arg.analyzer not in enabled_analyzers:
+            wrong_config_messages.append(
+                f"The analyzer '{cfg_arg.analyzer}' is not enabled. Enable it "
+                "first using --analyzers to set the config "
+                f"'{cfg_arg.option}'.")
+            continue
+
         if cfg_arg.analyzer not in available_analyzers:
             wrong_config_messages.append(
                 f"Invalid argument to --analyzer-config: '{cfg_arg.analyzer}' "
                 f"is not an available analyzer. Available analyzers are: "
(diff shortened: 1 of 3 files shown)


analyzer/codechecker_analyzer/cli/analyze.py
```diff
@@ -885,6 +900,9 @@ def is_analyzer_config_valid(
                 f"{', '.join(a for a in supported_analyzers)}.")
             continue
 
+        if enabled_analyzers and cfg_arg.analyzer not in enabled_analyzers:
```
bruntib, line 905:
```suggestion
        if cfg_arg.analyzer not in enabled_analyzers:
```

bruntib (changes requested):

analyzer/tests/unit/test_checker_handling.py
```diff
@@ -663,13 +663,16 @@ def test_analyze_wrong_parameters(self):
         --analyzer-config or --checker-config parameter is specified.
         """
 
+        args = create_analyze_argparse()
```
dkrupp, line 680: could you pleaese add a new test case when  you for example enable clang-tidy only, but you pass a clangsa config value. That should generate an error.

dkrupp (dismissed):
I think it would be better to give an error when someone passes a config parameter for an analyzer that is not enabled.

analyzer/codechecker_analyzer/cli/analyze.py
gulyasgergely902: At the beginning, enabled_analyzers is an empty list. If it stays so (meaning no analyzer is listed when calling CodeChecker) if the check for emptyness in the if statement is removed, then all set config will be discarded (since the config's analyzer is not in the empty list). Please correct me if I'm mistaken though.

bruntib (approved):

PR #4657, opened 2025-09-07
Page runs

xb058t:
Migrated runs page


web/server/vue-cli/config/webpack.common.js
```diff
@@ -147,34 +145,33 @@ module.exports = {
...
     }),
     new ProvidePlugin({
-      Buffer: ['buffer', 'Buffer'],
-    }),
-    new ESLintPlugin({
-      extensions: ['js', 'vue'],
-      overrideConfigFile: './.eslintrc.js'
+      Buffer: [ 'buffer', 'Buffer' ],
     }),
+    // new ESLintPlugin({
```
gulyasgergely902, line 170: Try not to leave commented out code in the final code. If needed, uncomment it, if not then remove it.

web/server/vue-cli/config/webpack.dev.js
```diff
@@ -27,15 +26,15 @@ module.exports = merge(common, {
     port: 8080,
     hot: true,
     historyApiFallback: {
-      // If the URL contains a product endpoint and we server a static file
```
gulyasgergely902: This comment might be informative in the future. You could keep it.

web/server/vue-cli/src/components/DateTimePicker.vue
```diff
@@ -4,17 +4,16 @@
     :content-class="dialogClass"
     width="400"
   >
-    <template v-slot:activator="{ on }">
+    <template #activator="{ props }">
```
gulyasgergely902, line 7: Is it a best practice to replace v-slot with a hash mark?

web/server/vue-cli/src/components/DateTimePicker.vue
```diff
@@ -147,34 +134,34 @@ export default {
...
-      this.date = format(initValue, this.dateFormat);
-      this.time = format(initValue, this.timeFormat);
+      if (initValue) {
+        this.date = format(initValue, this.dateFormat);
+        this.time = format(initValue, this.timeFormat);
+      } else {
+        this.resetDateTimes();
+      }
     },
-
```
gulyasgergely902, line 159: Deleting every single empty line will not make the code better but makes readability worse. Don't remove all of it but decide with the question: "Will it make the code readable?".

web/server/vue-cli/src/services/api/cc.service.js
```diff
@@ -27,7 +27,7 @@ function extractTagWithRunName(runWithTagName) {
 
 class CodeCheckerService extends BaseService {
   constructor() {
-    super("CodeCheckerService", ServiceClient,"v6.61", true);
```
gulyasgergely902, line 30: Didn't you remove the appended 'v' from an other place too? If I'm right, by removing this too, it will get rid of the 'v' in the whole version number.

web/server/vue-cli/src/views/RunList.vue
```diff
@@ -181,88 +183,96 @@ export default {
   mixins: [ VersionMixin ],
 
   data() {
```
gulyasgergely902, line 185: What happened to the identation?

PR #4660, opened 2025-09-12
fix: protect for missing clang binary

elupus:
When cli is validating analyzer configuration, it will call this method. This will fail with None error in file system functions since no binary is found.

This was introduced in 39666a7b71d0e79dd889791ab7a12770c7cb3a30


bruntib:
I close this ticket due to inactivity. I suppose, this issue has been fixed about `diagtool`.

PR #4661, opened 2025-09-15
Fix ClangTidy default hash type in codechecker_report_converter

barnabasdomozi:
Due to backward compatibility, the CodeChecker analyzer uses hash type PATH_SENSITIVE for ClangTidy by default. Meanwhile, the report-converter defaults to CONTEXT_FREE hash type.
This creates inconsistencies when running the analysis with the --makefile option, since it uses report-converter after executing clang-tidy.

To fix this issue, the report-converter now defaults to PATH_SENSITIVE hash type when processing ClangTidy results.

### DIFF
--- a/tools/report-converter/tests/unit/analyzers/tidy_yaml_output_test_files/tidy1.plist
+++ b/tools/report-converter/tests/unit/analyzers/tidy_yaml_output_test_files/tidy1.plist
@@ -12,7 +12,7 @@
 			<key>description</key>
 			<string>remainder by zero is undefined</string>
 			<key>issue_hash_content_of_line_in_context</key>
-			<string>5a0fa223fabaf319bc0ec3731e27d45a</string>
+			<string>9785900b66c87f89edcc9daaab59647a</string>
 			<key>location</key>
 			<dict>
 				<key>col</key>
--- a/tools/report-converter/tests/unit/analyzers/tidy_yaml_output_test_files/tidy2.plist
+++ b/tools/report-converter/tests/unit/analyzers/tidy_yaml_output_test_files/tidy2.plist
@@ -12,7 +12,7 @@
 			<key>description</key>
 			<string>unused variable 'y'</string>
 			<key>issue_hash_content_of_line_in_context</key>
-			<string>9f5469099db1a2765912c8254b32e90b</string>
+			<string>0afd6d1f5bf7f1856d9f01e4ac92534e</string>
 			<key>location</key>
 			<dict>
 				<key>col</key>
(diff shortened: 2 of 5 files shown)


tools/report-converter/codechecker_report_converter/analyzers/analyzer_result.py
```diff
@@ -121,12 +121,19 @@ def _post_process_result(self, reports: List[Report]):
...
+            self._add_report_hash(report)
 
     def _add_report_hash(self, report: Report):
         """ Generate report hash for the given plist data. """
-        report.report_hash = get_report_hash(report, HashType.CONTEXT_FREE)
+        hash_type = HashType.CONTEXT_FREE
+
+        # Due to backward compatibility, the CodeChecker analyzer
+        # uses hash type PATH_SENSITIVE for ClangTidy by default.
+        if report.analyzer_name == "clang-tidy":
```
bruntib: This should rather be implemented in an override method in `tools/report-converter/codechecker_report_converter/analyzers/clang_tidy/analyzer_result.py`.

Also, don't forget to change `clang_tidy_yaml`.

barnabasdomozi: Thanks for the suggestion! Implemented the changes in https://github.com/Ericsson/codechecker/commit/2bdda75699675241b81185ea5d0e6481eb79db7a

analyzer/codechecker_analyzer/analyzers/clangtidy/result_handler.py
```diff
@@ -58,6 +58,9 @@ def postprocess_result(
         # automatically when using this option we introduced a new choice for
         # --report-hash option ('context-free-v2') and we still do not use
         # context free hash for 'context-free' choice.
+
+        # When changing the default hash_type for ClangTidy, ensure
```
bruntib, line 62: If someone changes ClangTidy, they won't read this message. Also, the ClangTidy-related changes are now replaced to `clang_tidy/analyzer_result.py`.

By the way, in the long terms there should be a more consistent usage between ClangTidy analyzer and report-converter. For example, what happens if `--report-hash` flag changes the hash algorithm at `CodeChecker analyze` command?
Probably this is out-of-scope for this patch.

barnabasdomozi: > If someone changes ClangTidy, they won't read this message. Also, the ClangTidy-related changes are now replaced to clang_tidy/analyzer_result.py.

Unfortunately, we have two implementations for the same functionality for processing ClangTidy analysis reports. One for the `CodeChecker analyzer`: `analyzer/codechecker_analyzer/analyzers/clangtidy/result_handler.py` and another for the `report-converter`: `tools/report-converter/codechecker_report_converter/analyzers/clang_tidy/analyzer_result.py`. These  need to be kept in sync, otherwise different hash types result in inconsistent `plist` files. I'm not sure what would be the solution here.

> For example, what happens if --report-hash flag changes the hash algorithm at CodeChecker analyze command?

I would also say out of scope for this patch, first I think we should have a consistent behavior (report hash type) on both sides.

bruntib (approved):

PR #4662, opened 2025-09-15
Async store 3

bruntib:


web/api/codechecker_api_shared.thrift
```diff
@@ -14,14 +14,35 @@ enum Ternary {
...
-  AUTH_DENIED,   // Authentication denied. We do not allow access to the service.
-  UNAUTHORIZED,  // Authorization denied. User does not have right to perform an action.
-  API_MISMATCH,  // The client attempted to query an API version that is not supported by the server.
-  SOURCE_FILE,   // The client sent a source code which contains errors (e.g., source code comment errors).
-  REPORT_FORMAT, // The client sent a report with wrong format (e.g., report annotation has bad type in a .plist).
+  // Any other sort of error encountered during RPC execution.
+  GENERAL  = 2,
+
+  // Executing the request triggered a database-level fault, constraint violation.
+  DATABASE = 0,
```
gulyasgergely902, line 21: It's a bit confusing for me since usually return code 0 means there is no error and the execution was successful. I understand this enum is meant for error codes only but still, it is confusing this way.

web/client/codechecker_client/cli/store.py
```diff
@@ -958,77 +963,116 @@ def main(args):
...
+                    "operation.\n"
+                    "The server might not have finished processing "
+                    "everything at this point, so do NOT rely on querying "
+                    "the results just yet!\n"
+                    "To await the completion of the processing later, "
+                    "you can execute:\n\n"
+                    "\tCodeChecker cmd serverside-tasks --token %s "
+                    "--await",
+                    task_token)
+                # Print the token to stdout as well, so scripts can use
```
gulyasgergely902, line 1052: Is this comment important for future developers of this code?

web/server/codechecker_server/task_executors/task_manager.py
```diff
@@ -95,31 +124,78 @@ def allocate_task_record(self, kind: str, summary: str,
...
+        """
+        Sets all tasks in the database that were associated with the given
+        `machine_id` to ``"dropped"`` status, indicating that the status was
+        changed during the `action`.
+
+        Returns the number of `DBTask`s actually changed.
+        """
+        count: int = 0
+        with DBSession(self._database_factory) as session:
+            for t in session.query(DBTask) \
```
gulyasgergely902: Why not something more descriptive as a variable name like "db_task" instead of just "t"?

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -968,78 +1202,67 @@ def __realise_fake_checkers(self, session):
...
-                codechecker_api_shared.ttypes.ErrorCode.DATABASE,
-                str(ex))
+        grouped_by_checker: Dict[Tuple[str, str], List[int]] = \
+            defaultdict(list)
+        for _, (report, db_id) in \
+                self.__reports_with_fake_checkers.items():
+            checker: Tuple[str, str] = checker_name_for_report(report)
+            grouped_by_checker[checker].append(cast(int, db_id))
+
+        for chk, report_ids in grouped_by_checker.items():
```
gulyasgergely902: Instead of "chk", I'd use "checker" for a more descriptive naming.

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -968,78 +1202,67 @@ def __realise_fake_checkers(self, session):
...
-                        macro.range.end_line, macro.range.end_col,
-                        macro.message, file_path_to_id[macro.file.path],
-                        db_report.id, data_type))
-
-                if report.annotations:
-                    self.__validate_and_add_report_annotations(
-                        session, db_report.id, report.annotations)
+        for db_report, report in self.__added_reports:
+            LOG.debug("Storing bug path positions.")
+            for i, p in enumerate(report.bug_path_positions):
```
gulyasgergely902: Could you please add a more descriptive naming here too instead of just an "i" and "p"?

web/server/codechecker_server/api/report_server.py
```diff
@@ -1336,13 +1336,26 @@ def get_is_opened_case(subquery):
     )
 
 
+def remove_reports(session: DBSession,
+                   report_ids: Collection,
+                   chunk_size: int = SQLITE_MAX_VARIABLE_NUMBER):
+    """
+    Removes `Report`s in chunks.
+    """
+    for r_ids in util.chunks(iter(report_ids), chunk_size):
```
gulyasgergely902: Instead of "r_ids", could that be "report_ids"?

web/server/codechecker_server/api/report_server.py
```diff
@@ -3946,28 +3922,92 @@ def getMissingContentHashesForBlameInfo(self, file_hashes):
...
+                                         force=force,
+                                         trimPathPrefixes=trim_path_prefixes,
+                                         description=description,
+                                         )
+        token = self.__massStoreRun_common(False, b64zip, store_opts)
+
+        # To be compatible with older (<= 6.24, API <= 6.58) clients which
+        # may keep using the old API endpoint, simulate awaiting the
+        # background task in the API handler.
+        t = self._task_manager.get_task_record(token)
```
gulyasgergely902: The same concern as above: the not so descriptive variable namings.

web/server/codechecker_server/server.py
```diff
@@ -1253,6 +1062,38 @@ def _cleanup_incomplete_tasks(action: str) -> int:
...
     is_server_shutting_down = Value('B', False)
 
+    sync_manager = SyncManager()
+    sync_manager.start()
+    task_pipes = sync_manager.dict()
+
+    def _cleanup_incomplete_tasks(action: str) -> int:
+        config_db = config_sql_server.create_engine()
+        config_session_factory = sessionmaker(bind=config_db)
+        tm = BackgroundTaskManager(
```
gulyasgergely902: This should be "background_task_manager" or "task_manager" instead of "tm".

web/server/vue-cli/package-lock.json
gulyasgergely902, line 1: Was there a reason for downgrading all those packages or just when the PR was created back then those were the versions and since then those got updated but not this?

web/api/codechecker_api_shared.thrift
bruntib: These values were used even before this PR. Changing the values may cause backward incompatibility issues.

web/client/codechecker_client/cli/store.py
bruntib: Without this comment I almost removed the `print(task_token)` below it. First, because I thought, it was a redundant information, since the warning above also prints it. Second, because we print logs only through the logger library, conventionally.

web/server/codechecker_server/api/report_server.py
bruntib: The function parameter has the same name. I could rename all others.

PR #4663, opened 2025-09-19
[fix] Find diagtool based on ClangTidy

bruntib:
diagtool is used for gathering the list of compiler warnings. In some environments the name of this binary contains a version number, e.g. diagtool-18. Some heuristics try to find the proper binary based on the Clang version that is available in PATH. However, diagtool is used only by clang-tidy, so diagtool's version should also be deduced based on clang-tidy. In some environments clang itself is not even available.

This patch used clang-tidy instead of clang to find diagtool.


analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
```diff
@@ -125,44 +124,48 @@ def get_diagtool_bin():
...
-    clang_bin = context.analyzer_binaries.get(ClangSA.ANALYZER_NAME)
+    clang_tidy_bin = context.analyzer_binaries.get(ClangTidy.ANALYZER_NAME)
 
-    if not clang_bin:
+    if not clang_tidy_bin:
         return None
 
     path_env = os.environ.get('PATH', '').split(os.pathsep)
-    clang_path = Path(clang_bin)
+    clang_path = Path(clang_tidy_bin)
```
gulyasgergely902: If others are renamed, could this be renamed to `clang_tidy_path` too, please?

Issue #4601, opened 2025-06-12
Non-existent tmpfile while storing after analysis with bazel


Szelethus:
## Describe the bug
CodeChecker store crashes with this message:
```
Traceback (most recent call last):
  File "codechecker_client/cli/store.py", line 904, in main
    temp_dir = tempfile.mkdtemp(suffix="-store", dir=args.input[0])
  File "python3/lib/python3.10/tempfile.py", line 384, in mkdtemp
    _os.mkdir(file, 0o700)
PermissionError: [Errno 13] Permission denied: 'bazel-out/k8-fastbuild/bin/codechecker_test_flatc/codechecker-files/data/tmpf0ogqboy-store'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "codechecker_common/cli.py", line 230, in main
    sys.exit(args.func(args))
  File "codechecker_client/cli/store.py", line 908, in main
    f"permissions to create the {temp_dir} "
UnboundLocalError: local variable 'temp_dir' referenced before assignment
```

This file does not exist.

## CodeChecker version
```
[INFO 2025-06-12 10:31] - CodeChecker analyzer version:
---------------------------------------------------------------
Kind                 | Version                                 
---------------------------------------------------------------
Base package version | 6.26.0                                  
Package build date   | 2025-06-10T18:03                        
Git commit ID (hash) | 0bbf351e870c7056cd9cc0dc44a795c0fe805dfa
Git tag information  | 6.26                                    
---------------------------------------------------------------

[INFO 2025-06-12 10:31] - CodeChecker web version:
------------------------------------------------------------------------------
Kind                                | Version                                 
------------------------------------------------------------------------------
Base package version                | 6.26.0                                  
Package build date                  | 2025-06-10T18:03                        
Git commit ID (hash)                | 0bbf351e870c7056cd9cc0dc44a795c0fe805dfa
Git tag information                 | 6.26                                    
Server supported Thrift API version | 6.65                                    
Client Thrift API version           | 6.65                                    
------------------------------------------------------------------------------
```
bazel version:
```
bazel 6.5.0
```
codechecker_bazel version:
```
289b9cefca68c6f23e8a87ca332a9231de9c24ab
```

**To Reproduce**
Steps to reproduce the behaviour:

**Download a bazel 6 compatible codebase.**
```lang=bash
git clone https://github.com/google/flatbuffers
git checkout v22.10.26
```
**Apply this patch to enable CodeChecker analysis through bazel**

[...]

PR #4668, opened 2025-09-23
Add argument to set temporary directory during CodeChecker store

furtib:
Why:
When working on Bezel, we encounter a situation where the result files are placed into a read-only folder. By default, `CodeChecker store` creates a temporary directory for zipping the results before sending them to the server. This directory cannot be created in a read-only folder, causing CodeChecker to crash.

What:
This pull request adds an extra command-line argument, `--temp_dir`, to the `CodeChecker store` command, which specifies where temporary files should be created. When this argument is not set, it continues to use the previous implementation.

Fixes: #4601

### DIFF
--- a/web/client/codechecker_client/cli/store.py
+++ b/web/client/codechecker_client/cli/store.py
@@ -934,7 +934,7 @@ def main(args):
         LOG.debug(f"{temp_dir} directory created successfully!")
     except PermissionError:
         try:
-            # If the specified folder isn't writeable; fallback to temporary dir
+            # If the specified folder isn't writeable; fallback to temp dir
             LOG.debug(
                 "'%s' is readonly, falling back to temporary folder",
                 temp_dir_path,


web/client/codechecker_client/cli/store.py
```diff
@@ -269,6 +269,16 @@ def add_arguments_to_parser(parser):
...
                              "match will be removed. You may also use Unix "
                              "shell-like wildcards (e.g. '/*/jsmith/').")
+    
+    parser.add_argument('--temp_dir',
+                        type=str,
+                        metavar='PATH',
+                        required=False,
+                        default=None,
+                        dest="temp_dir",
+                        help="Specify a path to use as temporary folder "
```
Szelethus: , necessary to create a zip file upload. Defaults to the the results directory to be stored.

web/client/codechecker_client/cli/store.py
```diff
@@ -270,6 +270,19 @@ def add_arguments_to_parser(parser):
...
+                        metavar='PATH',
+                        required=False,
+                        default=None,
+                        dest="temp_dir",
+                        help="Specify a path to use as temporary folder. "
+                        "Necessary to have a place to compress files before "
+                        "uploading them. Defaults to the result directory. "
+                        "Normally it isn't necessary to specify. "
+                        "Useful when the folder containing the result folder "
+                        "is read only.")
```
Szelethus: Specify the location to write the compressed file used for storage. Useful if the results directory is read only. Defaults to the results directory.

web/client/codechecker_client/cli/store.py
```diff
@@ -909,8 +921,13 @@ def main(args):
                                                  port,
                                                  product_name=product_name)
 
+    temp_dir_path: str = args.temp_dir if args.temp_dir else args.input[0]
+    if not os.access(temp_dir_path, os.W_OK):
+        # Folder not writeable fallback to /tmp/
```
Szelethus:
```suggestion
        # Folder is not writeable; fallback to /tmp/
```

Szelethus: I understand that the more pythonic way in this scenario is [EAFP](https://realpython.com/python-lbyl-vs-eafp/), which would also be a bit more robust. Just try to create the file and fall back if it doesn't work.

furtib: This has been changed since. I will insert the `;`

furtib: The current implementation keeps the logic of path changing together. 
Using this more Pythonic way would create a confusing nested try-except tree and would make it harder to understand how the path selection happens.
I will push a change for this, but I ask you to reconsider this change.

Szelethus (commented):
Reflecting on our conversation yesterday, I think the term "temp" is way too ambigous. How about we call the flag `--zip_dir`, the variable that handles it same etc?

web/client/codechecker_client/cli/store.py
Szelethus: This isn't c++, so my it will never be the perfect sight for me eyes. To me, this solution doesn't look all that bad. Do you dislike it that much?

furtib: It turned out more tolerable than I expected. I guess it's fine this way. :/

furtib:
> Reflecting on our conversation yesterday, I think the term "temp" is way too ambigous. How about we call the flag `--zip_dir`, the variable that handles it same etc?

In my opinion, it would be more fitting to name it --zip_loc then.

web/client/codechecker_client/cli/store.py
```diff
@@ -257,6 +257,17 @@ def add_arguments_to_parser(parser):
                              "match will be removed. You may also use Unix "
                              "shell-like wildcards (e.g. '/*/jsmith/').")
 
+    parser.add_argument('--zip_loc',
```
bruntib:
```suggestion
    parser.add_argument('--zip-loc',
```
Conventionally we use dash in flag names.

web/client/codechecker_client/cli/store.py
```diff
@@ -914,14 +926,24 @@ def main(args):
                                                  port,
                                                  product_name=product_name)
 
+    # If the --temp_dir argument is specified set use that,
+    # else use the analyze result folder
+    temp_dir_path: str = args.zip_loc if args.zip_loc else args.input[0]
```
bruntib:
```suggestion
    temp_dir_path: str = args.zip_loc or args.input[0]
```

web/client/codechecker_client/cli/store.py
bruntib: Could you extend this documentation? https://github.com/Ericsson/codechecker/blob/master/docs/web/user_guide.md

bruntib (approved):

Issue #4664, opened 2025-09-22
Analyze with --file flag runs for multiple files


furtib:
**Describe the bug**
On a project with 2 files with names, such that one file's name is a prefix to the other file's name (e.g, `simple.c` and `simple.cc`), `CodeChecker analyze --file=simple.c` will analyze `simple.cc` too, not just `simple.c`

**CodeChecker version**
6.26.1

**To Reproduce**
1. Run the following script:
```bash
mkdir test_file_tag
cd test_file_tag

echo "int foo() { return 0/0; }" > simple.c
echo "int bar() { return 0/0; }" > simple.cc

echo "cmake_minimum_required(VERSION 3.10)" > CMakeLists.txt
echo "project(SimpleProject C CXX)" >> CMakeLists.txt
echo "add_executable(simple_app simple.c simple.cc)" >> CMakeLists.txt

mkdir build
cd build
cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
cd ..
CodeChecker analyze --analyzers clangsa clang-tidy --output=data --file=simple.c build/compile_commands.json
```
2. See that CodeChecker analyzed both `simple.c` and `simple.cc`
```bash
[INFO 2025-09-22 09:21] - Starting static analysis ...
[INFO 2025-09-22 09:21] - [1/4] clangsa analyzed simple.cc successfully.
[INFO 2025-09-22 09:21] - [1/4] clangsa analyzed simple.c successfully.
[INFO 2025-09-22 09:21] - [2/4] clang-tidy analyzed simple.cc successfully.
[INFO 2025-09-22 09:21] - [3/4] clang-tidy analyzed simple.c successfully.
```

**Expected behaviour**
I expected that only `simple.c` would be analyzed

**Desktop
 - OS: Linux - Ubuntu (ELX) 24.04
 - Browser: Firefox (143.0)

PR #4670, opened 2025-09-26
Fix over-eager file matching in skip files and `--file` option

furtib:
Why:
In the current implementation of the skipfiles, every file declared will match any other file that contains the declared file name at the beginning of its name. An example would be: `simple.c` and `simple.cpp`. Specifying `simple.c` with the `--flag` option would result in a regex: `/path/to/simple.c.*`

What:
Removed adding '*' to the end of any regex given through `--file` or the skipfile.

Notes:
According to the [documentation](https://codechecker.readthedocs.io/en/latest/analyzer/user_guide/#skip-file), this change is breaking, but I could see some users declaring directories in their skip file without placing a '*' at the end.

Fixes: #4664

### DIFF
--- a/codechecker_common/skiplist_handler.py
+++ b/codechecker_common/skiplist_handler.py
@@ -51,8 +51,6 @@ class SkipListHandler:
         the regular expressions.
         """
         for skip_line in skip_lines:
-            if skip_line[-1] == '/':
-                skip_line += '*'
             norm_skip_path = os.path.normpath(skip_line[1:].strip())
             rexpr = re.compile(
                 fnmatch.translate(norm_skip_path))
--- a/codechecker_common/skiplist_handler.py
+++ b/codechecker_common/skiplist_handler.py
@@ -66,7 +66,7 @@ class SkipListHandler:
             elif translated_glob.endswith(r"\Z"):
                 translated_glob = translated_glob[:-2]
             rexpr = re.compile(
-                translated_glob + r"(?:/.*)?$")
+                translated_glob + r"(?:" + os.path.normpath("/") + r".*)?$")
             self.__skip.append((skip_line, rexpr))
 
     def __check_line_format(self, skip_lines):
--- a/codechecker_common/skiplist_handler.py
+++ b/codechecker_common/skiplist_handler.py
@@ -10,6 +10,7 @@
 import fnmatch
 import re
 import os
+import sys
 
 from codechecker_common.logger import get_logger
 
@@ -52,15 +53,20 @@ class SkipListHandler:
         """
         for skip_line in skip_lines:
             norm_skip_path = os.path.normpath(skip_line[1:].strip())
-            # fnmatch places a '\Z' at the end, this  is equivalent to '$'
-            # (end of input) so it must be removed.
+            # fnmatch places a '\Z' (end of input) at the end, in this case,
+            # this is equivalent to '$' (end of line). This must be removed.
             # Optionally we match the user given path with a '/.*' ending.
             # This, if the given input is a folder will
             # resolve all files contained within.
             # Note: normalization removes '/' from the end, see:
             # https://docs.python.org/3/library/os.path.html#os.path.normpath
+            translated_glob = fnmatch.translate(norm_skip_path)
+            if sys.version_info.minor >= 9:
+                translated_glob = translated_glob.removesuffix(r"\Z")
+            elif translated_glob.endswith(r"\Z"):
+                translated_glob = translated_glob[:-2]
             rexpr = re.compile(
-                fnmatch.translate(norm_skip_path)[:-2] + r"(?:/.*)?\Z")
+                translated_glob + r"(?:/.*)?$")
             self.__skip.append((skip_line, rexpr))
 
     def __check_line_format(self, skip_lines):
--- a/codechecker_common/skiplist_handler.py
+++ b/codechecker_common/skiplist_handler.py
@@ -51,6 +51,8 @@ class SkipListHandler:
         the regular expressions.
         """
         for skip_line in skip_lines:
+            if skip_line[-1] == '/':
+                skip_line += '*'
             norm_skip_path = os.path.normpath(skip_line[1:].strip())
             rexpr = re.compile(
                 fnmatch.translate(norm_skip_path))
--- a/codechecker_common/skiplist_handler.py
+++ b/codechecker_common/skiplist_handler.py
@@ -61,9 +61,7 @@ class SkipListHandler:
             # Note: normalization removes '/' from the end, see:
             # https://docs.python.org/3/library/os.path.html#os.path.normpath
             translated_glob = fnmatch.translate(norm_skip_path)
-            if sys.version_info.minor >= 9:
-                translated_glob = translated_glob.removesuffix(r"\Z")
-            elif translated_glob.endswith(r"\Z"):
+            if translated_glob.endswith(r"\Z"):
                 translated_glob = translated_glob[:-2]
             rexpr = re.compile(
                 translated_glob + fr"(?:\{os.path.sep}.*)?$")
--- a/codechecker_common/skiplist_handler.py
+++ b/codechecker_common/skiplist_handler.py
@@ -52,8 +52,15 @@ class SkipListHandler:
         """
         for skip_line in skip_lines:
             norm_skip_path = os.path.normpath(skip_line[1:].strip())
+            # fnmatch places a '\Z' at the end, this  is equivalent to '$'
+            # (end of input) so it must be removed.
+            # Optionally we match the user given path with a '/.*' ending.
+            # This, if the given input is a folder will
+            # resolve all files contained within.
+            # Note: normalization removes '/' from the end, see:
+            # https://docs.python.org/3/library/os.path.html#os.path.normpath
             rexpr = re.compile(
-                fnmatch.translate(norm_skip_path) + r"(?:\/.*)?")
+                fnmatch.translate(norm_skip_path)[:-2] + r"(?:/.*)?\Z")
             self.__skip.append((skip_line, rexpr))
 
     def __check_line_format(self, skip_lines):
--- a/codechecker_common/skiplist_handler.py
+++ b/codechecker_common/skiplist_handler.py
@@ -66,7 +66,7 @@ class SkipListHandler:
             elif translated_glob.endswith(r"\Z"):
                 translated_glob = translated_glob[:-2]
             rexpr = re.compile(
-                translated_glob + r"(?:" + os.path.normpath("/") + r".*)?$")
+                translated_glob + fr"(?:{os.path.sep}.*)?$")
             self.__skip.append((skip_line, rexpr))
 
     def __check_line_format(self, skip_lines):
--- a/codechecker_common/skiplist_handler.py
+++ b/codechecker_common/skiplist_handler.py
@@ -66,7 +66,7 @@ class SkipListHandler:
             elif translated_glob.endswith(r"\Z"):
                 translated_glob = translated_glob[:-2]
             rexpr = re.compile(
-                translated_glob + fr"(?:{os.path.sep}.*)?$")
+                translated_glob + fr"(?:\{os.path.sep}.*)?$")
             self.__skip.append((skip_line, rexpr))
 
     def __check_line_format(self, skip_lines):
--- a/codechecker_common/skiplist_handler.py
+++ b/codechecker_common/skiplist_handler.py
@@ -53,7 +53,7 @@ class SkipListHandler:
         for skip_line in skip_lines:
             norm_skip_path = os.path.normpath(skip_line[1:].strip())
             rexpr = re.compile(
-                fnmatch.translate(norm_skip_path))
+                fnmatch.translate(norm_skip_path) + r"(?:\/.*)?")
             self.__skip.append((skip_line, rexpr))
 
     def __check_line_format(self, skip_lines):
--- a/codechecker_common/skiplist_handler.py
+++ b/codechecker_common/skiplist_handler.py
@@ -53,7 +53,7 @@ class SkipListHandler:
         for skip_line in skip_lines:
             norm_skip_path = os.path.normpath(skip_line[1:].strip())
             rexpr = re.compile(
-                fnmatch.translate(norm_skip_path + '*'))
+                fnmatch.translate(norm_skip_path))
             self.__skip.append((skip_line, rexpr))
 
     def __check_line_format(self, skip_lines):


furtib:
I have added a clause to add the star to entries that are evidently folders.
Users will use the '/' to point to folders even on Windows.
With this, I do respect that no filename *may* end with the '/' character.
Filenames can't contain the '/' character on [Windows](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#:~:text=Use%20any%20character%20in%20the%20current%20code%20page%20for%20a%20name%2C%20including%20Unicode%20characters%20and%20characters%20in%20the%20extended%20character%20set%20(128%E2%80%93255)%2C%20except%20for%20the%20following%3A). On linux I think common sense might stop users from ending their filenames with a '/'.

@barnabasdomozi, what are your thoughts?

furtib:
I cannot be sure if the path you gave is for a file named build or for a folder, so I do not handle that case.

During a conversation with Botond, I got the suggestion to instead of adding a '*' after a '/' we could add an optional '/.*' to the end of our regex. This would look something like this: `(?:\/.*)`?.

This would solve the issue of `simple.c` and `simple.cc` while also preserving the workings for a directory path defined like: `/workspace/myproject/build`

codechecker_common/skiplist_handler.py
```diff
@@ -52,8 +52,15 @@ def __gen_regex(self, skip_lines):
...
+            # fnmatch places a '\Z' at the end, this  is equivalent to '$'
+            # (end of input) so it must be removed.
+            # Optionally we match the user given path with a '/.*' ending.
+            # This, if the given input is a folder will
+            # resolve all files contained within.
+            # Note: normalization removes '/' from the end, see:
+            # https://docs.python.org/3/library/os.path.html#os.path.normpath
             rexpr = re.compile(
-                fnmatch.translate(norm_skip_path + '*'))
+                fnmatch.translate(norm_skip_path)[:-2] + r"(?:/.*)?\Z")
```
barnabasdomozi: If `fnmatch.translate()` function's implementation changes, e.g. it no longer puts `\Z` in the end, then `[:-2]` breaks the regex. So instead of `[:-2]` consider using `rstrip()`

Also, is there a reason we are using `\Z` in the end of the regex and not `$`?

barnabasdomozi (changes requested):
Great work!
See my comment about a minor issue.

We should also investigate which slash type is used on Windows.

codechecker_common/skiplist_handler.py
furtib: I wrote a simple check to see if \Z is there, before removing it.
rstrip(r"\Z") would have stripped away any path ending in z.
e.g: "abcz\Z" -> "abc"

furtib:
I have added a check to the fnmatch. translate output, so as not to depend so much on its implementation.
I also tried to support Windows systems. `os.path.normalize("/")` will return `"\\"` (an escaped backslash). This way, we will check for an optional `\\,*` on Windows

analyzer/tests/functional/skip/test_skip.py
```diff
@@ -170,13 +171,66 @@ def test_skip(self):
                             "Reports from headers should be kept"
                             " if the header is not on the skiplist")
 
+    def test_skip_directory(self):
+        """Test wether directories in skipfile entries works correctly"""
```
barnabasdomozi: Typo: "wether" should be "whether"

analyzer/tests/functional/skip/test_skip.py
```diff
@@ -482,11 +549,23 @@ def test_analyze_file_option(self):
                 any('skip_header.cpp' not in f for f in glob.glob(
                     os.path.join(self.report_dir, '*.plist'))))
 
+    def test_analyze_similar_file_option(self):
+        """
+        Test analyze command --file option for edge case,
+        where one filename is a prefix of an other.
```
barnabasdomozi: Typo: another

analyzer/tests/functional/skip/test_skip.py
```diff
@@ -170,13 +171,66 @@ def test_skip(self):
...
+            skip_file.flush()
+            self.__log_and_analyze("skip_folder", ["--ignore", skip_file.name])
+
+            # Check if the folder is skipped
+            # There shouldn't be any report generated
+            report_dir_files = os.listdir(self.report_dir)
+            for f in report_dir_files:
+                self.assertFalse("skipme.cpp" in f)
+
+        with tempfile.NamedTemporaryFile(
```
barnabasdomozi: These 3 code blocks seem redundant and only differ in skipfile content.
Consider creating a loop for the cases `['-*skipme/*', '-*skipme/', '-*skipme']` or define a helper function.

analyzer/tests/functional/skip/test_files/similar/Makefile
```diff
@@ -0,0 +1,14 @@
+OBJS = $(SRCS:.cpp=.o)
+
+CXXFLAGS = -Wno-all -Wno-extra -Wno-division-by-zero
+
+SRCS = simple.c \
+		simple.cc
```
barnabasdomozi: We usually use the `.cpp` extension rather than `.cc` throughout the C++ test files.

analyzer/tests/functional/skip/test_files/similar/Makefile
```diff
@@ -0,0 +1,14 @@
...
+
+.cpp.o:
```
barnabasdomozi: Currently there are no `.cpp` files in this folder. There is also no rule for `.c` files.

analyzer/tests/functional/skip/test_files/similar/Makefile
```diff
@@ -0,0 +1,14 @@
+OBJS = $(SRCS:.cpp=.o)
```
barnabasdomozi: This Makefile looks incomplete.
`OBJS` is not defined for `.c` or `.cc` files. Also note there are no `.cpp` files in this directory.

analyzer/tests/functional/skip/test_skip.py
```diff
@@ -170,13 +171,66 @@ def test_skip(self):
...
+            mode="w", suffix="skipfile", encoding="utf-8"
+        ) as skip_file:
+
+            # Skip the skipme folder, without any indication that its a folder
```
barnabasdomozi: Nit: it should be `it's` or `it is`. `its` is incorrect in this sentence and has a different meaning.

codechecker_common/skiplist_handler.py
```diff
@@ -52,8 +52,18 @@ def __gen_regex(self, skip_lines):
...
+            # This, if the given input is a folder will
+            # resolve all files contained within.
+            # Note: normalization removes '/' from the end, see:
+            # https://docs.python.org/3/library/os.path.html#os.path.normpath
+            translated_glob = fnmatch.translate(norm_skip_path)
+            if translated_glob.endswith(r"\Z"):
+                translated_glob = translated_glob[:-2]
             rexpr = re.compile(
-                fnmatch.translate(norm_skip_path + '*'))
+                translated_glob + fr"(?:\{os.path.sep}.*)?$")
```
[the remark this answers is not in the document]
furtib: There is a long comment before it explaining what it does, and its purpose.

analyzer/tests/functional/skip/test_skip.py
```diff
@@ -482,11 +521,23 @@ def test_analyze_file_option(self):
...
+    def test_analyze_similar_file_option(self):
+        """
+        Test analyze command --file option for edge case,
+        where one filename is a prefix of another.
+        """
+        self.__log_and_analyze("similar", ["--file", "*/simple.c"])
+        print(glob.glob(
+            os.path.join(self.report_dir, '*.plist')))
+        self.assertFalse(
+            any('simple.cpp' not in f for f in glob.glob(
```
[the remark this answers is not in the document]
furtib: You are right!

analyzer/tests/functional/skip/test_skip.py
```diff
@@ -170,13 +171,38 @@ def test_skip(self):
...
+            ["-*skipme/*"],  # The way its in the docs
+            ["-*skipme/"],  # If the user specified that this is a directory
+            ["-*skipme"],  # If we can't even be sure that this is a directory
+        ]
+        for skip_file_content in skip_file_contents:
+            with tempfile.NamedTemporaryFile(
+                mode="w", suffix="skipfile", encoding="utf-8"
+            ) as skip_file:
+
+                # Skip the `skipme` folder, the way its in the documentation
```
barnabasdomozi: Please update the comment here as well.

analyzer/tests/functional/skip/test_files/similar/Makefile
```diff
@@ -0,0 +1,20 @@
...
+CPP_OBJS = $(CPP_SRCS:.cpp=_cpp.o)
+
+OBJS =  $(C_OBJS) $(CPP_OBJS)
+
+CXXFLAGS = -Wno-all -Wno-extra -Wno-division-by-zero
+
+C_SRCS = simple.c
+CPP_SRCS = simple.cpp
+
+all: $(C_OBJS) $(CPP_OBJS)
```
barnabasdomozi: Use `$(OBJS)` here.

analyzer/tests/functional/skip/test_files/similar/Makefile
```diff
@@ -0,0 +1,20 @@
+C_OBJS = $(C_SRCS:.c=_c.o)
+CPP_OBJS = $(CPP_SRCS:.cpp=_cpp.o)
+
+OBJS =  $(C_OBJS) $(CPP_OBJS)
+
+CXXFLAGS = -Wno-all -Wno-extra -Wno-division-by-zero
```
barnabasdomozi: Instead of overwriting `$(CXXFLAGS)`, consider appending these flags to its default value.

analyzer/tests/functional/skip/test_files/similar/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+C_SRCS = simple.c
+CPP_SRCS = simple.cpp
+
+all: $(C_OBJS) $(CPP_OBJS)
+
+$(CPP_OBJS): %_cpp.o: %.cpp
+	$(CXX) $(CXXFLAGS) -c $<  -o $@
```
barnabasdomozi:
```suggestion
	$(CXX) $(CXXFLAGS) -c $< -o $@
```
Superfluous space.

analyzer/tests/functional/skip/test_files/similar/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+$(C_OBJS): %_c.o: %.c
+	$(CC) $(CFLAGS) -c $< -o $@
+
+clean:
+	rm -rf *.o
```
barnabasdomozi:
```suggestion
	rm -rf $(OBJS)
```
Don't delete unrelated object files.

analyzer/tests/functional/skip/test_files/skip_folder/Makefile
```diff
@@ -0,0 +1,13 @@
+OBJS = $(SRCS:.cpp=.o)
+
+CXXFLAGS = -Wno-all -Wno-extra -Wno-division-by-zero
+
+SRCS = skipme/skipme.cpp
+
+.cpp.o:
+	$(CXX) $(CXXFLAGS) -c $<  -o $@
+
+all: $(OBJS)
```
barnabasdomozi, line 7: Move `all` up to ensure that it's the first rule (which is executed when `make` is ran without any arguments). I'm not sure that the "magical" rule `.cpp.o` can be the default, but it's better to avoid confusion.

analyzer/tests/functional/skip/test_files/skip_folder/Makefile
```diff
@@ -0,0 +1,13 @@
+OBJS = $(SRCS:.cpp=.o)
+
+CXXFLAGS = -Wno-all -Wno-extra -Wno-division-by-zero
```
barnabasdomozi: As above, append these flags to the default `$(CXXFLAGS)`.

analyzer/tests/functional/skip/test_files/skip_folder/Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+SRCS = skipme/skipme.cpp
+
+.cpp.o:
+	$(CXX) $(CXXFLAGS) -c $<  -o $@
+
+all: $(OBJS)
+
+clean:
+	rm -rf skipme/*.o
```
barnabasdomozi:
```suggestion
	rm -rf $(OBJS)
```

analyzer/tests/functional/skip/test_files/similar/Makefile
```diff
@@ -0,0 +1,20 @@
...
+C_SRCS = simple.c
+CPP_SRCS = simple.cpp
+
+all: $(C_OBJS) $(CPP_OBJS)
+
+$(CPP_OBJS): %_cpp.o: %.cpp
+	$(CXX) $(CXXFLAGS) -c $<  -o $@
+
+$(C_OBJS): %_c.o: %.c
+	$(CC) $(CFLAGS) -c $< -o $@
```
barnabasdomozi, line 18: You modified `$(CXXFLAGS)`, but you are using the default `$(CFLAGS)`. Is this your intention? (It is a reasonable default, if you don't need any customization.)

analyzer/tests/functional/skip/test_files/skip_folder/Makefile
```diff
@@ -0,0 +1,13 @@
+OBJS = $(SRCS:.cpp=.o)
+
+CXXFLAGS = -Wno-all -Wno-extra -Wno-division-by-zero
+
+SRCS = skipme/skipme.cpp
+
+.cpp.o:
+	$(CXX) $(CXXFLAGS) -c $<  -o $@
```
barnabasdomozi:
```suggestion
	$(CXX) $(CXXFLAGS) -c $< -o $@
```
[nit] Remove the extra space.

analyzer/tests/functional/skip/test_skip.py
```diff
@@ -170,13 +171,38 @@ def test_skip(self):
...
+                skip_file.write('\n'.join(skip_file_content))
+                skip_file.flush()
+                self.__log_and_analyze(
+                    "skip_folder", ["--ignore", skip_file.name]
+                )
+
+                # Check if the folder is skipped
```
barnabasdomozi:
```suggestion
                # Check if the folder `skipme` is skipped
```

analyzer/tests/functional/skip/test_skip.py
```diff
@@ -170,13 +171,38 @@ def test_skip(self):
...
+                # There shouldn't be any report generated
```
barnabasdomozi:
```suggestion
                # There shouldn't be any report generated for `skipme/skipme.cpp`
```

analyzer/tests/functional/skip/test_skip.py
```diff
@@ -170,13 +171,39 @@ def test_skip(self):
...
+            ["-*skipme/*"],  # The way its in the docs
+            ["-*skipme/"],  # If the user specified that this is a directory
+            ["-*skipme"],  # If we can't even be sure that this is a directory
+        ]
+        for skip_file_content in skip_file_contents:
+            with tempfile.NamedTemporaryFile(
+                mode="w", suffix="skipfile", encoding="utf-8"
+            ) as skip_file:
+
+                # Skip the `skipme` folder, the way it is in the documentation
```
barnabasdomozi: `"the way it is in the documentation"`

This comment seems outdated since the code was modified to a loop which has 3 different cases.
```suggestion
                # Skip the `skipme` folder
```

analyzer/tests/functional/skip/test_skip.py
```diff
@@ -170,13 +171,39 @@ def test_skip(self):
                             "Reports from headers should be kept"
                             " if the header is not on the skiplist")
 
+    def test_skip_directory(self):
+        """Test whether directories in skipfile entries works correctly"""
+        skip_file_contents: list[list[str]] = [
+            ["-*skipme/*"],  # The way its in the docs
```
barnabasdomozi:
```suggestion
            ["-*skipme/*"],  # The way it is in the documentation
```

analyzer/tests/functional/skip/test_files/similar/Makefile
```diff
@@ -0,0 +1,21 @@
+C_OBJS = $(C_SRCS:.c=_c.o)
+CPP_OBJS = $(CPP_SRCS:.cpp=_cpp.o)
+
+OBJS =  $(C_OBJS) $(CPP_OBJS)
```
barnabasdomozi:
```suggestion
OBJS = $(C_OBJS) $(CPP_OBJS)
```

[nit] Remove extra space.

PR #4671, opened 2025-09-28
[fix] get analyzer name from SARIF report

Rayzedan:
**Describe the bug**
When saving the report in SARIF format, the analyzer name is not saved. 

**CodeChecker version**
6.26.0

**To Reproduce**

`CodeChecker store ./example_sarif -n test`

**Expected behaviour**

The analyzer name is displayed in the Web UI.

- OS: Linux
- Browser: Chrome

    
**Additional context**

error

### DIFF
--- a/tools/report-converter/codechecker_report_converter/report/parser/sarif.py
+++ b/tools/report-converter/codechecker_report_converter/report/parser/sarif.py
@@ -63,6 +63,7 @@ class Parser(BaseParser):
 
         for run in data.runs:
             rules = self._get_rules(run.run_data)
+            analyzer_name = self._get_analyzer_name(run.run_data)
             # $3.14.14
             self.original_uri_base_ids = None
             if "originalUriBaseIds" in run.run_data:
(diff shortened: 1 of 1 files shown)


tools/report-converter/codechecker_report_converter/report/parser/sarif.py
```diff
@@ -129,6 +131,14 @@ def _get_rules(self, data: Dict) -> Dict[str, Dict]:
 
         return rules
 
+    def _get_analyzer_name(self, data: Dict) -> str:
+        """ Get analyzer name from SARIF report. """
+        try:
+            driver = data.get("tool", {}).get("driver", {})
+            return driver.get("name") or "unknown"
```
bruntib: I think the `try` block is not needed because the `get()` method of `dict` type doesn't require the existence of the key. That's why a default value is provided in the 2nd parameter.
By the way, is this the same?
```.py
return data.get("tool", {}).get("driver", {}).get("name", "unknown")
```

bruntib (changes requested):
Thanks for the fix, I have just one minor comment.

tools/report-converter/codechecker_report_converter/report/parser/sarif.py
Rayzedan: Makes sense, thanks. I've applied the changes

PR #4672, opened 2025-09-29
Replace CTU query functions to ClangSA

bruntib:
CTU analysis is a feature of ClangSA. There are some query functions that check if CTU feature is available with the installed Clang binary. These query functions were implemented at several places (clangsa/analyzer.py, analyzer_types.py, analysis_manager.py). There was also an anomaly with naming: `is_ctu_on_demand_available` vs. `is_on_demand_ctu_available`.

This patch replaces these functions under ClangSA, so the common analyzer-related files are not polluted with this information.


barnabasdomozi (approved):

dkrupp:
@bruntib  you can even remove these checks alltogether as all clang these days contain CTU feature. 

This was only useful when we only had the CTU feature dowsntream. So better to remove every such detection.

PR #4680, opened 2025-10-06
Fix statistics numbers

gulyasgergely902:


web/server/vue-cli/src/components/Report/ReportFilter/ReportCount.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <span class="pa-1 text-no-wrap">
+  <span class="pa-1 text-no-wrap" title="Outstanding reports">
```
bruntib: This counter shows not only the outstanding reports, but the number of all reports that match the filter. "Number of reports" would be more adequate if any.

bruntib (changes requested):

bruntib (approved):

PR #4690, opened 2025-10-23
[fix] Cppcheck reports from checkers only

bruntib:
Sometimes Cppcheck emits informative diagnostics that are not related to defects, e.g. normalCheckLevelMaxBranches. These should be filtered out.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
@@ -292,8 +292,21 @@ class Cppcheck(analyzer_base.SourceAnalyzer):
         environ = analyzer_context.get_context().get_env_for_bin(
             command[0])
         try:
-            result = subprocess.check_output(command, env=environ)
-            return parse_checkers(result)
+            errorlist_output = subprocess.check_output(command, env=environ)
+            checkers = parse_checkers(errorlist_output)
+
+            # Cppcheck can and will report with checks that have a different
+            # name than marked in the --errorlist xml. To be able to suppress
+            # these reports, the checkerlist needs to be extended by those
+            # found in the label file.
+            checkers_from_label = analyzer_context \
+                .get_context().checker_labels.checkers("cppcheck")
+            parsed_checker_names = set(checker[0] for checker in checkers)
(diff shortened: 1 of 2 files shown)


analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
```diff
@@ -405,12 +418,12 @@ def construct_result_handler(self, buildaction, report_output,
                                             self.config_handler.report_hash)
 
         res_handler.skiplist_handler = skiplist_handler
+        res_handler.analyzer = self
```
barnabasdomozi, line 421: Why is setting the analyzer here needed?
There could be a valid reason, it's just not clear to me in this case.

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
```diff
@@ -292,8 +292,21 @@ def get_analyzer_checkers(cls):
...
+            errorlist_output = subprocess.check_output(command, env=environ)
+            checkers = parse_checkers(errorlist_output)
+
+            # Cppcheck can and will report with checks that have a different
+            # name than marked in the --errorlist xml. To be able to suppress
+            # these reports, the checkerlist needs to be extended by those
+            # found in the label file.
+            checkers_from_label = analyzer_context \
+                .get_context().checker_labels.checkers("cppcheck")
+            parsed_set = set(data[0] for data in checkers)
```
gulyasgergely902: Just a minor detail: data should be actually called a "checker" in single form. At the end of the day cherckers list contains object of that and it would better describe what is is. Also instead of parsed_set I'd call it a parsed_checker_names since the 0th element of a data (or checker) object is its name (according to parse_checkers function).

gulyasgergely902 (changes requested):

analyzer/codechecker_analyzer/analyzers/cppcheck/analyzer.py
bruntib: I wanted to fetch the list of checkers in the result handler from the analyzer. But it resulted a circular import, so it didn't work.

PR #4697, opened 2025-11-14
Fix clang-tidy 'clang-diagnostic-*' checkers

barnabasdomozi:
## Issue

When running
```bash
CodeChecker analyze --analyzers clang-tidy \
  -d profile:default -e guideline:memory-safety \
  ./compile_commands.json -o reports
```

we enable the following checkers: `clang-diagnostic-return-stack-address`, `clang-diagnostic-uninitialized*`

and then construct the following `clang-tidy` call:
```bash
/usr/lib/llvm-14/bin/clang-tidy \
-checks=-*,clang-diagnostic-return-stack-address,clang-diagnostic-uninitialized* \
...
```
When executing the above command, `clang-tidy` exits with the following message:
```
Error: no checks enabled.
USAGE: clang-tidy [options] <source0> [... <sourceN>]

OPTIONS:

Generic Options:

  --help                         - Display available options (--help-hidden for more)
  --version                      - Display the version of this program
```
## Solution
~~Checker names that begin with `clang-diagnostic-` are not valid `clang-tidy` checker names but compiler warning flags
 (e.g. `-Wreturn-stack-address`), and should not be added to the `checks` list.~~
 
 
 Enable a dummy checker if only `clang-diagnostic-` checkers were used.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangtidy/analyzer.py
@@ -417,20 +417,20 @@ class ClangTidy(analyzer_base.SourceAnalyzer):
         has_checker_config = \
             config.checker_config and config.checker_config != '{}'
 
-        CLANG_DIAGNOSTIC_PREFIX = 'clang-diagnostic-'
+        clang_diagnostic_prefix = 'clang-diagnostic-'
 
         # Config handler stores which checkers are enabled or disabled.
         for checker_name, value in config.checks().items():
             state, _ = value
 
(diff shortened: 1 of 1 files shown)


barnabasdomozi:
@dkrupp 
Added a dummy checker named `darwin-dispatch-once-nonstatic` (https://clang.llvm.org/extra/clang-tidy/checks/darwin/dispatch-once-nonstatic.html) if only `clang-diagnostic-*` checkers were used.

Also implemented a test case where all checkers turned off except `clang-diagnostic-return-stack-address`.

PR #4701, opened 2025-11-17
Remove default report filter values which only adds Detection status and Review status filters which does not required here.

gulyasgergely902:
This change removes the default report filter from link on the expanded runs page which links to the reports. This way Detection Status and Review Status filters are not getting applied to the resulting reports list.

### DIFF
--- a/web/server/vue-cli/src/components/Run/ExpandedRun.vue
+++ b/web/server/vue-cli/src/components/Run/ExpandedRun.vue
@@ -34,7 +34,6 @@
             <router-link
               :to="{ name: 'reports',
                      query: {
-                       ...defaultReportFilterValues,
                        run: history.runName,
                        'run-tag': history.id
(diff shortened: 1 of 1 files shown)


web/server/vue-cli/src/components/Run/ExpandedRun.vue
```diff
@@ -34,7 +34,6 @@
             <router-link
               :to="{ name: 'reports',
                      query: {
-                       ...defaultReportFilterValues,
```
bruntib, line 37: The `defaultReportFilterValues` in line 141 and the `data()` section in line 163 could also be removed, since they now become unused.

bruntib (changes requested):

dkrupp (approved):

bruntib (approved):

PR #4714, opened 2025-11-28
[refactor] Do not use pinned versions in requirements.txt

thorsten-klein:
Follow-up of #4501.

It is not recommended by the Python Packaging User Guide to pin dependencies. Instead, use version ranges or minimum/maximum versions wherever possible.

## Why?

With zephyr 4.3.0 I get some conflict:
```
The conflict is caused by:
    The user requested setuptools==70.2.0
    The user requested setuptools==70.2.0
    codechecker 6.26.2 depends on setuptools==70.2.0
    hidapi 0.14.0.post4 depends on setuptools>=19.0
    setuptools-scm 8.1.0 depends on setuptools
    spsdk 3.4.0 depends on setuptools<81 and >75
```

### DIFF
--- a/analyzer/requirements.txt
+++ b/analyzer/requirements.txt
@@ -1,3 +1,9 @@
+# These requirements are exported via setup.py and become the install
+# dependencies of the published pip package.
+# Avoid pinning exact versions!
+# Use version ranges or upper bounds instead (e.g. limited to a major release)
+# to let users select a specific compatible version.
+
 lxml<6
 portalocker<4
 psutil<8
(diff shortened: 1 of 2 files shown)


pdgendt (approved):

thorsten-klein:
If anybody else is stumbling across this: We have workarounded this issue by switching to `uv pip install` which supports `--overrides overrides.txt` argument to relax pinned versions.

The `overrides.txt` contains the same as done in this PR:
```
alembic<2
Authlib<2
gitpython<4
lxml<6
multiprocess<0.71
portalocker<4
psutil<8
PyYAML<7
requests<3
sarif-tools<4
setuptools<81
sqlalchemy<2
thrift<0.23
types-PyYAML<7
```

PR #4715, opened 2025-11-28
Dump compiler_info.json

bruntib:
A new flag is added to "CodeChecker analyze": --dump-compiler-info-file. This flag is given a filename as parameter. This filename will be created in the output folder. The file contains implicit include paths and some other gcc-specific data that can be used to finetune analysis later. When this flag is used, then analysis doesn't run.

The compiler_info.json was generated in the output directory by default. Due to some bug, this file is empty in recent CodeChecker versions. The reason is that parsing compile_commands.json is done in parallel with a process pool. The dict object that collects this data must be handled by multiprocessing.SyncManager() so it can be used in a process pool.

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -7,7 +7,7 @@
 # -------------------------------------------------------------------------
 
 
-from collections import namedtuple
+from dataclasses import dataclass
 from shutil import which
 from enum import Enum
 from functools import lru_cache
@@ -322,9 +322,25 @@ class ImplicitCompilerInfo:
     # attributes to the implicit settings. In the future we may find that some
     # other attributes are also dependencies of implicit compiler info in which
     # case this tuple should be extended.
-    ImplicitInfoSpecifierKey = namedtuple(
-        'ImplicitInfoSpecifierKey',
-        ['compiler', 'language', 'compiler_flags'])
+    @dataclass
+    class ImplicitInfoSpecifierKey:
+        compiler: str
+        language: str
+        compiler_flags: list[str]
+
+        def __str__(self):
+            return json.dumps([
+                self.compiler,
+                self.language,
+                self.compiler_flags])
+
+        def __eq__(self, other):
+            return (self.compiler, self.language, self.compiler_flags) == \
+                   (other.compiler, other.language, other.compiler_flags)
+
+        def __hash__(self):
+            return hash(
+                (self.compiler, self.language, tuple(self.compiler_flags)))
 
     compiler_info: Dict[ImplicitInfoSpecifierKey, dict] = {}
     compiler_isexecutable = {}
@@ -332,14 +348,6 @@ class ImplicitCompilerInfo:
     # If the value is False the compiler is not clang otherwise the value
     # should be a clang version information object.
     compiler_versions = {}
-
-    @staticmethod
-    def c():
-        return "c"
-
-    @staticmethod
-    def cpp():
-        return "c++"
 
     @staticmethod
     def is_executable_compiler(compiler):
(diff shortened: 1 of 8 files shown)


analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -322,9 +321,25 @@ class ImplicitCompilerInfo:
     # attributes to the implicit settings. In the future we may find that some
     # other attributes are also dependencies of implicit compiler info in which
     # case this tuple should be extended.
-    ImplicitInfoSpecifierKey = namedtuple(
-        'ImplicitInfoSpecifierKey',
-        ['compiler', 'language', 'compiler_flags'])
+    class ImplicitInfoSpecifierKey:
```
barnabasdomozi, line 326: Consider using a dataclass in this case which can reduce boilerplate code such as:
```py
def __init__(self, compiler, language, compiler_flags):
    self.compiler = compiler
    self.language = language
    self.compiler_flags = compiler_flags
```

analyzer/codechecker_analyzer/cli/analyze.py
```diff
@@ -1371,6 +1380,17 @@ def main(args):
                     "compilation database or all of them were skipped.")
         sys.exit(0)
 
+    if args.dump_compiler_info_file:
+        compiler_info = Path(args.output_path) / 'compiler_info.json'
+        try:
+            os.rename(
```
barnabasdomozi: From the code, it's not clear to me what this flag actually does. Renames an already existing file? Why?
Why not dump the file to output path `args.dump_compiler_info_file` in the first place?

The help description says:
```
Dump implicit gcc compiler info to a json file
that can be used for fine-tuning analysis later.
```

To me, it seems this file is already dumped regardless of this flag, just under a different name: `compiler_info.json`. Correct me if I'm wrong, but this part is not clear to me.

barnabasdomozi (changes requested):
I have left some questions when reviewing this patch.

Also note that analyzer tests are currently failing:
```
-   File "/home/runner/work/codechecker/codechecker/build/CodeChecker/lib/python3/codechecker_analyzer/cli/analyze.py", line 1383, in main
-     if args.dump_compiler_info_file:
- AttributeError: 'Namespace' object has no attribute 'dump_compiler_info_file'
```

analyzer/codechecker_analyzer/buildlog/log_parser.py
bruntib: Thanks for the suggestion, I changed it.

analyzer/codechecker_analyzer/cli/analyze.py
bruntib: The first idea was introducing this command:
```
CodeChecker analyze compile_commands.json --dump-compiler-info-file <filename>
```
This could have created the given file. However, `CodeChecker analyze` has a mandatory `-o flag`:
```
CodeChecker analyze compile_commands.json --dump-compiler-info-file <filename> -o reports
```
Intuitively this tells me that the compiler info file will be created into `reports` folder. However, `reports/compiler_info.json` is created anyway. I didn't find it elegant, that this file is created twice in the report folder: with the name `compiler_info.json` and the given `<filename>`.
This led me to the current solution which simply renames the `compiler_info.json` to the given filename.

Another solution could have been not dumping `compiler_info.json` in the `log_parser` module, but here in the `main()` function. Maybe we could refactor it in the future.

dkrupp: Can't we do this instead?

`CodeChecker analyze compile_commands.json --dump-compiler-info-file  -o <filename>
`

after the -o flag the user could directly specify the compiler_info.json. Then we could avoid this unfortunate renaming trick.

docs/analyzer/user_guide.md
```diff
@@ -1058,6 +1058,12 @@ optional arguments:
                         Read the compiler includes and target from the
                         specified file rather than invoke the compiler
                         executable.
+  --dump-compiler-info-file DUMP_COMPILER_INFO_FILE
```
dkrupp: DUMP_COMPILER_INFO_FILE can be removed as it does not require a parameter

PR #4717, opened 2025-11-28
support of out of tree built clang binary

dkrupp:


gulyasgergely902 (changes requested):
Check the issue with the tests otherwise looks good!

PR #4747, opened 2026-01-09
Fix the issue where the ReportTree component remains in a loading state when the result exceeds MAX_QUERY_SIZE

xJOAN:
The MAX_QUERY_SIZE constant is defined in the report_server.thrift file as const i64 MAX_QUERY_SIZE = 500. After being compiled into a JavaScript package, Thrift i64 constants are not of type number in JavaScript and therefore cannot be used directly.


bruntib:
Thanks for the fix!

PR #4751, opened 2026-01-12
Upgrade deps

bruntib:


bruntib:
This should be rebased after merging #4729.

PR #4758, opened 2026-01-21
Source component table

bruntib:


gulyasgergely902 (approved):

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -272,6 +274,50 @@ def get_file_content(file_path: str) -> bytes:
         return f.read()
 
 
+def assign_file_to_source_components(session, file_id, filepath):
```
barnabasdomozi: Please add type annotations to function parameters.

web/server/codechecker_server/api/report_server.py
```diff
@@ -146,41 +146,79 @@ def slugify(text):
...
+        if value[0] == '+':
+            include.append(v)
+        elif value[0] == '-':
+            skip.append(v)
+
+    return include, skip
+
+
+def update_source_component_files(
+    session,
```
barnabasdomozi: Please add type annotation for parameter `session`.

web/tests/functional/report_viewer_api/test_report_counting.py
```diff
@@ -283,30 +283,6 @@ def test_run1_all_file(self):
         self.assertEqual(len(res), len(self.run1_files))
         self.assertDictEqual(res, self.run1_files)
 
-    def test_filter_by_file_and_source_component(self):
```
barnabasdomozi, line 286: Can you explain why was this test case removed?

barnabasdomozi (changes requested):

web/tests/functional/report_viewer_api/test_report_counting.py
bruntib: This test was passing due to a bug in the test, The `ReportFilter.filepath` is supposed to be a list of file paths, however, this test is using a single file path as a string. The query generation iterates over the list (or the string by characters in this case) and constructs a condition with those elements.

The test intends to check if the report counts return for all file, even for those that don't have reports with the given filter. On the one hand this didn't work with even before this PR, on the other hand we don't have a functionality that needs to display those 0 numbers.

So, either we choose to rephrase `getFileCounts()` in a way that all files return even if they don't contain reports. Or, we specify this function as returning file counts with at least one report. I chose this 2nd option for now.

barnabasdomozi (approved):
LGTM

PR #4770, opened 2026-02-06
Remove store_time.log

barnabasdomozi:
With the introduction of the async store feature, it is no longer necessary to log store times in the workspace directory, as this information is now saved in the database.


dkrupp (approved):
LGTM

PR #4784, opened 2026-02-19
Fix CLI parse status filter

barnabasdomozi:
The parse status file filter now correctly handles relative file paths in the compile_commands.json file. Additonally, a bug was fixed when the filter was used for nonexistent files.

### DIFF
--- a/analyzer/codechecker_analyzer/cli/parse.py
+++ b/analyzer/codechecker_analyzer/cli/parse.py
@@ -299,7 +299,6 @@ def get_report_dir_status(compile_commands: List[dict[str, str]],
     for c in compile_commands:
         for analyzer in supported_analyzers:
             file, directory, cmd = c["file"], c["directory"], c["command"]
-            file = os.path.abspath(file)
 
             filename = os.path.basename(file)
             action_hash = analyzer_action_hash(file, directory, cmd)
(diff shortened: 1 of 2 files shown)


analyzer/codechecker_analyzer/cli/parse.py
```diff
@@ -407,10 +406,25 @@ def print_status(report_dir: str,
...
-        files_filter = [os.path.abspath(fp) for fp in files]
+        file_filter = [os.path.abspath(fp) for fp in files]
+
+        invalid_file_filter = [fp for fp in file_filter
+                               if not os.path.isfile(fp)]
+
+        if invalid_file_filter:
+            LOG.error("File filter (--file) contains nonexistent files:")
+            LOG.error(invalid_file_filter)
+            sys.exit(1)
```
bruntib, line 424: What is the purpose of checking the existence of the file? I think, it can be a valid scenario to ask the users to send me their report folder and I make some queries about their analyzed files on my machine where the project code is not available.

barnabasdomozi: In general, we assume the `report_dir/` and the source files are on the same system.
Typical use case for the `--status` command with `--file` option is to check whether a local file has been analysed or not. This is also used heavily in the VSCode plugin. Moreover, we would also like to know if a file has been recently analyzed or outdated, thus we check the modification time of the local source file in function `get_report_dir_status`. If the modification time of the source file is more recent than the creation of the `report_dir/`, we consider the analysis status outdated. 

This all assumes the report directory and the source files are on the same system. If this is the case, it is pointless to use the `--file` filter on a nonexisting file, thus an error was added.

> I think, it can be a valid scenario to ask the users to send me their report folder and I make some queries about their analyzed files on my machine where the project code is not available.

While this is valid, I believe this is a separate use case. As an alternative, maybe we could add a flag like `--ignore-nonexisting-files` or something like that. But in the above case, we should at least somehow inform the user that they tried to query analysis information of non-existing files.

PR #4787, opened 2026-02-24
[feat] Display guideline levels in the guideline statistics table

bruntib:
Some guidelines have their own severity ranking, like the levels at SEI-CERT. In this commit these are also displayed at the guideline statistics table.


web/api/report_server.thrift
```diff
@@ -572,10 +572,11 @@ struct Guideline {
...
 
 struct Rule {
-  1: string ruleId,        // The identifier of the rule.
-  2: string title,         // The rule summary.
-  3: string url,           // The link of the rule page.
-  4: list<string> checkers // List of checker names
+  1: string ruleId,          // The identifier of the rule.
+  2: string title,           // The rule summary.
+  3: string url,             // The link of the rule page.
+  4: list<string> checkers,  // List of checker names,
```
barnabasdomozi:
```suggestion
  4: list<string> checkers,  // List of checker names.
```

web/api/report_server.thrift
```diff
@@ -572,10 +572,11 @@ struct Guideline {
...
+  5: optional string level,
```
barnabasdomozi: Please add a description what `level` is and why is it `optional`.

barnabasdomozi (changes requested):

gulyasgergely902 (commented):
If not required in the next release, please, wait with this until the migration is complete to Vue 3.

web/api/report_server.thrift
bruntib: It is optional, because allegedly, thrift can provide backward compatibility if added members are optional. I haven't checked the truth of this statement, but we follow this convention. See any other structs or function parameters.

barnabasdomozi (approved):
LGTM

PR #4798, opened 2026-03-05
Include skiplist file in the analysis info

barnabasdomozi:
This patch introduces a new table called AnalysisInfoFile,
which is intended to store files related to analysis information.
The actual file contents are stored separately in the FileContent
table.

During CodeChecker analyze, we create a new directory called
conf/ in the report directory.
The entire conf/ directory will be added to the massStoreRun
ZIP file.

With this PR, only the skipfile is copied to the conf directory,
but this could be extended later.


web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -1000,6 +1004,28 @@ def __store_analysis_statistics(
...
+        analysis_info_id: int,
+        report_dir_path: str
+    ):
+        """ Store analyzer related config files (e.g. skipfile) """
+        zip_report_dir = os.path.join(
+            self._zip_dir, "reports",
+            hashlib.md5(report_dir_path.encode('utf-8')).hexdigest())
+
+        skip_file = os.path.join(zip_report_dir, 'skip_file')
+        if os.path.isfile(skip_file):
```
dkrupp: It would be great to preserve the original file name of the config files as the user sees the command with its oriignal file name.
Now the name skip_file is wired in. ... And I see that it is wired in at the analysis client.

I suggest creating a "conf" subdirectory in the report_directory where you collect all config files used by the CodeChecker analyze command with their original names.

Then the name could be preserved.

Image

dkrupp (dismissed):
Please keep the original config file name.
Otherwise looks good.

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -1012,37 +1042,30 @@ def __store_analysis_info(
                     analyzer_command.encode("utf-8"),
                     zlib.Z_BEST_COMPRESSION)
 
-                analysis_info_rows = session \
-                    .query(AnalysisInfo) \
-                    .filter(AnalysisInfo.analyzer_command == cmd) \
-                    .all()
-
-                if analysis_info_rows:
```
bruntib, line 1020: Why has this "if" statement been removed? In the new version, a new analysis info is added. What happens if it already exists in the database?

web/server/codechecker_server/migrations/report/versions/29e5047b6513_add_analysis_info_files_table.py
```diff
@@ -0,0 +1,51 @@
...
+
+from logging import getLogger
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# Revision identifiers, used by Alembic.
+revision = '29e5047b6513'
+down_revision = '198654dac219'
```
bruntib: This file needs to be regenerated, because the latest revision is newer than this.

bruntib (changes requested):

web/server/codechecker_server/migrations/report/versions/29e5047b6513_add_analysis_info_files_table.py
barnabasdomozi: Done in commit https://github.com/Ericsson/codechecker/commit/8659f56e2b498b51fe32af4d9afc1c8a54dc85c9

web/server/codechecker_server/api/mass_store_run.py
barnabasdomozi: Previously, the `if` check was used to determine whether a given analyzer command had already been inserted into the "AnalysisInfo" table, and if so, no new row was created. 

However, with the introduction of the AnalysisInfoFile table, the design now includes a mapping between an AnalysisInfo ID and the corresponding FileContent rows that should be displayed on the analysis info page. If we keep the old `if` check, this leads to incorrect behavior. 

Consider a scenario where 2 runs use the same analyzer command (for example, `CodeChecker analyze -i ./skipfile ...`), but the skipfile content differs between them. 

When storing the first run, a new row is inserted into the AnalysisInfo table (e.g., ID 1), and the associated file content is linked through the AnalysisInfoFile table. For the second run, the `if` check detects that the analyzer command already exists and reuses AnalysisInfo ID 1 instead of creating a new row, while still inserting a new entry into AnalysisInfoFile linked to that same ID. As a result, both runs share the same AnalysisInfo entry, and we will display both skipfile contents for both run 1 and run 2.

To avoid this issue, a new AnalysisInfo entry must be created for every stored run, even if the analyzer command itself is identical.

barnabasdomozi:
@bruntib Updated alembic schema file.

web/server/codechecker_server/api/mass_store_run.py
bruntib: I'm worried about growing the `AnalysisInfo` table. We have current issues with the size of `AnalysisInfoChecker` table. We should consider keeping the deduplication of `AnalysisInfo` and connecting the skipfiles to `RunHistory` table.

barnabasdomozi:
Before this change, the `analysis_info_checkers` table needs to be optimized

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -830,10 +830,11 @@ def __add_blame_info(
 
     def __add_file_content(
         self,
-        session: DBSession,
+        session: SA_Session,
         source_file_name: str,
-        content_hash: Optional[str]
-    ):
+        content_hash: Optional[str],
+        commit: bool
```
bruntib: We could commit at the places where it's necessary after calling this function. This way an extra parameter wouldn't be needed.

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -830,10 +830,11 @@ def __add_blame_info(
...
+    ) -> str:
         """
         Add the necessary file contents. If content_hash in None then this
         function calculates the content hash. Or if it's available at the
```
bruntib:
```suggestion
        function calculates and returns the content hash. Or if it's available at the
```

bruntib (changes requested):

bruntib (approved):

barnabasdomozi:
@dkrupp Can you please review this PR again?

PR #4800, opened 2026-03-10
Integrate mypy into CI

barnabasdomozi:
Added `pyproject.toml`, which defines the mypy configuration. As a first step, only the `codechecker_common/` folder is checked by mypy. This can be extended later.

The first issue detected by mypy is that `Report.report_hash` could be `None` in several cases in the codebase.
Also added installation of stub package `types-psutil` for mypy.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.9'
+          python-version: '3.10'
       - name: Install dependencies
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -27,6 +27,23 @@ jobs:
...
+  type-checker:
+    name: Type checker (mypy)
+
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.12'
```
bruntib: We should use the same python version for all test targets.

barnabasdomozi: The latest mypy version no longer supports Python 3.9, which has reached end of life:
https://devguide.python.org/versions/

PR #4801, opened 2026-03-11
Upgrade CodeChecker to use Vue 3

gulyasgergely902:
This PR upgrades the currently in production Vue 2 CodeChecker web interface to the newer and latest Vue 3. This brings other version upgrades as well to the used modules.

Most of the components are updated to use the newer Composition API which Vue 3 introduced instead of the old Options API. The components are also updated to be in line with the new Vuetify 3 version changes which is the component framework we use to add buttons, input fields and other elements to the pages.


bruntib (approved):

bruntib:
Well done, thank you for this huge work! :)

PR #4806, opened 2026-03-20
Add error on thrift version mismatch, add rebuild script.

### LINKED ISSUE #4807: Create warning for mismatching version of Thrift API.
Currently, after a Thrift API version bump, developers can forget to rebuild the virtual environment, which can lead to incompatible API calls at runtime. Additionally, the make process does not detect or warn about this mismatch.
A version consistency check should be added to the build process (before packaging) that compares the source API version with the virtual environment one and halts with a clear error if they differ. A rebuild script should also be provided so developers can quickly resolve the mismatch.

feyruzb:
This pull request introduces an automated check to ensure that the CodeChecker API version in the source matches the version installed in the project's virtual environment. If a mismatch is detected, the build process will now fail with a clear error message and instructions for resolving the issue. Additionally, a new script has been added to simplify rebuilding the Thrift API and virtual environment when needed.

**API version consistency enforcement:**

* Added variables in the `Makefile` to extract the API version from the source (`setup.py`) and from the installed package in either `venv` or `venv_dev` (`codechecker-api`).
* Introduced a new `check_api_version` target in the `Makefile` that compares the source and virtual environment API versions, halting the build with an error and instructions if they do not match.
* Updated the `package` target in the `Makefile` to run the `check_api_version` step before packaging, ensuring version consistency is always verified.

**Developer tooling:**

* Added a new script, `web/api/completly-rebuild-thrift.sh`, which automates rebuilding the Thrift API, cleaning previous builds, recreating the virtual environment, and rebuilding the package. This script provides clear messaging and usage instructions for developers.


Makefile
```diff
@@ -245,3 +248,12 @@ test_functional_in_env:
 
 set_git_commit_template:
 	if [ -d "$(CURRENT_DIR)/.git" ]; then git config --local commit.template .gitmessage; fi
+
+check_api_version:
+	@if [ "$(API_VERSION)" != "$(VENV_API_VERSION)" ]; then \
+		echo "ERROR: API version mismatch! Source: $(API_VERSION), venv: $(VENV_API_VERSION)"; \
+		echo "Please run './web/api/completly-rebuild-thrift.sh <venv|venv_dev>' to rebuild the Thrift API."; \
```
bruntib: This script could go somewhere under `/scripts`.

Makefile
```diff
@@ -245,3 +248,12 @@ test_functional_in_env:
 
 set_git_commit_template:
 	if [ -d "$(CURRENT_DIR)/.git" ]; then git config --local commit.template .gitmessage; fi
+
+check_api_version:
+	@if [ "$(API_VERSION)" != "$(VENV_API_VERSION)" ]; then \
```
bruntib: The API version in npm should also be checked. Not so long ago I was struggling with API version mismatch because I forgot to rewrite the version number in `package.json`.

scripts/thrift/completly-rebuild-thrift.sh
```diff
@@ -0,0 +1,58 @@
...
+echo "Cleaning previous builds..."
+make clean
+make "clean_${ENV_TYPE}"
+
+# Creating new virtual environment
+echo "Creating new virtual environment ($ENV_TYPE)..."
+make "$ENV_TYPE"
+
+echo "Building package..."
+make package
```
bruntib: I'm not sure if it's the script's job to build the package. I mean, I usually use `make dev_package`.

bruntib (changes requested):

Makefile
feyruzb: Fixed.

Makefile
feyruzb: Fixed, added another source of API version.

scripts/thrift/completly-rebuild-thrift.sh
feyruzb: Fixed.

PR #4809, opened 2026-03-23
[fix] Fix invisible chars in error plist

bruntib:
When either the standard input/output or compilation command contains an invisible character then plist dump fails.


gulyasgergely902 (approved):

barnabasdomozi (approved):
LGTM

PR #4828, opened 2026-04-23
Add PEP 517-compliant pip install workflow

gamesh411:
Rewrite setup.py to discover packages from source tree instead of requiring pre-built build_dist/ directory. Enables `pip install .` and `pip install -e .` as alternative to Makefile build.

Remove the ldlogger C Extension that was in the original setup.py. It used setuptools Extension to build an .so file, but the runtime needs a logger binary as well (which the setuptools Extension did not produce). Build logging on linux still requires `make package` (which builds ldlogger correctly).

Makefile build, Thrift API packages, and frontend build unchanged.


.github/workflows/pip-install.yml
```diff
@@ -0,0 +1,128 @@
...
+        run: |
+          CodeChecker --help
+          CodeChecker version
+          report-converter --help
+          merge-clang-extdef-mappings --help
+          post-process-stats --help
+          tu_collector --help
+
+      - name: Verify data files installed
+        run: |
```
gulyasgergely902: I'd move this part to a separate file to make it more readable and invoke it here. Also given the simplicity of this task, couldn't it be done with bash instead of python?

.github/workflows/pip-install.yml
```diff
@@ -0,0 +1,128 @@
...
+          sudo apt-get update -q
+          sudo apt-get install -y clang clang-tidy cppcheck
+
+      - name: pip install .
+        run: pip install .
+
+      - name: Test analyze and parse
+        run: |
+          WORK="$RUNNER_TEMP/analyze-test"
```
gulyasgergely902: Ideally I'd imagine this in a separate file too because of the length of the script. Also if JQ is available, counting reports could be done with it: `CodeChecker parse "$WORK/reports" -e json | jq ".reports | length"`.

gulyasgergely902 (approved):
LGTM apart from a few minor things.

PR #4830, opened 2026-04-24
Fix PostgreSQL product creation with special characters in DB name

### LINKED ISSUE #4805: PostgreSQL database name with special characters
Superusers can add new products via the web GUI. This product addition also creates the database itself. However, an error is thrown when a PostgreSQL database name is provided that is starting with a digit or contains a dash character. These are not invalid characters in PostgreSQL, but the string has to be quoted in the SQL query:

```.sql
CREATE DATABASE "42my-database";
```

If the database is created through SQLAlchemy, then quotation doesn't happen for some reason.

bishara74:
Closes #4805

Fixes `CREATE DATABASE` failing when the product database name contains a dash or starts with a digit (e.g. `test-product`, `1team`, `42my-database`). The old code interpolated the name into an f-string, and `text()` doesn't quote identifiers — so names that aren't legal unquoted PostgreSQL identifiers produced a syntax error.

**Two commits:**

1. **`routing: add PostgreSQL database name validator`** — new `is_valid_postgresql_db_name()` helper, permissively accepting anything safe to use as a quoted identifier, with 6 unit tests.
2. **`product: fix CREATE DATABASE failure on special characters`** — quote the identifier via SQLAlchemy's `IdentifierPreparer` in `_create_database`, and wire the new validator into `addProduct` so bad input fails early with a clear message.

**Verified manually via the GUI:** `test-product`, `1team`, `42my-database`, and normal names all create successfully; `bad"name` is rejected client-side with a readable error. All 31 existing unit tests still pass.

### DIFF
--- a/web/server/codechecker_server/routing.py
+++ b/web/server/codechecker_server/routing.py
@@ -63,6 +63,37 @@ def is_valid_product_endpoint(uripart):
     return True
 
 
+def is_valid_postgresql_db_name(db_name):
+    """
+    Returns whether or not the given string is a safe PostgreSQL database
+    name for CodeChecker to use.
+
+    CodeChecker quotes the database identifier when issuing CREATE DATABASE,
+    so dashes, leading digits, and PostgreSQL reserved keywords are all
+    allowed (e.g. "test-product", "1team", "user" are accepted). However,
+    characters that would break even a quoted identifier, or that are
+    plainly dangerous in an SQL context, are rejected here so we fail fast
+    with a clear error rather than producing broken SQL or an unusable
+    product.
+    """
+    if not db_name or not isinstance(db_name, str):
+        return False
+
+    # PostgreSQL identifiers (even quoted) cannot exceed 63 bytes by
+    # default. Names longer than this are silently truncated by the
+    # server, which would produce a product that cannot be reconnected
+    # to under the name the user provided. Reject them outright.
+    if len(db_name.encode('utf-8')) > 63:
+        return False
+
+    # Forbidden characters: anything that would prematurely terminate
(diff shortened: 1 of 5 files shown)


web/server/codechecker_server/routing.py
```diff
@@ -63,6 +63,37 @@ def is_valid_product_endpoint(uripart):
     return True
 
 
+def is_valid_postgresql_db_name(db_name):
```
bruntib: Please, replace this function to `codechecker_common.util` module. This `routing.py` is for other purposes.

bruntib (changes requested):
Thanks for the fix, it's a really nice implementation.
I have only a minor comment before it can be merged.

bishara74:
Done, moved is_valid_postgresql_db_name to codechecker_common.util and updated the imports in product_server.py and test_request_routing.py.
I kept the unit tests in test_request_routing.py for now since codechecker_common doesn't have a tests directory.

web/server/tests/unit/test_request_routing.py
```diff
@@ -61,3 +62,64 @@ def test_post(self):
 
         self.assertEqual(post('/DummyProduct/v6.0/FoobarService'),
                          ('DummyProduct', '6.0', 'FoobarService'))
+
+
+class PostgresqlDbNameValidationTest(unittest.TestCase):
```
bruntib: Sorry, I've just noticed that this test is also among the tests of the routing module. Could you place this to a new test file? Thank you!

bishara74:
Moved the test class to a new `web/server/tests/unit/test_db_name_validation.py`, and `test_request_routing.py` is back to just routing tests.

bruntib (approved):

PR #4836, opened 2026-05-04
Add gcc warnings to checkers

FDniel97:


barnabasdomozi (commented):
I discussed the plans about this patch with @bruntib.
The goal would be the following:
- gcc warnings e.g. `-Wfoo` should have checker name `gcc-foo`
- gcc analyzer checkers e.g. `-Wanalyzer-foo` should have checker name `gcc-analyzer-foo`.

Essentially, we should just replace `-W` with `gcc-` to make the checker name more clear to end users.

barnabasdomozi:
Additionally, this PR is still work in progress.

bruntib:
Has been merged in https://github.com/Ericsson/codechecker/pull/5056. Thank you @FDniel97 for the contribution!

PR #4838, opened 2026-05-05
Enable passing credentials from env var

dkrupp:
The user can pass authentication credentials
in the CC_PASSWORD environment variable
to the CodeChecker cmd login command.

### DIFF
--- a/docs/web/authentication.md
+++ b/docs/web/authentication.md
@@ -664,6 +664,9 @@ Is it possible to generate a token from command line which can be used to
 authenticate in the name of the given user. This way no need to store passwords
 in text files. For more information [see](#personal-access-token).
 
+The location of the password file can be configured by the `CC_PASS_FILE`
+environment variable.
+
 The location of the session file can be configured by the
 `CC_SESSION_FILE` environment variable. This can be useful if CodeChecker does
 not have the permission to create a session file under the user's home
--- a/docs/web/authentication.md
+++ b/docs/web/authentication.md
@@ -625,11 +625,12 @@ time a server is connected to, users can pre-configure their credentials to be
 used in authentication.
 
 You can pass the credential used as a password or personal access token
+in the `.codechecker.password.json` or
 directly in the `CC_PASSWORD` environment variable.
 If this environment variable is set, the credentials
 described in `.codechecker.password.json` will be ignored.
 
-If you wish to provide (multiple different) credentials is a password configuration file,
+If you wish to provide (multiple different) credentials in a password configuration file,
 first copy the `config/session_client.json` file from the CodeChecker
 package to your home directory and rename it to `.codechecker.passwords.json`
 After creating the new file open `~/.codechecker.passwords.json`.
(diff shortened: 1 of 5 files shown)


docs/web/authentication.md
```diff
@@ -622,7 +624,13 @@ To alleviate the need for supplying authentication in the command-line every
 time a server is connected to, users can pre-configure their credentials to be
 used in authentication.
 
-To do so first copy the `config/session_client.json` file from the CodeChecker
+You can pass the credential used as a password or personal access token
```
gulyasgergely902, line 627: I'd move this part after defining what .codechecker.password.json file is with the first sentence something like this: "You can also pass the credential...". This way the user knows already what the given file is which this way CodeChecker will ignore and does not have to read further to get the important knowledge.

docs/web/authentication.md
```diff
@@ -655,10 +663,6 @@ Is it possible to generate a token from command line which can be used to
 authenticate in the name of the given user. This way no need to store passwords
 in text files. For more information [see](#personal-access-token).
 
-The location of the password file can be configured by the `CC_PASS_FILE`
-environment variable. This environment variable can also be used to setup
-different credential files to login to the same server with a different user.
-
 Furthermore, the location of the session file can be configured by the
```
gulyasgergely902: It would sound slightly better if the sentence started with "The location of the session file...".

docs/web/authentication.md
```diff
@@ -622,7 +624,13 @@ To alleviate the need for supplying authentication in the command-line every
 time a server is connected to, users can pre-configure their credentials to be
 used in authentication.
 
-To do so first copy the `config/session_client.json` file from the CodeChecker
+You can pass the credential used as a password or personal access token
+directly in the `CC_PASSWORD` environment variable.
+If this environment variable is set, the credentials
+described in `.codechecker.password.json` will be ignored.
+
+If you wish to provide (multiple different) credentials is a password configuration file,
```
gulyasgergely902:
```suggestion
If you wish to provide (multiple different) credentials in a password configuration file,
```

gulyasgergely902 (changes requested):
Found some parts in the documentation which could be tweaked but aside from that LGTM.

docs/web/authentication.md
```diff
@@ -655,11 +664,7 @@ Is it possible to generate a token from command line which can be used to
 authenticate in the name of the given user. This way no need to store passwords
 in text files. For more information [see](#personal-access-token).
 
-The location of the password file can be configured by the `CC_PASS_FILE`
```
bruntib, line 658: Why has this paragraph been removed? `CC_PASS_FILE` still exists. Would we like to deprecate it?

dkrupp: It is moved a bit ahead to line 627. It describes the same thing as the removed part. I

bruntib (approved):

gulyasgergely902 (approved):
LGTM

PR #4842, opened 2026-05-06
Git blame

bruntib:


web/server/codechecker_server/api/report_server.py
```diff
@@ -2388,6 +2416,22 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
                         report.review_status_date,
                         report.review_status_is_in_source)
 
+                    blame_info = blame_infos[report.id] \
```
barnabasdomozi: This piece of code is redundant with lines between 2295-2309. Please create a helper function.

web/server/codechecker_server/api/report_server.py
```diff
@@ -2371,9 +2393,15 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
 
                 # Get report details if it is required.
                 report_details = {}
-                if get_details:
-                    report_ids = [r[0].id for r in query_result]
+                blame_infos = {}
```
barnabasdomozi, line 2387: This piece of code is redundant with lines between 2270-2276. Please create a helper function.

web/server/codechecker_server/api/report_server.py
```diff
@@ -2267,9 +2267,15 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
 
                 # Get report details if it is required.
                 report_details = {}
-                if get_details:
-                    report_ids = [r.id for r in query_result]
+                blame_infos = {}
+                if get_details and len(query_result):
+                    report_ids, blames = zip(*[
```
barnabasdomozi: Instead of `zip()` consider using a regular for loop and `.append()` to `report_ids` and to `blames` which is often faster and more clear to read.

web/server/codechecker_server/api/report_server.py
```diff
@@ -2267,9 +2267,15 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
 
                 # Get report details if it is required.
                 report_details = {}
-                if get_details:
-                    report_ids = [r.id for r in query_result]
+                blame_infos = {}
+                if get_details and len(query_result):
```
barnabasdomozi: To check if a list is non empty, simply use
`if query_result`, not needed to write `if len(query_result)`

web/server/codechecker_server/api/report_server.py
```diff
@@ -2286,6 +2292,21 @@ def getRunResults(self, run_ids, limit, offset, sort_types,
                         row.review_status_date,
                         row.review_status_is_in_source)
 
+                    blame_info = blame_infos.get(row.id)
+                    if blame_info and blame_info.commits and blame_info.blame:
+                        blame_data = [b for b in blame_info.blame
+                                      if row.line >= b.startLine
+                                      and row.line <= b.endLine]
+                        commitHash = blame_data[0].commitHash \
+                            if len(blame_data) else None
```
barnabasdomozi: To check if a list is non empty, simply use
`if blame_data`, not needed to write `if len(blame_data)`

web/tests/functional/report_viewer_api/test_get_run_results.py
```diff
@@ -472,3 +472,20 @@ def test_get_guideline_rules(self):
 
         self.assertEqual(con_54_cpp_rule.checkers,
                          ["bugprone-spuriously-wake-up-functions"])
+
+    def test_report_blame_info(self):
```
barnabasdomozi, line 481: Add a test case when blame info should not be present.

barnabasdomozi (changes requested):

web/server/codechecker_server/api/report_server.py
barnabasdomozi: This comment is still not resolved. A helper function would be better in this case too.

barnabasdomozi (changes requested):

dkrupp (changes requested):
1) The blame info only appears in the json output of
`CodeChecker cmd results --details tinxyml_sensitive2 -o json`

Why only when --details is given. I think we should give it in the json even if the --details not provided.

In fact there should be not difference in the output in json with or without --details flag

2) We could give a short blame info in the  the text output too only if `--details` is given.
commit date, commit number,  first few words of hte commit message.

3) How would a user know about this feature?

Please document the feature in web user guide and in the help, probably under the ---details  flag of `CodeChecker cmd results` we could describe what extra info it provides compared to when we dont have it.

dkrupp (commented):
One more minot thing.

Could you please add a slightly better description for the CodeChecker cmd results command help? 

Instead of this:
"Show the individual analysis reports' summary."

"Show the analysis reports which match the optional filters. 

In json  format all available information is shown. In plaintext/table only the report position, checker name, severity, analyzer name, report mesage, detection and review status and git blame information displayed."

dkrupp (approved):
LGTM

web/client/codechecker_client/cmd_line_client.py
```diff
@@ -820,8 +821,15 @@ def handle_list_results(args):
             report.runName = run_id2name[report.runId]
         print(CmdLineOutputEncoder().encode(all_results))
     else:
-        header = ['File', 'Checker', 'Severity', 'Message', 'Bug path length',
-                  'Analyzer name', 'Review status', 'Detection status']
+        if query_report_details:
```
barnabasdomozi: To simplify the code, this `if` can be merged with the other `if` on line 869.

web/client/codechecker_client/cmd_line_client.py
```diff
@@ -848,9 +856,26 @@ def handle_list_results(args):
             if len(msg) > max_msg_len:
                 msg = msg[:max_msg_len] + '...'
 
-            rows.append((checked_file, res.checkerId, sev, msg,
-                         res.bugPathLength, res.analyzerName, rw_status,
-                         dt_status))
+            if res.blameInfo and res.blameInfo.commits:
```
barnabasdomozi: Move this `if` to the other `if` on line 869.

barnabasdomozi (changes requested):

barnabasdomozi (approved):

PR #4843, opened 2026-05-07
fix in debug_tools "failure zip" handling scripts

balazske:
Include paths in commands were not detected correctly in all cases. This fixes problems with non-absolute paths and paths after '-I' option (which can start with any character).


dkrupp (approved):
LGTM

PR #4846, opened 2026-05-11
[test] Make test environment more robust

bruntib:
The test environment is quite sensitive on analyzer version changes. The test files assume a certain number of reports for a given project.

This patch makes the test environment more robust by enabling only given checkers and check their reports only.


web/tests/functional/source_change/test_source_change.py
```diff
@@ -176,9 +175,11 @@ def test_store(self):
 
         test_proj = os.path.join(self.test_workspace, 'test_proj')
 
+        print("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
```
barnabasdomozi: Avoid unnecessary prints in test cases.

web/tests/functional/source_change/test_source_change.py
```diff
@@ -176,9 +175,11 @@ def test_store(self):
...
         ret = codechecker.check_and_store(self._codechecker_cfg,
                                           'test_proj',
                                           test_proj)
+        print("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb", ret)
```
barnabasdomozi: Avoid unnecessary prints in test cases.

web/tests/functional/cmdline/test_filter_preset_cmd.py
```diff
@@ -441,7 +441,8 @@ def test_filter_preset_cmd_apply_preset(self):
                 '--url', str(self.server_url)]
         _, out3, _ = run_cmd(cmd3)
 
-        self.assertEqual(out3.count("HIGH"), 53)
+        severity_counts = self.__project_info["severity_counts"]
```
barnabasdomozi: Minor: The variable `severity_counts` is only used in one place and therefore can be omitted.
This line below 
```self.assertEqual(out3.count("HIGH"), severity_counts["HIGH"])``` 
could be simplified to  
```self.assertEqual(out3.count("HIGH"), self.__project_info["severity_counts"]["HIGH"])```

File: web/tests/functional/source_change/test_source_change.py [hunk 4]
```diff
@@ -73,8 +73,7 @@ def setup_class(self):
             'check_env': test_env,
             'workspace': TEST_WORKSPACE,
             'reportdir': os.path.join(TEST_WORKSPACE, 'reports'),
-            'checkers': [],
-            'analyzers': ['clangsa']
```
barnabasdomozi, line 77: Why was `clangsa` removed from here? Does it mean we run all analyzers?

web/tests/libtest/project.py
```diff
@@ -70,6 +70,10 @@ def get_clean_cmd(test_project):
         return ""
 
 
+def get_enabled_checkers(test_project):
```
barnabasdomozi: Please add type annotation for params and for return value.

barnabasdomozi (changes requested):

web/tests/functional/source_change/test_source_change.py
bruntib: Yes, it runs all analyzers. According to the new concept we require reports from the listed checkers in `project_info.json`. There are clang-tidy checkers among these, too.

barnabasdomozi (approved):
LGTM

PR #4848, opened 2026-05-14
Bugfix: no run names in ReportFilter.

### LINKED ISSUE #4816: No run-names in ReportFilter
Run names are intentionally not saved in ReportFilter, needs to be fixed for complete functionality of saving filter presets.

feyruzb:
This pull request makes improvements to how filter presets are saved and managed in the report filtering UI, particularly ensuring that selected run names are correctly serialized with presets and refining some filter lifecycle methods.

Improvements to preset saving and serialization:

* When saving a filter preset, the code now clones the current `reportFilter` and explicitly adds the selected run names (`runName`) to it, ensuring that run selections are preserved when serializing and restoring presets.
* The `selectedItems` property is now exposed from the `BaselineRunFilter` component, making it accessible to parent components for preset management.

Refinements to filter lifecycle and UI updates:

* The call to `updateUrl()` was removed from the `getFilterPreset` function, to prevent unnecessary URL updates when applying a preset.
* The call to `updateAllFilters()` was removed from `clearToolbarSilently`, to streamline the clearing process and avoid redundant updates.


gulyasgergely902:
Please check if this fix is needed with the run filter issue fix from PR #4850.

feyruzb:
Yes, this PR fixes the problem of run names not being saved in `FilterPreset object`, while the other PR is about losing the run id's during the initialization.

bruntib (approved):

PR #4850, opened 2026-05-15
[web] Fix timing issue in run filter

gulyasgergely902:
Run filter was getting the run name and tag from the global store which made it lose sync and not getting a valid result before setting the URL. This change returns the run id and tag to the update URL function without the need for global store to return the data in time.


web/server/vue-cli/src/components/Report/ReportFilter/Filters/BaselineRunFilter.vue
feyruzb, line 360: **Issue**: The run name persists in the URL even after the `CLEAR ALL FILTERS` button is pressed.

**Suggested fix**: Add these two lines to also clear the selected run names:

```
selectedRunIds.value = []; 
selectedTagIds.value = [];
```

bruntib (approved):

PR #4858, opened 2026-05-20
[gui] Fix oauth login button color

gulyasgergely902:
OAuth button color is not visible, this change makes is bolder increasing visibility.


barnabasdomozi (approved):
LGTM

PR #4859, opened 2026-05-20
[gui] Fix new features timeline alignment

gulyasgergely902:
New features timeline is misaligned in the new Vue 3 rework. This change fixes it by forcing all timeline item to be the same size disregarding its content.


barnabasdomozi (approved):
LGTM

PR #4861, opened 2026-05-23
Renaming functionality for Filter-presets

### LINKED ISSUE #4870: Filter Preset renaming button.
There is no way to rename a filter preset after creation.
There should be to rename  a filter after creation in the gui.

feyruzb:
This pull request adds the ability to rename filter presets throughout the CodeChecker system. It introduces a new backend API for renaming presets, updates the command-line interface and web UI to support this feature, and provides comprehensive tests for the new functionality. Additionally, the API and client versions are bumped to 6.71.0.

**Rename filter preset feature:**

* Added a new `renameFilterPreset` RPC method to the `codeCheckerDBAccess` Thrift service, allowing filter presets to be renamed by ID. The method validates the new name, checks for duplicate names, and ensures the preset exists.
* Implemented the `renameFilterPreset` method in the Python client and added a corresponding CLI command (`filter-preset rename`) with arguments for preset ID and new name. 
* Updated the Vue web UI to allow users to rename filter presets via a dialog, including UI changes and the necessary API call. 

**Testing improvements:**

* Added extensive functional tests for the new rename command, including tests for success, duplicate names, empty names, nonexistent IDs, and verifying the rename is reflected in the preset list.

**Version bumps:**

* Bumped the API and client versions to 6.71.0 to reflect the new feature.


web/server/codechecker_server/api/report_server.py
```diff
@@ -1717,6 +1717,61 @@ def storeFilterPreset(self, filterpreset):
...
+        Raises an error if id/name is empty or if
+        a preset with the new name already exists.
+        """
+        self.__require_admin()
+        try:
+            with DBSession(self._Session) as session:
+                if not name or not name.strip():
+                    raise codechecker_api_shared.ttypes.RequestFailed(
+                        codechecker_api_shared.ttypes.ErrorCode.DATABASE,
+                        "Preset name cannot be empty!")
```
gulyasgergely902, line 1738: Either use "xyz cannot be empty" or "Invalid xyz" but choose one and stick with it. I'd use "Preset name cannot be empty" and "Preset ID cannot be empty" since this error is thrown because a required argument was left out and so it cannot be invalid if it is empty.

web/server/codechecker_server/api/report_server.py
```diff
@@ -1717,6 +1717,61 @@ def storeFilterPreset(self, filterpreset):
...
+
+                if not preset_id:
+                    raise codechecker_api_shared.ttypes.RequestFailed(
+                        codechecker_api_shared.ttypes.ErrorCode.DATABASE,
+                        "Invalid preset ID!")
```
gulyasgergely902: Please, add an empty line between validation and the actual query of the filter preset for better readability.

web/server/codechecker_server/api/report_server.py
```diff
@@ -1717,6 +1717,61 @@ def storeFilterPreset(self, filterpreset):
                 codechecker_api_shared.ttypes.ErrorCode.DATABASE,
                 "CodeChecker could not store the filter preset: " + str(ex))
 
+    @exc_to_thrift_reqfail
+    @timeit
+    def renameFilterPreset(self, preset_id: int, name: str):
+        """
+        Rename a filter preset.
+        Returns the ID of the renamed preset.
```
gulyasgergely902, line 1725: Since you are supplying the preset_id which you want to rename, it is not much information to return it at the end too. I'd return a success instead like a 0 and raise (as currently implemented) if there was any error during. What was the motivation behind returning the preset_id?

web/server/vue-cli/src/components/Report/ReportFilter/ReportFilter.vue
```diff
@@ -31,15 +31,17 @@
         <v-btn
           color="primary"
           :disabled="!presetName"
-          @click="saveCurrentFilter(saveMode)"
+          @click="saveMode !== 'rename'
```
gulyasgergely902: I'd move this logic to the function "saveCurrentFilter()" instead. Also it could be renamed to savePreset and it could handle saving and renaming too.

gulyasgergely902 (changes requested):
Only left some minor questions / issues.

web/server/codechecker_server/api/report_server.py
feyruzb: Fixed, used `Preset ID cannot be empty!`

web/server/codechecker_server/api/report_server.py
feyruzb: Added.

web/server/codechecker_server/api/report_server.py
feyruzb: It returns the ID in case the operation was successful which makes it easier for testing. Additionally for applying the modified preset the ID which saves us 1 request of a list of all presets every time 1 is changed.

web/server/vue-cli/src/components/Report/ReportFilter/ReportFilter.vue
feyruzb: Nice idea! Applied.

web/server/vue-cli/src/components/Report/ReportFilter/ReportFilter.vue
```diff
@@ -688,42 +703,44 @@ onBeforeUnmount(() => {
...
-      ? activePresetId
-      : -1,
-    name: presetName.value,
-    reportFilter: presetReportFilter
-  };
+async function savePreset(mode) {
+  try {
+    let result;
+    if (mode === "rename") {
+      const preset_id = presetMenuRef.value?.activePresetId;
```
gulyasgergely902: Since this variable setting is the same for both branches of the if statement, it could be set before the statement. Also use camel case for all the variables, e.g. instead of new_name, use newName.

web/server/vue-cli/src/components/Report/ReportFilter/ReportFilter.vue
```diff
@@ -688,42 +703,44 @@ onBeforeUnmount(() => {
...
-    ).catch(
-      err => {
-        handleThriftError("FAILURE", err);
-      }
-    );
+    savePresetDialogOpen.value = false;
+    presetName.value = "";
+    presetMenuRef.value?.selectPresetAfterSave(result);
+  } catch (err) {
+    handleThriftError("FAILURE", err);
```
gulyasgergely902: Don't just use "FAILURE" for an error description, use something which tells where the issue happened or what happened, e.g. since this error happens when the save of the presets is failing, use something like this: "Failed to save preset: " and the error string comes after that. This makes debugging easier later.

web/server/vue-cli/src/components/Report/ReportFilter/ReportFilter.vue
gulyasgergely902: It was not part of this ticket, but could you please rename "overridePreset" to "overridePresetDialog, "cratePresetAsNew" to "createPresetAsNewDialog" and "createPreset" to "createPresetDialog" just like your "renamePresetDialog" to make it clear that is opens a dialog not overrides/creates a preset directly.

gulyasgergely902 (changes requested):

web/server/vue-cli/src/components/Report/ReportFilter/ReportFilter.vue
feyruzb: Fixed.

web/server/vue-cli/src/components/Report/ReportFilter/ReportFilter.vue
feyruzb: Fixed.

web/server/vue-cli/src/components/Report/ReportFilter/ReportFilter.vue
feyruzb: Fixed.

gulyasgergely902 (approved):

PR #4864, opened 2026-05-28
[doc] Fix sei-cert guideline references to sei-cert-c/sei-cert-cpp

gamesh411:
The bare 'sei-cert' guideline label does not exist. The actual labels are 'sei-cert-c' and 'sei-cert-cpp'.


dkrupp (approved):
LGTM

PR #4876, opened 2026-06-04
[fix] Fix analyzer env adding empty space

gulyasgergely902:
When LD_PRELOAD ENV variable is set but the value
is empty, CodeChecker log produces an LD_PRELOAD
value which contain an empty space (" ") too. This behaviour breaks the BBI codechecker wrapper.

This fix removes this empty space from the LD_PRELOAD value if LD_PRELOAD ENV variable is set but has
no value.


dkrupp (approved):
LGTM

PR #4878, opened 2026-06-05
[ci] Add fixup! commit check for pull requests

gulyasgergely902:
The fixup! commit check workflow will check any commits on a given pull request and fail if there is any. This way it is enforced to squash all fixup! commits which were made with the argument '--fixup=...' when committing.

### DIFF
--- a/.github/workflows/check_fixups.yml
+++ b/.github/workflows/check_fixups.yml
@@ -1,0 +1,30 @@
+name: Git Fixup Commit Check
+
+on:
+  pull_request:
+    branches: [ master ]
+    types: [ opened, synchronize, reopened ]
+
+jobs:
+  check_fixups:
+    name: Check for fixup commits
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Verify commit messages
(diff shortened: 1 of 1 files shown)


gamesh411:
Thanks for the attention to detail and for being conscious of code quality in this regard as well, @gulyasgergely902.
I would be interested in your experience that led you to propose this change. Did you find accidentally left-in fixup commits? Doesn't the "squash and merge" GitHub workflow eliminate those (for example, if there is a meaningful change with a meaningful commit message, and then there are some other commits on top of that)? Or is this targeting PRs where the only commit is a fixup commit?
I'm not against or for this, just curious :)

gulyasgergely902:
So basically, this workflow is for the specific case which the new git workflow is describing: After the developer opened the PR for the changes made and the reviewer found any bugs/mistakes/etc., the developer should fix those in a fixup! commit made with --fixup=<commit hash>. This makes checking what changes were pushed in a response to the requests easier. After everything was fixed and the reviewer is happy with the results, asks the developer to squash the changes with --autosquash in a rebase command. This makes every fixup squashed to the right commit where it belongs. This automation guards the git tree of an accidental merge with fixup commits still present.
You can read the new workflow in the CONTRIBUTING file.

bruntib (approved):

PR #4880, opened 2026-06-07
File-path and a tree-view migrated to vue3

xb058t:


web/server/vue-cli/src/views/Reports.vue
```diff
@@ -170,6 +193,81 @@
...
+          :items="treeItems"
+          item-value="fullPath"
+          open-on-click
+          density="compact"
+        >
+          <template #prepend="{ item, isOpen }">
+            <v-icon v-if="item.children.length > 0">
+              {{ isOpen ? 'mdi-folder-open' : 'mdi-folder' }}
+            </v-icon>
+            <v-icon v-else>mdi-file</v-icon>
```
gulyasgergely902: For every v-icon: put the icon name and also the end tag to a new line each like this:
<v-icon ...>
mdi-file
</v-icon>

web/server/vue-cli/src/components/Report/ReportFilter/Filters/FilePathFilter.vue
```diff
@@ -18,16 +17,78 @@
...
+            <v-icon v-if="item.children.length > 0" size="small">
+              {{ isOpen ? 'mdi-folder-open' : 'mdi-folder' }}
+            </v-icon>
+            <v-icon v-else size="small">mdi-file</v-icon>
+          </template>
+          <template #title="{ item }">
+            <span class="tree-item-label" @click.stop="toggleTreeItem(item)">
+              {{ item.name }}
+            </span>
+            <v-chip class="ml-2" size="x-small">{{ item.findings }}</v-chip>
```
gulyasgergely902: Put the content and the end tag to new lines:
<v-chip ...>
{{}}
</v-chip>

web/server/vue-cli/src/views/Reports.vue
```diff
@@ -14,7 +14,30 @@
         v-model="checkerDocDialog"
         :checker="selectedChecker"
       />
+      <v-btn-toggle
```
gulyasgergely902: Make this btn-toggle's variant outlined so the non-selected button can be seen better. Also I'd add a "View" label next to it to the left so the users will know what this button does.

web/server/vue-cli/src/views/Reports.vue
```diff
@@ -170,6 +193,81 @@
...
+            <review-status-icon :status="1" :size="14" />
+          </span>
+          <span class="tree-header-cell">
+            <review-status-icon :status="2" :size="14" />
+          </span>
+          <span class="tree-header-cell">
+            <review-status-icon :status="3" :size="14" />
+          </span>
+        </div>
+        <v-treeview
```
gulyasgergely902, line 232: What do you think, is it possible to remove the arrows from the deepest levels? There should be no arrow on a level which cannot be further expanded like for a single file at the end.

gulyasgergely902 (changes requested):
Great work overall with just minor easily fixable issues.

One thing came to my mind when testing: if the user clicks on a file on the tree view to see the results it navigates to the reports view but when going back to the tree, all the branches are collapsed. It might not be a big issue but for a project which has 15-20 deep folder structures it can be a hassle to go back to the previously opened place. It's - again - not a big issue but could be investigated on how this solution could bring back the user to the same place where it was before checking out any given file.

web/server/vue-cli/src/components/Report/ReportFilter/Filters/FilePathFilter.vue
```diff
@@ -18,16 +17,80 @@
...
+          item-value="fullPath"
+          open-on-click
+          density="compact"
+          class="file-path-tree"
+        >
+          <template #prepend="{ item, isOpen }">
+            <v-icon v-if="item.children.length > 0" size="small">
+              {{ isOpen ? 'mdi-folder-open' : 'mdi-folder' }}
+            </v-icon>
+            <v-icon v-else size="small">mdi-file</v-icon>
```
gulyasgergely902: These lines still don't have the new lines and the linter warns for it too. Please check and add new lines for every v-icon used.

gulyasgergely902 (changes requested):

web/server/vue-cli/src/components/Report/ReportFilter/Filters/FilePathFilter.vue
xb058t: Found 2 more places. All fixed now.

gulyasgergely902 (approved):
LGTM

PR #4883, opened 2026-06-08
Enhance macOS compatibility with spawn-based multiprocessing

gamesh411:
Enables the full analyzer and web test suites to run on macOS by switching from fork to spawn as the multiprocessing start method on non-Linux platforms, and making the test suite cross-platform.

CodeChecker currently only runs CI on Linux. macOS uses fork by default but it is unsafe (Obj-C runtime crashes in child processes, see #4893 ), and the test suite has several Linux-specific assumptions (/tmp path, LD_PRELOAD, ld_logger, compile_commands.json command format, fixed time.sleep for task state transitions). This PR addresses all of these so the existing test suite passes on macOS with no behavioral changes on Linux.

### Implementation

1. Remove compatibility.multiprocessing shim, use multiprocess directly.
   The custom compatibility wrapper is replaced with direct imports from the multiprocess package.
   The spawn start method is set explicitly in `cli.py:main()` on macOS/Windows.
   Spawn workers get explicit logger setup and shared state propagation (compiler_info dict) since they don't inherit parent state via fork.

2. Spawn-based worker processes in the server.
   On macOS/Windows, API and task worker processes cannot inherit the `HTTPServer` instance or shared `Queue`/`Value` objects.
   The server is reconstructed in spawned API workers from serializable config (`_build_worker_server`), and `SyncManager.Queue()` is used for the task queue so it can be pickled to spawned task workers.

3. arguments list format in compile_commands.json.
   The JSON compilation database spec allows entries to use either command (string) or arguments (list).
   The intercept-build wrapper used on macOS (via bear) produces the arguments form. tu_collector now normalizes both formats wherever the build action command is accessed.

4. CI worker count override.
   `CC_TEST_API_WORKERS` and `CC_TEST_TASK_WORKERS` env vars map to `--api-handler-processes` and `--task-worker-processes` server flags.
   On macOS CI, spawn workers are expensive (each re-imports the full Python stack), so constraining to 1 worker reduces resource pressure.
   Also improves wait_for_server_start with an HTTP port-based fallback and fail-fast diagnostics.

### Test infra

- `os.path.realpath('/tmp')` to handle macOS `/tmp` being a symlink to `/private/tmp`
- Skip `LD_PRELOAD/LD_LIBRARY_PATH`/`-m32` tests on macOS (not applicable)
- Skip gcc analyzer test when g++ is an Apple clang shim
- `-fsyntax-only` in buildcmd escaping tests (linking may fail with Homebrew clang@14)
- Normalize intercept-build vs ld-logger output strings for cross-platform comparison
- Normalize gcc/infer diagnostic quoting and type annotations across versions
- OAuth mock server: bind to 127.0.0.1, poll for readiness instead of fixed sleep,
  track and terminate in teardown
- Blame tests: restore cwd in finally block to prevent test pollution
- Task management tests: poll for state transitions instead of fixed sleeps
  (spawn workers take ~42s to start), use 5s tasks to reliably observe RUNNING state

 ### GH Actions

- install-deps-macos.sh: installs llvm@14, gcc@13, cppcheck, bear, and Infer.
  Creates an intercept-build wrapper around bear (LLVM's libear.dylib seems broken on macOS ARM64 with SIP),
  gcc/g++ symlinks matching Linux naming,
  and sets SDKROOT/MACOSX_DEPLOYMENT_TARGET/CC/CXX for correct compilation and SARIF output.
- test.yml: macOS matrix entries for analyzer and tools jobs,
   separate web-macos job (sqlite only, reduced worker counts). All macOS jobs use continue-on-error.


bruntib:
@gamesh411 Would you think that this ticket will be solved by this patch? https://github.com/Ericsson/codechecker/issues/3274
Thank you!

gamesh411:
> @gamesh411 Would you think that this ticket will be solved by this patch? #3274 Thank you!

@bruntib I have verified locally (on an OSX machine), and this change fixes #3274. The issue is still present on master, and this PR solves it.

PR #4887, opened 2026-06-11
Change all pre-endpoint product/version failures to ProductNotFoundError

Discookie:
This removes a couple unnecessary stack traces from the server console.

### DIFF
--- a/web/server/codechecker_server/server.py
+++ b/web/server/codechecker_server/server.py
@@ -476,11 +476,9 @@ class RequestHandler(SimpleHTTPRequestHandler):
                     elif request_endpoint == "CodeCheckerService":
                         # This endpoint is a product's report_server.
                         if not product:
-                            error_msg = \
-                                "Requested CodeCheckerService on a " \
-                                f"nonexistent product: '{product_endpoint}'."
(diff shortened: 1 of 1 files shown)


web/server/codechecker_server/server.py
```diff
@@ -479,8 +479,8 @@ def do_POST(self):
                             error_msg = \
                                 "Requested CodeCheckerService on a " \
                                 f"nonexistent product: '{product_endpoint}'."
-                            LOG.error(error_msg)
-                            raise ValueError(error_msg)
+                            LOG.debug(error_msg)
```
bruntib: The exception handler prints this message to the debug log already. I think, this could be removed.

web/server/codechecker_server/server.py
```diff
@@ -501,9 +501,9 @@ def do_POST(self):
                     else:
                         LOG.debug("This API endpoint does not exist.")
```
bruntib: The exception handler prints this message to the debug log already. I think, this could be removed.

bruntib (changes requested):

bruntib (approved):

PR #4897, opened 2026-06-15
Enhancement: Add confirmation dialog for deleting FilterPresets

### LINKED ISSUE #4898: Add confirmation dialog before deleting a FilterPreset
Currently clicking delete on a preset in PresetMenu.vue removes it immediately, with no way to undo. This makes accidental deletions easy.

Request:
Show a confirmation dialog before a preset is deleted. The user should have to explicitly confirm, and the dialog should show which preset is about to be deleted.

Acceptance criteria:
- Clicking delete opens a confirm dialog instead of deleting right away.
- Dialog displays the name of the preset pending deletion.
- Confirming deletes the preset; cancelling leaves it untouched.

feyruzb:
This pull request enhances the user experience for deleting presets in the `PresetMenu.vue` component by introducing a confirmation dialog before deletion. Instead of immediately deleting a preset when the delete button is clicked, users are now prompted to confirm their action, reducing the risk of accidental deletions.

**User experience improvements:**

* Added a `ConfirmDialog` component that appears when a user attempts to delete a preset, requiring explicit confirmation before the action proceeds.
* Introduced new state variables (`showDeleteConfirmDialog`, `pendingDeleteId`, and `pendingDeleteName`) to manage the dialog's visibility and display the name of the preset pending deletion.
* Replaced the direct call to `deletePreset` in the delete button's click handler with a new `promptDelete` method that triggers the confirmation dialog.
* Added `promptDelete` and `confirmDelete` methods to handle the confirmation workflow before actually deleting the preset.

### DIFF
--- a/web/server/vue-cli/src/components/Report/ReportFilter/Filters/PresetMenu.vue
+++ b/web/server/vue-cli/src/components/Report/ReportFilter/Filters/PresetMenu.vue
@@ -27,7 +27,7 @@
             <v-btn
               v-if="canSeeActions"
               variant="text"
-              @click.stop="deletePreset(item.value)"
+              @click.stop="promptDelete(item.value)"
             >
               <v-icon
                 v-if="deletingId === item.value"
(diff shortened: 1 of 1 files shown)


gulyasgergely902 (approved):
LGTM

PR #4900, opened 2026-06-17
Upgrade thrift version in npm package

### LINKED ISSUE #4891: The JavaScript thrift library has not been updated.
https://github.com/Ericsson/codechecker/pull/4581
The Thrift compiler and Python libraries have been updated to version 0.22.0,
but the JavaScript thrift library remains at version 0.13.hotfix.1.
Shouldn't the JavaScript thrift library also be updated to version 0.22.0?

web/api/js/codechecker-api-node/package.json
  "dependencies": {
    "thrift": "0.13.0-hotfix.1"
  }

bruntib:
Fixes #4891

### DIFF
--- a/web/api/js/codechecker-api-node/package.json
+++ b/web/api/js/codechecker-api-node/package.json
@@ -23,6 +23,6 @@
     "api"
   ],
   "dependencies": {
-    "thrift": "0.13.0-hotfix.1"
+    "thrift": "0.22.0"
   }
 }


barnabasdomozi (changes requested):
Reading the conversation in https://github.com/Ericsson/codechecker/issues/4891, if this file is not needed than it should be removed from the repository and not updated.

bruntib:
Sorry, that wasn't a true comment. This `package.json` is used in fact when building the Thrift stub. So, we'll really need this change.

barnabasdomozi (approved):

PR #4903, opened 2026-06-18
Include suppression information in sarif files

andyg319:
`CodeChecker parse` already populates Report objects with review status information, but currently only JSON export makes use of it. This PR adds suppression objects to SARIF reports, according to the [spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317733), including suppression source (source code comment or review config file) and justification message.

`location.physicalLocation` for suppressions may be added in a follow-up PR.


andyg319:
Closing in favor of https://github.com/Ericsson/codechecker/pull/4925.

PR #4905, opened 2026-06-19
[gui] Restore Vuetify 2 spacing for CodeChecker web UI

gulyasgergely902:
Vuetify 3 introduced new spacing values which creates a more spacious feel for the users. This is not ideal for CodeChecker since it allows less data to be present and users have to scroll and resize the page to view any info.

This change aims to restore the old Vuetify 2 spacings in some form to be able to display more information.

### DIFF
--- a/web/server/vue-cli/src/components/CopyBtn.vue
+++ b/web/server/vue-cli/src/components/CopyBtn.vue
@@ -6,8 +6,8 @@
   >
     <template v-slot:activator="{}">
       <v-btn
-        icon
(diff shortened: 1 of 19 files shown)


dkrupp (changes requested):
The report detail view is also too spacious compared to the last release. can you make it more compact too?

The fonts are too big in the left filter tree.
The spaces are too big between the lines in the code view.

new
Image

old

Image

dkrupp (approved):
LGTM

PR #4907, opened 2026-06-19
Refactor/replace report filter button icon

### LINKED ISSUE #4906: Replace reset report filter button icon
Currently resetting any report filter is done by a button with a trashcan icon. Deleting a report filter preset is done using a button with the same icon. This is misleading.

Replace all the trashcan icons on the buttons for the filters which resets them to a cross which is currently used for removing a selected report filter item. Trashcans should only be used for an action which deletes a saved entity.

feyruzb:
This pull request updates the `PresetMenu` component in the report filter UI to improve the user experience and clarify icon usage. The main changes focus on the appearance and behavior of the preset selection and deletion controls.

UI/UX improvements:

* The clear icon for the `v-select` component is now set to `mdi-delete` for better clarity when clearing a preset.
* The menu dropdown uses a scrim and closes on scroll for improved usability, and the order of some props is adjusted for clarity.

Icon and button updates:

* The delete/cancel icon in the preset list is now larger (`size="large"`) and uses `mdi-close-circle` instead of `mdi-delete` for clearer intent.

Closes #4906

### DIFF
--- a/web/server/vue-cli/src/components/Report/ReportFilter/ClearAllFilters.vue
+++ b/web/server/vue-cli/src/components/Report/ReportFilter/ClearAllFilters.vue
@@ -3,7 +3,7 @@
     id="clear-all-filter-btn"
     color="primary"
     variant="outlined"
-    prepend-icon="mdi-delete"
+    prepend-icon="mdi-close-circle"
     @click="clearAllFilters"
   >
     Clear All Filters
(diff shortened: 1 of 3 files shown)


gulyasgergely902 (changes requested):
This is not exactly what the ticket asked for but it might not be clear what it wanted. The goal of the ticket was to replace all buttons which deletes anything to a trashcan icon and use the cross icon to unselect a selection. For example the filter preset selection has it inverted now: the delete action (deleting a preset) has the cross icon and the unselection action has a trashcan. Please replace the icons which only unselects (for the whole filter sidebar) to a cross and add the trashcan to buttons which irreversibly deletes the given entity (currently only applicable for filter presets).

gulyasgergely902 (approved):
LGTM

PR #4908, opened 2026-06-22
[gui] Show the selected source component filter mode in toolbar

gulyasgergely902:
This change makes it easier to see which component filter mode is currently active by showing it on the toolbar.


dkrupp (approved):
LGTM

PR #4909, opened 2026-06-22
Share checker configurations across analysis runs via CheckerSet

barnabasdomozi:
Introduce CheckerSet and CheckerSetItem tables to deduplicate checker configurations across analysis runs. Instead of storing enabled/disabled checkers per AnalysisInfo row, checkers are
now grouped into reusable sets identified by a SHA-256 hash.

This avoids redundant rows when multiple runs share the same checker configuration.

### DIFF
--- a/web/server/codechecker_server/database/run_db_model.py
+++ b/web/server/codechecker_server/database/run_db_model.py
@@ -11,7 +11,9 @@ SQLAlchemy ORM model for the analysis run storage database.
 from datetime import datetime, timedelta
 from math import ceil
 import os
-from typing import Optional
+import json
+import hashlib
+from typing import Optional, List
 
 from sqlalchemy import BigInteger, Boolean, Column, DateTime, Enum, \
     ForeignKey, Integer, LargeBinary, MetaData, String, UniqueConstraint, \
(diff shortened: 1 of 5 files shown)


web/server/codechecker_server/database/run_db_model.py
```diff
@@ -53,29 +55,58 @@ def __init__(self, analyzer_name: str, checker_name: str, severity: int):
...
+    @staticmethod
+    def compute_hash(enabled_checkers: List[int],
+                     disabled_checkers: List[int]) -> str:
+        # Sort lists to create identical hashes.
+        enabled_checkers.sort()
+        disabled_checkers.sort()
+
+        checker_set_dict = {"e": enabled_checkers, "d": disabled_checkers}
+        return hashlib.sha256(
+                json.dumps(checker_set_dict).encode()).hexdigest()
```
bruntib: Python doesn't guarantee the same dumped JSON format over Python versions. If we still don't want to invent a specific format, then we should set `sort_keys=True` and `separators=(',', ';')` parameters for `json.dumps()` for a more stable format.

barnabasdomozi:
Due to other merged PRs, the schema hash needs to be changed.

barnabasdomozi:
@bruntib Rebased PR to master.

bruntib (changes requested):
We should do the garbage collection of checker sets on server restart.

barnabasdomozi:
@bruntib Implemented garbage collection in commit https://github.com/Ericsson/codechecker/commit/6a89c86290c3718d79d9c0083346869b3b553333

bruntib (approved):

PR #4910, opened 2026-06-22
Fix flakyness of the GuidelineStatistics Page

dkrupp:
Calling the onMounted() callback needs to be removed, because it is too early and by that time the filters are not ready.

1. Statistics.vue has a <ReportFilter> that emits @refresh="refresh".
2. refresh() in Statistics.vue calls refreshCurrentTab() which does bus.emit("refresh").
3. Each statistics child (like CheckerStatistics.vue) calls baseStats.setupRefreshListener(fetchStatistics) which registers bus.on("refresh", fetchStatistics) during onActivated.

So this way the fetchStatistics will be called after the filter has been initialized.


gulyasgergely902 (approved):
LGTM

PR #4911, opened 2026-06-23
[gui] Fix the issue where some report filters being cleared on dismiss event

gulyasgergely902:
This change fixes the issue where some of the report filters are being cleared when the user clicks outside of the element.

### DIFF
--- a/web/server/vue-cli/src/components/Report/ReportFilter/Filters/SelectOption/SelectOption.vue
+++ b/web/server/vue-cli/src/components/Report/ReportFilter/Filters/SelectOption/SelectOption.vue
@@ -162,6 +162,14 @@ watch(menu, async show => {
   }
 });
 
+watch(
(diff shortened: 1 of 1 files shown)


web/server/vue-cli/src/components/Report/ReportFilter/Filters/SelectOption/SelectOption.vue
```diff
@@ -162,6 +162,15 @@ watch(menu, async show => {
   }
 });
 
+watch(
+  () => props.selectedItems,
+  val => {
+    console.error("selectedItems changed: ", val);
```
dkrupp: Please remove debug log.

dkrupp (changes requested):

dkrupp (approved):
LGTM

PR #4912, opened 2026-06-23
Remove flakyness of the Reports view

dkrupp:
The reports view was flaky, because it loaded the reports too early (at page startup),
before the Report filter values were stored to the browser store. Now the page refresh is only called when the ReportFilter is updated.


gulyasgergely902 (approved):
LGTM

PR #4913, opened 2026-06-23
[gui] Fix the issue where source components were removed from the url

gulyasgergely902:
This change fixes an issue where the source component report filter was removed from the url after loading the page from the link. The issue was caused by the report filter id not getting passed to the report filter store on initialization.


dkrupp (approved):
LGTM.
Tested. works.

PR #4914, opened 2026-06-23
Fix post-login redirect dropping URL query parameters

barnabasdomozi:
Use router.replace() with a string path instead of { path: returnTo } so Vue Router correctly preserves the full query string.


dkrupp (approved):
LGTM

PR #4915, opened 2026-06-23
[gui] Fix pagination adding non-valid limit to url

gulyasgergely902:
This change fixes pagination adding a non-valid number to the url when changing the page size.


dkrupp (approved):
LGTM

PR #4916, opened 2026-06-23
Fix return_to being ignored after OAuth login

barnabasdomozi:
Persist the return_to query parameter in localStorage before redirecting to the OAuth provider, since the browser navigation loses query params. After successful OAuth callback, read it back and navigate accordingly instead of always redirecting to the app root.

Also set userName to null on OAuth login so the actual username is fetched from the server via getLoggedInUser instead of using a placeholder string.

### DIFF
--- a/web/server/vue-cli/src/store/modules/auth.js
+++ b/web/server/vue-cli/src/store/modules/auth.js
@@ -66,7 +66,7 @@ const actions = {
           "oauth", credentials.provider + "@" + credentials.url,
           handleThriftError(token => {
             context.commit(SET_AUTH, {
-              userName: "OAuth @" + credentials.provider,
+              userName: null,
               token: token
             });
             resolve(token);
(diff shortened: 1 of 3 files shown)


web/server/vue-cli/src/store/modules/auth.js
```diff
@@ -66,7 +66,7 @@ const actions = {
           "oauth", credentials.provider + "@" + credentials.url,
           handleThriftError(token => {
             context.commit(SET_AUTH, {
-              userName: "OAuth @" + credentials.provider,
+              userName: null,
```
dkrupp, line 69: Why this is needed?

dkrupp (commented):
I tested the feature and it worked. 

I am just not sure why replacing userName is needed in the context.commit call.

web/server/vue-cli/src/store/modules/auth.js
dkrupp: I see the answer in the commit message.

dkrupp (approved):
LGTM

barnabasdomozi:
If username is not replaced with null, it acted as a valid value and no user info is fetched from the server. With this change, if user info is null, the UI handling automatically fetches the user information after login again. Previously, after login, the page was reloaded twice, so new user info was fetched that way.

PR #4917, opened 2026-06-24
Fix report filter opening

gulyasgergely902:


web/server/vue-cli/src/components/Report/ReportFilter/ReportFilter.vue
```diff
@@ -798,6 +831,24 @@ function renamePresetDialog() {
...
+
+  activeCompareToPanelId.value =
+    comparedToRunFilterRef.value.panel ||
+    comparedToOpenReportsDateFilterRef.value.panel ||
+    comparedToDiffTypeFilterRef.value.panel ? 0 : undefined;
+
+  activeDatePanelId.value =
+    detectionDateFilterRef.value.panel ||
+    fixDateFilterRef.value.panel ? 0 : undefined;
+  console.error("detectionDateFilterRef=", detectionDateFilterRef.value.panel);
```
dkrupp: please remove debug logs.

web/server/vue-cli/src/components/Report/ReportFilter/ReportFilter.vue
dkrupp: can't we remove this commented out part?

dkrupp (changes requested):
Please remove the debug console logs.

I tested it and seems that it works. 

Iam just wondering. Isn't there a more compact solution to open the parent group panel when any of the children has a set value other than manually "watching" the value of the children elements one by one? 

It wrorks now, just looks like tedious to maintain.

dkrupp (approved):

PR #4919, opened 2026-06-25
[gui] Fix checker documentation dialog not showing data

gulyasgergely902:
This changes fixes a variable shadow issue in checker documentation dialog which made the dialog empty.

The fix also makes the dialog use the ConfirmDialog component to show the dialog itself.


dkrupp:
The fix works. Please make the checker name in the report table displayed as link-like, to make it apparent to the user that this is a link. It was like that in the old vue2 gui

image

dkrupp (changes requested):
The fix works, but please make the checker name link-like in the table as in the old gui.

dkrupp (approved):
LGTM

PR #4920, opened 2026-06-25
web: Remove "All" option from items-per-page in data tables

barnabasdomozi:
Replace Vuetify 2 footer-props pattern with Vuetify 3 items-per-page-options prop. Limit page size choices to 25, 50, and 100 across all data tables.


gulyasgergely902 (approved):
LGTM

PR #4921, opened 2026-06-25
[gui] Fix list review status rules pagination

gulyasgergely902:
This change replaces v-data-table with v-data-table-server to fix the pagination issue.


dkrupp (approved):
LGTM

PR #4925, opened 2026-06-26
[cmd] `parse` extends sarif output with extra fields

andyg319:
There are a few fields in the SARIF specification which would be meaningful to be filled out, but today are not generated into the exported files.

* `CodeChecker parse` already populates Report objects with review status information, but currently only JSON export makes use of it. This PR adds `suppression` objects to individual results in the SARIF file, according to the [spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790983), including suppression source (source code comment or review config file) and justification message.
`location.physicalLocation` may be added in a follow-up PR.

* SARIF `rules[]` (~checkers) are extended with `defaultConfiguration.level` and `helpUri` fields ([spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10541274)), corresponding to the `severity` and `doc_url` checker labels in CodeChecker config.
Note that this only applies to the `CodeChecker parse` command at the moment, which has the analyzer context (including a `CheckerLabel` object) readily available. Doing the same for the standalone report-converter is an orthogonal task, and might entail bigger refactorings.

### DIFF
--- a/analyzer/tests/functional/parse_sarif/test_files/Makefile
+++ b/analyzer/tests/functional/parse_sarif/test_files/Makefile
@@ -1,0 +1,7 @@
+default: a.o
+
+a.o: a.cpp
+	$(CXX) -c a.cpp -o /dev/null
+
+clean:
+	rm -f a.o
--- a/analyzer/tests/functional/parse_sarif/test_files/Makefile
+++ b/analyzer/tests/functional/parse_sarif/test_files/Makefile
@@ -1,0 +1,4 @@
+default: a.o
+
+a.o: a.cpp
+	$(CXX) -c a.cpp -o /dev/null
--- a/analyzer/tests/functional/parse_sarif/test_sarif.py
+++ b/analyzer/tests/functional/parse_sarif/test_sarif.py
@@ -1,0 +1,156 @@
+#
+# -------------------------------------------------------------------------
+#
+#  Part of the CodeChecker project, under the Apache License v2.0 with
+#  LLVM Exceptions. See LICENSE for license information.
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# -------------------------------------------------------------------------
+
+"""
+Test parse --export sarif command.
+"""
+
+import os
+import json
+import shutil
+import subprocess
+import unittest
+
+from libtest import env
+
+
+class TestParseSarif(unittest.TestCase):
+    _ccClient = None
+
+    def setup_class(self):
+        """Setup the environment for the tests."""
+
+        global TEST_WORKSPACE
+        TEST_WORKSPACE = env.get_workspace('skip')
+
+        report_dir = os.path.join(TEST_WORKSPACE, 'reports')
+        os.makedirs(report_dir)
+
+        os.environ['TEST_WORKSPACE'] = TEST_WORKSPACE
+
+    def teardown_class(self):
+        """Delete the workspace associated with this test"""
+
+        # TODO: If environment variable is set keep the workspace
+        # and print out the path.
+        global TEST_WORKSPACE
+
+        print("Removing: " + TEST_WORKSPACE)
+        shutil.rmtree(TEST_WORKSPACE)
+
+    def setup_method(self, _):
+
+        # TEST_WORKSPACE is automatically set by test package __init__.py .
+        self.test_workspace = os.environ['TEST_WORKSPACE']
+
+        test_class = self.__class__.__name__
+        print('Running ' + test_class + ' tests in ' + self.test_workspace)
+
+        # Get the CodeChecker cmd if needed for the tests.
+        self._codechecker_cmd = env.codechecker_cmd()
+        self._tu_collector_cmd = env.tu_collector_cmd()
+        self.report_dir = os.path.join(self.test_workspace, "reports")
+        self.test_dir = os.path.join(os.path.dirname(__file__), 'test_files')
+
+    def __run_cmd(self, cmd):
+        process = subprocess.Popen(
+            cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            cwd=self.test_dir,
+            encoding="utf-8",
+            errors="ignore")
+        out, err = process.communicate()
+
+        print(out)
+        print(err)
+
+        if process.returncode != (2 if "parse" in cmd else 0):
+            return err
+
+        return ''.join(out)
+
+    def __log_and_analyze(self, analyze_args=None):
+        """ Log and analyze the test project. """
+        build_json = os.path.join(self.test_workspace, "build.json")
+
+        clean_cmd = ["make", "clean"]
+        out = subprocess.check_output(clean_cmd,
+                                      cwd=self.test_dir,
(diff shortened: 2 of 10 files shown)


analyzer/tests/functional/parse_sarif/test_sarif.py
```diff
@@ -0,0 +1,179 @@
...
+            'defaultConfiguration': {
+                'level': 'error'
+            }
+        })
+
+    def test_parse_sarif_suppression_source(self):
+        self.__log_and_analyze()
+
+        parse_sarif_cmd = [self._codechecker_cmd, "parse", self.report_dir,
+                           "-e", "sarif", "--review-status", "false_positive"]
```
dkrupp, line 125: For sarif export I think the convenient default would be to export everything into sarif and not just the ['confirmed', 'unreviewed'] reports. For standard command line outptut and possible from html,  the current default is meaningful (you dont want to see dismissed reports).

Can you please add a filter option "everything" and make that as the default for sarif export?
This way the users would not need to list all review status possibilities at command line invocation if they want to export everything.

dkrupp (changes requested):
The implementation works nicely, Please check my comments related to the default filters.

analyzer/tests/functional/parse_sarif/test_sarif.py
dkrupp: This will be fixed in a followup TR

dkrupp (approved):
LGTM

analyzer/tests/functional/parse_sarif/test_files/Makefile
```diff
@@ -0,0 +1,7 @@
+default: a.o
+
+a.o: a.cpp
+	$(CXX) -c a.cpp -o /dev/null
+
+clean:
```
barnabasdomozi: This clean target is pointless. On the line above, the output of the compilation is sent to /dev/null.

tools/report-converter/codechecker_report_converter/report/parser/sarif.py
```diff
@@ -310,17 +310,35 @@ def convert(
...
+                    doc_url = checker_labels.label_of_checker(
+                            checker_name, "doc_url", report.analyzer_name)
+                    if doc_url:
+                        if isinstance(doc_url, list):
+                            doc_url = doc_url[0]
+                        rules[checker_name]["helpUri"] = doc_url
+
+                    severity = checker_labels.severity(
+                            checker_name,
+                            report.analyzer_name)  # type: ignore[call-arg]
```
barnabasdomozi: Instead of ignoring type warnings, the current type stub should be updated in file `tools/report-converter/codechecker_report_converter/report/checker_labels.py`.

tools/report-converter/codechecker_report_converter/report/parser/sarif.py
```diff
@@ -310,17 +310,35 @@ def convert(
...
 
+                if checker_labels:
+                    doc_url = checker_labels.label_of_checker(
+                            checker_name, "doc_url", report.analyzer_name)
+                    if doc_url:
+                        if isinstance(doc_url, list):
+                            doc_url = doc_url[0]
+                        rules[checker_name]["helpUri"] = doc_url
+
+                    severity = checker_labels.severity(
```
barnabasdomozi, line 334: `checker_labels.severity()` can return `UNSPECIFIED`, and in that case the `_to_level` function returns `warning`. Is that intentional?

analyzer/tests/functional/parse_sarif/test_sarif.py
```diff
@@ -0,0 +1,179 @@
...
+            return err
+
+        return ''.join(out)
+
+    def __log_and_analyze(self, analyze_args=None):
+        """ Log and analyze the test project. """
+        build_json = os.path.join(self.test_workspace, "build.json")
+
+        clean_cmd = ["make", "clean"]
+        out = subprocess.check_output(clean_cmd,
```
barnabasdomozi: In this case, why not reuse the existing `__run_cmd` function?

analyzer/tests/functional/parse_sarif/test_sarif.py
```diff
@@ -0,0 +1,179 @@
...
+            errors="ignore")
+        out, err = process.communicate()
+
+        print(out)
+        print(err)
+
+        if process.returncode != (2 if "parse" in cmd else 0):
+            return err
+
+        return ''.join(out)
```
barnabasdomozi: `out` is already a string, consider:
```suggestion
        return out
```

barnabasdomozi (changes requested):
The tests cases were not executed for some reason. Can you try force pushing a commit that would trigger the CI?

tools/report-converter/codechecker_report_converter/report/parser/sarif.py
andyg319: Yes, this is the Sarif [recommendation](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10541293).

analyzer/tests/functional/parse_sarif/test_files/Makefile
andyg319: `analyzer/tests/functional/parse_status` does the same, let's patch them together if we think this extra foolproofness is not necessary.

barnabasdomozi: I think we shouldn't commit code to the repository that would be eventually removed anyway.
Cleaning up `analyzer/tests/functional/parse_status` should be a separate task, unreleated to this change.

barnabasdomozi (changes requested):

barnabasdomozi (changes requested):
Currently, some of the test cases are failing in the CI.

tools/report-converter/codechecker_report_converter/report/parser/sarif.py
andyg319: Adding an `Optional` parameter to the stub does not satisfy mypy for some reason. It errors with "Too many|few arguments" either way. Note that the other method in the stub has the same problem in the first place, but `label_of_checker` is called with the same number of arguments everywhere (when typed with this stub at least), so the problem has not surfaced yet (if you try it, mypy complains right away).

andyg319: The problem is that you cannot specify the default `None` value for `Optional` arguments in type hints, and therefore this was an incorrect stub for the actual `CheckerLabels` implementation in `codechecker_common`. The only solution seems to be to provide placeholder implementations for the methods, instead of mere `Callable` hints.

barnabasdomozi:
Please rebase to master, the linter error is already fixed on it.

andyg319:
> Please rebase to master, the linter error is already fixed on it.

Still not okay.

barnabasdomozi:
@andyg319 I believe its fixed now. Try to rebase again.

barnabasdomozi (approved):
LGTM

PR #4929, opened 2026-06-29
[ci] Move CODEOWNERS to .github directory and fix the layout

gulyasgergely902:
CODEOWNERS file should be in .github directory in modern repositories. This change fixes the location and the layout of the file since GitHub reads the file frop top to bottom and applies the ownerships in this fashion. By moving the global rules to the top, it always applies to every PR and going down, it specifies the ownerships based on the modified files.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,14 +1,14 @@
 # Lines starting with '#' are comments.
 # Each line is a file pattern followed by one or more owners.
 
-# Any CODEOWNERS file modifications must be approved by repository owner.
-/.github/CODEOWNERS @dkrupp
-
 # These owners will be the default owners for everything in the repo.
-*      @bruntib @Vodorok
+* @bruntib
 
 # Documentations.
-/docs @dkrupp
+/docs @bruntib @dkrupp
 
 # Frontend.
-/web/server/vue-cli @gulyasgergely902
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,14 +1,14 @@
...
 # Documentations.
-/docs @dkrupp
+/docs @bruntib @dkrupp
 
 # Frontend.
-/web/server/vue-cli @gulyasgergely902
+/web/server/vue-cli @bruntib @gulyasgergely902
+
+# Any CODEOWNERS file modifications must be approved by repository owner.
+/.github/CODEOWNERS @dkrupp
```
bruntib, line 14: What is the motivation behind replacing this file to `.github` directory? Other standard files like license.txt, contributing.md, security.md etc. are also in the root. I think it would be better not to spread them in different directories.

bruntib (changes requested):

.github/CODEOWNERS
gulyasgergely902: The documentation on this suggests that for GitHub the default directory for storing these files is `.github/`. It can also read from the root and from `docs/`. I'm open for opposite suggestions, but I think we should store any git related files in the .github directory.

bruntib (approved):

PR #4931, opened 2026-06-29
Checkboxes and click-name selection(File-path-filter)

xb058t:


gulyasgergely902 (changes requested):
If I click on an entry it selects it automatically. The goal would be to open the tree leaves by clicking on them and select one or more by selecting the checkbox next to it.

xb058t:
> If I click on an entry it selects it automatically. The goal would be to open the tree leaves by clicking on them and select one or more by selecting the checkbox next to it.

fixed - clicking a row now only expands/collapses it (open-on-click), selection happens only via the checkbox.

gulyasgergely902 (approved):
LGTM

PR #4933, opened 2026-06-29
guard against null refs in syncGroupPanels

xb058t:


gulyasgergely902:
This change was already merged on a separate PR.

PR #4934, opened 2026-06-29
Fix vue warnings

xb058t:
Fixed:

TheHeader.vue - [Vue Router warn]: Discarded invalid param(s) "endpoint" when navigating - removed explicit endpoint param, Vue Router 4.1.4+ inherits it automatically

CheckerMessageFilter.vue - [Vue warn]: Property "setSelectedItems" was accessed during render but is not defined on instance - was referencing a non-existent local function instead of baseSelectOptionFilter.setSelectedItems

SingleLineWidget.vue - [Vue warn]: Invalid prop: type check failed for prop "value". Expected Array, got Null/Object/Number - prop type was too strict, broadened to [Array, Number, Object]

RunList.vue - [Vue warn]: Failed to resolve component: AnalysisInfoDialog - component was used in template but never imported

Statistics.vue - [Vue Router warn]: <router-view> can no longer be used directly inside <keep-alive> - updated to Vue 3 slot pattern

Statistics.vue - [Vue warn]: Runtime directive used on component with non-element root node - v-fill-height was on <ReportFilter> (fragment root), moved to a wrapping <div>

### DIFF
--- a/web/server/vue-cli/src/components/Statistics/Overview/SingleLineWidget.vue
+++ b/web/server/vue-cli/src/components/Statistics/Overview/SingleLineWidget.vue
@@ -62,7 +62,7 @@ defineProps({
   label: { type: String, required: true },
   helpMessage: { type: String, default: null },
   bus: { type: Object, required: true },
-  value: { type: Array, required: true },
+  value: { type: Number, default: null },
   loading: { type: Boolean, default: false }
 });
 
--- a/web/server/vue-cli/src/views/RunList.vue
+++ b/web/server/vue-cli/src/views/RunList.vue
@@ -9,12 +9,6 @@
       :run-history-id="selectedRunHistoryId"
     />
 
-    <AnalysisInfoDialog
-      v-model="analysisInfoDialog"
-      :run-id="selectedRunId"
-      :run-history-id="selectedRunHistoryId"
-    />
-
     <v-data-table-server
       v-model="selected"
       v-model:page="page"
@@ -59,7 +53,6 @@
               v-model:selected-compared-to-tags="selectedComparedToTags"
               :histories="item.$history.values"
               :run="item"
-              :open-analysis-info-dialog="openAnalysisInfoDialog"
               :open-analyzer-statistics-dialog="openAnalyzerStatisticsDialog"
             >
               <v-btn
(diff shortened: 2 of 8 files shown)


web/server/vue-cli/src/components/Statistics/Overview/SingleLineWidget.vue
```diff
@@ -62,7 +62,7 @@ defineProps({
   label: { type: String, required: true },
   helpMessage: { type: String, default: null },
   bus: { type: Object, required: true },
-  value: { type: Array, required: true },
+  value: { type: [ Array, Number, Object ], default: null },
```
gulyasgergely902: What was the reason setting it to accept arrays, numbers and objects? In the code this component is used 2 times and on both occasions it is getting a number so theoretically only numbers should be expected. Or did I miss something?

web/server/vue-cli/src/views/RunList.vue
```diff
@@ -170,6 +170,7 @@ import {
   RunSortType
 } from "@cc/report-server-types";
 
+import AnalysisInfoDialog from "@/components/AnalysisInfoDialog";
```
gulyasgergely902: This dialog was moved to the RunNameColumn itself where it gets activated. Because of this modification, the dialog can be removed from RunList entirely.

gulyasgergely902 (changes requested):

web/server/vue-cli/src/components/Statistics/Overview/SingleLineWidget.vue
xb058t: Added it because of the Expected Array, got Object/Number/Null warning, but fixed the actual cause instead, getNumberOfFailedFiles() wasn't calling .toNumber(). value is just Number now.

web/server/vue-cli/src/views/RunList.vue
xb058t: Removed it from RunList.vue entirely. Also cleaned up the unused openAnalysisInfoDialog prop in RunNameColumn.vue and ExpandedRun.vue.

gulyasgergely902 (approved):
LGTM

PR #4935, opened 2026-06-29
Updated linting profile

xb058t:
vue/component-api-style - "warn" to "error"
vue/block-order - "warn" to "error"
vue/define-macros-order - "warn" to "error"
vue/require-macro-variable-name - "warn" to "error"
vue/valid-v-slot - "off" to ["error", { "allowModifiers": true }]
vue/valid-next-tick - "off" to "error"
vue/no-mutating-props - removed entirely

package.json
test:lint script - removed obsolete --ext .js,.vue flag (not supported in ESLint v9, file extensions are handled by the files array in eslint.config.mjs)

### DIFF
--- a/web/server/vue-cli/eslint.config.mjs
+++ b/web/server/vue-cli/eslint.config.mjs
@@ -52,24 +52,23 @@ export default defineConfig([
         "multiline": 1
       }],
       "vue/v-slot-style": "off",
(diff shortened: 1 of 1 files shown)


gulyasgergely902 (approved):
LGTM

PR #4936, opened 2026-06-30
Change report ID columns to BIGINT to avoid 32-bit integer overflow

NikhilJeikar:
Report IDs were stored as INTEGER (max ~2.1B). Large deployments can hit this limit and fail on insert. Promote reports.id and all report_id FKs to BIGINT on PostgreSQL; SQLite already supports 64-bit integers.

### DIFF
--- a/web/server/codechecker_server/migrations/report/versions/a7f3c8e21b90_report_id_bigint.py
+++ b/web/server/codechecker_server/migrations/report/versions/a7f3c8e21b90_report_id_bigint.py
@@ -1,0 +1,73 @@
+"""
+Report id bigint
+
+Revision ID: a7f3c8e21b90
+Revises:     24c9660f82b1
+Create Date: 2026-06-30 00:00:00.000000
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# Revision identifiers, used by Alembic.
+revision = 'a7f3c8e21b90'
+down_revision = '24c9660f82b1'
+branch_labels = None
+depends_on = None
+
+
+REPORT_ID_FK_TABLES = [
+  ('bug_path_events', False),
+  ('bug_report_points', False),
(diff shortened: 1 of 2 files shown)


web/server/codechecker_server/migrations/report/versions/a7f3c8e21b90_report_id_bigint.py
```diff
@@ -0,0 +1,73 @@
...
+down_revision = '24c9660f82b1'
+branch_labels = None
+depends_on = None
+
+
+REPORT_ID_FK_TABLES = [
+  ('bug_path_events', False),
+  ('bug_report_points', False),
+  ('extended_report_data', False),
+  ('report_analysis_info', True),
```
bruntib: How can this column be NULL? Couldn't we set all `report_id` foreign key columns as not nullable?

bruntib (changes requested):
Thanks for the patch. I just have one question during the review.

web/server/codechecker_server/migrations/report/versions/a7f3c8e21b90_report_id_bigint.py
NikhilJeikar: Good catch. report_id being nullable is not intentional - it was defined that way in the original dabc6998b8f0 migration when report_analysis_info was added, but the application never creates rows without a report_id. If there's no analysis info for a report, no association row is created at all.
I agree we should tighten this. If you want I can do it in this PR or a seperate one.

bruntib: It would be great if this would be redesigned in this PR. Thank you in advance!

bruntib (approved):

bruntib:
Thanks for the improvement!

PR #4940, opened 2026-07-01
[gui] Upgrade npm packages

gulyasgergely902:
Upgrade outdated npm packages


bruntib:
The `^a.b.c` notation locks the major version only, so maybe in the future we won't need to upgrade minor and patch versions unless we must use a minimum minor release.

bruntib (approved):

PR #4941, opened 2026-07-01
[gui] Fix regex filters are not getting added to the URL

gulyasgergely902:
The change fixes an issue where the regex filters are not added to the URL on apply.


bruntib:
I did the following scenario:

1. Open "reports" page
2. Set *int* in "checker message" filter field
3. URL doesn't change

Is this patch supposed to fix all regex-based filter fields?

gulyasgergely902:
It was supposed to fix all and in fact it fixes them but during migration I left out a filter - CheckerMessage - so it does not add the selected filters to the URL. I've fixed this and it should now work.

bruntib (approved):

PR #4950, opened 2026-07-07
Fix unbound local variable bug

barnabasdomozi:
When running the
CodeChecker cmd results RUN --url URL --checker-msg MSG command,
the following error occurred:

File "codechecker/build/CodeChecker/lib/python3/codechecker_client/cmd_line_client.py", line 881, in handle_list_results
  print(twodim.to_str(args.output_format, header, rows))

UnboundLocalError: cannot access local variable 'header'
where it is not associated with a value

This patch fixes the above issue.


bruntib:
:heavy_check_mark:

bruntib (approved):

PR #4951, opened 2026-07-08
[perf] Improve analyze file filter for header

### LINKED ISSUE #4949: [perf] Improve analyze file filter for header
Automatic detection of source file affected by header is terribly slow. A complete analysis without the `--file` filter is 15 minutes. Specifying 40 headers with `--file` takes more than 1h to analyze. Our compilation database as ~1000 entry.

The bottle neck seem to be [get_dependent_sources](https://github.com/Ericsson/codechecker/blob/1a80dcf5743b1dd0d440becf938196632998e755/tools/tu_collector/tu_collector/tu_collector.py#L475). As far as I understood, this function is called for each header sequentially and iterate over each entry of the compilation database to compute a dependency lookup table.

We could save the dependency lookup table and parallelize this process.

Easily reproducible with following bash script:

```bash
#!/bin/bash

function generate_db {
    echo "Generating database..."

    echo "[" > compile_commands.json

    for _ in {1..1000}; do
        echo "{\"directory\": \"$(pwd)\", \"command\": \"g++ dummy.cpp -c -o dummy.o\", \"file\": \"dummy.cpp\", \"output\": \"dummy.o\"}," >> compile_commands.json
    done
    echo "{\"directory\": \"$(pwd)\", \"command\": \"g++ single_entry.cpp -c -o dummy.o\", \"file\": \"single_entry.cpp\", \"output\": \"dummy.o\"}" >> compile_commands.json
    echo "]" >> compile_commands.json
}

generate_db
touch dummy.cpp
touch single_entry.cpp
touch dummy.hpp
touch dummy_2.hpp

echo "Full analysis... ------------------------------------------"
time CodeChecker analyze compile_commands.json -o ./reports
echo "Single source analysis... ------------------------------------------"
time CodeChecker analyze compile_commands.json -o ./reports --file single_entry.cpp
echo "Single header analysis... ------------------------------------------"
time CodeChecker analyze compile_commands.json -o ./reports --file dummy.hpp
echo "Two header analysis... ------------------------------------------"
time CodeChecker analyze compile_commands.json -o ./reports --file dummy.hpp dummy_2.hpp
```

## Environment

```
❯ CodeChecker version
[INFO 2026-07-07 13:49:00] - CodeChecker analyzer version:
Base package version | 6.28.2                                  
Package build date   | 2026-06-29T08:05                        
Git commit ID (hash) | 13b2b26718497c13ad9987a5d71d7f2dcb9d270f
Git tag information  | 6.28.2                                  

[INFO 2026-07-07 13:49:00] - CodeChecker web version:
Base package version                | 6.28.2                                  
Package build date                  | 2026-06-29T08:05                        
Git commit ID (hash)                | 13b2b26718497c13ad9987a5d71d7f2dcb9d270f
Git tag information                 | 6.28.2                                  
Server supported Thrift API version | 6.71                                    
Client Thrift API version           | 6.71                                    
❯ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=24.04
DISTRIB_CODENAME=noble
DISTRIB_DESCRIPTION="Ubuntu 24.04.4 LTS"
```

gzzi:
In this PR:

 - Compute dependencies of each source file of the compilation database only once for all header provided
 - Parse the compilation database in parallel

Close #4949

### DIFF
--- a/analyzer/codechecker_analyzer/cli/analyze.py
+++ b/analyzer/codechecker_analyzer/cli/analyze.py
@@ -1012,7 +1012,8 @@ def is_checker_config_valid(
 
 def get_affected_file_paths(
     file_filters: List[str],
-    compile_commands: tu_collector.CompilationDB
+    compile_commands: tu_collector.CompilationDB,
+    jobs: int
 ) -> List[str]:
     """
     Returns a list of source files for existing header file otherwise returns
@@ -1028,7 +1029,7 @@ def get_affected_file_paths(
                 file_filter.endswith(header_file_extensions):
             LOG.info("Get dependent source files for '%s'...", file_filter)
             dependent_sources = tu_collector.get_dependent_sources(
-                compile_commands, file_filter, dependencies)
+                compile_commands, file_filter, dependencies, jobs)
 
             LOG.info("Get dependent source files for '%s' done.", file_filter)
             LOG.debug("Dependent source files: %s",
@@ -1047,7 +1048,7 @@ def __get_skip_handlers(args, compile_commands) -> SkipListHandlers:
     skip_handlers = SkipListHandlers()
     if 'files' in args:
         source_file_paths = get_affected_file_paths(
-            args.files, compile_commands)
+            args.files, compile_commands, args.jobs)
 
         # Creates a skip file where all source files will be skipped except
         # the given source files and all the header files.
--- a/analyzer/codechecker_analyzer/cli/analyze.py
+++ b/analyzer/codechecker_analyzer/cli/analyze.py
@@ -1019,6 +1019,7 @@ def get_affected_file_paths(
     with the same file path expression.
     """
     file_paths = []  # Use list to keep the order of the file paths.
+    dependencies: tu_collector.Dependencies = collections.defaultdict(set)
     for file_filter in file_filters:
         file_paths.append(str(Path(file_filter).resolve())
                           if '*' not in file_filter else file_filter)
@@ -1027,7 +1028,7 @@ def get_affected_file_paths(
                 file_filter.endswith(header_file_extensions):
             LOG.info("Get dependent source files for '%s'...", file_filter)
             dependent_sources = tu_collector.get_dependent_sources(
-                compile_commands, file_filter)
+                compile_commands, file_filter, dependencies)
 
             LOG.info("Get dependent source files for '%s' done.", file_filter)
             LOG.debug("Dependent source files: %s",
--- a/analyzer/codechecker_analyzer/cli/analyze.py
+++ b/analyzer/codechecker_analyzer/cli/analyze.py
@@ -1001,28 +1001,35 @@ def is_checker_config_valid(
 
 def get_affected_file_paths(
     file_filters: List[str],
-    compile_commands: tu_collector.CompilationDB
+    compile_commands: tu_collector.CompilationDB,
+    jobs: int
 ) -> List[str]:
     """
     Returns a list of source files for existing header file otherwise returns
     with the same file path expression.
     """
     file_paths = []  # Use list to keep the order of the file paths.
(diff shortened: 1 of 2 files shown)


bruntib (changes requested):
Hi @gzzi,

Thank you for the path, this is a really nice catch and a great improvement!

However, I have some concerns related to the solution design. It doesn't seem intuitive for a library to provide a "dependencies" as a parameter because it is supposed to be the result of the function, more or less. I mean, only an extra filtering happens, based on the file filter.

Couldn't `get_dependent_sources()` be redesigned in a way that `header_path` becomes `header_paths: Optional[List[str]]`. This way, this function could be called only once from `get_affected_file_paths()` instead of calling it many times in a loop.

What do you think?

gzzi:
Hello @bruntib,

Agree, I tried to keep my change on `analyze.py` low. Now I pushed a version where the loop iteration is done on the `tu_collector` lib. Let me know what you think about it.

Note that now the order of `file_paths` in `get_affected_file_paths` is not the same as before. All the source affected by header are placed at the end.

bruntib:
Hi @gzzi,

Thank you for the fixes. I have just two small comments:
1. Can `header_file` variable be renamed to `header_files`?
2. The default value of `jobs` in `get_dependent_sources()` function parameter could be 1 instead of `None`. This way there is no need to check `if jobs is None`.

Also, one more question. I understand the goal of the development, and it is really useful in case of big projects. However, if I `#include "dummy.hpp"` in the beginning of `single_entry.cpp` and run the command `CodeChecker analyze --file dummy.hpp ...` then it still doesn't analyze anything, because `--file` works as if one single file was provided in a skipfile with `--skip`. Since `single_entry.cpp` doesn't match `dummy.hpp`, it won't be analyzed.
This is an independent question from this PR, because this was the situation earlier, too. I'm just wondering, what is the usage of `--file` flag in your project in this regard?

Thank you again this development. Please, consider applying these two minor fixes, otherwise we can merge it.

gzzi:
Hello @bruntib,

Requested changes have been applied.

>  if I #include "dummy.hpp" in the beginning of single_entry.cpp and run the command CodeChecker analyze --file dummy.hpp ... then it still doesn't analyze anything

The path are absolute on the `compile_commands.json`. You need to provide `--file $(pwd)/dummy.hpp`. Here is a script showing that:

```bash
#!/bin/bash

function generate_db {
    echo "Generating database..."

    echo "[" > compile_commands.json

    for _ in {1..1000}; do
        echo "{\"directory\": \"$(pwd)\", \"command\": \"g++ dummy.cpp -c -o dummy.o\", \"file\": \"dummy.cpp\", \"output\": \"dummy.o\"}," >> compile_commands.json
    done
    echo "{\"directory\": \"$(pwd)\", \"command\": \"g++ single_entry.cpp -c -o dummy.o\", \"file\": \"single_entry.cpp\", \"output\": \"dummy.o\"}" >> compile_commands.json
    echo "]" >> compile_commands.json
}

generate_db
touch dummy.cpp
echo "#include \"dummy.hpp\"" > single_entry.cpp
touch dummy.hpp
touch dummy_2.hpp

echo "Relative path ------------------------------------------"
time CodeChecker analyze compile_commands.json -o ./reports --file dummy.hpp

echo "Absolute path ------------------------------------------"
time CodeChecker analyze compile_commands.json -o ./reports --file $(pwd)/dummy.hpp
```

> I'm just wondering, what is the usage of --file flag in your project in this regard?

On our CI, when an analysis of a pull request is made, we check only files that was modified compared to main branch fork point. This save us quite a lot of time. The full analysis are made only nightly.

bruntib:
Hi @gzzi,

I see now. Thank you for the explanation and thank you for the patch!

bruntib (approved):

PR #4952, opened 2026-07-08
Make file tree stat cells clickable to filter reports

xb058t:


gulyasgergely902 (changes requested):
Instead of directly modifying the storage which stores the filter settings (in function setReportFilter), you should update the URL then call an initByUrl to actually update the filters. The current solution does not put anything into the URL and fails to set the report filters severity and review status. Please check the proposed solution.

gulyasgergely902 (changes requested):
There were issues during the conflict resolution. Check the files and npm.

gulyasgergely902 (approved):
LGTM

PR #4954, opened 2026-07-08
Add wildcard pattern filter option to file path tree view

xb058t:
Type a pattern into the File path search box, check the "Filter by wildcard pattern:" box, and apply.


gulyasgergely902:
The old file path filter had the option to select multiple wildcard filters. Currently this solution only adds the latest and removes the previous. Please make it possible to add multiple wildcard filters.

gulyasgergely902 (changes requested):
Added the requests in the comments.

gulyasgergely902 (changes requested):
There were issues during conflict resolution. Check the code.

gulyasgergely902 (approved):
LGTM

PR #4955, opened 2026-07-08
Encode file tree view mode and expanded folders into the URL

xb058t:


gulyasgergely902:
Could you please add a list view option too and if the user opens up the reports page, it loads to the list view in the url too (view=list)? Also the opened path could be a single parameter if that is possible.

gulyasgergely902 (changes requested):
Left the requests in the comments.

gulyasgergely902 (approved):
LGTM

PR #4958, opened 2026-07-09
Rework guidelines selector component

### LINKED ISSUE #4959: Rework guideline statistics selector
Rework guideline statistics selector to show only one guideline type in a line. Currently it shows multiple if it fits into the line which looks bad.
As you can see the dropdown contains all the selected rules just in one line one after the other just the max width of the dropdown breaks them into multiple lines.

The goal would be to create a better solution to display what rules are selected currently.

feyruzb:
This PR makes the Guidelines selector be smaller by adding `chips` tag to component, making them individually and collectively clearable. 

New look:
image
Old look:
image


gulyasgergely902 (approved):
LGTM

PR #4962, opened 2026-07-14
[feat] Add memory safety reporter script

gulyasgergely902:
This change adds the new memory safety reporter script which collects all configs which were used for a given analysis.


scripts/memory_safety_reporter.py
```diff
@@ -0,0 +1,176 @@
...
+            reports_dir,
+            "-e",
+            "sarif",
+            "-o",
+            os.path.join(to_dir, "reports.sarif")
+        ],
+        check=False
+    )
+
+    if result.returncode != 2:
```
bruntib: In case of no reports `CodeChecker parse` returns 0.

tools/memory-safety-reporter/memory_safety_reporter.py
```diff
@@ -0,0 +1,176 @@
+#!/usr/bin/env python3
```
barnabasdomozi, line 1: Please license information to the top of the file.

tools/memory-safety-reporter/memory_safety_reporter.py
```diff
@@ -0,0 +1,176 @@
...
+import argparse
+import hashlib
+import os
+import shutil
+import subprocess
+import tempfile
+from pathlib import Path
+
+
+def generate_checker_details(to_dir: Path):
```
barnabasdomozi, line 32: While this works, writing the output of the `CodeChecker checkers` command to a JSON file should really be supported out of the box. E.g. with a command like:
```CodeChecker checkers --guideline memory-safety -o json --details --output-file checker_details.json```

File: tools/memory-safety-reporter/memory_safety_reporter.py [hunk 4]
```diff
@@ -0,0 +1,176 @@
...
+    originated from `path`"""
+    checksum_file = os.path.join(path, "CHECKSUMS.sha256")
+    with open(checksum_file, "w", encoding="utf-8") as file:
+        for root, _, files in os.walk(path):
+            for filename in files:
+                if filename == "CHECKSUMS.sha256":
+                    continue
+                filepath = os.path.join(root, filename)
+                sha256 = hashlib.sha256()
+                with open(filepath, "rb") as file_handle:
```
barnabasdomozi, line 91: Consider using:
https://docs.python.org/3/library/hashlib.html#hashlib.file_digest
which is natively supported and can be faster.

scripts/memory_safety_reporter.py
```diff
@@ -0,0 +1,176 @@
...
+    base_file_name = "_".join(set_arguments)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        temp_dir = Path(tmpdir)
+        base_dir = temp_dir / base_file_name
+        base_dir.mkdir(parents=True, exist_ok=True)
+
+        generate_reports_sarif(parsed_args.report_directory, base_dir)
+
+        shutil.copy2(
```
barnabasdomozi: Why use `shutil.copy2` and not `shutil.copy` ?

barnabasdomozi (changes requested):
Can you add a test case which uses this script?
In the future, flags for these CodeChecker commands could change and then this script could be broken by accident.

tools/memory-safety-reporter/memory_safety_reporter.py
gulyasgergely902: With the current setup it is not possible to write this file, just to print to the stdout. I completely agree that it should be part of the command to actually write a file. I think the best solution now is to create a ticket for it and proceed with this solution until the other one is complete.

tools/memory-safety-reporter/memory_safety_reporter.py
gulyasgergely902: It is only added in Python 3.11 but as I can see we use older versions too where this function is not yet added.

scripts/memory_safety_reporter.py
gulyasgergely902: In fact we can use the normal copy.

barnabasdomozi (changes requested):
As mentioned in my other comment, please add a simple test case which is executed in the GitHub CI.

barnabasdomozi (changes requested):
I think creating a whole python package with `setup.py` and `Makefile` is unnecessary for this simple script. And I also think it's unnecessary to add `Makefile` targets to upload this to the PyPi registry.

Moreover, when building CodeChecker with
`make venv` and then `make package`, this new `memory-safety-reporter` command is not available for some reason.

As a simpler solution, I think it would suffice to add `#!/usr/bin/env python3` to the top of the file, and then it could be executed as `memory-safety-reporter`. 
An alternative would be to add this script as a `CodeChecker` subcommand, e.g. `CodeChecker memory-safety-converter`.  I think this would make more sense and it would worth a discussion with the team.

As for the test cases, the `tools/memory-safety-reporter/tests/functional/cmdline/test_files/` directory should be generated by the test case and not commited to this repository. Ideally, the test case should not include binary files like `tools/memory-safety-reporter/tests/functional/cmdline/test_files/test-report/failed/xmltest.cpp_clangsa_382c4897e0fb3b0c4fe94ece86cee8ca.plist_CTU_compile_error.zip` and should also not include development machine absolute paths like `/local/test_projects/tinyxml2/xmltest.cpp`.

gulyasgergely902:
After a short meeting agreed to:
- Use a simpler solution instead of the module approach since the script will not be imported by others.
- Simplify the tests to use a simpler recreation of a report directory.

Makefile
```diff
@@ -41,7 +41,10 @@ mkdocs_build:
 package_gerrit_skiplist: package_dir_structure
 	cp -p scripts/gerrit_changed_files_to_skipfile.py $(CC_BUILD_BIN_DIR)
 
-package: package_dir_structure set_git_commit_template package_gerrit_skiplist
+package_memory_safety_reporter: package_dir_structure
+	cp -p tools/memory-safety-reporter/memory_safety_reporter.py $(CC_BUILD_BIN_DIR)
```
barnabasdomozi: Minor suggestion:
I think it would be more clean if we copy this python file to the bin folder and name it `memor_safety_reporter` (without the `.py`). 

This way, users can just call it by typing `memory_safety_reporter`.

tools/memory-safety-reporter/test-report/tinyxml2.cpp_clang-tidy_fa24366a09dbdda17fbcdca1785fc072.plist
barnabasdomozi, line 1: Please remove these large `.plist` and `.plist.bak` files from this patch. As an example, this file is 4500 lines long.

I think we should keep these test cases as small as possible with only few plists files that are below 100 lines.

barnabasdomozi (changes requested):
Overall looks good to me, but I left a few comments.

tools/memory-safety-reporter/test-report/tinyxml2.cpp_clang-tidy_fa24366a09dbdda17fbcdca1785fc072.plist
gulyasgergely902: I've created an even smaller report by analyzing a single cpp file. This contains only the necessary files to test the script.

barnabasdomozi (approved):
LGTM

PR #4971, opened 2026-07-16
[gui] Add checker documentation link to report info dialog

gulyasgergely902:
This change adds a documentation link button to the report info dialog.

### DIFF
--- a/web/server/vue-cli/src/components/Report/ReportInfo/ReportInfo.vue
+++ b/web/server/vue-cli/src/components/Report/ReportInfo/ReportInfo.vue
@@ -56,6 +56,7 @@
                 :title="formatLabel('checkerId')"
                 :value="value['checkerId']"
                 :query="links['checkerId']"
+                :doc-url="docUrl"
               />
             </v-row>
             <v-row>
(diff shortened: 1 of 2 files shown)


barnabasdomozi (changes requested):
While this works, it's actually hard to find the button to open the documentation.
Can you add the link next to the checker name like shown on this screenshot? Of course, I meant the same styling as label "checker name".

Image

gulyasgergely902:
I put the button so it fits the style of the whole window as a link would disturb the whole aesthetics. We can talk about a middle solution of course.

bruntib:
What are the aesthetic considerations here? Couldn't there be a "Checker documentation" item in that box with the link itself that could be clicked? If you think of the height of the gray boxes, then even in the screenshot above, the left box has some empty space already, at the bottom due to the 2-lines checker message.

gulyasgergely902:
The checker documentation is "tied" to the checker name itself. So if the checker name is already there, why can't there be a button to open the corresponding documentation? I think it is a cleaner approach to this than to add another field to this.

bruntib:
The only concern is that it's hard to find. Maybe a label "(documentation)" besides the button would draw the attention better. But I don't think that severity or checker message ties less to the checker than the documentation.

gulyasgergely902:
I've added a v-chip component which shows if there is a documentation available. It should be visible enough for the users to see.

barnabasdomozi:
Great work, now it's more clear to find it.

barnabasdomozi (approved):

PR #4973, opened 2026-07-17
[gui] Mitigate frontend namespaces

gulyasgergely902:
This change removes the namespaced report filtering which was implemented way before to allow a separately set report filter selection for Reports page and for Statistics page.

This change uniforms this and makes the same report filter set applied for all pages.


bruntib (approved):

barnabasdomozi (approved):
LGTM

PR #4978, opened 2026-07-20
[FIX] Remove hash generation for malformed plists

Makkareal:
For historical reasons if a plist does not have bug hashes we used to generate them on the fly. We also opened the plist for writing which breaks storing from Bazel where these files are readonly.

Since we no longer accept non compliant plists it makes no sense to preserve this functionality. This also fixes the permission issue in Bazel.


web/tests/functional/cppcheck/0_hashes_test_proj/divide_zero_0_hashes.plist
Szelethus, line 1: The name could reflect that this is a non-compliant plist file. How about `divide_zero_cppcheck_incorrect.plist`?

web/tests/functional/cppcheck/test_cppcheck.py
```diff
@@ -96,11 +94,17 @@ def test_cppcheck_report_storage(self):
...
-        shutil.copytree(report_dir, temp_workspace)
+                                      'test_proj_code')
+
+        shutil.copytree(report_dir, temp_workspace,
+                        ignore=shutil.ignore_patterns
+                        ('divide_zero_cppcheck.plist'))
 
-        report_file = os.path.join(temp_workspace, 'divide_zero.plist')
+        report_file = os.path.join(temp_workspace,
+                                   'divide_zero_codechecker.plist')
```
Szelethus: I see what you are going for here, but I think its a tad inelegant. How about creating two result directories? Something like this:

```
tests/functional/cppcheck
├── correct_plist_proj
│   ├── cppcheck_result.plist
│   └── divide_zero.cpp
├── incorrect_plist_proj
│   ├── cppcheck_result_0_hashes.plist
│   └── divide_zero.cpp
└── test_cppcheck.pyp
```
Just an idea, not necesserily with the same names.

web/tests/functional/cppcheck/test_cppcheck.py
```diff
@@ -121,11 +125,43 @@ def test_cppcheck_report_storage(self):
...
+        store_cmd = [env.codechecker_cmd(), 'store', '--name', run_name,
+                     # Use the 'Default' product.
+                     '--url', env.parts_to_url(codechecker_cfg),
+                     temp_workspace]
+        # As the plist file contains 0 as Hash the expected
+        # behaviour is to terminate with error code 1.
+        result = subprocess.run(
+            store_cmd, encoding="utf-8",
+            errors="ignore", check=False)
+        self.assertEqual(result.returncode, 1)
```
Szelethus, line 170: We could do one better and check for the specific error message we just added.

Szelethus (changes requested):
The patch is great overall, some nitpicking.

Szelethus (approved):
LGTM

PR #4979, opened 2026-07-20
Introduce Personal Access Token creation date column

barnabasdomozi:
Previously, the personal_access_tokens table's
last_access column was not updated after the token's creation and therefore it reflected its creation date.
With this change, last_access column is now updated after each login, and a new column is added to save the creation date of the Personal Access Token.

The previous schema migration could not finish because the BackgroundTasks status Enum was missing a name, also added that in this patch.

### DIFF
--- a/web/server/codechecker_server/database/config_db_model.py
+++ b/web/server/codechecker_server/database/config_db_model.py
@@ -155,6 +155,7 @@ class PersonalAccessToken(Base):
     # List of group names separated by semicolons.
     groups = Column(String)
 
+    creation_date = Column(DateTime, nullable=False)
     last_access = Column(DateTime, nullable=False)
     expiration = Column(DateTime)
 
@@ -176,7 +177,8 @@ class PersonalAccessToken(Base):
         self.token = token
         self.description = description
         self.groups = groups
-        self.last_access = datetime.now()
+        self.creation_date = datetime.now()
+        self.last_access = self.creation_date
         self.expiration = self.last_access + timedelta(days=expiration)
 
 
@@ -277,7 +279,7 @@ class BackgroundTask(Base):
         # The job never started, or its execution was terminated due to a
         # system-level reason (such as the server's foced shutdown).
         "dropped",
-        ),
+        name="background_task_statuses"),
                     nullable=False,
                     default="enqueued",
                     index=True)
--- a/web/server/codechecker_server/migrations/config/versions/511b1b37de2e_add_pat_creation_date_column.py
+++ b/web/server/codechecker_server/migrations/config/versions/511b1b37de2e_add_pat_creation_date_column.py
@@ -1,0 +1,48 @@
+"""
+Add PAT creation date column
+
+Revision ID: 511b1b37de2e
+Revises:     5e1501dfd333
+Create Date: 2026-07-20 15:35:59.392866
+"""
+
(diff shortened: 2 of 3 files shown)


web/server/codechecker_server/database/config_db_model.py
```diff
@@ -176,7 +177,8 @@ def __init__(
         self.token = token
         self.description = description
         self.groups = groups
-        self.last_access = datetime.now()
+        self.creation_date = datetime.now()
+        self.last_access = self.creation_date
```
bruntib, line 181: Does the creation count as an access? I remember it as if other systems print something like "this token hasn't been used, yet".

web/server/codechecker_server/database/config_db_model.py
```diff
@@ -277,7 +279,7 @@ class BackgroundTask(Base):
         # The job never started, or its execution was terminated due to a
         # system-level reason (such as the server's foced shutdown).
         "dropped",
-        ),
+        name="background_task_statuses"),
```
bruntib, line 282: What does this `name` parameter mean? How is it related to this PR?

bruntib (changes requested):

web/server/codechecker_server/database/config_db_model.py
barnabasdomozi: It was the default behavior before this patch, and I kept it as is.
See the removed line above:
```
self.last_access = datetime.now()
```

web/server/codechecker_server/database/config_db_model.py
barnabasdomozi: See the commit message:
> The previous schema migration could not finish because the BackgroundTasks status Enum was missing a name, also added that in this patch.

web/server/codechecker_server/database/config_db_model.py
bruntib: Earlier the `last_access` was the creation date in fact, because its value never changed. If there would be any backward incompatibility (which there isn't, I think) then it's because now the `last_access` changes compared to previous CodeChecker versions. I think it's fair to say that a newly created personal access token has not been used at.

I'm rather worried about the Thrift interface, because the `lastAccess` is not an optional field. I don't know if it would be backward compatible to make it optional. Intuitively it seems compatible.

Also, wouldn't we like to expose the creation date and display them in the clients? (GUI, CLI) If not, then we should create a follow-up ticket to do it in the future.

barnabasdomozi: > I don't know if it would be backward compatible to make it optional.

It would not be backward compatible. Consider the following scenario:
Let's take a server that was upgraded to a newer release and now initially the `lastAccess` column is NULL.
At the same time, older clients still manage Personal Access Tokens via the CLI. However, due to their older Thrift specification, it requires a valid `string` as the `lastAccess` column and not a nullable value. Therefore, the older clients expect a valid `lastAccess` field when listing their Personal Access Tokens.

> Also, wouldn't we like to expose the creation date and display them in the clients? (GUI, CLI) If not, then we should create a follow-up ticket to do it in the future.

Currently, creation date is only used for administrative purposes only. We can extend it in the future, but that is out of scope of this patch.

bruntib (approved):

PR #4980, opened 2026-07-21
[fix] Eliminate silent query limits

bruntib:
Many report_server API functions have a limit parameter for limiting the number of results. When the parameter is 0 or None, then the MAX_QUERY_SIZE is used which is currently 500.

Silently limiting the results may cause practical problems, too. For example, when we filter on runs with a joker character (e.g. myrun*) then the results from the first 500 runs return only.

### DIFF
--- a/web/client/codechecker_client/cmd_line_client.py
+++ b/web/client/codechecker_client/cmd_line_client.py
@@ -132,8 +132,8 @@ def get_run_tag(client, run_ids: List[int], tag_name: str):
     """Return run tag information for the given tag name in the given runs."""
     run_history_filter = ttypes.RunHistoryFilter()
     run_history_filter.tagNames = [tag_name]
-    run_histories = client.getRunHistory(run_ids, None, None,
-                                         run_history_filter)
+    run_histories = client.getRunHistory(
+        run_ids, constants.MAX_QUERY_SIZE, None, run_history_filter)
 
     return run_histories[0] if run_histories else None
 
@@ -385,13 +385,13 @@ def check_deprecated_arg_usage(args):
                     'help.')
 
 
-def get_run_data(client, run_filter, sort_mode=None,
-                 limit=constants.MAX_QUERY_SIZE):
+def get_run_data(client, run_filter, sort_mode=None):
     """Get all runs based on the given run filter."""
 
     all_runs = []
 
     offset = 0
+    limit = constants.MAX_QUERY_SIZE
     while True:
         runs = runs = client.getRunData(run_filter, limit, offset, sort_mode)
         all_runs.extend(runs)
@@ -522,8 +522,8 @@ def parse_report_filter(client, args):
         # Handle tags (requires API call to resolve tag names to IDs)
         if 'tag' in args:
             run_history_filter = ttypes.RunHistoryFilter(tagNames=args.tag)
-            run_histories = client.getRunHistory(None, None, None,
-                                                 run_history_filter)
+            run_histories = client.getRunHistory(
+                None, constants.MAX_QUERY_SIZE, None, run_history_filter)
             if run_histories:
                 report_filter.runTag = [t.id for t in run_histories]
 
@@ -1555,7 +1555,7 @@ def handle_list_result_types(args):
         checkers = client.getCheckerCounts(run_ids,
                                            report_filter,
                                            None,
-                                           None,
+                                           constants.MAX_QUERY_SIZE,
                                            0)
 
         return dict((res.name, res.count) for res in checkers)
@@ -1588,7 +1588,7 @@ def handle_list_result_types(args):
     all_checkers = client.getCheckerCounts(run_ids,
                                            all_checkers_report_filter,
                                            None,
-                                           None,
+                                           constants.MAX_QUERY_SIZE,
                                            0)
     all_checkers_dict = dict((res.name, res) for res in all_checkers)
 
@@ -1846,7 +1846,8 @@ def handle_list_run_histories(args):
         runs = get_run_data(client, run_filter)
         run_ids = [r.runId for r in runs]
 
-    run_history = client.getRunHistory(run_ids, None, None, None)
+    run_history = client.getRunHistory(
+        run_ids, constants.MAX_QUERY_SIZE, None, None)
 
     if args.output_format == 'json':
         print(CmdLineOutputEncoder().encode(run_history))
--- a/web/server/codechecker_server/api/report_server.py
+++ b/web/server/codechecker_server/api/report_server.py
@@ -121,6 +121,13 @@ def verify_limit_range(limit):
 
     Query limit should not be larger than the max allowed value.
     Max is returned if the value is larger than max.
+
+    TODO: It would be nice in the long terms to raise an error when the limit
+    parameter is too big, otherwise the clients aren't notified about capping
+    the results.
+    The client implementation is providing a limit from CodeChecker 6.29. This
+    behavior could be strictly enforced when this version becomes quite
+    outdated.
     """
     max_query_limit = constants.MAX_QUERY_SIZE
     if not limit:
--- a/web/server/vue-cli/src/components/Report/ReportInfo/ReportInfo.vue
+++ b/web/server/vue-cli/src/components/Report/ReportInfo/ReportInfo.vue
@@ -220,7 +220,7 @@ import { useReviewStatus } from "@/composables/useReviewStatus";
 import { useSeverity } from "@/composables/useSeverity";
 import { ccService, handleThriftError } from "@cc-api";
 import ReportInfoItem from "@/components/Report/ReportInfo/ReportInfoItem";
-import { Checker } from "@cc/report-server-types";
+import { Checker, RunFilter } from "@cc/report-server-types";
 
 const props = defineProps({
   value: { type: Object, default: null }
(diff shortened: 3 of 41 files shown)


barnabasdomozi (changes requested):
In general, it's considered a good practice to have some hard limits on the amount of data and API endpoint can return from the database.
As an example, it's not uncommon to have production databases with more than 1 million rows in the `RunHistory` table. Without any enforced limit, the client could request all information at once which is an excessive amount of data.

> Silently limiting the results may cause practical problems, too. For example, when we filter on runs with a joker character (e.g. myrun*) then the results from the first 500 runs return only.

For which API endpoints is this a problem?

bruntib:
> In general, it's considered a good practice to have some hard limits on the amount of data and API endpoint can return from the database. As an example, it's not uncommon to have production databases with more than 1 million rows in the `RunHistory` table. Without any enforced limit, the client could request all information at once which is an excessive amount of data.

I think, that's why these functions have a `limit` parameter. Capping this number arbitrarily on the server side is not an intuitive behavior and leads to bugs like in this specific case.

> 
> > Silently limiting the results may cause practical problems, too. For example, when we filter on runs with a joker character (e.g. myrun*) then the results from the first 500 runs return only.
> 
> For which API endpoints is this a problem?

In this specific case the client queried the run IDs matching a run name pattern `run_name*`. The server returned only 500 IDs even if there was more. These were passed to `getRunResultCount()` (and to several other functions) through the `run_ids` parameter.

barnabasdomozi:
> I think, that's why these functions have a limit parameter. Capping this number arbitrarily on the server side is not an intuitive behavior and leads to bugs like in this specific case.

Yes, but if we remove the guard, essentially the client controls this limit parameter. If the limit set by the client is `0` or `None` or a large number, the server will try to return all rows with that limit. That means, in case of `getRunHistory` function, to load millions of rows of data into the server's memory and then it tries to return that excessive amount of data via Thrift / HTTP. 

> These were passed to getRunResultCount() (and to several other functions) through the run_ids parameter.

Maybe you meant a different function, but as of today, `getRunResultCount` has no enforced limit.

I think this server-side limit enforcement should be investigated on a case by case basis, and not removed from everywhere.

web/server/vue-cli/src/services/api/cc.service.js
```diff
@@ -58,33 +60,38 @@ class CodeCheckerService extends BaseService {
...
+    const runs = [];
+    let offset = 0;
 
-    return new Promise(resolve => {
-      this.getClient().getRunData(runFilter, null, 0, null,
-        handleThriftError(res => {
-          resolve(res);
-        }));
-    });
+    while (true) {
```
barnabasdomozi, line 71: Why do we need to return all runs in this case?

bruntib: This is just a helper function for those who need to get all runs. The users of this function would make several requests with increasing offset values either way.

The semantics of the code hasn't been changed. Even in the previous version, all runs were queried. So the question is: why do the callers of this function would need all runs?

The answer for this specific question is that `getRuns()` is used at 2 places:
- In `getSameReports()` the run names are queried that contain a specific bug hash. In the worst case scenario a bug hash is found in all runs. Since we don't have a paging functionality for displaying them, we show them all. It could be paginated in the future.
- In `ReportInfo.vue` this function is used for querying a run for a specific run ID. Yes, it's unnecessary to use this `getRuns()` in that case. I can fix that.

bruntib: Sorry, I was looking at a wrong branch. This function is used at several places. At those places all runs were queried with pagination even before.

web/server/codechecker_server/api/report_server.py
```diff
@@ -116,23 +116,17 @@ def comment_kind_to_thrift_type(kind):
...
-        return max_query_limit
-    if limit > max_query_limit:
-        LOG.warning('Query limit %d was larger than max query limit %d, '
-                    'setting limit to %d',
-                    limit,
-                    max_query_limit,
-                    max_query_limit)
-        limit = max_query_limit
-    return limit
+    if not limit or limit not in range(constants.MAX_QUERY_SIZE + 1):
```
barnabasdomozi: Consider the following `if` expression which is more Pythonic and easier to read:
```suggestion
    if limit not in range(1, constants.MAX_QUERY_SIZE + 1):
```

web/server/vue-cli/src/composables/useAnalysisInfo.js
```diff
@@ -39,14 +39,13 @@ function decideNegativeCheckerStatusAvailability(
     const filter = new RunFilter({
       ids: [ runId ]
     });
-    ccService.getClient().getRunData(filter, null, null, null,
-      handleThriftError(runDataList => {
-        if (runDataList.length !== 1) return;
+    ccService.getRunData(filter).then(runDataList => {
```
barnabasdomozi: Why was `.getClient()` dropped from here? `ccService` has no exported function named `getRunData`.

barnabasdomozi (changes requested):

web/server/vue-cli/src/components/Report/ReportInfo/ReportInfo.vue
```diff
@@ -273,8 +273,11 @@ onMounted(function() {
 });
 
 async function fetchRunName() {
-  const runs = await ccService.getRuns([ props.value.runId ]);
-  runName.value = runs[0].name;
+  const runFilter = new RunFilter({ ids: [ props.value.runId ] });
+  ccService.getClient().getRunData(runFilter, 1).then(runs => {
+    runName.value = runs[0].name;
+    console.warn(runName.value);
```
barnabasdomozi: Debug prints should be removed.

web/server/vue-cli/src/views/Statistics.vue
```diff
@@ -197,7 +197,7 @@ body {
 }
 
 .splitpanes__pane {
-  overflow-y: auto;
+  overflow-y: hidden;
```
barnabasdomozi, line 200: Why was this CSS property changed? Looks unrelated to this patch.

web/server/vue-cli/src/composables/useAnalysisInfo.js
```diff
@@ -39,14 +39,13 @@ function decideNegativeCheckerStatusAvailability(
     const filter = new RunFilter({
       ids: [ runId ]
     });
-    ccService.getClient().getRunData(filter, null, null, null,
-      handleThriftError(runDataList => {
-        if (runDataList.length !== 1) return;
+    ccService.getClient().getRunData(filter).then(runDataList => {
```
barnabasdomozi: The `limit` parameter is not supplied in this case and due to the new server-side design this will fail.

web/client/codechecker_client/cmd_line_client.py
```diff
@@ -1846,7 +1846,8 @@ def handle_list_run_histories(args):
         runs = get_run_data(client, run_filter)
         run_ids = [r.runId for r in runs]
 
-    run_history = client.getRunHistory(run_ids, None, None, None)
+    run_history = client.getRunHistory(
```
barnabasdomozi, line 1849: This change breaks backward compatibility. Many users still rely on older CodeChecker clients which will still pass `None` as the `limit` parameter to the server. Due to this new server design, the API requests e.g. `getRunHistory` will fail for these older clients.

barnabasdomozi (changes requested):

barnabasdomozi:
The issue with this new enforcement is that older CodeChecker clients will still rely on passing `None` as the `limit` parameter and therefore these API endpoints will break for those older clients.

web/server/vue-cli/src/views/Statistics.vue
bruntib: This really is unrelated. It eliminates double scroll bar in the statistics view. Should I place this to a separate commit?

bruntib:
I reverted the `verify_limit_range()` function and added a TODO message about the danger of this implementation. Now the old clients work as before.

web/server/codechecker_server/api/report_server.py
```diff
@@ -1543,7 +1547,7 @@ def __add_comment(self, bug_id, message, kind=CommentKindValue.USER,
     def getRunData(self, run_filter, limit, offset, sort_mode):
         self.__require_view()
 
-        limit = verify_limit_range(limit)
```
barnabasdomozi, line 1546: The return value of `verify_limit_range` is not used anywhere. This means clients can set arbitrary limits.
This comment applies to all cases where this function is used.

web/server/vue-cli/src/views/Statistics.vue
barnabasdomozi: We can keep this as is.

web/server/codechecker_server/api/report_server.py
```diff
@@ -121,6 +121,10 @@ def verify_limit_range(limit):
 
     Query limit should not be larger than the max allowed value.
     Max is returned if the value is larger than max.
+
```
barnabasdomozi, line 124: Add a comment here that the CodeChecker cmd client is now patched to not pass `None` as the `limit` parameter. This way, in the future we could drop this legacy serverside limit verification and apply the enforcement.

barnabasdomozi (changes requested):

barnabasdomozi (approved):
LGTM

PR #4981, opened 2026-07-21
[gui] Fix statistics table coloring

gulyasgergely902:
This change fixes the coloring issues of the statistics tables. All tables are now colored as it was before the update without the shifted "Total" row. The guideline statistics table highlights the rows only with outstanding reports greater than 0.


barnabasdomozi (commented):
Is the `components/Statistics/style.scss` file still used by other components or can be deleted?

dkrupp (dismissed):
Can you please add the same coloring ot the checker coverage table too?
that table is generated very similarly to the guidelinestatistics table.

A lot of copy paste code unfortunately.

bruntib (approved):

Issue #4090, opened 2023-11-20
Uplift the nodejs install in the server Dockerfile


vodorok:
When building the docker image, the following warning message appears:

================================================================================
================================================================================

                           SCRIPT DEPRECATION WARNING                    

  
  This script, located at https://deb.nodesource.com/setup_X, used to
  install Node.js is deprecated now and will eventually be made inactive.

  Please visit the NodeSource distributions Github and follow the
  instructions to migrate your repo.
  https://github.com/nodesource/distributions

  The NodeSource Node.js Linux distributions GitHub repository contains
  information about which versions of Node.js and which Linux distributions
  are supported and how to install it.
  https://github.com/nodesource/distributions

                          SCRIPT DEPRECATION WARNING

================================================================================
================================================================================

TO AVOID THIS WAIT MIGRATE THE SCRIPT
Continuing in 60 seconds (press Ctrl-C to abort) ...

## Installing the NodeSource Node.js 16.x repo...

As per instructions in the warning message, the nodejs installation method should be uplifted.

bruntib:
NodeJS has been upgraded to version 24.

PR #4984, opened 2026-07-22
Uplift the nodejs install in the server Dockerfile

hornokattila:
## Addresses #4090: Uplift Node.js installation in server Dockerfile

Replace the deprecated `deb.nodesource.com/setup_X.x` shell script with the new NodeSource signed-repository method. The old `setup_X.x` scripts are deprecated by NodeSource, show a 60-second warning during build, and will eventually be removed entirely. The new method is their only supported approach. 

Also upgrades Node.js from 20 to 22 LTS.

### Changes

- **`web/docker/Dockerfile`** — Replaced `curl | bash` install with GPG keyring + apt sources list approach. Node version is now a `NODE_MAJOR` build arg for easy future upgrades.
- **`docs/README.md`** — Updated developer setup instructions to match.


bruntib:
I'm sorry to miss this PR. We have merged another one that upgrades to Node 24: https://github.com/Ericsson/codechecker/pull/5029

PR #4987, opened 2026-07-25
[fix][report-converter] Honour trim-path-prefix in sarif export

### LINKED ISSUE #4659: Sarif output ignores trim-path-prefix
**Describe the bug**
The artifactLocation in the sarif report does not strip the path prefix.

**CodeChecker version**
6.26.0

**To Reproduce**

CodeChecker parse report --trim-path-prefix --export sarif  --output report.sarif

**Expected behaviour**

Paths in sarif output is relative instead of absolute

 - OS: Linux
 - Browser: N/A
 - Version: N/A

**Additional context**

This looks to need similar fix as https://github.com/Ericsson/codechecker/pull/4387

Sanjays2402:
Closes #4659

The sarif exporter built `artifactLocation` URIs from `report.file.original_path`, which always holds the untrimmed path, so `CodeChecker parse --trim-path-prefix ... --export sarif` still emitted absolute paths. Using `report.file.path` returns the trimmed path once `Report.trim_path_prefixes()` has run and the original path otherwise — the same fix applied to the HTML export in #4387.

Added a regression test covering both the primary result location and the code flow locations; it fails on master with `file:///home/user/project/src/main.c` and passes with the change.

### DIFF
--- a/tools/report-converter/codechecker_report_converter/report/parser/sarif.py
+++ b/tools/report-converter/codechecker_report_converter/report/parser/sarif.py
@@ -349,7 +349,7 @@ class Parser(BaseParser):
             "locations": [{
                 "physicalLocation": {
                     "artifactLocation": {
-                        "uri": f"file://{report.file.original_path}"
+                        "uri": f"file://{report.file.path}"
                     },
                     "region": {
                         "startLine": report.line,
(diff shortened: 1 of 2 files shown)


dkrupp (approved):
LGTM.

Thanks for this PR!

PR #4994, opened 2026-07-29
Use PrettyTable library for human-readable table output

Muteeb-Haider:
Replace the custom to_table() implementation in twodim.py with the PrettyTable library (prettytable~=3.0). The old implementation produced tables that fell apart on long lines. PrettyTable handles long content gracefully with proper box-drawing borders and correct column alignment.

- Rewrite to_table() using PrettyTable with SINGLE_BORDER style
- Left-align all columns to match previous formatting convention
- Support separate_head and headerless table modes
- Add prettytable~=3.0 to analyzer/requirements.txt
- Add prettytable~=3.0 to report-converter dev requirements

Other output formats (csv, json, rows, dictlist) are unchanged.


analyzer/tests/functional/analyze/test_analyze.py
```diff
@@ -536,7 +536,7 @@ def test_reproducer_postprocesses_failures(self):
         # The key bug: with --generate-reproducer, postprocessing was skipped
         # so parse would show 0 processed files. After the fix, it should
         # show 1 processed file.
-        self.assertIn("Number of processed analyzer result files | 1", out)
+        self.assertIn("Number of processed analyzer result files", out)
```
bruntib: Please, keep the `| 1` part, because this test checks the number of analyzer result files.

bruntib (changes requested):
Thanks for the improvement. It's nice that we can use an existing library for table creation.
The main purpose of this development would be to make a table fit to the screen. For example, if I store the reports to the server and query them by the following command, then the table breaks to multiple lines if the columns are too wide:
```
CodeChecker cnd results <run name>
```
I'm not sure if this library is able to detect the screen width. Also, the proportion of the columns could be set reasonably. For example, the "severity" column is always short. But we shouldn't hard-code the column types. But we could compute the maximum width of each column and the long ones could be shortened somehow.

Thank you!

analyzer/tests/functional/analyze/test_analyze.py
Muteeb-Haider: Thanks for the feedback! I've addressed all the points you raised.
   The table now fits to the terminal width by detecting it via
  shutil.get_terminal_size(), falling back to 80 columns when output is piped.
  Column widths are computed from the actual data  no column types are
  hard-coded. Short columns like "severity" naturally keep their full width, and
  only the wider ones get shortened proportionally.
    For the test, I restored the | 1 check as you requested. Since PrettyTable
  uses Unicode │ instead of ASCII |, the assertion is now assertIn("Number of
  processed analyzer result files │ 1", out), which checks both the label and
  the value

tools/report-converter/codechecker_report_converter/twodim.py
```diff
@@ -94,59 +97,197 @@ def to_rows(lines: Iterable[str]) -> str:
...
-                            "columns than the others")
-        if i == 0 and separate_head:
-            str_parts.append("-" * (sum(widths) + 3 * (len(widths) - 1)))
-        if separate_footer and i == len(lns) - 2:
-            str_parts.append("-" * (sum(widths) + 3 * (len(widths) - 1)))
+    # Detect the current terminal width. honour the COLUMNS environment
+    # variable first (allows the user to override), then fall back to the
+    # OS-reported terminal size, and finally to 80 columns when stdout is
+    # redirected to a file or pipe.
+    import os  # pylint: disable=import-outside-toplevel
```
bruntib: Import libs at the top of the file.

tools/report-converter/codechecker_report_converter/twodim.py
```diff
@@ -94,59 +97,197 @@ def to_rows(lines: Iterable[str]) -> str:
...
-    str_parts.append("-" * (sum(widths) + 3 * (len(widths) - 1)))
+    if separate_footer and len(data_rows) > 1:
+        table = _make_table(field_names, data_rows, show_header,
+                            hrules=HRuleStyle.ALL)
+    else:
+        table = _make_table(field_names, data_rows, show_header)
 
-    return '\n'.join(str_parts)
+    nat_widths = _compute_natural_widths(field_names, data_rows, show_header)
+    return _fit_table_to_width(table, nat_widths, terminal_width)
```
bruntib: The `_compute_natural_widths()` should be called by `_fit_table_to_width()`, because `nat_widths` isn't used elsewhere.

web/tests/functional/store/test_store.py
```diff
@@ -511,5 +511,4 @@ def test_store_stats(self):
         _, out, _ = _call_cmd(store_cmd)
         # There are 9 individual reports, but only 6 unique.
         # The statistics should only print the unique ones.
-        self.assertIn("Number of analyzer reports                       | 6",
-                      out)
+        self.assertIn("Number of analyzer reports", out)
```
bruntib: Keep the number, because the goal of this test is to check the number.

bruntib (changes requested):
This is a really nice work, thank you! I have a few minor comments. After fixing those we can merge this PR.

Side note: I'm just thinking about execution in Jenkins. That is a popular platform for CodeChecker integration. In that environment there is no "terminal width", so it defaults to 80. It might make it difficult to visualize big tables there. Users can set the `COLUMNS` environment variable as a workaround, but I'm thinking of detecting Jenkins based on Jenkins-specific environment variables as some heuristics. It shouldn't be done in this PR, I'm just writing this note for the future.

bruntib (approved):
Thank you for the development! It's much more convenient to read this output.

PR #4995, opened 2026-07-29
Fix: recognize --std=/--stdlib= double-dash compiler flags (#4926)

### LINKED ISSUE #4926: simple example with c++23 std::print is not found, error with double dash vs single dash
**Describe the bug**
I tried a simple example with g++ and c++23:

makefile

```
all:
	g++-16 --std=c++23 main.cpp
``` 
main.cpp

```
#include <print>
int main() {
  std::print("oi");
  return 0;
}
```

This works fine with "make".

However, when executing CodeChecker on linux with gcc and clang-tidy installed, it will fail!

	CodeChecker log --build "make" --output ./compile_commands.json
	CodeChecker analyze ./compile_commands.json --enable sensitive --output ./reports

ERROR: Analyzing main.cpp with clangsa  failed!

The reason, it took some time to understand, is that double dash in --std=c++23 is the problem! It only works with -std=c++23
Maybe it's worth putting that on documentation, or trying to detect and fixing that!
In any case, I couldn't find any issue on internet mentioning this fact, so I'm registring this here.

**CodeChecker version**

```
CodeChecker version
[INFO 2026-06-27 14:21:12] - CodeChecker analyzer version:
Base package version | 6.28.1                                  
Package build date   | 2026-06-17T14:16                        
Git commit ID (hash) | eb60480ae1858da2b3a7490939e673cc4885ec20
Git tag information  | 6.28.1                                  

[INFO 2026-06-27 14:21:12] - CodeChecker web version:
Base package version                | 6.28.1                                  
Package build date                  | 2026-06-17T14:16                        
Git commit ID (hash)                | eb60480ae1858da2b3a7490939e673cc4885ec20
Git tag information                 | 6.28.1                                  
Server supported Thrift API version | 6.71                                    
Client Thrift API version           | 6.71                                    
```

and my compilers:

```
 CodeChecker analyzers --details
 clangsa /usr/lib/llvm-21/bin/clang       21.1.8
 clang-tidy /usr/lib/llvm-21/bin/clang-tidy  21.1.8
 cppcheck /usr/bin/cppcheck                2.19.0
 gcc /usr/bin/x86_64-linux-gnu-g++-15 15.2.0
 infer                                  NOT FOUND
[WARNING 2026-06-27 14:21:44] - Can't analyze with 'infer': Failed to detect analyzer binary!
```

**To Reproduce**

Already explained above.

**Expected behaviour**

I expected it to work, or at least TO WARN that double dash notation is UNSUPPORTED! (it could be when generating the compile commands json, perhaps!).

 - OS: Ubuntu 26.04
 
**Additional context**
It could help new users like me.

mmido6039:
Fixes #4926

## Problem
`CodeChecker log` + `CodeChecker analyze` fail when a build command uses
the double-dash spelling of the standard flag, e.g. `g++ --std=c++23`.
GCC accepts both `-std=` and `--std=` as equivalent, but CodeChecker's
flag-recognition regex only matched the single-dash form. As a result,
`--std=c++23` was silently dropped while translating the build command
into the analyzer invocation, and CodeChecker fell back to querying the
compiler's own default standard (e.g. `gnu++17`), causing spurious
"not found" errors for anything newer.

## Root cause
In `analyzer/codechecker_analyzer/buildlog/log_parser.py`:
- `COMPILE_OPTIONS_LIST` included `'-std='` / `'-stdlib='`, matched via
  `re.match()`, which anchors at the start of the string — so
  `--std=c++23` (starting with `--s`, not `-s`) never matched.
- The same anchoring issue existed in the separate pattern used inside
  `filter_compiler_includes_extra_args()`.

Verified with `--verbose debug_analyzer`: the assembled clangsa
invocation contained `-std=gnu++17` with no trace of the user's
`--std=c++23` at all (not duplicated, just dropped).

## Fix
Made both patterns tolerant of one or two leading dashes:
`-std=` → `-{1,2}std=`, `-stdlib=` → `-{1,2}stdlib=`.

## Testing
- Added two regression tests in `test_log_parser.py` covering both the
  `COMPILE_OPTIONS` match and `filter_compiler_includes_extra_args()`.
- Ran the full `test_log_parser.py` suite (25 tests) — all pass, no
  regressions.
- Reproduced the original bug end-to-end with a minimal project using
  `g++ --std=c++23` and `<print>`, confirmed it fails before the fix
  and `clangsa analyzed main.cpp successfully` after the fix.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
@@ -594,13 +594,14 @@ class ClangSA(analyzer_base.SourceAnalyzer):
                     self.buildaction.arch != "":
                 analyzer_cmd.extend(["-arch", self.buildaction.arch])
 
-            if not has_flag('-std', analyzer_cmd) and \
-                    self.buildaction.compiler_standard != "":
-                analyzer_cmd.append(self.buildaction.compiler_standard)
-
             analyzer_cmd.extend(config.analyzer_extra_arguments)
 
             analyzer_cmd.extend(self.buildaction.analyzer_options)
+
+            if not has_flag('-std', analyzer_cmd) and \
+                    not has_flag('--std', analyzer_cmd) and \
+                    self.buildaction.compiler_standard != "":
+                analyzer_cmd.append(self.buildaction.compiler_standard)
 
             analyzer_cmd.extend(prepend_all(
                 '-isystem' if config.add_gcc_include_dirs_with_isystem else
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -224,8 +224,8 @@ COMPILE_OPTIONS_LIST = [
     '-pedantic',
     '-O[1-3]',
     '-Os',
-    '-std=',
-    '-stdlib=',
+    '-{1,2}std=',
+    '-{1,2}stdlib=',
     '-f',
     '-m',
     '-Wno-',
@@ -293,7 +293,8 @@ def filter_compiler_includes_extra_args(compiler_flags):
     # If these options are present in the original build command, they must
     # be forwarded to get_compiler_includes and get_compiler_defines so the
     # resulting includes point to the target that was used in the build.
-    pattern = re.compile('-m(32|64)|-std=|-stdlib=|-nostdinc|--driver-mode=')
+    pattern = re.compile(
+        '-m(32|64)|-{1,2}std=|-{1,2}stdlib=|-nostdinc|--driver-mode=')
     extra_opts = list(filter(pattern.match, compiler_flags))
 
     pos = next((pos for pos, val in enumerate(compiler_flags)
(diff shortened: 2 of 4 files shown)


bruntib (changes requested):
Thank you for the fix, it's a really nice work!

At the clangsa and cppcheck analyzers the default standard version is added with the original one. The clang-tidy seems fine in this regard. Also, I haven't checked infer and gcc. Could you, please, check the assembly of the analyzer commands, why both `-std` and `--std` are added? In case of clangsa, for example, the implicit `-std` is added here: https://github.com/Ericsson/codechecker/blob/efbf12883f62ad0d061e0cae98a172bfcd2530e2/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py#L597
Thank you!

mmido6039:
Thanks for catching that! You were right - for clangsa, the has_flag('-std', analyzer_cmd) check was running before self.buildaction.analyzer_options (which contains the original --std=c++23) got merged into the command, so it could never see it and always appended the implicit default. clang-tidy didn't have this bug because it checks in the right order already. I checked cppcheck directly with --verbose debug_analyzer and it wasn't actually duplicating - its default-standard append is currently commented out in the code. Haven't looked at infer/gcc yet, happy to check those too if useful. Pushed a fix that reorders the check in clangsa to match clang-tidy, plus regression tests covering both cases (explicit --std= isn't duplicated, and the implicit default still gets added when no -std is given at all).

bruntib (approved):

bruntib:
Perfect, thank you for the fix!

PR #4996, opened 2026-07-30
Add report-converter --example flag and EXAMPLE_CMD interface (#4992)

### LINKED ISSUE #4992: report-converter prints help about the input's format
`report-converter` tool converts the output of an analyzer tool to plist format which is a common representation for CodeChecker commands. However, analyzers can print their results in different formats like json, plaintext, etc. Users need to read the documentation to get information about how to gather an output from these analyzers in a format that `report-converter` is able to consume. However, this documentation goes out of sync easily: https://github.com/Ericsson/codechecker/blob/master/docs/tools/report-converter.md

For example, at the time of this ticket's creation, Ruff parser section is missing. The solution to this problem could be extending the interface in `report-converter` so it is forced new parser creators to display the corresponding documentation to the users.

mmido6039:
## Summary

Fixes #4992 by adding a standardized way for `report-converter` analyzers to expose example commands for generating compatible input.

### What changed

* Added the `EXAMPLE_CMD` interface field to `AnalyzerResultBase`.
* Added `EXAMPLE_CMD` to the Ruff analyzer parser.
* Added a new `--example <ANALYZER>` CLI option to `report-converter`.
* `--example` works independently of the other required CLI arguments, similarly to `--help` and `--version`.
* Added documentation for Ruff and the new `--example` functionality.
* Added a documentation note recommending `--example` as the up-to-date source for analyzer example commands.
* Added CLI/interface tests covering:

  * An analyzer with an `EXAMPLE_CMD`.
  * An analyzer without an `EXAMPLE_CMD`.
  * Invalid analyzer names.
  * Enforcement that newly added analyzer parsers define `EXAMPLE_CMD`.

### Validation

Manually verified:

* `report-converter --example ruff` prints the expected example command and exits successfully.
* `report-converter --example pylint` handles an analyzer without an example gracefully without crashing.
* Invalid analyzer names are rejected cleanly by argparse.

All 5 newly added tests pass.

The broader `report-converter` test suite also contains pre-existing/environment-related failures unrelated to this change, including fixture corruption and package metadata mismatches. These were verified against the unmodified codebase.

### Motivation

This provides a consistent interface for analyzer-specific example commands and prevents new analyzer parsers from being added without documenting how their compatible input can be generated.

### DIFF
--- a/docs/tools/report-converter.md
+++ b/docs/tools/report-converter.md
@@ -22,6 +22,7 @@ a CodeChecker server.
   * [TSLint](#tslint)
   * [Golint](#golint)
   * [Pyflakes](#pyflakes)
+  * [Ruff](#ruff)
   * [PVS-Studio](#PVS-Studio)
   * [Markdownlint](#markdownlint)
   * [Coccinelle](#coccinelle)
@@ -132,6 +133,13 @@ Supported analyzers:
 </details>
 
 ## Supported analyzer outputs
+For an always up-to-date usage example for a given analyzer straight from
+the source, run:
+```sh
+report-converter --example <TYPE>
+# e.g. report-converter --example ruff
+```
+
 ### Sanitizers
 #### Undefined Behaviour Sanitizer
 [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)
@@ -407,6 +415,28 @@ report-converter -t pyflakes -o ./codechecker_pyflakes_reports ./pyflakes_report
 
 # Store the Pyflakes reports with CodeChecker.
 CodeChecker store ./codechecker_pyflakes_reports -n pyflakes
+```
+
+### Ruff
+[Ruff](https://docs.astral.sh/ruff/) is an extremely fast static analysis
+tool (linter) for `Python` code.
+
+report-converter only accepts a `json` output file from Ruff, so you must
+generate one.
+
+The following example shows you how to run Ruff and store the results
+found by Ruff to the CodeChecker database.
+
+```sh
+# Run ruff and generate a json output file.
+ruff check --output-format json /path/to/my/project > ./ruff_reports.json
+
+# Use 'report-converter' to create a CodeChecker report directory from the
+# analyzer result of ruff.
+report-converter -t ruff -o ./codechecker_ruff_reports ./ruff_reports.json
+
+# Store the ruff reports with CodeChecker.
+CodeChecker store ./codechecker_ruff_reports -n ruff
 ```
 
 ### PVS-Studio
--- a/docs/tools/report-converter.md
+++ b/docs/tools/report-converter.md
@@ -22,6 +22,7 @@ a CodeChecker server.
   * [TSLint](#tslint)
   * [Golint](#golint)
   * [Pyflakes](#pyflakes)
+  * [Ruff](#ruff)
   * [PVS-Studio](#PVS-Studio)
   * [Markdownlint](#markdownlint)
   * [Coccinelle](#coccinelle)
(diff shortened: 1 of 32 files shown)


docs/tools/report-converter.md
```diff
@@ -132,6 +133,13 @@ Supported analyzers:
 </details>
 
 ## Supported analyzer outputs
+The list below can go out of sync with the actual code. For an always
```
bruntib: I think, the first sentence is unnecessary here.

docs/tools/report-converter.md
```diff
@@ -409,6 +417,27 @@ report-converter -t pyflakes -o ./codechecker_pyflakes_reports ./pyflakes_report
 CodeChecker store ./codechecker_pyflakes_reports -n pyflakes
 ```
 
+### Ruff
+[Ruff](https://docs.astral.sh/ruff/) is an extremely fast static analysis
+tool (linter) for `Python` code.
+
+The recommended way of running Ruff is to generate a `json` output file.
```
bruntib: report-converter accepts only `json` from rust. So, it is not simply "recommended", but "required" by report-converter.

tools/report-converter/tests/unit/test_cli.py
```diff
@@ -0,0 +1,115 @@
...
+from codechecker_report_converter import cli
+
+
+# TOOL_NAMEs of analyzers that do not yet provide an EXAMPLE_CMD.
+#
+# This list must not grow: any newly added parser is required to define a
+# non-empty EXAMPLE_CMD (see AnalyzerResultBase.EXAMPLE_CMD). Existing
+# parsers on this list predate that requirement and should have one added
+# as a follow-up; when a parser is updated, remove it from this list.
+LEGACY_ANALYZERS_WITHOUT_EXAMPLE_CMD = {
```
bruntib: We should include the examples for all analyzer tools in this PR.
In fact, I'm thinking if we could force the existence of `EXAMPLE` member variable for each analyzer. Would you think, that it could rather be a virtual function? Or maybe the test could fail if any of the supported analyzers don't fill in the `EXAMPLE` variable.

bruntib (changes requested):

docs/tools/report-converter.md
mmido6039: Removed.

docs/tools/report-converter.md
mmido6039: Good point, fixed - changed to "report-converter only accepts a json output file from Ruff, so you must generate one."

tools/report-converter/tests/unit/test_cli.py
mmido6039: Done - all 29 supported analyzers now have an EXAMPLE_CMD, no exemptions left. I removed the legacy list entirely and simplified the test to just assert every supported analyzer has a non-empty EXAMPLE_CMD.

On the virtual/abstract property idea: I'd lean against making it a hard abstract member (e.g. @abstractmethod) enforced at class-definition or instantiation time. The risk is that if it's ever forgotten on a new parser, the failure mode is the whole report-converter tool crashing on import/use of that specific analyzer in production, which is a worse experience than catching it in CI. A test that iterates supported_converters and asserts EXAMPLE_CMD is non-empty (which is what I did) gives the same guarantee - a new parser without one fails CI immediately - but fails safely in review rather than at runtime for a user. Happy to switch to the abstract-property approach if you'd still prefer it though.

bruntib (changes requested):
Please, don't remove the `.gitignore` file. Otherwise looks good, thank you!

tools/report-converter/tests/unit/test_cli.py
bruntib: Alright, forcing the documentation through the CI is a good idea. Thank you!

mmido6039: Glad that works for you!, also my mistake for the gitignore file.

bruntib (approved):

PR #5001, opened 2026-08-02
Recognize --target=<value>/-target=<value> equals-sign form (#1158)

### LINKED ISSUE #1158: Multiple target build with Clang is not handled
We do not differentiate entries of the compile db for the same source file with different `-target` switches if the compiler is Clang.
See `log_parser.py` / `get_compiler_target()`.

mmido6039:
Fixes #1158

## Problem
CodeChecker did not differentiate compile database entries with
different Clang `-target` switches, as described in the issue.

## Root cause
`__get_target()` in log_parser.py only matched the space-separated
form of the target flag ('-target <value>' / '--target <value>').
The equals-sign form ('--target=<value>' / '-target=<value>'), a
very common cross-compilation syntax, was not recognized at all and
fell through unhandled. As a result, the explicitly requested target
was silently dropped, and the compiler's own native default target
was used instead (verified directly: `clang++ --target=aarch64-linux-gnu`
ended up analyzed as `x86_64-pc-linux-gnu` on the test machine).

## Fix
Extended `__get_target()` to also recognize and strip the
'--target='/'-target=' prefix form, in addition to the existing
space-separated form.

## Testing
Added a regression test (test_target_equals_sign_form) with a new
fixture covering both the equals-sign and space-separated forms for
the same source file, asserting the exact resolved target for each -
the existing similarly-named test only asserted `len(target) > 0`,
which would not have caught this regression. All tests in
test_log_parser.py pass; pycodestyle clean.

## Known follow-up (out of scope for this PR)
The compiler-info caching key (used to cache implicit include paths/
standard/target per compiler+language+flags) also does not account
for an explicit target, so two different `--target=` values on the
same compiler binary could still share cached compiler_includes.
Fixing that touches the persisted compiler_info.json cache format
and felt like a separate, riskier change - happy to follow up in a
second PR if useful.

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -226,6 +226,7 @@ COMPILE_OPTIONS_LIST = [
     '-Os',
     '-{1,2}std=',
     '-{1,2}stdlib=',
+    '-{1,2}target=',
     '-f',
     '-m',
     '-Wno-',
--- a/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzers/clangsa/analyzer.py
@@ -586,10 +586,6 @@ class ClangSA(analyzer_base.SourceAnalyzer):
             if not has_flag('-x', analyzer_cmd):
                 analyzer_cmd.extend(['-x', compile_lang])
 
-            if not has_flag('--target', analyzer_cmd) and \
-                    self.buildaction.target != "":
-                analyzer_cmd.append(f"--target={self.buildaction.target}")
-
             if not has_flag('-arch', analyzer_cmd) and \
                     self.buildaction.arch != "":
                 analyzer_cmd.extend(["-arch", self.buildaction.arch])
@@ -597,6 +593,11 @@ class ClangSA(analyzer_base.SourceAnalyzer):
             analyzer_cmd.extend(config.analyzer_extra_arguments)
 
             analyzer_cmd.extend(self.buildaction.analyzer_options)
+
+            if not has_flag('--target', analyzer_cmd) and \
+                    not has_flag('-target', analyzer_cmd) and \
+                    self.buildaction.target != "":
+                analyzer_cmd.append(f"--target={self.buildaction.target}")
 
             if not has_flag('-std', analyzer_cmd) and \
                     not has_flag('--std', analyzer_cmd) and \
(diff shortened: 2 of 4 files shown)


analyzer/codechecker_analyzer/buildlog/log_parser.py
```diff
@@ -892,6 +892,14 @@ def __get_target(flag_iterator, details):
         details['compilation_target'] = flag_iterator.item
         return True
 
+    for prefix in ['--target=', '-target=']:
+        if flag_iterator.item.startswith(prefix):
+            details['compilation_target'] = \
+                flag_iterator.item[len(prefix):]
+            details['analyzer_options'].append(flag_iterator.item)
```
bruntib: In the first branch where `-target` is separated from its parameter, the flag is not added to `analyzer_options`. What is the motivation of adding it here? Either both or none of the branches should have this logic. I vote for not adding it to `analyzer_options`, because this was the original behavior.

bruntib (changes requested):
This is a good catch and good bugfix. However, the original ticket (https://github.com/Ericsson/codechecker/issues/1158) is about something else. Maybe I'm wrong, but I interpret the ticket in a way that there are two compilation actions: one with `-target=<target1>` and another with `-target=<target2>`. I understand like CodeChecker wouldn't analyze them separately. However, this is not the case, so let's accept this PR as a fix for this ticket.

analyzer/codechecker_analyzer/buildlog/log_parser.py
mmido6039: you're right that the two branches were inconsistent - that was left over from an earlier attempt at fixing a CI failure and I should have cleaned it up before pushing.

I looked into this more and I don't think either branch should append to analyzer_options directly. The reason the equals-sign form needs special handling at all is that it was being silently dropped entirely on the GCC compile-command path (COMPILE_OPTIONS_LIST doesn't include a target pattern), unlike the space-separated form or the Clang path (which has a catch-all that already preserves unrecognized flags). So instead of touching __get_target() at all, I added '-{1,2}target=' to COMPILE_OPTIONS_LIST, which lets the equals-sign form get preserved the same way it already naturally survives elsewhere - no special-casing inside __get_target, no inconsistency between branches.

While digging into this I also found the actual duplicate-flag risk this creates: in clangsa/analyzer.py, the '--target' check ran before analyzer_options got merged into the command, so it couldn't detect an already-present explicit target and would append a wrong default alongside it. Fixed that by reordering it to run after the merge, same pattern as the --std fix from #4926.

Verified against a fresh clone of current master: all existing tests pass including test_keep_clang_flags and test_target_parsing_clang, plus new regression tests for both issues. Pushed the corrected version.

bruntib (changes requested):
Please, don't remove `.gitignore`.

Also, could you, please, add a test for the "space" scenario, too? I mean this: `clang -target aarch64-linux-gnu main.c`.

Thank you!

mmido6039:
Sorry about that - restored .gitignore back to the original, and added the space-separated test case as requested (test_target_space_separated_form, covering exactly `clang -target aarch64-linux-gnu main.c`). Verified all 54 relevant unit tests pass, including test_keep_clang_flags and test_target_parsing_clang.

bruntib (approved):

PR #5005, opened 2026-08-04
Add OWASP Top 10 2025 rules

gulyasgergely902:
Add the OWASP Top 10 2025 rules to the guidelines.


config/labels/analyzers/clang-tidy.json
```diff
@@ -144,7 +144,10 @@
     "android-cloexec-open": [
       "doc_url:https://clang.llvm.org/extra/clang-tidy/checks/android/cloexec-open.html",
       "guideline:owasp-top-10-2021",
+      "guideline:owasp-top-10-2025",
       "owasp-top-10-2021:owasp-A01-2021",
+      "owasp-top-10-2025:owasp-A01-2025",
+      "cwe-mapping:cwe-668",
```
dkrupp: cwe-mapping sounds like very implementation detail oriented.

Let's call it "cwe-vulnerabilities" instead.

config/labels/descriptions.json
```diff
@@ -20,6 +20,7 @@
     "sei-cert-cpp": "https://cmu-sei.github.io/secure-coding-standards/sei-cert-cpp-coding-standard/",
     "misra-c": "https://www.misra.org.uk/",
     "memory-safety": "https://cwe.mitre.org/data/definitions/1399.html",
-    "owasp-top-10-2021": "https://owasp.org/Top10/2021/"
+    "owasp-top-10-2021": "https://owasp.org/Top10/2021/",
+    "owasp-top-10-2025": "https://owasp.org/Top10/2025/"
```
dkrupp, line 24: add the new cwe-vulnerabilities guideline also to the descriptions.

"cwe-vulnerabilities" : "Maps checkers to CWE vulnerabilities https://cwe.mitre.org/"

dkrupp (changes requested):
I think this is great.
I think it would be better to call the "cwe-mapping" guideline as "cwe-vulnerabilities" instead. Please see my detailed commenets. otherwise good!

dkrupp (commented):
Please also include the description of the rules
e.g.
A01:2025 - Broken Access Control

dkrupp (changes requested):
1)

Please check the possible additional clang static analyzer cwe pairings.

For example

for  cwe-78 os command injection there is the generictaint checker
cwe-416 Use After Free  https://clang.llvm.org/docs/analyzer/checkers.html#cplusplus-newdelete-c
and  https://clang.llvm.org/docs/analyzer/checkers.html#unix-malloc-c

please note that "user after free" is word by word can be found in the clang static analyzers' documentation, so the llms should have found it. please check.

2)
Also: There are other cwe labels in the label files (added with the memory safety guidelines) which are not added to this new cwe-vulnerabilities for completeness?
for example: cwe-562

Interestingly cwe-416 if  you access them from the memory safety guideline already has these associations: cplusplus.newdeleteleaks, unix.malloc, cppcheck-deallocuse

I suspect codechecker does not handle properly the case when one rule (e.g. cwe-416) is included in more than 1 guideline. Please check this too.

dkrupp (approved):
LGTM

PR #5006, opened 2026-08-04
[config] Checker label fixes and 6.29 profile update

gamesh411:
clang renamed alpha.unix.cstring.UninitializedRead to
unix.cstring.UninitializedRead since the last release. Add an entry for
the new name, and give the alpha one the same EXP33-C mapping, since we
still support the clang versions that only have it.

Three misspellings left the affected checkers without a severity and
outside the CWE Top 25 guideline, with nothing to signal it: a
misspelled key is silently ignored, and so is a guideline name that
descriptions.json does not define.

alpha.cplusplus.MisusedMovedObject carried only a doc url. It is the
pre-rename name of cplusplus.Move, and we still support the clang
versions that only have the old name, so it takes that entry's mapping.
cplusplus. Move itself had no doc url at all, so it gets one.

alpha.core.StdVariant also carried only a doc url. Retrieving an
inactive alternative with std::get throws std::bad_variant_access, so a
true positive means the code will fail at run time, which is HIGH by the
descriptions.json definition.

cppcheck-redundantContinue was in default but in neither sensitive nor
extreme. The profiles are nested by definition.

Eighteen checkers new in clang-tidy and cppcheck had no entry at all,
added those as well.


barnabasdomozi (dismissed):
Currently, there is a merge conflict. Please rebase to master.

config/labels/analyzers/clang-tidy.json
```diff
@@ -2149,7 +2166,6 @@
       "doc_url:https://clang.llvm.org/docs/DiagnosticsReference.html#wconditional-uninitialized",
       "guideline:sei-cert-c",
       "label-tool-skip:severity",
-      "profile:default",
```
dkrupp, line 2219: this looks like a pretty important check for me. using an unitialized variable in a condition... is it that noisy really?

dkrupp (changes requested):
I am afraid that with the current methodology we are moving checkers from and to the profiles back and forth. Let's label up only the new checkers now and start recategorizing the checkers only after the labeling methodology is better developed in a folloup PR.

config/labels/analyzers/clangsa.json
```diff
@@ -216,6 +222,8 @@
     ],
     "alpha.unix.cstring.UninitializedRead": [
       "doc_url:https://clang.llvm.org/docs/analyzer/checkers.html#alpha-unix-cstring-uninitializedread-c",
+      "guideline:sei-cert-c",
```
dkrupp: let's not add alpha checker to the sei cert profile, becuase we dont' recommend to run them in production.

config/labels/analyzers/clang-tidy.json
gamesh411: I have reverted all profile changes for existing checkers.

config/labels/analyzers/clangsa.json
gamesh411: Removed guidelines from alpha checkers.

dkrupp (approved):

dkrupp (approved):
LGTM

PR #5007, opened 2026-08-04
[config] Point doc urls at live documentation

gamesh411:
Checked every doc_url in the label files and 146 were dead: upstream
rebuilds its docs from trunk, so when a checker is renamed or dropped,
the link rots silently.
Where the checker is still documented but under a changed anchor, the
url is re-anchored.
Where it is gone from trunk, the url is pinned to the last version's url
that had the docs for that checker.
The hicpp check pages are no longer generated on trunk at all, so those
are pinned as well.
There was also one mistyped host and two wrong markdownlint anchors.

Five infer entries had CodeChecker's severity token substituted into the
anchor as well as into the checker name, so they get infer's own token
back, and two more carried a stray empty fragment.
The two entries left alone point at issue types that no published
version of the infer docs describes.

### DIFF
--- a/config/labels/analyzers/clang-tidy.json
+++ b/config/labels/analyzers/clang-tidy.json
@@ -1210,7 +1210,7 @@
       "severity:MEDIUM"
     ],
     "cert-int09-c": [
-      "doc_url:https://clang.llmv.org/extra/clang-tidy/checks/cert/int09-c.html",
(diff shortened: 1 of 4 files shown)


gamesh411:
The config coverage is fixed in #5006, so I did not implement it here. Once that PR is merged, this PR will merge in master.

bruntib (approved):

PR #5009, opened 2026-08-04
fix: missing username on re-login & keep session alive oauth

### LINKED ISSUE #5010: Username dissappears after session timeout relogin
Steps to reproduce:
0. This bug is related to session timeouts, so in server_config.json change "session_lifetime" to something like 60 seconds.
1. Open the codechecker server in a new incognito window.
2. Login using OAuth.
3. Navigate a few links.
4. Do nothing and wait until session timeout (e.g. 60 seconds).
5. Click a link and you will be navigated to the login page.

First bug appears here: on the top right corner, the username is still visible despite the session was timed out.

6. Login again using OAuth.

We see the bug: the login page appears twice.
After login, the username is not visible
If the page is reloaded the username is correct again

### LINKED ISSUE #5047: Keep CodeChecker GUI session alive until OAuth session is active
This way we don't need to click on "Login with Microsoft" button in every 20mins.

xb058t:
Closes #5047.
Closes #5010.

### DIFF
--- a/web/server/vue-cli/src/main.js
+++ b/web/server/vue-cli/src/main.js
@@ -84,7 +84,8 @@ router.beforeResolve((to, from, next) => {
         (!store.getters.authParams.sessionStillActive ||
          !store.getters.isAuthenticated)
       ) {
-        // Clear stale auth state so the UI doesn't keep showing it as logged in.
+        // Clear stale auth state so the UI doesn't keep
+        // showing it as logged in.
         store.commit(PURGE_AUTH);
 
         // Redirect the user to the login page but keep the original path to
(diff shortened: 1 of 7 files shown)


web/server/vue-cli/src/main.js
```diff
@@ -80,6 +84,9 @@ router.beforeResolve((to, from, next) => {
         (!store.getters.authParams.sessionStillActive ||
          !store.getters.isAuthenticated)
       ) {
+        // Clear stale auth state so the UI doesn't keep showing it as logged in.
```
gulyasgergely902: Shorten the commit to fit within the 80 char limit.
Check if your IDE supports rulers which is essentially a line which marks certain char length. If so (e.g. VSCode does support this), set it to 80 for js files so you can see how many characters you have to work with.

gulyasgergely902 (changes requested):
Fix the char count issue

PR #5011, opened 2026-08-04
[analyzer] Deprecate --z3 constraint solver flag

gamesh411:
Upstream Clang removed -analyzer-constraints=z3, replacing it with
unsupported-z3 (llvm/llvm-project#205370). CodeChecker no longer
injects this flag into the analyzer command line.

- --z3 flag is kept but deprecated (no-op with warning), removal in 6.30
- is_z3_capable() probes both unsupported-z3 and z3 for older clangs
- Z3 refutation (--z3-refutation) is unaffected

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/analyzer_types.py
+++ b/analyzer/codechecker_analyzer/analyzers/analyzer_types.py
@@ -61,9 +61,17 @@ def is_z3_capable():
     if not enabled_analyzers:
         return False
 
-    return host_check.has_analyzer_option(ClangSA.analyzer_binary(),
-                                          ['-Xclang',
-                                           '-analyzer-constraints=z3'])
+    # Newer clang (>= 23) renamed -analyzer-constraints=z3 to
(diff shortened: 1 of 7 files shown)


bruntib (changes requested):
In CodeChecker we usually support all analyzer versions as much as possible. I think, we should keep `-analyzer-constraints=z3` or `-analyzer-constraints=unsupported-z3` depending on the clang version. I'm aware of a few users who use z3.

dkrupp (changes requested):
I  think since --z3 solver is way too slow  to be useful for clangsa analyis. I think we can just remove this option from CodeChecker (don't need deprecation).

The refutation will and should stay. That is useful and used.

I will make a for of this PR and remove the z3 flag from codechecker. Users will be able to access it through saargs.

dkrupp:
I removed the whole --z3 flag in this PR https://github.com/Ericsson/codechecker/pull/5055

Please consider that PR instead.

PR #5023, opened 2026-08-12
[gui] Make report filters hideable

gulyasgergely902:
This change adds the ability to hide certain filters for Reports and for Statistics page. On the Statistics page it is broken down to every tab which is shown.

This change also introduces the hiding of most filters for CheckerCoverage and for GuidelineStatistics pages which are not usable on those pages.


barnabasdomozi (approved):
LGTM

PR #5024, opened 2026-08-12
log_parser: forward target/arch flags to implicit compiler info detection

tpambor:
CodeChecker detects a compiler's implicit include paths and target triple by invoking the compiler, but `filter_compiler_includes_extra_args()` only forwards a small whitelist of flags (`-m32`/`-m64`, `-std=`, `-stdlib=`, `-nostdinc`, `--driver-mode=`, `--sysroot`). Target-selection flags such as `--target`, `-mcpu`, `-march`, `-mfloat-abi`, `-mfpu`, `-mabi` and `-mthumb` are dropped.

For a multi-target compiler (e.g. clang) this makes the include/target probe run for the compiler's default (usually host) target instead of the target used in the build. The multilib system headers of the real target are therefore never added, and analyzing a cross-compiled translation unit fails with errors like `unknown type name 'uint64_t'` because `#include <stdint.h>` cannot be resolved.

Forward the target and architecture selection flags to both `get_compiler_includes()` and `get_compiler_target()` so the reported include paths and triple match the build.

### DIFF
--- a/analyzer/codechecker_analyzer/buildlog/log_parser.py
+++ b/analyzer/codechecker_analyzer/buildlog/log_parser.py
@@ -289,13 +289,23 @@ def filter_compiler_includes_extra_args(compiler_flags):
 
     compiler_flags -- A list of compiler flags which may affect the list
                       of implicit compiler include paths, like -std=,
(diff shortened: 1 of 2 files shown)


tpambor:
@bruntib Could you take a look please?

PR #5026, opened 2026-08-13
Print server banner in store and diff CLI commands

### LINKED ISSUE #1916: server notification in the command line
Print the server side banner information #1861 viewed in the banner in the command line too.
Commands which should print the extra information:
- store
- login
- diff? (should not mess up the return value (json, csv))

amina825:
Fixes #1916   - The server announcement banner was already printed in the login command. This change extends the same behaviour to the store and diff commands (when a remote server is involved).

- Extract print_banner() helper in cmd_line_client.py to avoid duplicating the get/decode/log logic
- Call print_banner() in store.main() after setup_client()
- Call print_banner() in handle_diff_results() after setup_client(), outside the try/except so network errors are not silently swallowed
- Extend test_announcement_showing_in_cli to cover store and diff

### DIFF
--- a/web/client/codechecker_client/cli/store.py
+++ b/web/client/codechecker_client/cli/store.py
@@ -51,6 +51,7 @@ except ImportError:
         raise NotImplementedError()
 
 from codechecker_client import client as libclient, product
+from codechecker_client.cmd_line_client import print_banner
 from codechecker_client.task_client import await_task_termination
 from codechecker_common import arg, logger, cmd_config
 from codechecker_common.checker_labels import CheckerLabels
@@ -937,6 +938,7 @@ def main(args):
     client = libclient.setup_client(args.product_url)
     protocol, host, port, product_name = \
         product.split_product_url(args.product_url)
+    print_banner(protocol, host, port)
     prod_client = libclient.setup_product_client(protocol,
                                                  host,
                                                  port,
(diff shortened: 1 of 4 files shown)


bruntib (changes requested):
Thank you, it's a nice enhancement.

However, I have some suggestions for the implementation:

`get_announcement_msg()` is called only from `print_banner()`. Couldn't these two functions be merged?
Also, `split_server_url()` could also be called from `print_banner()`, because it's called either way at every occurrence. This way we can save the `protocol`, `host` and `port` parameters. The `args.server_url` should be enough.
Could we print the log message to `LOG`? I think, the `log=None` parameter could be eliminated.

Finally, it would be useful if this banner text would be printed right at the very beginning of each command. The problem is that it gets lost among the lot of other log messages. If you look at `handle_login()`, the `print_banner()` call is at the very beginning, but in case of `handle_diff_results()` it's quite low. Maybe, this could be the first function call after `init_logger()` in every case.

Please, look at the help page of `Codechecker cmd --help`. Here you can see, that right now only `CodeChecker cmd login` and `CodeChecker cmd diff` are covered. But, there are many other sub-commands, too. A `handle_<blabla>` function belongs to each. These are also candidates for printing the banner.

Thank you!

Issue #1445, opened 2018-03-10
indicated locked database


gyorb:
Give the user some warning if deleting a run is not possible because it is locked.

bruntib:
How can a database be locked?

whisperity:
@bruntib
> How can a database be locked?

The user specified in the connection string does not have write access, or a schema transaction (such as an ongoing migration?) or the database server (e.g. for a journaling recovery or after sensing a corrupted physical layer (data storage files)) put an exclusive system lock on the database itself.

It *SHOULD* be rare in normal circumstances, but not something that can't happen at all.

I don't know the exact details on how it can happen with Postgres. MySQL was really trigger happy on locking you out of your own database if the backing files were corrupted (sometimes due to some COW or RAID falling out of sync).

The most simple answer could be the backing files (or file for SQLite) having the write permission removed?

bruntib:
Now I tested and the behavior is really questionable:
1. Start storing a run (it is useful to print some debug logs when RunLock is added and removed)
2. While RunLock is active, remove the run in the command line interface.
3. The run storage finishes but the run is in fact removed.

Run deletion should be rejected when RunLock is active for that run.

PR #5035, opened 2026-08-16
Fix TOCTOU race in removeRun's deletion query (#5031)

### LINKED ISSUE #5031: removeRun can delete a locked run: lock check and deletion use two different queries
**Describe the bug**

A run that is protected by an active run lock can still be permanently deleted. The lock check in `removeRun()` and the actual deletion resolve the runs to act on with two separate queries, so a run that starts matching the filter between the two queries is deleted without its lock ever being checked.

In `removeRun()` (`web/server/codechecker_server/api/report_server.py`), the preflight resolves the matching run IDs and checks their locks:

https://github.com/Ericsson/codechecker/blob/ac2c3ac12e7037aead059b30f1109c369cdd263c/web/server/codechecker_server/api/report_server.py#L4195-L4197

```python
            matched_run_ids = get_run_ids_for_filter(session, run_filter)
            if matched_run_ids:
                check_remove_runs_lock(session, matched_run_ids)
```

But the destructive query is then built from the `RunFilter` again, not from `matched_run_ids`, and the loop deletes and commits each returned row with no further lock check:

https://github.com/Ericsson/codechecker/blob/ac2c3ac12e7037aead059b30f1109c369cdd263c/web/server/codechecker_server/api/report_server.py#L4199-L4212

```python
            q = process_run_filter(session, session.query(Run), run_filter)
            ...
            for run in q.all():
                try:
                    runs.append(run.name)
                    session.delete(run)
                    session.commit()
```

Because each delete is committed in its own transaction, the set of rows returned by `q.all()` can differ from the set that was lock-checked a moment earlier. Any run that newly matches the filter in that window is deleted even if it holds an active `RunLock`.

**CodeChecker version**

master at ac2c3ac (code inspection; the window is timing-dependent so it is easiest to see in the source).

**To Reproduce**

Minimal failure scenario with a name-based filter:

1. A run named `myrun` exists and is not locked.
2. Client A calls `removeRun(None, RunFilter(names=['myrun']))`. The preflight resolves `myrun`'s ID and `check_remove_runs_lock` passes.
3. Before `q.all()` executes, client B deletes the original `myrun` and starts a fresh store under the same name, which creates a new `Run` row holding an active `RunLock`.
4. Client A's second query matches the new row by name and deletes it, along with the in-progress store's data, even though it was locked.

**Expected behaviour**

Only the runs that passed the lock check should be deletable. The simplest fix is to build the destructive query from the already-resolved `matched_run_ids` (for example `session.query(Run).filter(Run.id.in_(matched_run_ids))`) instead of re-running `process_run_filter`. Alternatively, re-check `check_remove_runs_lock` on the exact rows returned by `q.all()` right before deleting them.

 - OS: Linux (server-side issue, not client or browser specific)

**Additional context**

This is a follow-up to the fix merged in #4999, which correctly moved the lock check from `[run_id]` to the filter-resolved IDs. The remaining gap is that the check and the deletion still resolve the filter independently, so the guarantee only holds for rows that matched at preflight time.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs. Full analysis: https://app.ito.ai/share/d4939f58-5894-45e5-b1dc-9b62c14b9e4e.

mmido6039:
Fixes #5031

## Problem
A run protected by an active RunLock could still be permanently
deleted. removeRun() resolves matched_run_ids from the filter and
lock-checks those specific ids, but the actual deletion query re-ran
process_run_filter() with the original filter instead of reusing the
already-resolved ids.

## Root cause
Since each delete commits in its own transaction, there's a window
between the lock check and the deletion query where a run matching
the same filter (e.g. same name, freshly recreated with a new active
lock - as when a store restarts under the same run name) could be
picked up by the deletion query without ever being lock-checked. This
is a follow-up to #1445/#4999, which fixed the lock check itself but
left the deletion query resolving the filter independently, so the
guarantee only held for rows that matched at preflight time.

## Fix
Build the destructive query directly from the already lock-checked
matched_run_ids:

  q = session.query(Run).filter(Run.id.in_(matched_run_ids))

instead of re-running process_run_filter(). This guarantees only the
exact rows that passed the lock check can be deleted, regardless of
what the filter would match afterward.

## Testing
Added test_deletion_query_uses_resolved_ids_not_the_filter_again to
web/server/tests/unit/test_run_removal_lock.py, simulating the exact
race from the issue: resolve and lock-check an unlocked run, then
simulate a concurrent client deleting it and creating a new run under
the same name with an active lock. Confirms the fixed query
(Run.id.in_(matched_run_ids)) correctly excludes the new row, while
explicitly demonstrating that re-running the filter (the old buggy
pattern) would have matched it. All 5 tests in the file pass; no
regressions in the rest of the web unit suite; pycodestyle clean.

### DIFF
--- a/web/server/tests/unit/test_run_removal_lock.py
+++ b/web/server/tests/unit/test_run_removal_lock.py
@@ -22,24 +22,30 @@ whether the targeted run(s) were actually locked.
 
 import unittest
 from datetime import datetime, timedelta
+from unittest.mock import MagicMock, patch
 
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
+from sqlalchemy.pool import StaticPool
 
 from codechecker_api.codeCheckerDBAccess_v6.ttypes import RunFilter
 from codechecker_api_shared.ttypes import RequestFailed
 
 from codechecker_server.api.report_server import \
-    check_remove_runs_lock, get_run_ids_for_filter, process_run_filter
+    ThriftRequestHandler, check_remove_runs_lock, get_run_ids_for_filter
 from codechecker_server.database.run_db_model import \
     Base, Run, RunLock
 
 
 class RunRemovalLockTest(unittest.TestCase):
     def setUp(self):
-        self.engine = create_engine('sqlite:///:memory:')
+        self.engine = create_engine(
+            'sqlite:///:memory:',
+            connect_args={'check_same_thread': False},
+            poolclass=StaticPool)
         Base.metadata.create_all(self.engine)
-        self.session = sessionmaker(bind=self.engine)()
+        self.session_factory = sessionmaker(bind=self.engine)
+        self.session = self.session_factory()
 
         run1 = Run('locked_run', '1.0')
         run1.id = 1
@@ -102,7 +108,8 @@ class RunRemovalLockTest(unittest.TestCase):
         # expired/stale.
         check_remove_runs_lock(self.session, matched_run_ids)
 
-    def test_deletion_query_uses_resolved_ids_not_the_filter_again(self):
+    def test_remove_run_does_not_delete_a_run_recreated_during_the_call(
+            self):
         """
         Regression test for
         https://github.com/Ericsson/codechecker/issues/5031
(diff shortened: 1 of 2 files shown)


web/server/tests/unit/test_run_removal_lock.py
```diff
@@ -102,6 +102,60 @@ def test_expired_lock_does_not_block_removal(self):
...
+        new_run.id = 3
+        self.session.add(new_run)
+        self.session.add(RunLock('unlocked_run'))
+        self.session.commit()
+
+        # The fixed deletion query: built from the already-resolved,
+        # already lock-checked ids. It must not match the new, locked
+        # run.
+        fixed_query = self.session.query(Run) \
+            .filter(Run.id.in_(matched_run_ids))
```
bruntib: This test duplicates the code that is supposed to be tested. So, the code-base itself is not covered by this test. This is not to best approach, because if the implementation in `removeRun()` changes, then this query gets out of sync. I know, it's really not elegant to say, but we shouldn't add this test function. If we add a test, then that should simulate the concurrent removal of a run through its name.

bruntib (changes requested):
The fix looks good to me, but I have some concerns about the test.

web/server/tests/unit/test_run_removal_lock.py
```diff
@@ -102,6 +108,73 @@ def test_expired_lock_does_not_block_removal(self):
...
+        racing in between. The new, locked run must survive the call.
+        """
+        run_filter = RunFilter(names=['unlocked_run'], exactMatch=True)
+        real_get_run_ids_for_filter = get_run_ids_for_filter
+
+        def resolve_then_simulate_concurrent_removal(session, rf):
+            ids = real_get_run_ids_for_filter(session, rf)
+
+            concurrent_session = self.session_factory()
+            concurrent_session.query(Run).filter(Run.id == 2).delete()
```
bruntib, line 141: Thank you, this looks better now. One last request. Could you not to hard-code the run ID here? How can we make sure that the run in question has ID 2 in the long terms? What if the tests are executed in a different order, or another test is added later which also inserts runs? In that case this hard-coded number would break. Maybe, the run could be filtered by its name.

Also, when inserting a new run a few lines below, the ID 3 may conflict with another test in the future. Could we rely on having an autoincrement ID?

Thank you!

bruntib:
Merged in https://github.com/Ericsson/codechecker/pull/5061

PR #5038, opened 2026-08-18
[gui] Add the missing rules and guidelines to the checkers

gulyasgergely902:
Guideline Statistics shows different checkers for the same rules across different guidelines. This is not how it should be.

This change fixes this and adds the same rules for all guidelines which applicable to the checkers.


dkrupp (approved):
LGTM

Issue #1761, opened 2018-10-09
non ascii character in plist


gyorb:
Add plist parser test files with non ascii characters.

PR #5040, opened 2026-08-18
Fix/non ascii plist

amina825:
Add regression coverage for plist parsing with non-ASCII characters in diagnostic text and bug path messages.

Validated with the plist parser test suite.


bruntib (changes requested):
This is a good patch, but please, put the commit on a separate branch so the other pull request's content is not added to this one.

Also, please add the test "Fixes #1761" in the commit message so GitHub can connect it with the corresponding ticket.

Thank you!

amina825:
I created a new clean PR containing only the fix for #1761, as requested. Closing this PR to avoid including unrelated commits.

PR #5042, opened 2026-08-18
[gui] Auto-expand file tree to the first branching node

### LINKED ISSUE #5041: Open the file tree leaves until the first branch with more than one child
Make the file tree open up automatically until the first node with more than one child.

xb058t:
Closes #5041 .

### DIFF
--- a/web/server/vue-cli/src/views/Reports.vue
+++ b/web/server/vue-cli/src/views/Reports.vue
@@ -33,7 +33,7 @@
           <v-btn
             value="tree"
             size="small"
-            @click="setReportFilter({ filepath: null })"
+            @click="onFileTreeViewClick"
           >
             File Tree
           </v-btn>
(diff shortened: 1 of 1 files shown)


web/server/vue-cli/src/views/Reports.vue
```diff
@@ -904,6 +904,28 @@ function buildTreeItems() {
   treeItems.value = items;
 }
 
+// Open nodes down to the first branching (multi-child) node.
```
gulyasgergely902: This comment is unnecessary, please remove

web/server/vue-cli/src/views/Reports.vue
```diff
@@ -904,6 +904,28 @@ function buildTreeItems() {
...
+function collectAutoOpenPaths(nodes, opened) {
```
gulyasgergely902: I'd call this function 'collectSingleChildedNodes' since it essentially does that: collects the nodes which has only one child and adds them to the opened list.

gulyasgergely902 (changes requested):

web/server/vue-cli/src/views/Reports.vue
xb058t: Done

web/server/vue-cli/src/views/Reports.vue
xb058t: Done

gulyasgergely902 (approved):
LGTM

PR #5043, opened 2026-08-19
[gui] Fix comment btn spacing

gulyasgergely902:


dkrupp (approved):
I like the new functionality.
Please consider rephrasing the commit message for better understandability:
"After this change, if the review status is anything other than unreviewed, then the comment field is displayed in a well visible place in the code details view."

bruntib:
The label is shown only if the comment is not empty. If I change the review status without providing a message, then the label doesn't appear. Is it intentional?

PR #5049, opened 2026-08-25
Execution order of the static analyzers

### LINKED ISSUE #4362: Execution order of the static analyzers
The order of execution of the analyzers seems random. Sometimes clangsa is executed first, sometimes clang-tidy.

Some of my files take more then 90 minutes to analyze with clangsa. If I start to proces this file as soon as possible it will reduce the time for the total run.

I want to have it in fixed order, first clangsa and then clang-tidy.

I've already asked on the discussions forum: https://github.com/Ericsson/codechecker/discussions/4361

hornokattila:
The order in which analyzers (clangsa, clang-tidy, etc.) were scheduled for execution was non-deterministic because check_supported_analyzers() used a set to collect enabled analyzers. This meant that on some runs clang-tidy tasks would be dispatched before clangsa tasks.

This change replaces the set with a list, preserving the iteration order from the supported_analyzers dict. This guarantees clangsa actions are always dispatched to the worker pool first, followed by clang-tidy, cppcheck, etc.

For users with long-running clangsa analyses (e.g. 90+ minutes per file), this ensures those heavy tasks start immediately on available cores, reducing total wall-clock time.

Fixes #4362

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/analyzer_types.py
+++ b/analyzer/codechecker_analyzer/analyzers/analyzer_types.py
@@ -29,6 +29,11 @@ from .infer.analyzer import Infer
 
 LOG = get_logger('analyzer')
 
+# The order of this dictionary determines the default execution order of
+# analyzers. Longer-running analyzers (e.g. clangsa) should come first so
+# that they are scheduled earlier in parallel execution, reducing total
+# wall-clock time. Do not change the order without considering the impact
+# on analysis scheduling.
 supported_analyzers = {ClangSA.ANALYZER_NAME: ClangSA,
                        ClangTidy.ANALYZER_NAME: ClangTidy,
                        Cppcheck.ANALYZER_NAME: Cppcheck,
--- a/analyzer/codechecker_analyzer/analyzer.py
+++ b/analyzer/codechecker_analyzer/analyzer.py
@@ -149,6 +149,14 @@ def perform_analysis(args, skip_handlers, filter_handlers,
     analyzers, errored = \
         analyzer_types.check_available_analyzers(args.analyzers)
 
+    # Sort analyzers according to the order defined in supported_analyzers.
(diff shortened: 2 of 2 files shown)


analyzer/codechecker_analyzer/analyzers/analyzer_types.py
```diff
@@ -149,7 +149,7 @@ def check_supported_analyzers(analyzers):
 
     analyzer_binaries = context.analyzer_binaries
 
-    enabled_analyzers = set()
+    enabled_analyzers = []
     failed_analyzers = set()
 
     for analyzer_name in analyzers:
```
bruntib, line 160: If I see correctly, the `analyzers` variable comes from the `--enable` command line arguments and not from the `supported_analyzers`. This way the user determines the execution order with this flag.

Nevertheless, I wouldn't think, this is the right place of implementing this feature. The purpose of this `check_supported_analyzers()` is really performing some check and not ordering the analyzers. I think, an explicit sorting should go to `perform_analysis()` function in `analyzer.py` based on `supported_analyzers` so it is explicitly visible that sorting analyzers is an intentional feature. Also, we could write a comment above `supported_analyzers` that says something like "please, don't change the order of this list, because analysis execution is intended to happen in this order".

bruntib (changes requested):
Thank you for the fix!

analyzer/codechecker_analyzer/analyzers/analyzer_types.py
hornokattila: Good point! Moved the ordering logic into perform_analysis() where it's explicit. The analyzers are now sorted based on the key order of supported_analyzers.

bruntib: Thank you!

bruntib (approved):

PR #5052, opened 2026-08-27
[gui] Only call fetchStatistics if the filters are ready

gulyasgergely902:
This change fixes an issue where fetchStatistics were called multiple times even before the reports filters were ready. This made the UI unresponsive at times.

### DIFF
--- a/web/server/vue-cli/src/views/Statistics.vue
+++ b/web/server/vue-cli/src/views/Statistics.vue
@@ -12,7 +12,7 @@
           :report-count="reportCount"
           :refresh-filter="refreshFilterState"
           :hidden-filters="hiddenFilters"
-          @refresh="refresh"
+          @refresh="refreshByReportFilter"
           @set-refresh-filter-state="setRefreshFilterState"
         />
       </div>
@@ -146,6 +146,7 @@ const tabs = [
 const refreshFilterState = ref(false);
 const reportCount = ref(0);
 const tab = ref(null);
+const reportFiltersReady = ref(false);
 
 const bus = mitt();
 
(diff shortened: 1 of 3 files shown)


dkrupp (changes requested):
I am not so fan of timers, because it can introduce flaky behaviour depending on (browser/machine).

Also the debounce mechanism hides multiple sequential events and selects teh first one randomly. Are the filters already loaded then?

I think a much clearer solution would be to define a 1 callback which is fired once all filter is loaded and the data need to be refetched.

1) filter change
2) page reload
3) tab change

web/server/vue-cli/src/views/Statistics.vue
```diff
@@ -183,11 +180,13 @@ watch(() => tab.value, async () => {
...
+  if (!reportFiltersReady.value) return;
+
   await nextTick();
-  refreshCurrentTab();
+  emitRefresh();
 });
 
-function refresh() {
+function refresh(reason) {
   ccService.getClient().getRunResultCount(
```
dkrupp, line 206: please consider protecting these calls with early return for the case when the filters are not ready yet.

web/server/vue-cli/src/views/Statistics.vue
```diff
@@ -203,10 +202,18 @@ function refresh() {
...
+  }
+
+  if (reason === "filter-change" && reportFiltersReady.value) {
+    emitRefresh();
+  }
 }
 
-function refreshCurrentTab() {
+function emitRefresh() {
   bus.emit("refresh");
```
dkrupp, line 223: consider clarifying the naming of the refresh to distinguish between "filter refresh" and "statistics refresh" events. This will make the code more understandable.

dkrupp (changes requested):
please check my minor comments. otherwise looks good.

dkrupp (approved):
LGTM

PR #5054, opened 2026-08-28
Username dissappears after session timeout relogin

### LINKED ISSUE #5010: Username dissappears after session timeout relogin

xb058t:
Closes #5010.

### DIFF
--- a/web/server/vue-cli/src/components/Layout/UserInfoMenu.vue
+++ b/web/server/vue-cli/src/components/Layout/UserInfoMenu.vue
@@ -83,7 +83,7 @@
 </template>
 
 <script setup>
-import { computed, onMounted, ref, watch } from "vue";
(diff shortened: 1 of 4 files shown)


gulyasgergely902 (approved):
LGTM

PR #5056, opened 2026-08-30
Introduce gcc warnings as checkers

bruntib:
Gcc has its own checkers that can be enabled as warnings: -Wanalyzer-<checker_name>. However, all other warnings can be considered as checkers.

Until this patch the -Wanalyzer-<checker_name> checkers could have been enabled in CodeChecker as --enable gcc-<checker_name>. This patch makes it possible to enable all other warnings as checkers: --enable gcc-<warnings_name>.

The problem is that both gcc static analyzers and gcc warning names have the following transformation in CodeChecker:

-Wanalyzer-<name>  ->  gcc-<name>
-W<name>           ->  gcc-<name>

Theoretically it's possible to generate the inverse transformation, because static analyzer names and warning names are disjoint. But we decided to preserve all original names for the checkers:

-Wanalyzer-<name>  ->  gcc-analyzer-<name>
-W<name>           ->  gcc-<name>

This is a backward incompatible change.

### DIFF
--- a/analyzer/codechecker_analyzer/analyzers/gcc/result_handler.py
+++ b/analyzer/codechecker_analyzer/analyzers/gcc/result_handler.py
@@ -29,18 +29,18 @@ LOG = get_logger('analyzer.gcc')
 
 
 def actual_name_to_codechecker_name(actual_name: str):
-    assert actual_name.startswith('-Wanalyzer')
-    return actual_name.replace("-Wanalyzer", "gcc")
+    assert actual_name.startswith('-W')
+    return actual_name.replace("-W", "gcc-", 1)
 
 
 def codechecker_name_to_actual_name(codechecker_name: str):
-    assert codechecker_name.startswith('gcc')
-    return codechecker_name.replace("gcc", "-Wanalyzer")
+    assert codechecker_name.startswith('gcc-')
+    return codechecker_name.replace("gcc-", "-W", 1)
 
 
 def codechecker_name_to_actual_name_disabled(codechecker_name: str):
-    assert codechecker_name.startswith('gcc')
-    return codechecker_name.replace("gcc", "-Wno-analyzer")
+    assert codechecker_name.startswith('gcc-')
+    return codechecker_name.replace("gcc-", "-Wno-", 1)
 
 
 class GccResultHandler(ResultHandler):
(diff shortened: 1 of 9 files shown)


analyzer/codechecker_analyzer/analyzers/gcc/result_handler.py
```diff
@@ -29,18 +29,18 @@
 
 
 def actual_name_to_codechecker_name(actual_name: str):
-    assert actual_name.startswith('-Wanalyzer')
-    return actual_name.replace("-Wanalyzer", "gcc")
+    assert actual_name.startswith('-W')
+    return actual_name.replace("-W", "gcc-")
```
barnabasdomozi: Only the first occurrence should be replaced, not all.
Consider:

```suggestion
    return actual_name.replace("-W", "gcc-", 1)
```

analyzer/codechecker_analyzer/analyzers/gcc/result_handler.py
```diff
@@ -29,18 +29,18 @@
...
 
 
 def codechecker_name_to_actual_name(codechecker_name: str):
-    assert codechecker_name.startswith('gcc')
-    return codechecker_name.replace("gcc", "-Wanalyzer")
+    assert codechecker_name.startswith('gcc-')
+    return codechecker_name.replace("gcc-", "-W")
```
barnabasdomozi:
```suggestion
    return codechecker_name.replace("gcc-", "-W", 1)
```

analyzer/codechecker_analyzer/analyzers/gcc/result_handler.py
```diff
@@ -29,18 +29,18 @@
...
 
 
 def codechecker_name_to_actual_name_disabled(codechecker_name: str):
-    assert codechecker_name.startswith('gcc')
-    return codechecker_name.replace("gcc", "-Wno-analyzer")
+    assert codechecker_name.startswith('gcc-')
+    return codechecker_name.replace("gcc-", "-Wno-")
```
barnabasdomozi:
```suggestion
    return codechecker_name.replace("gcc-", "-Wno-", 1)
```

barnabasdomozi (changes requested):
Please update the documentation in `docs/analyzer/checker_and_analyzer_configuration.md` as well.
E.g. it still refers to `gcc-double-free` instead of  `gcc-analyzer-double-free`.

barnabasdomozi (approved):
LGTM

PR #5075, opened 2026-09-06
Re-introduce --details flag for "CodeChecker cmd results"

bruntib:
"CodeChecker cmd results" contains the git blame info by default. However, this can be a really slow query on the server side in case of big result set. For this reason, the --details flag has been re-introduced. The JSON output format contains these detailed information.

### DIFF
--- a/web/client/codechecker_client/cmd_line_client.py
+++ b/web/client/codechecker_client/cmd_line_client.py
@@ -811,9 +811,8 @@ def handle_list_results(args):
 
     report_filter = parse_report_filter(client, args)
 
-    # TODO: JSON format contains detailed information either way. --details
-    # flag is deprecated in 6.28.0 and should be removed in 6.29.0.
-    query_report_details = 'details' in args or args.output_format == 'json'
+    query_report_details = 'details' in args and args.details and \
+        args.output_format == 'json'
 
     all_results = get_run_results(client,
                                   run_ids,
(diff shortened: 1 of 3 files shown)


web/client/codechecker_client/cmd_line_client.py
```diff
@@ -813,7 +813,8 @@ def handle_list_results(args):
 
     # TODO: JSON format contains detailed information either way. --details
```
barnabasdomozi: This comment seems no longer relevant.

web/client/codechecker_client/cmd_line_client.py
```diff
@@ -830,6 +831,9 @@ def handle_list_results(args):
             report.runName = run_id2name[report.runId]
         print(CmdLineOutputEncoder().encode(all_results))
     else:
+        header = ['File', 'Checker', 'Severity', 'Message', 'Bug path length',
```
barnabasdomozi: This `header` variable is already set on line 874 and 885. Consider removing this line.

web/client/codechecker_client/cmd_line_client.py
```diff
@@ -813,7 +813,8 @@ def handle_list_results(args):
 
     # TODO: JSON format contains detailed information either way. --details
     # flag is deprecated in 6.28.0 and should be removed in 6.29.0.
-    query_report_details = 'details' in args or args.output_format == 'json'
+    query_report_details = args.details and args.output_format == 'json' \
```
barnabasdomozi: Consider:

```suggestion
    query_report_details = (if 'details' in args and args.details and args.output_format == 'json')
```

Which is more clear and easier to read.

barnabasdomozi (changes requested):

bruntib:
Thanks for the review, good catches!

barnabasdomozi (approved):
LGTM

PR #5087, opened 2026-09-09
[nfc] Uplift Pylint's Python base version, use modern type annotations

barnabasdomozi:


codechecker_common/checker_labels.py
```diff
@@ -7,13 +7,13 @@
 # -------------------------------------------------------------------------
 from collections import defaultdict
 import os
-from typing import Any, cast, DefaultDict, Dict, Iterable, List, Optional, \
-    Set, Tuple, Union
+from typing import Any, cast, DefaultDict, Iterable, Optional, \
```
bruntib: `defaultdict` inherits from `dict` as far as I know. Couldn't we apply the `dict` type-hint to `defaultdict`? Or, does `defaultdict` behave like a type-hint?

.pylintrc
```diff
@@ -93,7 +93,7 @@ prefer-stubs=no
 
 # Minimum Python version to use for version dependent checks. Will default to
 # the version used to run pylint.
-py-version=3.8
+py-version=3.11
```
bruntib, line 96: We use 3.10 in the github actions. It may be useful to use the same version, because sometimes pylint reports depend on the python version running it.

bruntib (changes requested):

.pylintrc
barnabasdomozi: Python 3.10 will be deprecated this October. I think we should also updates the GitHub actions to now use 3.11.

codechecker_common/checker_labels.py
barnabasdomozi: We can indeed use the built-in `defaultdict` as a type annotation instead of `DefaultDict`.

bruntib (approved):

PR #5089, opened 2026-09-10
Replace hashlib.md5 with hashlib.sha256 for FIPS compliance

### LINKED ISSUE #4308: Add FIPS support to CodeChecker
CodeChecker uses hashlib md5() in several areas. This causes problems on FIPS enabled systems, since MD5 is not supported in that mode.

The hashlib md5() signature was updated to include a [workaround in Python 3.9](https://docs.python.org/3/library/hashlib.html#hashlib.new). Implementing this workaround in CodeChecker should allow it to function on FIPS systems.

If changing how md5() is called is not feasible, then it may also be possible to use sha256(), which is FIPS compliant.

We are using CodeChecker for our CD/CI, which is transitioning to a FIPS platform. This will become a blocker for that process.

Update: Simply replacing 'md5' with 'sha256' seems to work, but I don't know if there are any unintended side effects.

wadehunk:
Modified calls to `hashlib.md5()` to use `hashlib.sha256()` in order to be FIPS compliant. Closes #4308.


wadehunk:
Closing in deference to #3751 as it seems to impact less code.

PR #5091, opened 2026-09-10
[server] Store SQLite product database paths relative to workspace

### LINKED ISSUE #3081: Server sqlite database files are stored as absolute paths
I create a new instance of codechecker server, running as a different user, in order to test new version or new settings. To do that I copy all old config to the new user:
```sh
sudo cp -r  /home/user1/.codechecker  /home/user2/
sudo chown -R user2  /home/user2/.codechecker
```
Then I start the server as user2 on a new port, and I see that it is trying to access the original databases at `/home/user1` for all databases except `config.sqlite`.

Reason: `config.sqlite` contains absolute paths of the databases for each product.

Solution: Store relative paths when the db is under the config directory.

Muteeb-Haider:
## Summary
- `config.sqlite` stored absolute paths for each product's SQLite database. Copying/moving a server's config or workspace directory elsewhere (e.g. to test as a different user, per the reproduction steps in #3081) left the server trying to reach the old, now-inaccessible location for every product except `config.sqlite` itself.
- New and edited SQLite-backed products now have their database path stored relative to the workspace directory, and it's resolved back to an absolute path everywhere the server actually connects to a product: normal startup (`add_product`), background config-db cleanup, background `massStoreRun` tasks, and the CLI's migration/status-check paths.
- Existing rows with an absolute path are left untouched and continue to work (no DB migration needed); editing such a product will naturally migrate it to a relative path going forward.

## Test plan
- [x] Added `web/server/tests/unit/test_sqlite_relative_path.py` covering the new `SQLServer.resolve_sqlite_relative_path` helper (relative sqlite path resolves against workspace dir; absolute sqlite path and postgres connection strings pass through unchanged).
- [x] Full existing unit suite (`web/server/tests/unit`) passes.
- [x] Functional test `tests/functional/products/test_products.py::test_add_invalid_product` passes against a live server.
- [x] Manually reproduced the exact issue scenario: started a server, confirmed `config.sqlite` stores a relative path (`sqlite+pysqlite:///Default.sqlite`), copied the whole workspace directory to a new path (simulating copying to a different user's home), and confirmed a fresh server instance pointed at the copy connects to its own database with no reference to the original path anywhere in the logs.

Fixes #3081

### DIFF
--- a/web/server/codechecker_server/api/mass_store_run.py
+++ b/web/server/codechecker_server/api/mass_store_run.py
@@ -45,7 +45,7 @@ from codechecker_report_converter.report.hash import get_report_path_hash
 
 from ..database import db_cleanup
 from ..database.config_db_model import Product
-from ..database.database import DBSession
+from ..database.database import DBSession, SQLServer
 from ..database.run_db_model import \
     AnalysisInfo, AnalysisInfoFile, AnalyzerStatistic, \
     ReportPathData, ReportPathDataFile, \
@@ -649,10 +649,14 @@ class MassStoreRunTask(AbstractTask):
             if not db_product:
                 raise KeyError(f"No product with ID '{self._product_id}'")
 
+            connection_string = SQLServer.resolve_sqlite_relative_path(
+                db_product.connection,
+                self._package_context.codechecker_workspace)
+
             self._product = ServerProduct(db_product.id,
                                           db_product.endpoint,
                                           db_product.display_name,
-                                          db_product.connection,
+                                          connection_string,
                                           self._package_context,
                                           tm.environment)
 
--- a/web/server/codechecker_server/api/product_server.py
+++ b/web/server/codechecker_server/api/product_server.py
@@ -31,7 +31,8 @@ from codechecker_web.shared import convert
 
 from .. import permissions
 from ..database.config_db_model import IDENTIFIER, Product, ProductPermission
-from ..database.database import DBSession, SQLServer, conv, escape_like
+from ..database.database import DBSession, SQLServer, SQLiteDatabase, \
+    conv, escape_like
 from ..routing import is_valid_product_endpoint
 
 from .thrift_enum_helper import confidentiality_enum, \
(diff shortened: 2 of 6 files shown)


web/server/codechecker_server/database/database.py
```diff
@@ -440,6 +440,25 @@ def connection_string_to_args(connection_string):
...
+        strings, and SQLite ones that already use an absolute path, are
+        returned unchanged.
+        """
+        url = make_url(connection_string)
+        if 'sqlite' not in url.drivername or not url.database or \
+                os.path.isabs(url.database):
+            return connection_string
+
+        return str(url.set(
+            database=os.path.join(workspace_directory, url.database)))
```
bruntib:
```suggestion
            database=os.path.relpath(url.database, workspace_directory)))
```

web/server/codechecker_server/api/product_server.py
```diff
@@ -592,8 +593,10 @@ def addProduct(self, product):
                     codechecker_api_shared.ttypes.ErrorCode.DATABASE,
                     "SQLite database must be given by relative path!")
 
-            dbc.database = path_for_fake_root(os.path.join(
+            resolved_path = path_for_fake_root(os.path.join(
                 "/", dbc.database), self.__server.workspace_directory)
+            dbc.database = os.path.relpath(
+                resolved_path, self.__server.workspace_directory)
```
bruntib: The function `path_for_fake_root()` creates an absolute path that ensures that the given path of the `.sqlite` file is under the `workspace_directory`. The `os.path.relpath` in the next line transforms it back to a relative path.

I think, we should simplify the specification: We should require the user in case of SQLite usage to provide a file name, only. We should reject everything else, like `/hello.sqlite`, `../hello.sqlite`, `my/path/hello.sqlite`. Only `hello.sqlite` should be accepted. It could be checked in line 591: `if os.path.split(dbc.database)[0] != ''`.

If we ensure having a filename only, then these 4 lines are not necessary at all. The `path_for_fake_root()` function could be eliminated from the code base.

web/server/codechecker_server/api/mass_store_run.py
```diff
@@ -649,10 +649,14 @@ def _implementation(self, tm: TaskManager):
             if not db_product:
                 raise KeyError(f"No product with ID '{self._product_id}'")
 
+            connection_string = SQLServer.resolve_sqlite_relative_path(
```
bruntib: This operation could go to the `Product` class. That class has the workspace location anyway, and this path manipulation would be applied automatically at every product server instantiation.

PR #5098, opened 2026-09-11
Introduce new Thrift workflow

barnabasdomozi:
Previously, changing the Thrift API from one version to another
was cumbersome and prone to mistakes. This patch aims
to address this.

Changes:
- The `codechecker_api` PyPI and npm packages were removed.
- Existing Thrift API tar.gz files were removed.
- The entire `web/api/` and `scripts/thrift/` directories were
removed.
- The Thrift API files are now moved to a newly created
`codechecker_api/` directory.
- Running `make package` or `make dev_package` now generates the stubs into
`codechecker_api/python/` and `codechecker_api/js/`. Executing the Thrift
compiler requires Docker; therefore, Docker is now a hard dependency of
the project.
- Existing Python imports were rewritten to reflect the above change.
- The codechecker_api version is defined in only one file:
`codechecker_api/api_version.json` and is read by the webserver context
class to provide it for both the client and server implementations.
- The `codechecker_api_shared` Thrift interface file was renamed to
`shared`, so it can be imported as `codechecker_api.python.shared`
- The relevant Thrift documentation was moved to the file
`codechecker_api/README.md`
- Previously, `web/api` contained OAuth-related documentation, which
was moved to `docs/web/oauth_dev_guide.md`
- The Docker CI build pipeline was modified to build the Thrift
stubs before building the CodeChecker web docker image.


codechecker_api/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore built Thrift stubs
```
bruntib, line 1: The `codechecker_api` should be somewhere under `web` directory. All things that do some API communication are under `web`.

codechecker_api/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Table of Contents
+* [API change workflow](#api-change-workflow)
+* [Generating the stubs](#generating-the-stubs)
+* [API versioning](#api-versioning)
+* [Available APIs](#available-apis)
+
+## API change workflow
+- Modify the `.thrift` API files.
+- Update the Thrift API version in `api_version.json`.
```
bruntib, line 15: The original documentation had a workflow where `.thrift` files are changed without increasing the version number. Does this scenario work, too?

codechecker_api/server_info.thrift
```diff
@@ -4,7 +4,7 @@
 //  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 // -------------------------------------------------------------------------
 
-namespace py ServerInfo_v6
+namespace py codechecker_api.python.ServerInfo_v6
 namespace js codeCheckerServerInfo_v6
```
bruntib, line 8: What is the purpose of changing the namespace for Python but not for JavaScript?

codechecker_api/Makefile
```diff
@@ -0,0 +1,48 @@
...
+TARGET_PY = --gen py
+TARGET_JS = --gen js:node
+
+API_DIR := $(CURDIR)
+uid = $(shell id -u)
+guid = $(shell id -g)
+
+clean_thrift_stubs:
+	# Remove the generated Thrift stubs
+	rm -rf $(API_DIR)/python/ $(API_DIR)/js/ $(API_DIR)/gen-*/
```
bruntib, line 12: I don't have `gen-*` folders. They are deleted by `thrift_stubs` target.

docs/web/api/README.md
```diff
@@ -1,132 +0,0 @@
...
-Minor version changes should not change the behaviour and the invocation of
-already existing functions, only extend the API:
-
- 1. After implementing the changes to the existing API files and handlers,
- you need to increase the `SUPPORTED_VERSIONS` tag in
- `codechecker_web/shared/version.py`. This dict stores for each major
- version, which is the highest supported minor version. In this case, simply
- increase the number by `1`.
-
-### Major API changes <a name="major-api-changes"></a>
```
bruntib, line 94: The new documentation doesn't allow changing the major documentation. It looks like that it's intentionally not supported to increase the major version. However, it is theoretically possible that we introduce a completely new API with new major version in the URL.
I'm OK with eliminating this section, but it would be nice to have a word in the new documentation, about why we don't support major version change.

docs/deps.md
```diff
@@ -26,6 +25,7 @@ Javascript dependencies are automatically downloaded based on the ext_source_dep
   * [Marked](https://github.com/chjj/marked) (BSD) - view documentation for checkers written in markdown (generated dynamically)
   * [Dojotoolkit](https://dojotoolkit.org/) (BSD) - main framework for the web UI
```
bruntib, line 26: Wow, it's fun to see that Dojo is still in the docs :D

scripts/thrift/client.py
```diff
@@ -1,190 +0,0 @@
-#!/usr/bin/env python3
```
bruntib, line 1: This is a "Hello World" example on API usage for those who would like to create their own client application for querying custom data through the API. This script should be kept.

scripts/thrift/README.md
```diff
@@ -1,41 +0,0 @@
-# Python client for CodeChecker
```
bruntib, line 1: This is a documentation for the "Hello World" client. This should be kept, too.

bruntib (changes requested):

bruntib:
`make_dist` target fails for me. We should make sure that this new infrastructure works in the `pip install` scenario, too.

codechecker_api/.gitignore
barnabasdomozi: While it is used to be under `web/`, placing it there breaks Python's packaging conventions. The convention is that import statements should translate to the directory structure and not be flattened or rearranged by a build pipeline. Breaking this convention leads to unnecessary complications when setting up type checkers, linters, IDEs or any other tooling that assumes this basic principle.

codechecker_api/README.md
barnabasdomozi: Yes. In this new workflow, the `api_version.json` is totally separated from any Thrift stub generation. This JSON file is simply read from disk when the program starts and it gets appended in the request URL.
E.g. `https://example.com:8001/Default/v6.74/CodeCheckerService`

codechecker_api/server_info.thrift
barnabasdomozi: In Thrift, the namespace defines the resulting packaging structure it will generate.
E.g. `codechecker_api.py.ServerInfo_v6` creates directory structure `codechecker_api/py/ServerInfo_v6`.
More importantly, it also defines how the import statements will be generated to the Python stubs.
Since we have a `shared.thrift` file, other generated stubs will need to import the generated `shared` module as well. Therefore, the resulting imports should reflect a valid directory structure that will be created by the `make package` target.

docs/web/api/README.md
barnabasdomozi: We actually do support it. You can increase it to `7.0`, and then the server would simply no longer accept requests from `6.x` clients.

docs/deps.md
barnabasdomozi: Should be updated in separate PR.

barnabasdomozi:
@bruntib 
Readded the Hello World Thrift script and fixed `make dist` in commit https://github.com/Ericsson/codechecker/commit/068985ffefc2502fb134ffab3451fc90f111ef41

codechecker_api/Makefile
barnabasdomozi: By default, Thrift generates `gen-*` directories. These should be gitignored. Accidentally, you can have a `gen-*` directory if you press CTRL+C during `thrift_stubs` target is built, before the `mv` command executes.

codechecker_api/.gitignore
bruntib: No, unfortunately these are not the current conventions. Here are some examples on the paths:

web/codechecker_web
web/client/codechecker_client
web/server/codechecker_server

Still we import all `codechecker_*` modules without the full path, because the build system flattens them, currently.
I think, we should follow the current conventions, until we change them.

bruntib:
I still get some error when running `CodeChecker server` installed from dist package:
```
# Create the dist package
make dist

# Create a clean environment
python3 -m venv venv
source venv/bin/activate

# Install the package
pip install /path/to/codechecker/dist/codechecker-6.30.0.tar.gz

# Start CodeChecker server
CodeChecker server
```

barnabasdomozi:
@bruntib 

Fixed in commit https://github.com/Ericsson/codechecker/commit/aa5dbacefa33a0e26318a934d96ee658ba1d4712
Also added a server log which prints the API version.

bruntib (approved):

PR #5103, opened 2026-09-16
Use non-cryptographic MD5 hashing for FIPS compliance

### LINKED ISSUE #4308: Add FIPS support to CodeChecker

wadehunk:
Resolves runtime errors on FIPS-enforced hosts where calling `hashlib.md5()` throws exceptions due to MD5 being blocked for cryptographic operations by default.

#### Changes Made

* **Centralized Hashing Helper:** Added `md5_hexdigest(data: bytes)` to `codechecker_common/util.py` and `codechecker_report_converter/util.py`.
* **FIPS Compliance Flag:** Enabled `usedforsecurity=False` in `hashlib.md5()` calls for Python 3.9+ environments to explicitly denote non-cryptographic checksum usage (e.g., source path hashing, report directory naming).
* **Backwards Compatibility:** Added a `TypeError` fallback for Python versions prior to 3.9 where the `usedforsecurity` keyword argument is unsupported.
* **Code Refactoring:** Replaced individual `hashlib.md5()` calls across analyzer modules, test runners, report converters, and web client commands with the new helper function.
* **Test Optimization:** Cached intermediate line hashes (`suppress_line_hash`) in `test_suppress.py` and `test_suppress_generation.py` to eliminate redundant MD5 re-computations.

Fixes #4308


codechecker_common/util.py
```diff
@@ -194,6 +194,20 @@ def load_args_from_file(filepath: str) -> list[str]:
...
 
+def md5_hexdigest(data: bytes) -> str:
+    """
+    Computes MD5 hash with non-cryptographic flag for FIPS compliance.
+    """
+    try:
+        # 'usedforsecurity=False' allows non-cryptographic MD5 hashing on
+        # FIPS-enforced hosts
+        return hashlib.md5(data, usedforsecurity=False).hexdigest()
+    except TypeError:
```
barnabasdomozi: CodeChecker no longer supports Python versions below 3.11 so handling this case can be dropped.

tools/report-converter/codechecker_report_converter/util.py
```diff
@@ -121,3 +122,17 @@ def dump_json_output(
         out.write(f"{data_str}\n")
 
     return data_str
+
+
+def md5_hexdigest(data: bytes) -> str:
```
barnabasdomozi: This utility function should only be defined once - in codechecker_common module.

Unfortunately, as of now, in codechecker_report_converter we cannot import codechecker_common module.

Instead, I suggest:
- modify existing `hashlib.md5` calls and also pass `usedforsecurity=False`.
- add a `FIXME` comment above each line that in the future, `md5_hexdigest` should be imported from `codechecker_common` instead.

tools/tu_collector/tu_collector/tu_collector.py
```diff
@@ -260,7 +260,15 @@ def __analyzer_action_hash(build_action: CompileAction) -> str:
 
     build_info = source_file + '_' + ' '.join(args)
 
-    return hashlib.md5(build_info.encode(errors='ignore')).hexdigest()
+    try:
```
barnabasdomozi: The same applies here as for the codechecker_report_converter module. This utility function should only be defined once.

codechecker_common/util.py
```diff
@@ -194,6 +194,20 @@ def load_args_from_file(filepath: str) -> list[str]:
         return shlex.split(content)
 
 
+def md5_hexdigest(data: bytes) -> str:
+    """
+    Computes MD5 hash with non-cryptographic flag for FIPS compliance.
+    """
+    try:
+        # 'usedforsecurity=False' allows non-cryptographic MD5 hashing on
+        # FIPS-enforced hosts
```
barnabasdomozi: Please also add a comment linking Python hashlib documentation:
https://docs.python.org/3/library/hashlib.html

barnabasdomozi (changes requested):
Thank you for this contribution.
I have left some comments.

barnabasdomozi (changes requested):
Currently, the CI test are failing:

> Found 'fixup!' commits in this pull request. Please squash them before merging!

barnabasdomozi (approved):
LGTM

